PR #42, opened 2021-01-12
refactor: replace GSON-based Json reader with pure Kotlin version

kiiadi:
Replace the platform-specific GSON based Json parser with a pure Kotlin version based on a forward-read only stream of characters - moving closer to being able to stream parsing rather than having to buffer full responses into memory.


aajtodd:
closing in favor of #477

Issue #124, opened 2021-03-04
Automatic pagination


kggilmer:
## Community Note
  1. Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
  2. Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
  3. If you are interested in working on this issue, please leave a comment
       
## Describe the feature
Provide API's that support automatic pagination of operations marked with the [paginated trait](https://awslabs.github.io/smithy/1.0/spec/core/behavior-traits.html#paginated-trait)

## Proposed Solution

See the proposed [design doc](https://github.com/awslabs/smithy-kotlin/blob/main/docs/design/paginators.md)

aajtodd:
Note: `SdkAsyncIterable` ought to probably be renamed to `SdkAsyncSequence` since it is lazy like `Sequence` (`Iterable` on the other hand implies eager which it is not)

Update design as necessary when implementing to make this distinction clear.

PR #462, opened 2021-08-23
docs: pagination design

aajtodd:
## Issue \#

Design for #124 

## Description of changes

Design docs for pagination. These have been converted from quip but have yet to be approved for implementation. 

[Rendered](https://github.com/awslabs/smithy-kotlin/blob/docs-pagination/docs/design/paginators.md)

The largest outstanding question I have is what to do about the `Flow` extension. 


docs/design/paginators.md
```diff
@@ -68,71 +68,87 @@ structure FunctionConfiguration { ... }
...
+fun<T> SdkAsyncIterable<Iterable<T>>.flatten(): SdkAsyncIterable<T> = TODO("not-implemented")
+fun <T> SdkAsyncIterable<T?>.filterNotNull(): SdkAsyncIterable<T> = TODO("not-implemented")
+
+suspend fun<T> SdkAsyncIterable<T>.toList(): List<T> = TODO("not-implemented")
+suspend fun<T> SdkAsyncIterable<T>.toSet(): Set<T> = TODO("not-implemented")
+
+/**
+ * Terminal operator that collects the given iterable and calls [action] for each result
+ */
+inline fun <T> SdkAsyncIterable<T>.collect(crossinline action: suspend (T) -> Unit):Unit = TODO()
```
ianbotsf, line 111: **Nit**: Missing whitespace after colon in `Unit):Unit`

ianbotsf (approved):

docs/design/paginators.md
```diff
@@ -68,71 +68,87 @@ structure FunctionConfiguration { ... }
...
+      }
+
+      val result = client.listFunctions(req)
+      isFirstPage = false
+      cursor = result.nextMarker
+      return result
+   }
+
+   // "functions" is the member targeted by the trait's `items` property
+   val functions: SdkAsyncIterable<FunctionConfiguration> = mapNotNull(ListFunctionsResponse::functions).flatten()
```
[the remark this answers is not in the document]
aajtodd: If it were a sparse list then yes it would be `SdkAsyncIterable<T?>`. So no it wouldn't be every circumstance. I'm not necessarily convinced this is the right thing to do but I agree it is probably the most ergonomic to flatten and remove nulls.

docs/design/paginators.md
```diff
@@ -68,71 +68,87 @@ structure FunctionConfiguration { ... }
...
+   private var isFirstPage: Boolean = true
 
-    override val hasMorePages: Boolean
-        get() = isFirstPage || (cursor?.isNotEmpty() ?: false)
+   override operator fun iterator(): SdkAsyncIterator<ListFunctionsResponse> = this
 
-    override suspend fun next(): ListFunctionsResponse {
-        if (!hasMorePages) throw NoSuchElementException("no pages remaining")
+   override operator fun hasNext(): Boolean
+           = isFirstPage || (cursor?.isNotEmpty() ?: false)
```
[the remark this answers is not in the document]
aajtodd: Hmm that's a good point. It may be that `hasNext()` needs to be `suspend` as well because transformations of any kind off the original paginator will all suffer the same issue where we don't actually know if there is an item or not without getting the first page.

docs/design/paginators.md
```diff
@@ -68,71 +68,87 @@ structure FunctionConfiguration { ... }
...
+   /**
+    * Returns the results for the next page or null when no more results are available.
+    * @throws NoSuchElementException if [hasMorePages()] is false
+    */
+   suspend operator fun next(): T
+}
+
+inline fun<T, R> SdkAsyncIterable<T>.map(transform: (T) -> R): SdkAsyncIterable<R> = TODO("not-implemented")
+inline fun<T, R> SdkAsyncIterable<T>.mapNotNull(transform: (T) -> R?): SdkAsyncIterable<R> = TODO("not-implemented")
+inline fun<T, R> SdkAsyncIterable<Iterable<T>>.flatMap(transform: (T) -> Iterable<R>): SdkAsyncIterable<R> = TODO("not-implemented")
```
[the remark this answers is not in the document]
aajtodd: I would not consider it complete, we can add whichever we feel is appropriate and may extend it in the future. I imagine when I or whomever gets to implementing this we can hash it out in a PR what the right starting set of transforms is.

docs/design/paginators.md
```diff
@@ -0,0 +1,297 @@
...
+    operations: [ListFunctions]
+}
+
+@paginated(
+    inputToken: "Marker",
+    outputToken: "NextMarker",
+    pageSize: "MaxItems",
+    items: "Functions"
+)
+@http(method: "PUT", uri: "/2015-03-31/functions", code: 200)
```
kggilmer, line 34: nit

uri path probably doesn't need the date snapshot prefix, internal detail.

docs/design/paginators.md
```diff
@@ -0,0 +1,297 @@
...
+}
+
+list FunctionConfigurationList {
+    member: FunctionConfiguration
+}
+
+structure FunctionConfiguration { ... }
+```
+
+**NOTE**: The trait may be applied at the service level to set default pagination. See the trait definition and examples for the ways it can be used in a model.
```
kggilmer, line 59: nit

consider adding link to trait def

docs/design/paginators.md
```diff
@@ -0,0 +1,297 @@
...
+## Considerations
+
+1. **Discoverability** - pagination is mostly a convenience for common code customers are able to implement themselves. We want them to be able to discover and use the already implemented pagination logic easily so that they don’t end up writing it themselves unnecessarily. 
+2. **Forwards compatibility** - The use of `items` in the `paginated` trait is not required. ~~If a service team later updates the trait to target a specific member using `items` then previous pagination code should continue to work (which means you can’t really generate a paginator for only the member of `items` as adding `items` later would change the generated pagination code in a way that isn’t backwards compatible).~~
+    1. Update 2/12/2021: The Smithy team has indicated this would not be a valid model evolution. The recommendation was still to provide a way to paginate over the raw responses as well as providing a flatMap like operation for the `items` if specified.
+
+
+## API
+
+The SDK runtime will provide a `SdkAsyncIterable` interface that represents the public API for a paginated operation.
```
kggilmer, line 71: nit

Maybe more descriptive:

The SDK runtime will provide a `SdkAsyncIterable` interface that represents the public API **for the response of a** paginated operation.

docs/design/paginators.md
```diff
@@ -0,0 +1,297 @@
...
+```kotlin
+/**
+ * An asynchronous data stream. This type is an asynchronous version of [Iterable].
+ */
+interface SdkAsyncIterable<out T> {
+   operator fun iterator(): SdkAsyncIterator<T>
+}
+
+/**
+ * An asynchronous [Iterator].
```
kggilmer, line 83: suggestion

`An [Iterator] which asynchronously returns the next value` since it doesn't look like `hasNext()` suspends.

docs/design/paginators.md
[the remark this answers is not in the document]
kggilmer: Maybe I'm missing something but can't we know definitively via the [presence of `sparse`](https://awslabs.github.io/smithy/1.0/spec/core/model.html#list)?

kggilmer (approved):
Other that ongoing issue of suspendability of `hasNext()`, LGTM

docs/design/paginators.md
```diff
@@ -0,0 +1,297 @@
...
+
+
+
+### Example Usage - Manual Pagination
+
+An example of driving a paginator manually and processing results:
+
+```kotlin
+suspend fun rawPaginationExample(client: LambdaClient) {
+    val req = ListFunctionsRequest{}
```
kggilmer, line 179: nit

whitespace

docs/design/paginators.md
```diff
@@ -0,0 +1,297 @@
...
+```
+
+
+
+### Example Usage - Manual Pagination
+
+An example of driving a paginator manually and processing results:
+
+```kotlin
+suspend fun rawPaginationExample(client: LambdaClient) {
```
kggilmer, line 178: suggestion

Not specific to this PR but in general it may be better if we specify types in code snippets in documentation since readers will not have the benefit of the IDE to tell them.  In particular to this example it may be helpful for the reader to know that `.functions` returns `SdkAsyncIterable<FunctionConfiguration>`

docs/design/paginators.md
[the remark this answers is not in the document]
ianbotsf: The more I think about this, the more I wonder if there are actually _two_ types here:

A page-fetching iterator that knows _without suspending_ when more items are available:
```kotlin
interface AsyncResponseIterator<out T> {
    operator fun hasNext(): Boolean
    suspend operator fun next(): T
}
```

An item-fetching iterator that _may have to fetch one or more pages_ (i.e., suspend) to determine if there are more items:
```kotlin
interface AsyncItemsIterator<out T> {
    suspend operator fun hasNext(): Boolean
    suspend operator fun next(): T
}
```

docs/design/paginators.md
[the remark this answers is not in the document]
aajtodd: Yes we can, I _think_ Ian was just asking if this would always be generated this way or if there are cases (like sparse) where we would generate something different. 

I'll try and clarify in the doc this call out.

docs/design/paginators.md
aajtodd: I'm not sure what you're suggesting, the whitespace looks good to my 👀

PR #468, opened 2021-08-30
refactor: remove raw serde functions

### LINKED ISSUE #265: remove the raw serde interfaces
We should remove the serialize "raw" function(s) from the serde interfaces:

```kt
serializeRaw(value: String)
rawField(descriptor: SdkFieldDescriptor, value: String)
rawEntry(key: String, value: String)
```

These methods was added primarily to allow for epoch timestamps to be formatted with the correct precision (without supporting Instant in the serde interface directly). We could have gone through conversion to double but we risk losing precision in the conversion. This API allows us to format the value however we want beforehand and serialize it (can't go through string, it needs to be a JSON number).

The idea though was this is a more general mechanism to allow anything to be formatted arbitrarily (an escape hatch).

This was a bandaid though and we should remove this before it gets us into trouble and come up with a better approach for handling `Instant`.

aajtodd:
## Issue \#

closes #265

## Description of changes

* Removes the `raw*` functions from serializer in favor of an explicit `serializeInstant` function. Each format can decide what they need to do to support serializing timestamps (which mostly consists of JSON serializing epoch seconds to a JSON number). 


ianbotsf (approved):

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/SerializeStructGenerator.kt
```diff
@@ -679,3 +668,10 @@ open class SerializeStructGenerator(
         return "serialize$suffix"
     }
 }
+
+fun TimestampFormatTrait.Format.toRuntimeEnum(): String = when (this) {
```
kggilmer, line 672: suggestion

would be nice to link to the spec that defines this mapping for future readers.  https://awslabs.github.io/smithy/1.0/spec/core/protocol-traits.html#timestampformat-trait

kggilmer (approved):

PR #472, opened 2021-08-30
Chore kotlin version bumps

kggilmer:
## Description of changes

* Update kotlin, coroutine, and kotest dependencies to latest version

## Scope

- [ ] New feature (non-breaking change which adds functionality)


runtime/protocol/http-client-engines/http-client-engine-ktor/jvm/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngine.kt
```diff
@@ -38,6 +39,7 @@ class KtorEngine(val config: HttpClientEngineConfig) : HttpClientEngineBase("kto
     }
     private val logger = Logger.getLogger<KtorEngine>()
 
+    @OptIn(InternalAPI::class)
```
ianbotsf, line 42: **Question**: Isn't opting into using their internal API a bad thing? Shouldn't we look to rid ourselves of Ktor internal usage instead?

kggilmer, line 42: FYI I was unable to get other variations of this to work.  Annotating at the class level or function level of `@InternalApi` throws an error saying that super needs the same annotation but super does not depend on ktor.  I found this to be a better than adding a dependency on ktor.

ianbotsf: Ktor's internal APIs are not stable and are subject to change. Presumably, the reason they're public is to allow sharing between subprojects within Ktor, not to allow 3P callers to reuse them. Shouldn't we replace these calls with either stable APIs or our own custom implementations?

kggilmer: After looking at the ktor code for awhile, I'm not seeing a way of doing that.  Given the updated policy on experimental APIs, it would seem that all of `StringValues` becomes experimental (the interface itself references `StringValuesBuilder` which is annotated as Experimental.  Both headers and (query) parameters are based on this type. I'm probably missing something obvious but as it stands I don't see how to construct an instance of Ktor's `HttpRequest` without opting in to internal API usage.

aajtodd: This is more than likely a bug in ktor. The [HttpRequestBuilder](https://github.com/ktorio/ktor/blob/main/ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt#L63) type is a pretty fundamental type used throughout the `HttpClient` and `HttpRequestPipeline/HttpResponsePipeline`.

aajtodd: Indeed: https://youtrack.jetbrains.com/issue/KTOR-3001

They basically just said to optin and it will be removed in a future release.

runtime/protocol/http-client-engines/http-client-engine-ktor/jvm/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorRequestAdapter.kt
```diff
@@ -50,6 +51,7 @@ internal class KtorRequestAdapter(
         }
     }
 
+    @InternalAPI
     fun toBuilder(): KtorRequestBuilder {
```
aajtodd, line 55: We shouldn't propagate `@InternalAPI` we should optin at the sites that it is used.

runtime/protocol/http-client-engines/http-client-engine-ktor/jvm/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngine.kt
ianbotsf: Ah, I missed that in my searching yesterday. Then I agree accepting opt-in seems like our only recourse for now.

ianbotsf: After seeing [the KTOR issue](https://youtrack.jetbrains.com/issue/KTOR-3001) pointed out by @aajtodd I agree we have to opt in. Can we please comment these opt-ins with a link to the issue and note that once the issue is resolved, we can remove the opt-ins?

runtime/protocol/http-client-engines/http-client-engine-ktor/jvm/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorRequestAdapter.kt
kggilmer: Thanks, yeah wrong annotations here.

runtime/protocol/http-client-engines/http-client-engine-ktor/jvm/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngine.kt
kggilmer: I'm fine with that and it looks like we won't have to wait long:

"We’re aiming to release 2.0 hopefully by October."
- https://blog.jetbrains.com/ktor/2021/08/10/an-update-on-our-2020-2021-roadmap/

kggilmer: Sure, will do.

ianbotsf (approved):

aajtodd (approved):

PR #473, opened 2021-08-31
feat: serde benchmarks

aajtodd:
## Issue \#

N/A

## Description of changes

Adds some benchmarks for serialization. I created a Smithy model for the twitter feed JSON document, generated the model, and copied the types + deserializers into the `model.twitter` folder. Did some minor cleanup on package names and what not. I don't expect to need to update this often but if we do we can create a script to automate it. 


runtime/serde/serde-benchmarks/jvm/src/aws/smithy/kotlin/serde/benchmarks/json/CitmBenchmark.kt
```diff
@@ -0,0 +1,28 @@
...
+import aws.smithy.kotlin.runtime.serde.json.jsonStreamReader
+import kotlinx.benchmark.*
+import kotlinx.coroutines.runBlocking
+
+@Warmup(iterations = 7, time = 1)
+@Measurement(iterations = 5, time = 1, timeUnit = BenchmarkTimeUnit.MILLISECONDS)
+@OutputTimeUnit(BenchmarkTimeUnit.MILLISECONDS)
+@BenchmarkMode(Mode.AverageTime)
+@State(Scope.Benchmark)
+open class CitmBenchmark {
```
ianbotsf, line 18: **Style**: These annotations are repeated in both benchmark classes and presumably would be on future benchmarks as well. Can we abstract a base class for them? Alternatively, can we define `7`, `5`, `1`, etc. as constants?

runtime/serde/serde-benchmarks/jvm/src/aws/smithy/kotlin/serde/benchmarks/json/twitter/model/Entities.kt
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+// Code generated by smithy-kotlin-codegen. DO NOT EDIT!
+
+package aws.smithy.kotlin.serde.benchmarks.json.twitter.model
+
+class Entities private constructor(builder: BuilderImpl) {
```
ianbotsf, line 10: **Comment**: Checking in this generated code means it'll get out of sync with our codegen, thus separating the benchmark results from reality. Can we codegen this as part of the build instead?

settings.gradle.kts
```diff
@@ -37,6 +39,7 @@ include(":runtime:serde")
 include(":runtime:serde:serde-json")
 include(":runtime:serde:serde-xml")
 include(":runtime:serde:serde-form-url")
+include(":runtime:serde:serde-benchmarks")
 include(":runtime:protocol:http")
```
ianbotsf, line 43: **Question**: Why does this new test package belong in the **runtime** subproject? We have other non-publishing test packages like [**android-test**](https://github.com/awslabs/smithy-kotlin/tree/main/android-test) and [**compile-tests**](https://github.com/awslabs/smithy-kotlin/tree/main/compile-tests) as their own top-level projects. Seems like this could go at top-level or under a new **tests** subproject.

ianbotsf (approved):

settings.gradle.kts
aajtodd: Is there a reason not to? Having it close to what it tests seems like the most obvious location to me. The other two test projects aren't specific to one thing (also `android-test` is likely to just be deleted) which is why they don't live next to something more specific.

We could restructure it and have a root `tests` or `bechmarks` folder though. I think if we are to move it my preference would be for `benchmarks` folder with `serde-benchmarks` beneath that.

runtime/serde/serde-benchmarks/jvm/src/aws/smithy/kotlin/serde/benchmarks/json/twitter/model/Entities.kt
aajtodd: If I could have done it easily I would have.

* we don't have an easy way to just spit out _only_ the document serializers/deserializers + model types
* there is no protocol other than the AWS ones to choose from which means
    * we would have to define a separate package, implement a dummy json/xml/whatever format protocol. 
    * depend on it from this benchmark build
    * implement logic to have `compileKotlin` tasks run after codegen tasks
    * copy/move the bits we need from codegen into the main sourceSet

It's certainly possible but the amount of time to invest in this didn't seem worth it right now. If we're that worried I'd rather just remove the benchmarks for serialize/deserialize and only add them once we have the time to invest in something like above.

runtime/serde/serde-benchmarks/jvm/src/aws/smithy/kotlin/serde/benchmarks/json/CitmBenchmark.kt
aajtodd: I think we can define it in the gradle settings as the defaults, will take a look. I don't think constants are going to help, choosing these numbers is likely not going to always be appropriate for every test. The number of warmups, benchmark mode, etc.

settings.gradle.kts
ianbotsf: One reason not to is to avoid inheriting config we don't want from parent projects, as you had to do for the new `skipPublish` configuration.

I think having a top-level **benchmarks** subproject makes sense (and even a top-level **tests** subproject). That'll facilitate a future where we can `./gradlew :benchmarks:all` (or something similar) to run every benchmark we have.

runtime/serde/serde-benchmarks/jvm/src/aws/smithy/kotlin/serde/benchmarks/json/twitter/model/Entities.kt
ianbotsf: Hmm, I see the problems. If we can't keep the types and structurers/destructurers up to date easily, that curbs the value of checking in benchmarks for those things (for me). I now lean towards removing them but my opinion isn't strong enough to make it a requirement.

I bet some refactoring of our codegen could get us a clean way to generate just the layers we need for this test. This also seems like a reason why it may be good for non-AWS protocols to be defined and for **smithy-kotlin** to support them.

settings.gradle.kts
aajtodd: That's the thing though we do want to inherit _almost_ all of that config. Also thats basically how gradle works anyway, you can just use an unqualified task name like `./gradlew jvmBenchmark` and it will run every task with that name.

But at any rate I'm fine moving it to a top level project under a `benchmarks` folder.

runtime/serde/serde-benchmarks/jvm/src/aws/smithy/kotlin/serde/benchmarks/json/twitter/model/Entities.kt
aajtodd: It would be useful to be able to just generate entities OR entities + "document" serializers/deserializers for some format (json/xml/etc). This wouldn't require a protocol and I could see people finding it useful. 

The gradle problems would be solvable it just takes time to invest in getting the build setup.

runtime/serde/serde-benchmarks/models/twitter.smithy
```diff
@@ -0,0 +1,205 @@
+$version: "1.0"
+
+namespace aws.benchmarks.twitter
+
+use aws.api#service
+use aws.protocols#restJson1
+
+@service(sdkId: "Twitter JSON Benchmark")
```
kggilmer, line 8: comment

I wonder if any of this work is upstreamable to Smithy.  It would seem that any client SDK generator would want to measure performance of at least serde.  OTOH things outside of the scope of codegen would also want to be benchmarked so not a perfect fit for a Smithy-based solution.  :shrug:

kggilmer (approved):
Impressive work!  As we've discussed offline I'm curious to explore other ways of expressing benchmark tests.  Ideally we would have:
* benchmark test execution automation
* ability to save results (mapped to sdk version?  commit hash?) and see trends
* test the output of codegen (or rt) without hand rolling anything

I'm a little concerned by the amount of code we need to add to achieve this (although that's not a criticism of the PR, just the nature of how our SDK works atm), but am guessing that this particular benchmark may be one of the more verbose use cases.

In any case, none of these thoughts should block proceeding of this work.

settings.gradle.kts
```diff
@@ -5,7 +5,6 @@
 
 pluginManagement {
     repositories {
-        maven("https://dl.bintray.com/kotlin/kotlin-eap")
```
ianbotsf, line 8: **Question**: What was this for? Why can we now remove it?

benchmarks/serde-benchmarks-codegen/src/main/kotlin/SerdeBenchmarkJsonProtocolGenerator.kt
```diff
@@ -0,0 +1,65 @@
...
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.model.shapes.ServiceShape
+import software.amazon.smithy.model.shapes.Shape
+import software.amazon.smithy.model.shapes.ShapeId
+import software.amazon.smithy.model.traits.TimestampFormatTrait
+
+/**
+ * Protocol generator for benchmark protocol [SerdeBenchmarkJsonProtocol]
+ */
+class SerdeBenchmarkJsonProtocolGenerator : HttpBindingProtocolGenerator() {
```
ianbotsf, line 20: **Comment**: Very clean. I like it!

ianbotsf (approved):
Fantastic work here!

settings.gradle.kts
aajtodd: I don't recall what it was for, it's from very early on. All references to `jcenter` and/or `bintray` should be removed though because they are [shutting down](https://blog.gradle.org/jcenter-shutdown). I actually just recently removed `jcenter` from this repo but missed this reference. Happened to catch it debugging something else. Having it present slows our builds down because it tries to resolve plugins/dependencies and gets an error response and then resolves using a different repo.

benchmarks/serde-benchmarks-codegen/src/main/kotlin/SerdeBenchmarkJsonProtocol.kt
```diff
@@ -0,0 +1,20 @@
...
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+package software.amazon.smithy.kotlin.codegen.protocols
+
+import software.amazon.smithy.kotlin.codegen.integration.KotlinIntegration
+import software.amazon.smithy.kotlin.codegen.rendering.protocol.ProtocolGenerator
+import software.amazon.smithy.model.shapes.ShapeId
+
+/**
+ * Dummy protocol for use in serde-benchmark project models. Generates JSON based serializers/deserializers
```
kggilmer, line 12: nit

Maybe there is something conveyed in the term 'dummy' but 'test protocol' sounds more descriptive to me.

benchmarks/serde-benchmarks/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+This project contains micro benchmarks for the serialization implementation(s).
+
+## Testing
+
+```sh
+./gradlew :runtime:serde:serde-benchmarks:jvmBenchmark
+```
+
+Baseline `0.4.0-alpha`
```
kggilmer, line 11: nit

for convenience in reproducability maybe note the commit hash from which these baseline measurements since obviously this wasn't run on the commit of the alpha release.

settings.gradle.kts
```diff
@@ -41,6 +44,9 @@ include(":runtime:protocol:http")
 include(":runtime:protocol:http-client-engines:http-client-engine-ktor")
 
 include(":compile-tests")
+include(":benchmarks")
```
kggilmer, line 47: question

given the note above about the module failing to compile until `assemble` is called, perhaps we want to gate the inclusion of these modules behind some kind of `local.properties` setting?  Making it opt-in essentially.  Do we expect there to be actions performed in these modules as part of routine SDK development?

benchmarks/serde-benchmarks/build.gradle.kts
```diff
@@ -0,0 +1,129 @@
...
+    id("software.amazon.smithy")
+}
+
+buildscript {
+    val smithyVersion: String by project
+    dependencies {
+        classpath("software.amazon.smithy:smithy-cli:$smithyVersion")
+    }
+}
+
```
kggilmer, line 19: nit

extra whitespaces in a few places in this file

benchmarks/serde-benchmarks/build.gradle.kts
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+}
+
+benchmark {
+    targets {
+        register("jvm")
+    }
+
+    configurations {
```
kggilmer, line 65: question

Is the idea here that we may define multiple benchmarking configurations, like say one perhaps specifically for CI?  How are the configurations specified at execution time? (I searched but was unable to come up w/ suitably specific terms)

benchmarks/serde-benchmarks/build.gradle.kts
```diff
@@ -0,0 +1,129 @@
...
+    group = "codegen"
+    dependsOn(generateSdk)
+    doLast {
+        benchmarkModels.forEach {
+            copy {
+                from("${it.projectionRootDir}")
+                into("${it.sourceSetRootDir}")
+                include("**/model/*.kt")
+                include("**/transform/*.kt")
+                exclude("**/transform/*OperationSerializer.kt")
```
kggilmer, line 120: (curious) question

Is the reason for exclusion here that these generated files are invalid due to implementation details of the benchmark protocol, or just cleanliness?  (if you remove these exclusions does the benchmarking still work?)

benchmarks/serde-benchmarks/jvm/src/aws/smithy/kotlin/benchmarks/serde/json/CitmBenchmark.kt
```diff
@@ -0,0 +1,25 @@
...
+
+@BenchmarkMode(Mode.AverageTime)
+@State(Scope.Benchmark)
+open class CitmBenchmark {
+    private val input = CitmBenchmark::class.java.getResource("/citm_catalog.json")!!.readBytes()
+
+    @Benchmark
+    fun tokensBenchmark() = runBlocking {
+        val tokenizer = jsonStreamReader(input)
+        do {
```
kggilmer, line 21: nit

perhaps test the iteration count matches an expected value, in case a bug somehow causes no tokens to be executed.

benchmarks/serde-benchmarks/model/serde-protocols.smithy
```diff
@@ -0,0 +1,8 @@
+$version: "1.0"
+
+namespace aws.benchmarks.protocols
+
+// dummy protocol just for benchmarking purposes
```
kggilmer, line 5: nit/style

again i think 'test' is more descriptive than 'dummy'.

benchmarks/serde-benchmarks/model/twitter.smithy
```diff
@@ -0,0 +1,203 @@
...
+    count: Integer,
+    max_id: Long,
+    max_id_str: String,
+    next_results: String,
+    query: String,
+    refresh_url: String,
+    since_id: Integer,
+    since_id_str: String
+}
+
```
kggilmer, line 45: pedantic nit

extra whitespaces in model files

gradle.properties
```diff
@@ -1,7 +1,6 @@
 kotlin.code.style=official
 kotlin.incremental.js=true
 kotlin.incremental.multiplatform=true
-kotlin.parallel.tasks.in.project=true
```
kggilmer, line 4: question

is this going to slow things down?  this was removed to support benchmarks?

kggilmer (approved):
nice work!

benchmarks/serde-benchmarks/README.md
aajtodd: it wasn't but it also shouldn't matter we haven't made changes to serde since then

settings.gradle.kts
aajtodd: What note are you referring to? The module should compile fine out the gate.

benchmarks/serde-benchmarks/build.gradle.kts
aajtodd: Had to look it up myself, closest documentation I could find on it:
```
    configurations {
        main { 
            // configure default configuration
        }
        smoke { 
            // create and configure "smoke" configuration, e.g. with several fast benchmarks to quickly check
            // if code changes result in something very wrong, or very right. 
        }       
    }
```

So yeah it looks like if you wanted to multiple benchmark classifications you could group them into a single config and run them independent of the other.

benchmarks/serde-benchmarks/build.gradle.kts
aajtodd: Sort of multifaceted. (1) The benchmark protocol generator doesn't generate anything other than the "document" serializers/deserializers and (2) the benchmark test dependencies don't include any of the other modules other than serde. We would have to add dependencies on e.g. `http`, etc for these to compile.

gradle.properties
aajtodd: No this was removed because of a warning being generated, this property is taken by default from the `org.gradle.parallel` (`--parallel`) flag now by default and does not need to be specified separately

PR #474, opened 2021-09-07
chore(ci): remove tokens and add windows to ci

aajtodd:
## Issue \#

N/A

## Description of changes

* Add windows target to CI
* Removes unnecessary tokens that were required to checkout dependencies when repos were private

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -5,9 +5,6 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
-    paths-ignore:
-      - 'docs/**'
-      - '*.md'
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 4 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -5,9 +5,6 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
-    paths-ignore:
```
kggilmer, line 8: question

why did you remove this?

kggilmer (approved):
Nice work!  I was going to need windows coverage as part of my current task so perfect timing.

.github/workflows/continuous-integration.yml
aajtodd: Because as we discovered in a different branch if we actually trigger this then CI won't run (as you would expect). The issue is all of our checks for this workflow are marked as required status checks. You can't merge the branch without the status check passing which will never be generated because the workflow never runs.

ianbotsf (approved):

PR #477, opened 2021-09-14
feat(rt): replace GSON based JSON serde with KMP compatible impl

aajtodd:
## Issue \#

N/A

## Description of changes

Removes GSON in favor of a hand rolled JSON encoder and lexer.

This was based on early work from @kiiadi in #42 but updated to handle Unicode, escape sequences, and improve overall correctness/strictness of the parser. The state machine was adapted from work in smithy-rs by @jdisanti which drastically simplified handling various conditions as well as improved the error messages generated.

Implementation notes:
* The tokenizer is based on `peek()` which requires state mutations to be delayed until the token that is peeked is actually consumed. This allowed removing the `RawJsonToken` type.
* The original `CharStream` abstraction was removed for performance reasons. 
    * It turns out that `suspend` is incredibly slow when invoked on every character (not entirely surprising).
    * It is also is 3x slower to decode UTF-8 chars on the fly from raw byte sequences than just calling `decodeToString()` on the appropriate slice of data. I'm not entirely sure why but suspect it's either a JVM intrinsic somewhere and/or has an optimized ASCII loop that is just faster when done in bulk.
    * This lead to keeping the lexer/deserializer to work off the raw `ByteArray` payload which requires us to keep generated deserializers as they are now where they read the entire payload contents into memory (`val payload = httpResponse.body.readAll()`). This trades memory for CPU. The team discussed and agreed that most service responses will be small enough that this is the right trade off.

The tokenizer was ran through JSON test suite as described in [TESTING.md](https://github.com/awslabs/smithy-kotlin/blob/f45a4a1dd5c3d0be9bb098c4054767d82cd610b2/runtime/serde/serde-json/TESTING.md). 


runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
```diff
@@ -0,0 +1,433 @@
...
+
+    private fun unexpectedToken(found: Char?, vararg expected: String): Nothing {
+        val pluralModifier = if (expected.size > 1) " one of" else ""
+        val formatted = expected.joinToString(separator = ", ") { "'$it'" }
+        fail("found '$found', expected$pluralModifier $formatted")
+    }
+
+    private fun fail(message: String, offset: Int = idx, cause: Throwable? = null): Nothing {
+        throw DeserializationException("Unexpected JSON token at offset $offset; $message", cause)
+    }
```
ianbotsf, line 348: **Comment**: Raw offsets are not easy to use for debugging when dealing with multiline text. How feasible is it to support line number and position?

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
```diff
@@ -0,0 +1,433 @@
...
+        val value = data.decodeToString(start, idx)
+        consume('"')
+        return if (needsUnescaped) value.unescape() else value
+    }
+
+    private fun readKeyword(): JsonToken = when (val ch = peekOrThrow()) {
+        't' -> readLiteral("true", JsonToken.Bool(true))
+        'f' -> readLiteral("false", JsonToken.Bool(false))
+        'n' -> readLiteral("null", JsonToken.Null)
+        else -> fail("Unable to handle keyword starting with '$ch'")
```
ianbotsf, line 280: **Question**: This doesn't give the user a lot to go on. Can we give the full keyword we failed to parse?

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
```diff
@@ -0,0 +1,433 @@
...
+     * Invoke [consume] for each character in [expected]
+     */
+    private fun consume(expected: String) = expected.forEach { consume(it) }
+
+    /**
+     * Assert that the next character is [expected] and advance
+     */
+    private fun consume(expected: Char) {
+        val chr = data[idx].toInt().toChar()
+        check(chr == expected) { "Unexpected char '$chr' expected '$expected'" }
```
ianbotsf, line 309: **Question**: Should this be `lexerCheck` instead of `check`?

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
```diff
@@ -0,0 +1,433 @@
...
+                        }
+                        '\\' -> append('\\')
+                        '/' -> append('/')
+                        '"' -> append('"')
+                        'b' -> append('\b')
+                        'f' -> append("\u000C")
+                        'r' -> append('\r')
+                        'n' -> append('\n')
+                        't' -> append('\t')
+                        else -> throw DeserializationException("Invalid escape character: `$byte`")
```
ianbotsf, line 382: **Question**: Can we find a way for this to include position information?

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/LexerState.kt
```diff
@@ -0,0 +1,21 @@
...
+    // Expecting the next token to be the *first* value in an array, or the end of the array.
+    ArrayFirstValueOrEnd,
+    // Expecting the next token to the next value in an array, or the end of the array.
+    ArrayNextValueOrEnd,
+    // Expecting the next token to be the *first* key in the object, or the end of the object.
+    ObjectFirstKeyOrEnd,
+    // Expecting the next token to the next object key, or the end of the object.
+    ObjectNextKeyOrEnd,
+    // Expecting the next token to be the value of a field in an object.
+    ObjectFieldValue,
```
ianbotsf, line 20: **Nit**: Using KDoc comments will make these notes available in the IDE (e.g., Ctrl/Cmd-Q

runtime/serde/serde-json/common/test/aws/smithy/kotlin/runtime/serde/json/JsonStreamWriterTest.kt
```diff
@@ -9,68 +9,231 @@ import kotlin.test.assertEquals
...
+        "text": "@json_newb just use JsonWriter!",
+        "geo": [
+            50.454722,
+            -104.606667
+        ],
+        "user": {
+            "name": "jesse",
+            "followers_count": 2
+        }
     }
```
ianbotsf, line 36: **Nit**: We should probably write our own test data instead of reusing examples from https://www.javadoc.io/doc/com.google.code.gson/gson/2.8.0/com/google/gson/stream/JsonWriter.html

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/text/Utf8.kt
```diff
@@ -0,0 +1,53 @@
...
+@InternalApi
+fun Char.Companion.codePointToChars(codePoint: Int): CharArray = when (codePoint) {
+    in 0 until SUPPLEMENTARY_PLANE_LOW -> charArrayOf(codePoint.toChar())
+    in SUPPLEMENTARY_PLANE_LOW..MAX_CODEPOINT -> {
+        val low = MIN_LOW_SURROGATE.code + ((codePoint - 0x10000) and 0x3FF)
+        val high = MIN_HIGH_SURROGATE.code + (((codePoint - 0x10000) ushr 10) and 0x3FF)
+        charArrayOf(high.toChar(), low.toChar())
+    }
+    else -> throw IllegalArgumentException("invalid codepoint $codePoint")
+}
```
ianbotsf, line 53: **Style**: I don't think I've ever seen (effectively) static extension methods before. Given that there's no state to use/reuse in `Char.Companion`, I think these might be better as top-level non-extension methods or extensions on `Int`.

ianbotsf (commented):
Heroic!

runtime/serde/build.gradle.kts
```diff
@@ -19,7 +19,7 @@ kotlin {
     }
 }
 
-subprojects {
+allprojects {
```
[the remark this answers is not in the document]
aajtodd: It might not be now. It was because we added code to `serde/common` that needed it. I'll check if it's still required.

runtime/serde/serde-json/TESTING.md
```diff
@@ -0,0 +1,178 @@
+How to run JSONTestSuite against serde-json deserialize
+========================================================
+
+When making changes to the lexer it is a good idea to run the
+changes against the [JSONTestSuite](https://github.com/nst/JSONTestSuite) and manually examine the test results.
+
+### How to setup the JSONTestSuite
```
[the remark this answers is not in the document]
aajtodd: JSONTestSuite is a bit interesting/onerous to setup and use. It does not lend itself to this kind of integration and I see it as an operational cost not worth paying.

runtime/serde/serde-json/TESTING.md
```diff
@@ -0,0 +1,178 @@
...
+allprojects {
+    repositories {
+        mavenLocal()
+        mavenCentral()
+    }
+}
+
+
+// NOTE: set to whatever locally published version you are working on
+val smithyKotlinVersion: String = "0.4.1-kmp-json"
```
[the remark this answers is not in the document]
aajtodd: This readme is targeted towards developers (i.e. us). Not end users.

runtime/serde/serde-json/TESTING.md
```diff
@@ -0,0 +1,178 @@
...
+dependencies {
+   implementation(kotlin("stdlib"))
+   implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.0")
```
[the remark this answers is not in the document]
aajtodd: I'm not sure that's important, again this is targeted at developers of `smithy-kotlin` though so I would expect us to make changes here as necessary.

runtime/serde/serde-json/TESTING.md
```diff
@@ -0,0 +1,178 @@
...
+    manifest {
+        attributes["Main-Class"] = "aws.smithy.kotlin.jsontest.MainKt"
+    }
+}
+```
+
+4. Add the following code to `src/main/kotlin/Main.kt` with:
+
+```kotlin
+package aws.smithy.kotlin.jsontest
```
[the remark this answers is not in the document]
aajtodd: It isn't really important since this is creating an application jar. Also technically it would be in line with the recommended [directory structure](https://kotlinlang.org/docs/coding-conventions.html#directory-structure) if `aws.smithy.kotlin.jsontest` is the "root".

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonEncoder.kt
```diff
@@ -0,0 +1,157 @@
...
+import aws.smithy.kotlin.runtime.util.*
+
+// character code points
+private const val CP_QUOTATION: Int = 0x22
+private const val CP_BACKSLASH: Int = 0x5C
+private const val CP_NEWLINE: Int = 0x0A
+private const val CP_CARRIAGE_RETURN: Int = 0x0D
+private const val CP_TAB: Int = 0x09
+private const val CP_BACKSPACE: Int = 0x08
+private const val CP_FORMFEED: Int = 0x0C
```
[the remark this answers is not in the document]
aajtodd: they are not required, habit/copy-pasta

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonEncoder.kt
```diff
@@ -0,0 +1,157 @@
...
+        }
+        depth--
+        writeIndent()
+        buffer.append(token)
+        state.pop()
+    }
+
+    private fun writeIndent() {
+        if (pretty && depth > 0) {
+            val indent = " ".repeat(depth * 4)
```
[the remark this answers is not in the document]
aajtodd: Actually that brings about a good question. We had an option for `pretty` before so I made an attempt to keep it. I'm not sure we actually need to support a pretty output though...any thoughts?

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonEncoder.kt
```diff
@@ -0,0 +1,157 @@
...
+    private var depth: Int = 0
+
+    override fun beginObject() = openStructure("{", LexerState.ObjectFirstKeyOrEnd)
+    override fun endObject() = closeStructure("}")
+    override fun beginArray() = openStructure("[", LexerState.ArrayFirstValueOrEnd)
+    override fun endArray() = closeStructure("]")
+
+    private fun openStructure(token: String, nextState: LexerState) {
+        encodeValue(token)
+        if (pretty) writeNewline()
```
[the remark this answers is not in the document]
aajtodd: Ha so it is. Good catch

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonEncoder.kt
```diff
@@ -0,0 +1,157 @@
...
+        buffer.appendQuoted(name.escape())
+        state.replaceTop(LexerState.ObjectFieldValue)
+    }
+
+    override fun writeValue(value: String) = encodeValue("\"${value.escape()}\"")
+
+    override fun writeValue(bool: Boolean) = when (bool) {
+        true -> encodeValue("true")
+        false -> encodeValue("false")
+    }
```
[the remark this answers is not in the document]
aajtodd: Is it guaranteed to always spit out lowercase "true/false"? I would assume so...will update.

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
```diff
@@ -0,0 +1,433 @@
...
+
+    /**
+     * Push a pending mutation
+     */
+    fun mutate(mutation: StateMutation) { pendingMutations.add(mutation) }
+
+    /**
+     * Get the top of the state stack
+     */
+    fun current(): LexerState = state.top()
```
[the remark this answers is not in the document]
aajtodd: Sure will update. This is probably a result of numerous refactors.

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
```diff
@@ -0,0 +1,433 @@
...
+                LexerState.ObjectFirstKeyOrEnd -> stateObjectFirstKeyOrEnd()
+                LexerState.ObjectNextKeyOrEnd -> stateObjectNextKeyOrEnd()
+                LexerState.ObjectFieldValue -> stateObjectFieldValue()
+            }
+        } catch (ex: DeserializationException) {
+            throw ex
+        } catch (ex: Exception) {
+            throw DeserializationException(cause = ex)
+        }
+    }
```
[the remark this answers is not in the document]
aajtodd: I find this hurts my eyes personally. Braces can be a good thing you know :)

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
aajtodd: I gave it some thought but think we would be maybe better off trying to implement better error messages in the deserializer using a path like syntax:

```
Error deserializing path=`.foo.bar[1].blah`; <original cause from lexer>
```

I wasn't sure it was worth the overhead or not to go down this route. I do think in general I agree that the better our error messages are it will help us if we have bugs in the either the lexer/deserializer or in codegen. 

I'm open to ideas.

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
aajtodd: well it gives us the position and the character. We don't know what the word is since it wouldn't be a keyword. Perhaps updating the error message to `Invalid literal starting with $ch`?

runtime/serde/serde-json/common/test/aws/smithy/kotlin/runtime/serde/json/JsonStreamWriterTest.kt
aajtodd: Ha didn't even know that is where it was from. This test was originally written by Zeynep who was briefly on the team. I'll remove and write our own

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/text/Utf8.kt
aajtodd: > Given that there's no state to use/reuse in Char.Companion

Not sure why this matters. It's more about scoping the methods to their intended use. I could see them as extensions off `Int` though if you feel strongly.

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonEncoder.kt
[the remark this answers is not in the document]
ianbotsf: Pretty output is definitely helpful for debugging. I could even see us using it for request logging. I'd vote to retain it.

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
```diff
@@ -0,0 +1,433 @@
...
+
+    override suspend fun peek(): JsonToken = peeked ?: doPeek().also { peeked = it }
+
+    override suspend fun skipNext() {
+        val startDepth = state.size
+        nextToken()
+        while (state.size > startDepth) {
+            nextToken()
+        }
```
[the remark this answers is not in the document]
ianbotsf: Ah got it. That's now clear from re-reading the code but when I was in the weeds yesterday that intent somehow escaped me.

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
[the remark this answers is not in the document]
ianbotsf: I personally find them to be unnecessary boilerplate when they wrap a single expression (like parentheses around a scalar). I don't care hard enough in this case though...we can leave it as is.

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
ianbotsf: I like the idea of including path information where available but it may not localize the problem very well in some cases. Because the errors represent a failure to parse a well-formed document, the paths cannot localize exactly where the error site is.

In this example:

```json
{
  "foo": "bar",
  "baz": [
    { "key": "val1" },
    { "key": "val2" },
    { "key": "val3" },
    ...
    { "key": "val1023" }
    { "key": "whoops!   ^ I forgot the comma after the last object" },
  ]
}
```

The best path we could derive is `".baz"` right?

In any event, giving a path may help for short documents that can be quickly read by humans. For longer documents that don't fit in an window or require significant mental effort to parse, I still think line:pos is a helpful locator to have alongsize path.

aajtodd: Yeah so that's a good point (although I think path _may_ be able to be something like `.baz[n]` where `n` is the closest element). I don't disagree that line:pos may be more helpful but in practice services don't return pretty JSON anyway so there won't be a line number. For a general purpose JSON parser though you are correct that it would probably be more useful. Let me think on it, I definitely want the errors to be useful...

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
ianbotsf: I think the starting character doesn't tell us much. You're right that position will be helpful (all the moreso if we can augment it to be line:pos and/or include the path) but the starting character doesn't help with searching or getting more context for debugging. Ideally, I'd like to see something like:

```json
{
  "a": true,
  "b": false,
  "c": trueish
}
```

Lead to the error message: `"Expected one of ['true', 'false', 'null'] but found 'trueish'"`. I believe that'd be the most helpful message we could give debuggers.

aajtodd: If you don't know what you're parsing how do you know what to parse and when to stop? At this point in lexical analysis we have an invalid JSON document, it could be anything that follows.

ianbotsf: Yeah, once the document's known to be invalid then it gets tricky for sure.

I'm just hoping we can gobble up the text we find until we encounter a whitespace or punctuation. A regex like `"[-_a-zA-Z0-9]+"` encapsulates something that would get us the next literal word even though the document is invalid and we're going to fail parsing. If you're worried about length (i.e., potentially grabbing the remainder of some very long document) we could even cap the size of the word we'll attempt to include in the error message (e.g., no more than 20 chars).

I think this would improve debuggability if it's possible. If it's not or introduces other concerns then we'll just have to rely on whatever positioning/pathing we can get.

runtime/serde/serde-json/TESTING.md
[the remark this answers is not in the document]
kggilmer: In this line of thinking, can this code as-is serve as a general purpose JSON parser?  If so perhaps it would make sense to break out as a separate dependency at some point in the future such that JSONTestSuite could run against it as the others they have in their repo.  Obviously not something for this PR.

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonEncoder.kt
```diff
@@ -0,0 +1,157 @@
...
+    override fun writeValue(value: Long) = writeNumber(value)
+    override fun writeValue(value: Short) = writeNumber(value)
+    override fun writeValue(value: Int) = writeNumber(value)
+    override fun writeValue(value: Float) = writeNumber(value)
+    override fun writeValue(value: Double) = writeNumber(value)
+
+    override fun writeRawValue(value: String) = encodeValue(value)
+}
+
+internal fun String.escape(): String {
```
kggilmer, line 125: suggestion

This may be more efficient in time and concise if implemented with a map.  Here we have two iterations on the input string, once to check to see if anything should be done, and then again to do a conditional transform.  This could be collapsed to one iteration if there is a static `Map<Char, String>` which contains the escape values and then simply calling `Map.getOrElse(chr)` on each character.

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
```diff
@@ -0,0 +1,433 @@
...
+ */
+private data class StateManager(
+    private val state: StateStack = mutableListOf(LexerState.Initial),
+    private val pendingMutations: MutableList<StateMutation> = mutableListOf()
+) {
+
+    /**
+     * The size of the state stack
+     */
+    val size: Int
```
kggilmer, line 32: nit

consider `UInt`

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
```diff
@@ -0,0 +1,433 @@
...
+    private var idx = 0
+
+    override suspend fun nextToken(): JsonToken {
+        val next = peek()
+        peeked = null
+        state.update()
+        return next
+    }
+
+    override suspend fun peek(): JsonToken = peeked ?: doPeek().also { peeked = it }
```
kggilmer, line 71: :pinching_hand:

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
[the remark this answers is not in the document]
kggilmer: I also prefer the more concise style to pull the `return` out of the `try`, and would argue it's more readable than the nested `return` FWIW.

kggilmer (commented):
I spent some time but only got maybe a quarter way through this.  I'll continue but will share what I have so far.  Very clean, nice work.

runtime/serde/serde-json/TESTING.md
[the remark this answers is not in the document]
aajtodd: Unless we are planning to support it as a general purpose parser I would think it probably isn't in our interest (or anyone else) to do that. Perhaps we fork it and set it up so that it's easier to run...

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/Stack.kt
```diff
@@ -0,0 +1,45 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package aws.smithy.kotlin.runtime.util
+
+/**
+ * Convenience type for treating a list like stack
+ */
+typealias ListStack<T> = MutableList<T>
```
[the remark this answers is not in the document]
aajtodd: forgot it existed there, yes we can remove the one defined in `XmlNode`

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonEncoder.kt
aajtodd: hmm ya I'll look into it

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
```diff
@@ -0,0 +1,433 @@
...
+    private fun readLiteral(expectedString: String, token: JsonToken): JsonToken {
+        consume(expectedString)
+        return token
+    }
+
+    /**
+     * Advance the cursor until next non-whitespace character is encountered
+     * @param peek Flag indicating if the next non-whitespace character should be consumed or peeked
+     */
+    private fun nextNonWhitespace(peek: Boolean = false): Char? {
```
kggilmer, line 292: suggestion

I don't find any place where this function is called where `peek` is `false`, maybe unneeded.

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
```diff
@@ -0,0 +1,433 @@
...
+ * @return The number of characters consumed
+ */
+private fun readEscapedUnicode(s: String, start: Int, sb: StringBuilder): Int {
+    // already consumed \u escape, take next 4 bytes as high
+    val high = s.substring(start, start + 4).decodeEscapedCodePoint()
+    var consumed = 4
+    if (high.isHighSurrogate()) {
+        val lowStart = start + consumed
+        val escapedLow = s.substring(lowStart, lowStart + 6)
+        check(escapedLow.startsWith("\\u")) { "Expected surrogate pair, found `$escapedLow`" }
```
kggilmer, line 410: nit

This PR mixes ``` and `'` characters to represent quoted literals in error messages. I'm not aware of any convention we have defined as a team but should probably be consistent.

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
```diff
@@ -0,0 +1,433 @@
...
+ * been consumed and [start] is pointing to the first hex digit. If the code point represents a surrogate pair
+ * an additional escaped code point will be consumed from the string.
+ * @param s The string to decode from
+ * @param start The starting index to start reading from
+ * @param sb The string builder to append unescaped unicode characters to
+ * @return The number of characters consumed
+ */
+private fun readEscapedUnicode(s: String, start: Int, sb: StringBuilder): Int {
+    // already consumed \u escape, take next 4 bytes as high
+    val high = s.substring(start, start + 4).decodeEscapedCodePoint()
```
kggilmer, line 405: suggestion

guard that `start + 4` is not out of bounds for the string.

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonStreamReader.kt
```diff
@@ -23,11 +24,25 @@ interface JsonStreamReader {
...
-    suspend fun peek(): RawJsonToken
+    suspend fun peek(): JsonToken
 }
 
 /*
 * Creates a [JsonStreamReader] instance
 */
 @InternalApi
-expect fun jsonStreamReader(payload: ByteArray): JsonStreamReader
+fun jsonStreamReader(payload: ByteArray): JsonStreamReader = JsonLexer(payload)
```
kggilmer, line 34: :boom:

kggilmer (approved):
Impressive work!  I don't have any ship-blocking concerns.  Some nits and questions.

runtime/utils/common/test/aws/smithy/kotlin/runtime/util/text/Utf8Test.kt
```diff
@@ -0,0 +1,45 @@
...
+import kotlin.test.*
+
+class Utf8Test {
+
+    @Test
+    fun testUtf8ByteCount() {
+        assertEquals(1, byteCountUtf8("$".encodeToByteArray()[0]))
+        assertEquals(2, byteCountUtf8("¢".encodeToByteArray()[0]))
+        assertEquals(3, byteCountUtf8("€".encodeToByteArray()[0]))
+        assertEquals(4, byteCountUtf8("\uD834\uDD22".encodeToByteArray()[0]))
```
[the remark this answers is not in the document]
aajtodd: I don't believe so because it's a surrogate pair (you can try in the editor to copy and paste and it will encode it for you).

runtime/serde/build.gradle.kts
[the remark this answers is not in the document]
aajtodd: It was added for `CharStream` tests (since removed). I think I'll keep it since it allows tests to be added to serde common if needed.

runtime/utils/common/test/aws/smithy/kotlin/runtime/util/text/Utf8Test.kt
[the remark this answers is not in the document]
ianbotsf: I was able to copy/paste the 𝄢 symbol directly into source code in the editor and it seemed to render fine as a single character. Maybe that behavior is platform/font dependent.

It's a minor question so I can live with it as-is, it merely seemed conspicuous to render three Unicode characters and then fallback to Unicode escape sequences for the last test case.

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonEncoder.kt
```diff
@@ -0,0 +1,157 @@
...
+    override fun endArray() = closeStructure("]")
+
+    private fun openStructure(token: String, nextState: LexerState) {
+        encodeValue(token)
+        if (pretty) writeNewline()
+        depth++
+        state.push(nextState)
+    }
+
+    private fun closeStructure(token: String) {
```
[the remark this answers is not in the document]
aajtodd: updated

runtime/utils/common/test/aws/smithy/kotlin/runtime/util/text/Utf8Test.kt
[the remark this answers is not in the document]
aajtodd: That is likely the case. I can't paste it directly, it auto formats it to the code points

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
kiiadi: ?

kiiadi (approved):

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
kggilmer: was the closest thing I could make to `chefs-kiss`

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonLexer.kt
aajtodd: We actually don't ever enter this else clause because we only get called when we match `t`, `f,` or `n`. The error we see when trying to read:

* `trueish` -> we will consume `true` just fine and then we'll see unexpected token `i` with offset info
* `ture` -> `Unexpected JSON token at offset 1; Unexpected char 'u' expected 'r'`

PR #482, opened 2021-09-24
feat(rt)!: refactor SdkBuffer to implement immutable and mutable interfaces; use unsigned types for cursors

aajtodd:
## Issue \#

Support for https://github.com/awslabs/aws-sdk-kotlin/pull/320

## Description of changes

* `SdkBuffer` -> `SdkByteBuffer`
* Defined `Buffer` and `MutableBuffer` as interfaces and now have `SdkByteBuffer` implementing those. This will allow us to define other types of read/write buffers in the future if needed and prevent downstream code from being hard coded to a particular implementation
* Updated all the internal read/write positions from `Int` ->  `ULong`. This has large ramifications (as you can see) not only internally but externally. I think this is for the best (and how I would implement it any other language with unsigned types) but so many APIs internally (especially on JVM) only accept `Int`. There are lots of conversions. I think I've got them right but a second set of eyes would be good. 
    * It's important to probably note that in practice we are usually actually constrained to an underlying memory (ByteBuffer/ByteArray/etc) that can only go up to `Int.MAX_VALUE` as it's maximum capacity. Some of the conversions take advantage of this fact

## Scope
* This is a breaking change. `SdkBuffer` was renamed to `SdkByteBuffer` and several of the cursors changed from `Int` to `ULong`


runtime/io/common/src/aws/smithy/kotlin/runtime/io/Buffer.kt
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/**
+ * Reads at most [length] bytes from this buffer or `-1` if no bytes are available for read.
+ * @return the number of bytes read or -1 if the buffer is empty
+ */
+fun Buffer.readAvailable(dst: SdkByteBuffer, length: Long = dst.writeRemaining.toLong()): Long {
+    if (this is SdkByteBuffer) return this.readAvailable(dst, length)
+    TODO("readAvailable fallback not implemented")
+}
```
ianbotsf, line 100: **Question**: Why signed types for `offset`? Why signed types for `length`?

runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteBuffer.kt
```diff
@@ -0,0 +1,452 @@
...
+     */
+    override val writeRemaining: ULong
+        get() = capacity - writePosition
+
+    /**
+     * Reserve capacity for at least [count] bytes to be written.
+     *
+     * More than [count] bytes may be allocated in order to avoid frequent re-allocations.
+     */
+    fun reserve(count: Long) {
```
ianbotsf, line 96: **Question**: Why signed type for `count`?

ianbotsf (approved):
Generally looks good but I still see a lot of signed types where it seems negative values aren't what we want. Called out a couple examples below...

runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteBuffer.kt
aajtodd: There's not good choice here actually. As I mentioned in the other comment the real constraints for the underlying memory are usually `Int`. Both `Long` and `ULong` could be unsatisfiable when exceeding `Int.MAX_VALUE`.

I don't remember the details on why I changed this to `Long` now that I'm looking at it. I'm open to ideas but want to hear how we want to handle unsatisfiable requests. My inclination is almost to return this to `Int`.

kggilmer (approved):
Nice work.  It's unfortunate that unsigned numeric types add so much noise.

runtime/io/common/src/aws/smithy/kotlin/runtime/io/Buffer.kt
aajtodd: @ianbotsf Any opinions here?

runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteBuffer.kt
aajtodd: @ianbotsf Same, any opinion or final input?

runtime/io/common/src/aws/smithy/kotlin/runtime/io/Buffer.kt
ianbotsf: We've gone to the trouble of wrapping the underlying JVM implementation in our own interface so I think that frees us a little from the JVM type's constraints. My recommendation would be to desig the best possible interface with the clearest semantics and then make the translations happen in the JVM wrapper.

For me, `-1` isn't a clear indication that no bytes are available. `null` is better and a self-documenting union type would be even better.

I don't feel strongly about this and if it's truly unworkable to deviate so much from the JVM-specific implementation, then I suppose we'll have to live with it.

Issue #224, opened 2021-03-04
Implement Request Retry Support


kggilmer:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue, please leave a comment

       
## Describe the Feature

Add support for handling automatic request retries (look at resulting errors to see if retryable, retry w/backoff with some # max retries). Also has to consider if this is a streaming request and can actually be retried.

## Additional Context
See also:
* /seps/accepted/shared/new-retries/
* /seps/accepted/shared/retry-after-header/
* /seps/accepted/shared/retry-header/

*internal id*: 173407402

PR #487, opened 2021-09-30
feat: implement basic retry support in runtime

ianbotsf:
## Issue \#

Addresses #224.

## Description of changes

This change adds support for retries in the runtime (but does not yet integrate them into codegen—separate PR coming). See the included design doc (**docs/design/retries.md**) for more details.

## Revision 2

Addressed the main points of PR feedback:
* `BackoffDelayer` → `DelayProvider`
* added integration tests sourced from spec
* `Throttling` → `ThrottlingError`
* retry exceptions now inherit from `ClientException`
* updated retry defaults to 20s total, 3 tries
* refactored `StandardRetryPolicy` evaluation strategies for clarity
* added handling of `CancellationException`
* cleaned up token failure handling
* improved non-retryable exception message
* moved companion object constant to top-level

Additionally, while adding the integration tests:
* added circuit-breaker mode for token bucket
* added max backoff configuration option

## Scope

- [X] New feature (non-breaking change which adds functionality)


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/Exceptions.kt
```diff
@@ -0,0 +1,67 @@
...
+
+package aws.smithy.kotlin.runtime.retries
+
+/**
+ * Indicates some failure happened while retrying.
+ * @param message A message indicating the failure mode.
+ * @param cause An underlying exception which caused this exception.
+ * @param attempts The number of attempts made before this failure was encountered.
+ * @param lastResponse The last response received from the retry strategy before this failure was encountered.
+ * @param lastException The exception caught by the retry strategy before this failure was encountered. Note that the
```
kggilmer, line 14: suggestion

if there is a way to further distinguish between `cause` and `lastException` for clarity...

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/RetryPolicy.kt
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * A policy that evaluates a [Result] from a retry attempt and indicates action a [RetryStrategy] should take next.
+ */
+interface RetryPolicy<in R> {
+    /**
+     * Evaluate the given retry attempt.
+     * @param result The [Result] of the retry attempt.
+     * @return A [RetryDirective] indicating what action the [RetryStrategy] should take next.
+     */
+    fun evaluate(result: Result<R>): RetryDirective
```
kggilmer, line 17: question

I feel that this was brought up in discussion but I forget, what's the reasoning behind which functions suspend vs which do not?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/ExponentialBackoffWithJitter.kt
```diff
@@ -0,0 +1,65 @@
...
+ * The configuration options for a [BackoffDelayer].
+ * @param initialDelayMs The initial amount of delay
+ */
+data class ExponentialBackoffWithJitterOptions(
+    val initialDelayMs: Int,
+    val scaleFactor: Double,
+    val jitter: Double,
+) {
+    init {
+        require(initialDelayMs > 0) { "initialDelayMs must be at least 0" }
```
kggilmer, line 50: comment

Do we want to be more conservative here?  Setting initial delay to 0 and scale factor to 1 seems like it something people may want to do (for reasons) and may be a bad trip server side.  I wonder if other SDKs constrain the lower bounds more...

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/ExponentialBackoffWithJitter.kt
```diff
@@ -0,0 +1,65 @@
...
+        require(scaleFactor >= 1.0) { "scaleFactor must be at least 1" }
+        require(jitter in 0.0..1.0) { "jitter must be between 0 and 1" }
+    }
+
+    companion object {
+        /**
+         * The default backoff configuration to use.
+         */
+        val Default = ExponentialBackoffWithJitterOptions(
+            initialDelayMs = 10, // start with 10ms
```
kggilmer, line 60: comment

Thank you for the clarifying comment, very helpful

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicy.kt
```diff
@@ -0,0 +1,56 @@
...
+            isRetryable && errorType == ErrorType.Server -> RetryDirective.RetryError(RetryErrorType.ServerSide)
+            isRetryable && errorType == ErrorType.Client -> RetryDirective.RetryError(RetryErrorType.ClientSide)
+            else -> null
+        }
+    }
+
+    protected open fun evaluateOtherExceptions(ex: Throwable): RetryDirective? = null
+
+    private fun evaluateNonSdkException(ex: Throwable): RetryDirective? =
+        // TODO Write logic to find connection errors, timeouts, stream faults, etc.
```
kggilmer, line 54: question

is this follow up work or blocked on something else?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryStrategy.kt
```diff
@@ -0,0 +1,160 @@
...
+                is RetryDirective.TerminateAndFail ->
+                    throwFailure(fromToken, attempt, callResult)
+
+                is RetryDirective.RetryError ->
+                    if (attempt >= options.maxAttempts) {
+                        throwTooManyAttempts(fromToken, attempt, callResult)
+                    } else {
+                        // Prep for another loop
+                        backoffDelayer.backoff(attempt)
+                        fromToken.scheduleRetry(evaluation.reason)
```
kggilmer, line 73: question

So this is the only condition that would result in a recursive call?

kggilmer (commented):
PR review in progress...

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicy.kt
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Evaluate the given retry attempt.
+     */
+    override fun evaluate(result: Result<Any?>): RetryDirective = when {
+        result.isSuccess -> RetryDirective.TerminateAndSucceed
+        else -> evaluate(result.exceptionOrNull()!!)
+    }
+
+    private fun evaluate(ex: Throwable): RetryDirective =
+        (ex as? SdkBaseException)?.run(::evaluateBaseException)
```
aajtodd, line 29: comment

I find this to be a mess to read and debug. Isn't there a bit cleaner (if not more verbose) way of expressing this?

I believe you had this as a `when` expression at one point in time that IMO is much easier to read and set breakpoints on.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicy.kt
```diff
@@ -0,0 +1,56 @@
...
+import aws.smithy.kotlin.runtime.ServiceException
+import aws.smithy.kotlin.runtime.ServiceException.*
+import aws.smithy.kotlin.runtime.retries.RetryDirective
+import aws.smithy.kotlin.runtime.retries.RetryErrorType
+import aws.smithy.kotlin.runtime.retries.RetryPolicy
+
+/**
+ * A standard retry policy which attempts to derive information from the Smithy exception hierarchy.
+ */
+open class StandardRetryPolicy : RetryPolicy<Any?> {
```
aajtodd, line 19: question/comment 
I'm not sure this should be in `impl`. It seems like something that is part of the public interface that someone may want to re-use or specify themselves?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryStrategy.kt
```diff
@@ -0,0 +1,160 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package aws.smithy.kotlin.runtime.retries.impl
```
aajtodd, line 6: comment/question

I think this probably shouldn't be in `impl` but just part of the top level package/interface

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryStrategy.kt
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * Implements a retry strategy utilizing backoff delayer and a token bucket for rate limiting and circuit breaking. Note
+ * that the backoff delayer and token bucket work independently of each other. Either can delay retries (and the token
+ * bucket can delay the initial try). The delayer is called first so that the token bucket can refill as appropriate.
+ * @param options The options that control the functionality of this strategy.
+ * @param tokenBucket The token bucket instance. Utilizing an existing token bucket will share call capacity between
+ * those scopes.
+ * @param backoffDelayer A delayer that can back off after the initial try to spread out the retries.
+ */
+class StandardRetryStrategy(
```
aajtodd, line 21: suggestion

Rather than `StandardRetryStrategy` why not just name it `TokenBucketRetryStrategy`?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/Exceptions.kt
```diff
@@ -0,0 +1,67 @@
...
+ * [Throwable] in this parameter isn't necessarily the cause of this exception.
+ */
+sealed class RetryException(
+    message: String,
+    cause: Throwable?,
+    val attempts: Int,
+    val lastResponse: Any?,
+    val lastException: Throwable?,
+) : Exception(message, cause)
```
aajtodd, line 23: should this inherit from the SDK exception hierarchy? Probably `ClientException` since it's a client side error?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryStrategy.kt
```diff
@@ -0,0 +1,160 @@
...
+     */
+    private tailrec suspend fun <R> doTryLoop(
+        block: suspend () -> R,
+        policy: RetryPolicy<R>,
+        attempt: Int,
+        fromToken: RetryToken,
+        previousResult: Result<R>?,
+    ): R {
+        val callResult = runCatching { block() }
+        if (callResult.exceptionOrNull() is TimeoutCancellationException) {
```
aajtodd, line 55: comment

We also probably want to check for cancellation and re-throw the same exception unmodified.

```kotlin
if (callResult.exceptionOrNull() is CancellationException) { // re-throw }
```

Unless we are sure that we will end up re-throwing some exception when encountered.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryStrategy.kt
```diff
@@ -0,0 +1,160 @@
...
+                is RetryDirective.RetryError ->
+                    if (attempt >= options.maxAttempts) {
+                        throwTooManyAttempts(fromToken, attempt, callResult)
+                    } else {
+                        // Prep for another loop
+                        backoffDelayer.backoff(attempt)
+                        fromToken.scheduleRetry(evaluation.reason)
+                    }
+            }
+        } catch (e: TimeoutCancellationException) {
```
aajtodd, line 76: fix

We probably need to catch any exception that might happen from the backoff delayer / token retry such that we always call `token.notifyFailure()`. For instance if the coroutine `Job` is cancelled you won't get a `TimeoutCancellationException`, you would just get `CancellationException` and we would fail to call `notifyFailure()`

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/Exceptions.kt
ianbotsf: It's a difficult nuance to convey. `cause` is pretty much the standard exception re-wrapping mechanism that exists throughout Java/Kotlin. `lastException` is more analogous to `lastResponse`...it's the last _output_ we received from the block but it's not necessarily the underlying cause of this exception.

Take for example a retry loop that keeps getting a throttling exception (which is retryable). Eventually, we're going to give up and throw a `TimedOutException` or `TooManyAttemptsException`. In those cases, the most recent retryable exception we got from our code block _isn't the cause_ of the timeout or whatever, it's merely the last thing we saw and may be interesting to code that catches the timeout.

Can you think of a succinct way to capture that in KDoc?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryStrategy.kt
```diff
@@ -0,0 +1,160 @@
...
+ * Defines configuration for a [StandardRetryStrategy].
+ * @param maxTimeMs The maximum amount of time to retry (in milliseconds).
+ * @param maxAttempts The maximum number of attempts to make (including the first attempt).
+ */
+data class StandardRetryStrategyOptions(val maxTimeMs: Long, val maxAttempts: Int) {
+    companion object {
+        /**
+         * The default retry strategy configuration.
+         */
+        val Default = StandardRetryStrategyOptions(maxTimeMs = 10_000, maxAttempts = 10)
```
aajtodd, line 158: question

Where do these values come from? The SEP indicates 3 as default for max attempts.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/RetryPolicy.kt
ianbotsf: For retries, I put `suspend` around functions that would either need to execute the code block (which probably makes IO calls) or delay (e.g., backoff, token bucket interactions, etc.). All the retry policies that I've imagined so far all happen in straightforward, on-box logic. While I suppose it's technically possible some retry policy could involve network calls or local storage calls, it seems unlikely that we'll ever need that for the bare smithy or SDK use cases.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryTokenBucket.kt
```diff
@@ -0,0 +1,141 @@
...
+    private val clock: Clock = Clock.System,
+) : RetryTokenBucket {
+    internal var capacity = options.maxCapacity
+        private set
+
+    private var lastTimestamp = now()
+    private val mutex = Mutex()
+
+    companion object {
+        private const val MS_PER_S = 1_000
```
aajtodd, line 35: style

Why define this in a `companion object`. If it's a `const val` just define it at the top of the file.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryTokenBucket.kt
```diff
@@ -0,0 +1,141 @@
...
+ * @param maxCapacity The maximum capacity for the bucket.
+ * @param refillUnitsPerSecond The amount of capacity to return per second.
+ * @param retryCost The amount of capacity to decrement for standard retries.
+ * @param timeoutRetryCost The amount of capacity to decrement for timeout or throttling retries.
+ * @param initialTryCost The amount of capacity to decrement for the initial try.
+ * @param initialTrySuccessIncrement The amount of capacity to return if the initial try is successful.
+ */
+data class StandardRetryTokenBucketOptions(
+    val maxCapacity: Int,
+    val refillUnitsPerSecond: Int,
```
aajtodd, line 122: suggestion

`refillRate`

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries.impl/ExponentialBackoffWithJitterTest.kt
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package aws.smithy.kotlin.runtime.retries.impl
```
aajtodd, line 6: fix

The file location of this looks to be `runtime/retries.impl` rather than `runtime/retries/impl`

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries.impl/StandardRetryPolicyTest.kt
```diff
@@ -0,0 +1,60 @@
...
+        ex.sdkErrorMetadata.attributes[ErrorMetadata.Retryable] = true
+        ex.sdkErrorMetadata.attributes[ServiceErrorMetadata.ErrorType] = ServiceException.ErrorType.Client
+        val result = test(ex)
+        assertEquals(RetryDirective.RetryError(RetryErrorType.ClientSide), result)
+    }
+
+    @Test
+    fun testClientException() {
+        val result = test(ClientException())
+        assertEquals(RetryDirective.RetryError(RetryErrorType.ClientSide), result)
```
aajtodd, line 52: question

should this attempt to retry any `ClientException` or should this terminate and fail? I would assume most client side (if not all) are not actually retryable.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryTokenBucket.kt
```diff
@@ -0,0 +1,141 @@
...
+import kotlinx.coroutines.sync.withLock
+import kotlin.math.ceil
+import kotlin.math.floor
+import kotlin.math.min
+
+/**
+ * The standard implementation of a [RetryTokenBucket].
+ * @param options The configuration to use for this bucket.
+ */
+class StandardRetryTokenBucket(
```
aajtodd, line 24: question

What makes this `Standard`?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/ExponentialBackoffWithJitter.kt
ianbotsf: My default posture is to document every public type and member, although I concede some members' very name and placement is the all documentation many will need. Do you think there's better documentation to be written here?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryTokenBucket.kt
aajtodd: question

This only supports `standard` not `adaptive` behavior correct?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicy.kt
ianbotsf: I'd hoped this very PR would spawn a discussion about how to fill this in. Do we have a sane way of detecting unmodeled exceptions for things like IO?

I'd imagine the `HttpClientEngine` implementations are what will be throwing most of these but the only engine-level exception I can find is [`HttpClientEngineClosedException`](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/engine/HttpClientEngine.kt#L63). Absent a common exception layer, it seems like the retry policy would have to know the individual exceptions thrown by implementations (which sounds too tightly coupled).

Do we need a more clearly defined set of exceptions thrown by engines? Are there non-engine sources of useful exceptions?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryStrategy.kt
ianbotsf: Affirmative.

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries.impl/StandardRetryTokenBucketTest.kt
```diff
@@ -0,0 +1,99 @@
...
+
+import aws.smithy.kotlin.runtime.retries.RetryErrorType
+import aws.smithy.kotlin.runtime.time.ManualClock
+import kotlinx.coroutines.test.runBlockingTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.time.Duration
+import kotlin.time.ExperimentalTime
+
+class StandardRetryTokenBucketTest {
```
aajtodd, line 16: question

Can we get the SEP standard mode test cases in here?

aajtodd (changes requested):
Fantastic. Some overall minor comments/suggestions/questions.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicy.kt
aajtodd: It's true that we probably need to have more taxonomy for particular exceptions but I'd start with at least classifying transient errors off raw response codes (we can leave this to `aws-sdk-kotlin` if we like since the SEP for this behavior is AWS specific)

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/BackoffDelayer.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package aws.smithy.kotlin.runtime.retries
+
+/**
+ * An object that can be used to delay between iterations of code.
+ */
+interface BackoffDelayer {
```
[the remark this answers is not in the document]
ianbotsf: > style
> 
> `fun interface`?

Unfortunately, `suspend` functions are not allowed in SAM interfaces.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/RetryStrategy.kt
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package aws.smithy.kotlin.runtime.retries
```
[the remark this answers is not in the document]
ianbotsf: I thought about finding a more general home for these but eventually I'd created 11 different runtime/retry-related files. Seemed like that was past the threshold where a shared namespace was cool.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/BackoffDelayer.kt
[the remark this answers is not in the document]
aajtodd: That was true at one point but I thought it was lifted? It seems to compile fine locally

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/ExponentialBackoffWithJitter.kt
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package aws.smithy.kotlin.runtime.retries.impl
```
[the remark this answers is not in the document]
ianbotsf: No, it's not intended to be an internal package. If we eventually provide customizability of the default retry configuration, callers will need to use/instantiate these classes.

The division I wanted here was between the _specification_ of how retries work (`retries`) and the _**smithy-kotlin** default implementation_ (`retries.impl`). I originally grouped them all in the same namespace but eventually I had 4 different implementation files and that seemed like a good time to separate them.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicy.kt
ianbotsf: `when` would be preferable if it allowed fall-through. The key here is that we want to try each strategy until we find something that works.

That said, yes there are alternate ways of writing this which may be cleaner...like an actual list of strategies to try in a loop.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/RetryStrategy.kt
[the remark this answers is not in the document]
aajtodd: Ya that's fair. Like I said not a big deal.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/ExponentialBackoffWithJitter.kt
kggilmer: No, I was just saying that these comments let me quickly understand how the parameters work.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicy.kt
ianbotsf: Yes, `StandardRetryPolicy` is part of the public interface (in the non-`interface` sense of the word) but as I alluded elsewhere, the `impl` namespace is not intended to be internal-only...it's intended to hold implementations.

Is the distinction between interface and implementation at the namespace level perhaps not useful here?

aajtodd: Now that I understand what you were shooting for it makes a _little_ more sense. 

> Is the distinction between interface and implementation at the namespace level perhaps not useful here?

Thinking about it for a second I don't know that we gain much from it. I don't know of anywhere else we do this either. I think _if_ we wanted more taxonomy it would make more sense to have subpackages according to type. e.g.

```
retries/
    strategy/
    policy/
```

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryStrategy.kt
ianbotsf: I liked the name "standard" because it implied _recommended_. Also, the retry specification calls it "standard". I don't mind renaming it to something clearer but I don't like `TokenBucketRetryStrategy` given that there's a `StandardRetryTokenBucket`...any other ideas?

aajtodd: Yeah I kind of started to understand the naming came from the SEP as I was reading. 

That being said it's only "standard" because of the values chosen for the configuration and the implementation of `TokenBucket` given as a parameter. Other than that it's an implementation of a retry strategy using a token bucket. I'm not sure it matters all that much but my preference I guess is to rename this to `TokenBucketRetryStrategy`. The `StandardRetryTokenBucket` makes more sense because it is what actually implements the "standard" logic.

Not a huge deal just my 2 cents

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryStrategy.kt
ianbotsf: They are my own attempt at establishing sensible defaults. I can update to the SEP-specified values.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryTokenBucket.kt
ianbotsf: I wanted to be explicit about the units here. `refillRate = 10` doesn't really indicate what unit of thing is being refilled nor how often.

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries.impl/StandardRetryPolicyTest.kt
ianbotsf: That's a good question and I suppose it depends on what else is a `ClientException`. We just decided to make all of the retryer exceptions (e.g., `TimedOutException`) inherit from `ClientException` and those are technically retryable (in that they _could_ succeed with more tries).

Taking a look through the runtime, looks like most other instances are serialization (likely not retryable) or configuration related (possibly retryable if the configuration is coming from somewhere remote).  Any other instances we can think of?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryTokenBucket.kt
ianbotsf: > What makes this `Standard`?

This is standard in that it's recommended and the only implementation provided by **smithy-kotlin**.

> This only supports `standard` not `adaptive` behavior correct?

For now, yes that's correct.

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries.impl/StandardRetryPolicyTest.kt
aajtodd: No but in general `ClientException` is basically reserved for runtime related exceptions (not to be confused with `ServiceException` where the `ErrorType` is set to `Client`)

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryTokenBucket.kt
aajtodd: yeah I mean you could do `refillRatePerSecond` or be fancy and define a type. Something like:



```kotlin
value class TokenRefillRate(val tokensPerSecond: Int)
```

Not a big deal, my brain just kept saying "theres a name for xyz measured per unit of time...RATE"

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/Exceptions.kt
kggilmer: Yeah it's a tough nut to crack.  In what cases will both `lastResponse` and `lastException` contain a value?  One minor improvement may be to use the word `previous` rather than `last`.   IMO `last` is somewhat ambiguous as it can mean just the last time something happened, which could be multiple iterations in the past, whereas `previous` to me implies something that only could have occurred in the last iteration.  As I write this it's pretty obvious how subjective this all so :shrug: .

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries.impl/TimingUtils.kt
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package aws.smithy.kotlin.runtime.retries.impl
+
+import kotlinx.coroutines.test.TestCoroutineScope
+import kotlin.test.assertTrue
+
+private const val timeToleranceMs = 20
+
+suspend fun <T> TestCoroutineScope.assertTime(expectedMs: Int, block: suspend () -> T): T {
```
kggilmer, line 13: <3

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryStrategy.kt
```diff
@@ -0,0 +1,160 @@
...
+     * Handles the termination of the retry loop because of a non-retryable failure by marking the [RetryToken] as
+     * failed and throwing a [RetryFailureException].
+     * @param token The [RetryToken] used in the attempt that was unsuccessful.
+     * @param attempt The number of attempts completed.
+     * @param result The [Result] that yielded the non-retryable condition.
+     */
+    private suspend fun <R> throwFailure(token: RetryToken, attempt: Int, result: Result<R>): Nothing {
+        token.notifyFailure()
+        throw RetryFailureException(
+            "The operation failed",
```
kggilmer, line 105: suggestion

under the assumption that this exception's message will be visible to customers, perhaps adding some of the kdoc content into the message would be clarifying.  eg "The operation resulted in a non-retryable failure"

kggilmer (approved):

kggilmer:
Incredible effort, very clean and well thought out.  Bravo @ianbotsf !

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/Exceptions.kt
ianbotsf: > In what cases will both `lastResponse` and `lastException` contain a value?

Under no circumstances will both fields contain a value. The last _result_ from a block execution can only be a response or an exception, not both.

kggilmer: In that case would it make sense to represent the combination as a union via a sealed class?

ianbotsf: We had a union called `Outcome` in the design review but it seemed like popular opinion was against it. I can create a union here but it will conflict conceptually and semantically with [`Result`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/) (which is not a union).

Do you have any suggestion on names for the union and the members?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/BackoffDelayer.kt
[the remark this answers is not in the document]
ianbotsf: It appears you're right. My IDE lied to me.

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/retries/impl/StandardRetryIntegrationTestResources.kt
```diff
@@ -0,0 +1,188 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package aws.smithy.kotlin.runtime.retries.impl
+
+val standardRetryIntegrationTestCases = mapOf(
```
aajtodd, line 8: nice

aajtodd (approved):
Other than previous suggestions overall looks good. Nothing blocking from my end.

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/retries/impl/StandardRetryIntegrationTest.kt
```diff
@@ -0,0 +1,105 @@
...
+
+            val finalState = tc.responses.last().expected
+            when (finalState.outcome) {
+                Outcome.Success -> assertEquals(200, result.getOrNull(), "Unexpected outcome for $name")
+                Outcome.MaxAttemptsExceeded -> assertIs<TooManyAttemptsException>(result.exceptionOrNull())
+                Outcome.RetryQuotaExceeded -> assertIs<TooManyAttemptsException>(result.exceptionOrNull())
+                else -> fail("Unexpected outcome for $name: ${finalState.outcome}")
+            }
+
+            if (finalState.outcome != Outcome.RetryQuotaExceeded) {
```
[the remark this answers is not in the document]
ianbotsf: Good idea, I can add that.

PR #489, opened 2021-10-05
refactor(rt): relocate and extend test utils

aajtodd:
## Issue \#

Support for https://github.com/awslabs/aws-sdk-kotlin/pull/330

## Description of changes

* New `http-test` module
    * Adds a `TestConnection` type that implements client engine that can be used to assert multiple requests/responses
    * Relocates `TestWithLocalServer` from `aws-sdk-kotlin`
* Relocates several test classes from `aws-sdk-kotlin` into `runtime:testing`
* Defines new interfaces over `Platform` to allow easier testing/mocking. 


runtime/protocol/http-test/common/src/aws/smithy/kotlin/runtime/httptest/TestConnection.kt
```diff
@@ -0,0 +1,139 @@
...
+ * ```kotlin
+ * val testEngine = buildTestConnection {
+ *     expect {
+ *         request {
+ *             url.host = "myhost"
+ *             headers.append("x-foo", "bar")
+ *         }
+ *         response = HttpResponse(HttpStatusCode.OK, Headers.Empty, HttpBody.Empty)
+ *     }
+ * }
```
ianbotsf, line 133: **Comment**: Nice, I like the clarity and expressiveness of the DSL here.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/client/SdkClientOption.kt
```diff
@@ -32,6 +33,11 @@ public object SdkClientOption {
      * The client logging mode (see [SdkLogMode]
      */
     public val LogMode: ClientOption<SdkLogMode> = ClientOption("LogMode")
+
+    /**
+     * The source of time for an operation
+     */
+    public val Clock: ClientOption<Clock> = ClientOption("Clock")
```
ianbotsf, line 40: **Question**: While I see the benefit of a specifiable `Clock` to facilitate testing, I don't understand the use case for a `ClientOption` for it. What do you see this providing?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/time/Clock.kt
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package aws.smithy.kotlin.runtime.time
+
+/**
+ * A source of time
+ */
+public interface Clock {
```
ianbotsf, line 11: **Comment**: My brazen re-appropriation of this class is [now pushed](https://github.com/awslabs/smithy-kotlin/commit/76f754d0950c5581a675d7804254acb976e1899a#diff-712258690a49644e675a8ef45306a56880884706c5bec294841d0e9a0e17db20). Feel free to rebase/merge or whatever.

runtime/testing/jvm/src/aws/smithy/kotlin/runtime/testing/ManualDispatchTestBase.kt
```diff
@@ -0,0 +1,79 @@
...
+import kotlinx.coroutines.test.runBlockingTest
+import kotlin.test.assertEquals
+import kotlin.test.assertNotEquals
+
+// modeled after https://github.com/ktorio/ktor/blob/78e36790cdbb30313dfbd23b174bffe805d26dca/ktor-io/common/test/io/ktor/utils/io/ByteChannelTestBase.kt
+// but implemented using [kotlinx-coroutines-test](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-test)
+// rather than rolling our own dummy coroutines dispatcher
+/**
+ * Test suspend functions with precise control over coordination. This allows testing that suspension happens at the
+ * expected point in time using the provided [expect] function, e.g.:
```
ianbotsf, line 19: **Question**: Why a language-level comment and a KDoc? Why not just include the provenance in the KDoc itself?

runtime/testing/jvm/src/aws/smithy/kotlin/runtime/testing/ManualDispatchTestBase.kt
```diff
@@ -0,0 +1,79 @@
...
+     * will not be executed immediately and instead be scheduled for dispatch. Explicit calls to `yield()`
+     * will advance the dispatcher.
+     */
+    protected fun runTest(block: suspend TestCoroutineScope.() -> Unit) {
+        runBlockingTest {
+            // ensure launch/async calls are coordinated with yield() points
+            pauseDispatcher()
+            block()
+        }
+    }
```
ianbotsf, line 59: **Nit**: You can get rid of the outer braces:

```kotlin
protected fun runTest(block: suspend TestCoroutineScope.() -> Unit) = runBlockingTest {
    // ensure launch/async calls are coordinated with yield() points
    pauseDispatcher()
    block()
}
```

runtime/testing/jvm/src/aws/smithy/kotlin/runtime/testing/RandomInputStream.kt
```diff
@@ -0,0 +1,72 @@
...
+ * Test utility InputStream implementation that generates random ASCII data when
+ * read, up to the size specified when constructed.
+ */
+public class RandomInputStream constructor(
+    /** The requested amount of data contained in this random stream.  */
+    private val lengthInBytes: Long,
+
+    /** Flag controlling whether binary or character data is used.  */
+    private val binaryData: Boolean = false
+) : InputStream() {
```
ianbotsf, line 21: **Question**: Is the `constructor` keyword necessary?

ianbotsf, line 21: **Nit**: I don't think the KDoc directly on constructor parameters works (at least I couldn't get anything to show up in my IDE). I think those docs should be specified in `@param` tags on the class-level KDoc:

```kotlin
/**
 * Test utility InputStream implementation that generates random ASCII data when
 * read, up to the size specified when constructed.
 * @param lengthInBytes The requested amount of data contained in this random stream.
 * @param binaryData Flag controlling whether binary or character data is used.
 */
```

runtime/testing/jvm/src/aws/smithy/kotlin/runtime/testing/RandomTempFile.kt
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @throws IOException
+     * If any problems were encountered creating the new temp file.
+     */
+    public constructor(filename: String, sizeInBytes: Long, binaryData: Boolean = false) : super(
+        TEMP_DIR + separator + System.currentTimeMillis().toString() + "-" + filename
+    ) {
+        this.binaryData = binaryData
+        createFile(sizeInBytes)
+    }
```
ianbotsf, line 51: **Nit**: These could be a single constructor with default values for `filename` and `binaryData`.

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
```diff
@@ -0,0 +1,40 @@
...
+        fun fromMap(data: Map<String, ByteArray>, filePathSeparator: String = "/"): Filesystem = MapFilesystem(data, filePathSeparator)
+    }
+}
+
+internal class MapFilesystem(
+    private val data: Map<String, ByteArray>,
+    override val filePathSeparator: String
+) : Filesystem {
+    override suspend fun readFileOrNull(path: String): ByteArray? = data[path]
+}
```
ianbotsf, line 40: **Question**: It seems like these are test conveniences. Should be they in the **test** source instead?

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/Platform.kt
```diff
@@ -5,45 +5,24 @@
...
-    /**
-     * Get an environment variable or null
-     *
-     * @param key name of environment variable
-     * @return value of environment variable or null if undefined
-     */
-    fun getenv(key: String): String?
-
+public interface PlatformEnvironProvider : EnvironmentProvider, PropertyProvider
+public interface PlatformProvider : PlatformEnvironProvider, Filesystem {
```
ianbotsf, line 9: **Question**: The `PlatformEnvironProvider` interface seems unnecessary. Why not just add `EnvironmentProvider` and `PropertyProvider` directly to `PlatformProvider`?

ianbotsf (approved):
Looks good overall.

runtime/protocol/http-test/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+description = "Utilities for testing HTTP requests"
```
kggilmer, line 5: note

at some point maybe we want to add some kind of post build test that validates that none of our test modules accidentally made it into our rt or service artifacts, unless there is already a mechanism in place that guards this?

runtime/protocol/http-test/common/src/aws/smithy/kotlin/runtime/httptest/TestConnection.kt
```diff
@@ -0,0 +1,139 @@
...
+import aws.smithy.kotlin.runtime.time.Instant
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+/**
+ * An expected HttpRequest with the response that should be returned by the engine
+ * @param request the expected request. If null no assertions are made on the request
+ * @param response the response to return for this request. If null it defaults to an empty 200-OK response
+ */
+data class ExpectedHttpRequest(val request: HttpRequest?, val response: HttpResponse? = null)
```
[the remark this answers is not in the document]
kggilmer: Yeah `roundTrip` in the parlance of our (ktor?) times.

runtime/protocol/http-test/common/src/aws/smithy/kotlin/runtime/httptest/TestConnection.kt
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Assert that all of the components set on [expected] are also the same on [actual]. The actual request
+     * may have additional headers, only the ones set in [expected] are compared.
+     */
+    internal suspend fun assertRequest(idx: Int) {
+        if (expected == null) return
+        assertEquals(expected.url.toString(), actual.url.toString(), "[request#$idx]: URL mismatch")
+        expected.headers.forEach { name, values ->
+            values.forEach {
+                assertTrue(actual.headers.contains(name, it), "[request#$idx]: header $name missing value $it")
```
kggilmer, line 42: nit

enclose key/values in single quotes to make debugging w/ empty values easier.

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
```diff
@@ -0,0 +1,40 @@
...
+     * @param path fully qualified path encoded specifically to the target platform's filesystem.
+     * @return contents of file or null if error (file does not exist, etc.)
+     */
+    public suspend fun readFileOrNull(path: String): ByteArray?
+
+    companion object {
+        /**
+         * Construct a fake filesystem from a mapping of paths to contents
+         */
+        fun fromMap(data: Map<String, ByteArray>, filePathSeparator: String = "/"): Filesystem = MapFilesystem(data, filePathSeparator)
```
kggilmer, line 31: nice

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+internal class MapFilesystem(
```
kggilmer, line 35: comment

I agree after seeing this that decomposing `Platform` into smaller pieces yields more manageable code.

kggilmer (approved):

runtime/testing/jvm/src/aws/smithy/kotlin/runtime/testing/ManualDispatchTestBase.kt
aajtodd: Because it's unimportant to how it's used but it is of interest to someone that needs to maintain it. Thats the line I draw at least

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
aajtodd: If they are in the *test* source then you can't depend on them. Test sources aren't part of a published artifact. 

They are indeed test conveniences. We could put them in `runtime:testing` I suppose if you prefer.

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/Platform.kt
aajtodd: Glad you asked (figured/hoped someone would).

That is how I had it originally and I back tracked a bit simply because smaller interfaces mean I can express what I depend on more precisely. An example is `AwsSdkSetting` which only needs `PlatformEnvironProvider` not all of the capabilities of `PlatformProvider`. 

That being said it's a minor thing and if you can make a case for just a single `PlatformProvider` interface that must implement all 3 then I'm happy to change it. If not and we want to keep it I'm open to hearing better names...not sure what to call something that provides environment variables and properties

runtime/protocol/http-test/build.gradle.kts
aajtodd: If it's not a runtime dependency of them then we are good. I'm not sure of a mechanism for it but the only time you would need it is testing so you would have to accidentally add it to say `commonMain` instead of `commonTest` source set.

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/Platform.kt
ianbotsf: Yeah, this is a time where [intersection types](https://en.wikipedia.org/wiki/Intersection_type) would be really handy in the language.

I see what you're going for with minimal dependency scoping. I guess now, seeing how it's used in the companion PR, that I'd suggest calling this a `ConfigMapProvider` or a `PlatformSettingProvider` or something like that. Something that doesn't give undue weight to only one of the sub-interfaces.

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
aajtodd: I know why I didn't move them there. `testing` would need to take an `api` dependency on `utils`. which would mean we couldn't use `testing` for testing the `utils` package at some point in the future (we don't currently depend on it it seems).

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/client/SdkClientOption.kt
aajtodd: We can remove it for now probably, was just thinking it would be convenient to be able to specify the source of time for an operation. Until we have further need for this though I think we can just isolate it to where it is used for now.

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/Platform.kt
aajtodd: `EnvironmentPropertyFactorySettingPlatformProvider`?

PR #490, opened 2021-10-05
feat: add codegen wrappers for retries

### LINKED ISSUE #224: Implement Request Retry Support

ianbotsf:
## Issue \#

Closes #224 

## Description of changes

Implements the codegen support and wrappers for retrying all operations. Runtime components were implemented in #487.

This PR includes the first example of **replaceable integrations** which may be controversial. I chose that to allow a default strategy/policy in **smithy-kotlin** but allow overriding for specific codegen use cases (like **aws-sdk-kotlin**). Comments on this approach are welcome.

## Scope

- [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
@@ -47,10 +47,18 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Unit>() {
     init {
         val classLoader = context.pluginClassLoader.orElse(javaClass.classLoader)
         LOGGER.info("Discovering KotlinIntegration providers...")
-        integrations = ServiceLoader.load(KotlinIntegration::class.java, classLoader)
-            .also { integration -> LOGGER.info("Loaded KotlinIntegration: ${integration.javaClass.name}") }
+
+        val integrationsBeforeReplacement = ServiceLoader.load(KotlinIntegration::class.java, classLoader)
+            .onEach { integration -> LOGGER.info("Loaded KotlinIntegration: ${integration.javaClass.name}") }
             .filter { integration -> integration.enabledForService(context.model, settings) }
-            .also { integration -> LOGGER.info("Enabled KotlinIntegration: ${integration.javaClass.name}") }
+            .onEach { integration -> LOGGER.info("Enabled KotlinIntegration: ${integration.javaClass.name}") }
+
+        val replaceIntegrations = integrationsBeforeReplacement
+            .flatMap { it.replacesIntegrations }
+            .onEach { integration -> LOGGER.info("Replacing KotlinIntegration: ${integration.javaClass.name}") }
+
+        integrations = integrationsBeforeReplacement
+            .filter { integration -> integration.javaClass !in replaceIntegrations }
             .sortedBy(KotlinIntegration::order)
             .toList()
 
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/RuntimeTypes.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/RuntimeTypes.kt
@@ -48,6 +48,10 @@ object RuntimeTypes {
             val HttpResponse = runtimeSymbol("HttpResponse", KotlinDependency.HTTP, "response")
         }
 
+        object Retries {
+            val StandardRetryFeature = runtimeSymbol("StandardRetryFeature", KotlinDependency.HTTP, "retries")
+        }
+
         object Operation {
             val HttpDeserialize = runtimeSymbol("HttpDeserialize", KotlinDependency.HTTP, "operation")
             val HttpSerialize = runtimeSymbol("HttpSerialize", KotlinDependency.HTTP, "operation")
@@ -77,6 +81,18 @@ object RuntimeTypes {
             val StringContent = runtimeSymbol("StringContent", KotlinDependency.CORE, "content")
             val toByteArray = runtimeSymbol("toByteArray", KotlinDependency.CORE, "content")
             val decodeToString = runtimeSymbol("decodeToString", KotlinDependency.CORE, "content")
+        }
+
+        object Retries {
+            object Impl {
+                val ExponentialBackoffWithJitter = runtimeSymbol("ExponentialBackoffWithJitter", KotlinDependency.CORE, "retries.impl")
+                val ExponentialBackoffWithJitterOptions = runtimeSymbol("ExponentialBackoffWithJitterOptions", KotlinDependency.CORE, "retries.impl")
+                val StandardRetryPolicy = runtimeSymbol("StandardRetryPolicy", KotlinDependency.CORE, "retries.impl")
+                val StandardRetryStrategy = runtimeSymbol("StandardRetryStrategy", KotlinDependency.CORE, "retries.impl")
+                val StandardRetryStrategyOptions = runtimeSymbol("StandardRetryStrategyOptions", KotlinDependency.CORE, "retries.impl")
+                val StandardRetryTokenBucket = runtimeSymbol("StandardRetryTokenBucket", KotlinDependency.CORE, "retries.impl")
+                val StandardRetryTokenBucketOptions = runtimeSymbol("StandardRetryTokenBucketOptions", KotlinDependency.CORE, "retries.impl")
+            }
         }
     }
 
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/retries/StandardRetryIntegration.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/retries/StandardRetryIntegration.kt
@@ -1,0 +1,98 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package software.amazon.smithy.kotlin.codegen.retries
+
+import software.amazon.smithy.codegen.core.Symbol
+import software.amazon.smithy.codegen.core.SymbolReference.*
+import software.amazon.smithy.kotlin.codegen.core.*
+import software.amazon.smithy.kotlin.codegen.integration.KotlinIntegration
+import software.amazon.smithy.kotlin.codegen.model.buildSymbol
+import software.amazon.smithy.kotlin.codegen.model.namespace
+import software.amazon.smithy.kotlin.codegen.rendering.ClientConfigProperty
+import software.amazon.smithy.kotlin.codegen.rendering.protocol.HttpFeatureMiddleware
+import software.amazon.smithy.kotlin.codegen.rendering.protocol.ProtocolGenerator
+import software.amazon.smithy.kotlin.codegen.rendering.protocol.ProtocolMiddleware
+
+/**
+ * Adds retry wrappers around operation invocations, driven by client config (which specifies the [RetryStrategy]) and
+ * a middleware (which specifies the [RetryPolicy][aws.smithy.kotlin.runtime.retries.RetryPolicy]). This class is
+ * `open` and may be subclassed to customize the strategy and/or policy. Note that codegen which specifies a subclass of
+ * this should probably configure [replacesIntegrations] to remove the base implementation—otherwise, multiple copies of
+ * the config properties may be included in client config.
+ */
+open class StandardRetryIntegration : KotlinIntegration {
+    override fun additionalServiceConfigProps(ctx: CodegenContext): List<ClientConfigProperty> =
+        listOf(
+            ClientConfigProperty {
+                val retryStrategyBlock = """
+                    run {
+                        val strategyOptions = StandardRetryStrategyOptions.Default
+                        val tokenBucket = StandardRetryTokenBucket(StandardRetryTokenBucketOptions.Default)
+                        val delayer = ExponentialBackoffWithJitter(ExponentialBackoffWithJitterOptions.Default)
+                        StandardRetryStrategy(strategyOptions, tokenBucket, delayer)
+                    }
+                """.trimIndent()
+
+                symbol = buildSymbol {
+                    name = "RetryStrategy"
+                    namespace(KotlinDependency.CORE, "retries")
+                    nullable = false
+                    reference(RuntimeTypes.Core.Retries.Impl.StandardRetryStrategy, ContextOption.USE)
+                    reference(RuntimeTypes.Core.Retries.Impl.StandardRetryStrategyOptions, ContextOption.USE)
+                    reference(RuntimeTypes.Core.Retries.Impl.StandardRetryTokenBucket, ContextOption.USE)
+                    reference(RuntimeTypes.Core.Retries.Impl.StandardRetryTokenBucketOptions, ContextOption.USE)
+                    reference(RuntimeTypes.Core.Retries.Impl.ExponentialBackoffWithJitter, ContextOption.USE)
+                    reference(RuntimeTypes.Core.Retries.Impl.ExponentialBackoffWithJitterOptions, ContextOption.USE)
+                }
+                name = "retryStrategy"
+                documentation = """
+                    The [RetryStrategy] implementation to use for service calls. All API calls will be wrapped by the
+                    strategy.
+                """.trimIndent()
+                constantValue = retryStrategyBlock
+            }
+        )
+
+    override fun customizeMiddleware(
+        ctx: ProtocolGenerator.GenerationContext,
+        resolved: List<ProtocolMiddleware>
+    ): List<ProtocolMiddleware> = resolved.filter { it !is StandardRetryIntegration } + middleware
+
+    private val middleware = object : HttpFeatureMiddleware() {
+        override val name: String = "StandardRetryFeature"
+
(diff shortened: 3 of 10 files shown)


smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
```diff
@@ -47,10 +47,18 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Unit>() {
...
         val classLoader = context.pluginClassLoader.orElse(javaClass.classLoader)
         LOGGER.info("Discovering KotlinIntegration providers...")
-        integrations = ServiceLoader.load(KotlinIntegration::class.java, classLoader)
-            .also { integration -> LOGGER.info("Loaded KotlinIntegration: ${integration.javaClass.name}") }
+
+        val integrationsBeforeReplacement = ServiceLoader.load(KotlinIntegration::class.java, classLoader)
+            .onEach { integration -> LOGGER.info("Loaded KotlinIntegration: ${integration.javaClass.name}") }
             .filter { integration -> integration.enabledForService(context.model, settings) }
-            .also { integration -> LOGGER.info("Enabled KotlinIntegration: ${integration.javaClass.name}") }
+            .onEach { integration -> LOGGER.info("Enabled KotlinIntegration: ${integration.javaClass.name}") }
```
kggilmer, line 54: correctness

I think this log message is misleading as any particular integration may be removed in the subsequent code.  Maybe "loaded" and then "enabled" in the final stanza?  Generally I think it's noisy if we are saying anything about an unused integration except that it's unused in logs.

kggilmer (approved):
Yes regarding the introduction of this new extension capability of replacing integrations..it's not clear to me from these PRs why defining/overriding a section would not suffice.  Can you briefly outline why you choose this approach over sections?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/RuntimeTypes.kt
```diff
@@ -48,6 +48,10 @@ object RuntimeTypes {
             val HttpResponse = runtimeSymbol("HttpResponse", KotlinDependency.HTTP, "response")
         }
 
+        object Retries {
+            val StandardRetryFeature = runtimeSymbol("StandardRetryFeature", KotlinDependency.HTTP, "retries")
```
aajtodd, line 52: nit

I should have caught this before but this probably should have lived in `aws.smithy.kotlin.runtime.http.middleware` which is an existing package. Also since it takes a strategy and a policy nothing makes it "Standard". It could probably just be `RetryMiddleware`

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/retries/StandardRetryIntegration.kt
```diff
@@ -0,0 +1,98 @@
...
+                        val tokenBucket = StandardRetryTokenBucket(StandardRetryTokenBucketOptions.Default)
+                        val delayer = ExponentialBackoffWithJitter(ExponentialBackoffWithJitterOptions.Default)
+                        StandardRetryStrategy(strategyOptions, tokenBucket, delayer)
+                    }
+                """.trimIndent()
+
+                symbol = buildSymbol {
+                    name = "RetryStrategy"
+                    namespace(KotlinDependency.CORE, "retries")
+                    nullable = false
```
aajtodd, line 42: question

This implies we wouldn't be able to set the strategy to `null` someday correct? To get "no retry" behavior someone would have to explicitly pass `retryStrategy = RetryStrategy.None` or something. I think that's probably ok (and more clear) but just clarifying our stance.

aajtodd (commented):
Overall it looks good. I too am not quite convinced that we need the new "replace" capability though.

In fact we already have a [section](https://github.com/awslabs/smithy-kotlin/blob/v0.4.1-alpha/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ServiceGenerator.kt#L77) in the `ServiceGenerator` to override the config properties if needed. We have precedence for default config properties. Both the idempotency token provider and http client engine are [registered](https://github.com/awslabs/smithy-kotlin/blob/v0.4.1-alpha/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigGenerator.kt#L45) as default properties. 

We also have a way of specifying [default middleware](https://github.com/awslabs/smithy-kotlin/blob/v0.4.1-alpha/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt#L63). (Alternatively you could insert it into `getHttpMiddleware()` such that it is _always_ present and a customization would have to remove it if they really wanted it gone). 

I almost want to just say let's always render a retry strategy and a retry middleware and if you want to not have retries then pass in a "no retry" strategy.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/RuntimeTypes.kt
ianbotsf: Well you couldn't have caught this before since it's new in this PR. 😃 I can relocate it.

ianbotsf:
> I almost want to just say let's always render a retry strategy and a retry middleware and if you want to not have retries then pass in a "no retry" strategy.

OK, let me see what I can work up.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/QueryParameters.kt
```diff
@@ -33,13 +31,12 @@ private object EmptyQueryParameters : QueryParameters {
...
     override fun toString(): String = "QueryParametersBuilder ${entries()} "
-    override fun build(): QueryParameters {
-        require(!built) { "QueryParametersBuilder can only build a single instance" }
-        built = true
-        return QueryParametersImpl(values)
-    }
+    override fun build(): QueryParameters = QueryParametersImpl(values)
+
+    override fun deepCopy(): QueryParametersBuilder =
+        QueryParametersBuilder().apply { values.putAll(this@QueryParametersBuilder.values.deepCopy()) }
```
aajtodd, line 39: style

I found this hard to read/follow due to `this` both referring to the same type `QueryParametersBuilder`. I guess it's fine as is if it works but it took me staring at this for a second to verify it was correct.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/RetryFeature.kt
```diff
@@ -3,35 +3,44 @@
...
     }
 
     override fun <I, O> install(operation: SdkHttpOperation<I, O>) {
         operation.execution.finalize.intercept { req, next ->
-            strategy.retry(policy) { next.call(req) }
+            if (req.subject.isRetryable) {
+                strategy.retry(policy) {
+                    // Deep copy the request because later middlewares (e.g., signing) mutate it
+                    val reqCopy = req.deepCopy()
+                    next.call(reqCopy)
```
aajtodd, line 39: correctness

Need to rewind streaming bodies

aajtodd (commented):
I think I may have jumped the gun, will wait for the rest of the changes before continuing to review. Let me know when it's ready

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/util/StringValuesMap.kt
```diff
@@ -64,8 +64,10 @@ internal open class StringValuesMapImpl(
     override val caseInsensitiveName: Boolean = false,
     initialValues: Map<String, List<String>> = emptyMap()
 ) : StringValuesMap {
-    protected val values: Map<String, List<String>> by lazy {
-        if (caseInsensitiveName) CaseInsensitiveMap<List<String>>().apply { putAll(initialValues) } else values.toMap()
+    protected val values: Map<String, List<String>> = run {
+        // Make a defensive copy so modifications to the initialValues don't mutate our internal copy
+        val copiedValues = initialValues.deepCopy()
```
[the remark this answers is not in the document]
ianbotsf: I'm afraid it is necessary. The type of the _parameter_ may be the immutable interface type `Map` but the actual _argument_ supplied is a mutable implementation. Modifications to that map in the builder will cause already-built instances to mutate unless copied.

Moreover, the map values passed in are mutable lists. They too need to be defensively copied to prevent post hoc mutation.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/WaiterGenerator.kt
```diff
@@ -0,0 +1,133 @@
...
+import software.amazon.smithy.kotlin.codegen.core.RenderingContext
+import software.amazon.smithy.kotlin.codegen.core.defaultName
+import software.amazon.smithy.kotlin.codegen.core.withBlock
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.waiters.Matcher
+import software.amazon.smithy.waiters.PathMatcher
+import software.amazon.smithy.waiters.WaitableTrait
+import software.amazon.smithy.waiters.Waiter
+
+class WaiterGenerator(
```
ianbotsf, line 21: **Comment**: This shouldn't be here and I'll pull it out shortly...ignore it for now!

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/util/CanDeepCopy.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package aws.smithy.kotlin.runtime.http.util
+
+interface CanDeepCopy<T> {
```
aajtodd, line 8: style

`out T`

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/util/StringValuesMap.kt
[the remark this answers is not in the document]
aajtodd: Which argument where?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/retries/StandardRetryIntegration.kt
```diff
@@ -0,0 +1,79 @@
...
+                    reference(RuntimeTypes.Core.Retries.Impl.StandardRetryStrategy, ContextOption.USE)
+                    reference(RuntimeTypes.Core.Retries.Impl.StandardRetryStrategyOptions, ContextOption.USE)
+                    reference(RuntimeTypes.Core.Retries.Impl.StandardRetryTokenBucket, ContextOption.USE)
+                    reference(RuntimeTypes.Core.Retries.Impl.StandardRetryTokenBucketOptions, ContextOption.USE)
+                    reference(RuntimeTypes.Core.Retries.Impl.ExponentialBackoffWithJitter, ContextOption.USE)
+                    reference(RuntimeTypes.Core.Retries.Impl.ExponentialBackoffWithJitterOptions, ContextOption.USE)
+                }
+                name = "retryStrategy"
+                documentation = """
+                    The [RetryStrategy] implementation to use for service calls. All API calls will be wrapped by the
```
aajtodd, line 54: correctness

This won't resolve in documentation I don't think. I believe you'll need to give the full import path `[aws.smithy.kotlin.*.RetryStrategy]`. Ignore if I'm wrong.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/retries/StandardRetryIntegration.kt
```diff
@@ -0,0 +1,79 @@
...
+    override fun customizeMiddleware(
+        ctx: ProtocolGenerator.GenerationContext,
+        resolved: List<ProtocolMiddleware>,
+    ): List<ProtocolMiddleware> = resolved + middleware
+
+    private val middleware = object : HttpFeatureMiddleware() {
+        override val name: String = "RetryFeature"
+
+        override fun renderConfigure(writer: KotlinWriter) {
+            writer.declareSection(RetryConfigSection) {
```
aajtodd, line 70: this isn't quite what I had in mind..

There's two things here. The (1) config property and the (2) middleware. 

* For (1) just add the config property to the list of [default properties](https://github.com/awslabs/smithy-kotlin/blob/bd016aea94abafa48a977c1574398e04384144db/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigGenerator.kt#L34) if we are taking a stance that all SDK clients should be generated with a retry strategy (and I think we should). We have a way to override this if needed already to remove default props. No need for this to be in an integration.

* For (2) you don't need a `Section`. Since it's a middleware just wholesale [replace](https://github.com/awslabs/smithy-kotlin/blob/f715840b97230deb9c42ffbf920e8bbc5b63a822/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/ProtocolMiddleware.kt#L52) the thing in `aws-sdk-kotlin`. You can make it re-usable if you want or just duplicate it to a new `HttpFeatureMiddleware` type there. Since it's not a lot of code I'd probably just duplicate it personally.


I'm not going to die on this hill, if you think they belong as a `KotlinIntegration` I'll live with it. It just seems at odds with what we've done before.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/retries/StandardRetryIntegration.kt
```diff
@@ -0,0 +1,79 @@
...
+                name = "retryStrategy"
+                documentation = """
+                    The [RetryStrategy] implementation to use for service calls. All API calls will be wrapped by the
+                    strategy.
+                """.trimIndent()
+                constantValue = retryStrategyBlock
+            }
+        )
+
+    override fun customizeMiddleware(
```
aajtodd, line 61: correctness

Should probably only customize the middleware if it's an HTTP protocol. Another reason to maybe move this to a set of [default HTTP middleware](https://github.com/awslabs/smithy-kotlin/blob/45ff27e749f7454b232d2846b96ec2c8c40d48fa/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt#L63) instead of an integration.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HttpBody.kt
```diff
@@ -35,6 +44,14 @@ sealed class HttpBody {
          * Provides [ByteArray] to be sent to peer
          */
         abstract fun bytes(): ByteArray
+
+        override fun deepCopy(): Bytes = object : Bytes() {
```
aajtodd, line 48: comment

I don't love that we are going to deep copy a payload. I'm not convinced this type should implement `CanDeepCopy`

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/RetryFeature.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    override fun <I, O> install(operation: SdkHttpOperation<I, O>) {
+        operation.execution.finalize.intercept { req, next ->
+            if (req.subject.isRetryable) {
+                strategy.retry(policy) {
+                    // Deep copy the request because later middlewares (e.g., signing) mutate it
+                    val reqCopy = req.deepCopy()
+
+                    // Reset bodies back to beginning (mainly for streaming bodies)
+                    reqCopy.subject.body.reset()
```
aajtodd, line 42: comment

I don't know that I love how much of retry has seeped into `HttpBody`. It seems like a lot of these concerns are localized to retry middleware.

aajtodd (commented):
Couple questions on correctness. I'm also not sure about `HttpBody` changing to support `CanDeepCopy`. My inclination is to really isolate this as a retry concern and not have it supported in the type hierarchy.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/retries/StandardRetryIntegration.kt
ianbotsf: Hmm, looks like the documentation isn't used at all right now because we only write KDocs on the `DslBuilder` and we only put mutable properties in the `DslBuilder`.

If and when we do make this mutable, I believe it should work as-is because the symbol type of `RetryStrategy` gets imported (which I can verify by manually inspecting a generated client).

PR #493, opened 2021-10-08
feat(rt): http client engine config

### LINKED ISSUE #198: Define Advanced HttpClient Config
We generate service clients with the ability to specify an http client engine to use instead of the default. We currently don't take this into account in our generated close() call. 

Tasks are roughly:
* Implement some kind of SPI/loading from a factory rather than hard coding it. An example from ktor of what might work:
[common factory](https://github.com/ktorio/ktor/blob/master/ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt#L42)
[jvm spi implementation](https://github.com/ktorio/ktor/blob/master/ktor-client/ktor-client-core/jvm/src/io/ktor/client/HttpClientJvm.kt#L42)
[native implementation](https://github.com/ktorio/ktor/blob/master/ktor-client/ktor-client-core/posix/src/io/ktor/client/engine/Loader.kt#L18)

* Update the generated `close()` method to take into account whether we manage the underlying `HttpClientEngine` or not to determine if we should close it. If a user passes a custom engine in - we do not own it and shouldn't close it on the user. They are expected to manage it's lifetime and cleanup any resources associated with it.

*internal id*: 176011935

aajtodd:
## Issue \#

closes #198 

## Description of changes

* Removes `HttpClientConfig` which is a relic from a simpler time. It has not real purpose for us now.
* Fleshes out an initial set of HTTP client configuration settings that most engines can/should support.

### DIFF
--- a/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/engine/HttpClientEngineConfig.kt
+++ b/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/engine/HttpClientEngineConfig.kt
@@ -52,7 +52,7 @@ open class HttpClientEngineConfig constructor(builder: Builder) {
     /**
      * The amount of time to wait for an already-established connection from a connection pool
      */
-    var connectionAcquireTimeout: Duration = builder.connectionAcquireTimeout
+    val connectionAcquireTimeout: Duration = builder.connectionAcquireTimeout
 
     /**
      * The amount of time before an idle connection should be reaped from a connection pool. Zero indicates that


runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/engine/HttpClientEngineConfig.kt
```diff
@@ -4,7 +4,123 @@
...
+
 /**
  * Common configuration options to be interpreted by an underlying engine
+ *
+ * NOTE: Not all engines will support every option! Engines *SHOULD* log a warning when given a configuration
+ * option they don't understand/support
  */
-class HttpClientEngineConfig
+@OptIn(ExperimentalTime::class)
+open class HttpClientEngineConfig constructor(builder: Builder) {
```
ianbotsf, line 20: **Question**: Do we see the members of this class growing over time? If so, I think a layer of organization might be in order. Otherwise this may turn into a difficult class to manage by users.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/engine/HttpClientEngineConfig.kt
```diff
@@ -4,7 +4,123 @@
...
+
+    /**
+     * The amount of time to wait for a connection to be established
+     */
+    val connectTimeout: Duration = builder.connectTimeout
+
+    /**
+     * The amount of time to wait for an already-established connection from a connection pool
+     */
+    var connectionAcquireTimeout: Duration = builder.connectionAcquireTimeout
```
ianbotsf, line 55: **Question**: Why `var`? Every other field in this class is `val`.

PR #504, opened 2021-10-19
refactor!: overhaul client config property generation

### LINKED ISSUE #498: Service client DslBuilder should not expose a build() method
The service config DSL builder should not expose `build()` method. It need not be accessible from this context and would allow customers to call build at the wrong time accidentally.

Example:

```kotlin
    val client = LambdaClient {

        this.build()
    }

```

See [here](https://github.com/awslabs/smithy-kotlin/blob/bd016aea94abafa48a977c1574398e04384144db/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigGenerator.kt#L146)

Structures also do this for their `DslBuilder` implementation. This is so that we can construct a builder and use it for deserialization. We could probably find another way to make it such that `build()` isn't accessible publicly (e.g. making the `BuilderImpl` generated `internal`)

We also expose a `FluentBuilder` interface on service client config with no way of constructing a fluent builder (unlike structures which do).

aajtodd:
## Issue \#

fixes #498 

## Description of changes

* (refactor): Define a sealed class to represent different types of configuration properties and how they are rendered
* (refactor): remove `build()` from `Config.DslBuilder`. Before a customer could do `Client { build() }`. The `build()` method is still available on the `FluentBuilder` interface as well as the implementation
* (fix): add a way to create a client given a configuration. This fixes 498, before there was no way to use the fluent builder to construct a client.

### DIFF
--- a/runtime/smithy-test/common/src/aws/smithy/kotlin/runtime/smithy/test/HttpResponseTest.kt
+++ b/runtime/smithy-test/common/src/aws/smithy/kotlin/runtime/smithy/test/HttpResponseTest.kt
@@ -7,12 +7,12 @@ package aws.smithy.kotlin.runtime.smithy.test
 import aws.smithy.kotlin.runtime.http.Headers
 import aws.smithy.kotlin.runtime.http.HttpBody
 import aws.smithy.kotlin.runtime.http.HttpStatusCode
-import aws.smithy.kotlin.runtime.http.content.ByteArrayContent
 import aws.smithy.kotlin.runtime.http.engine.HttpClientEngine
 import aws.smithy.kotlin.runtime.http.engine.HttpClientEngineBase
 import aws.smithy.kotlin.runtime.http.request.HttpRequest
 import aws.smithy.kotlin.runtime.http.response.HttpCall
 import aws.smithy.kotlin.runtime.http.response.HttpResponse
+import aws.smithy.kotlin.runtime.io.SdkByteReadChannel
 import aws.smithy.kotlin.runtime.testing.runSuspendTest
 import aws.smithy.kotlin.runtime.time.Instant
 
@@ -88,7 +88,16 @@ fun <T> httpResponseTest(block: HttpResponseTestBuilder<T>.() -> Unit) = runSusp
             }
 
             val body: HttpBody = testBuilder.expected.body?.let {
-                ByteArrayContent(it.encodeToByteArray())
+                // emulate a real response stream which typically can only be consumed once
+                // see: https://github.com/awslabs/aws-sdk-kotlin/issues/356
+                object : HttpBody.Streaming() {
+                    private var consumed = false
+                    override fun readFrom(): SdkByteReadChannel {
+                        val content = if (consumed) ByteArray(0) else it.encodeToByteArray()
+                        consumed = true
+                        return SdkByteReadChannel(content)
+                    }
+                }
             } ?: HttpBody.Empty
 
             val resp = HttpResponse(testBuilder.expected.statusCode, headers, body)


smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
```diff
@@ -137,13 +133,47 @@ class ClientConfigProperty private constructor(builder: Builder) {
...
+        var propertyType: ClientConfigPropertyType = ClientConfigPropertyType.SymbolDefault
 
         fun build(): ClientConfigProperty = ClientConfigProperty(this)
     }
 }
 
+/**
+ * Descriptor for how a configuration property is rendered when the configuration is built
+ */
+sealed class ClientConfigPropertyType {
```
kggilmer, line 145: Nice!

ianbotsf:
**Nit**: The title of this commit/PR should probably start with "fix!" instead of "refactor!" since it does fix #498. This will be meaningful once we use release scripts to automatically create the changelog.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
```diff
@@ -137,13 +133,47 @@ class ClientConfigProperty private constructor(builder: Builder) {
...
+     */
+    object SymbolDefault : ClientConfigPropertyType()
+
+    /**
+     * Specifies that the value should be populated with a constant value that cannot be overridden in the builder.
+     * These are effectively read-only properties that will show up in the configuration type but not the builder.
+     *
+     * @param value the value to assign to the property at construction time
+     */
+    data class ConstantValue(val value: String) : ClientConfigPropertyType()
```
ianbotsf, line 157: **Question**: This new ADT is generally very good and I love the encapsulation it provides. It continues the pattern, however, of requiring imports necessary for the default value to be placed on the symbol itself rather than closer to the value. For example, the retry strategy default value requires six imports but the only way to bring them along is to attach them to a custom-built symbol (where otherwise a clean, global definition would be sufficient).

Is there a clean way to move these value-specific imports down into this type hierarchy?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
```diff
@@ -93,7 +93,9 @@ class StructureGenerator(
             write("@JvmStatic")
             write("fun fluentBuilder(): FluentBuilder = BuilderImpl()")
             write("")
-            write("fun builder(): DslBuilder = BuilderImpl()")
+            // the manual construction of a DslBuilder is mostly to support serde, end users should go through
+            // invoke syntax for kotlin or fluentBuilder for Java consumers
+            write("internal fun builder(): DslBuilder = BuilderImpl()")
```
ianbotsf, line 98: **Question**: Do we need to decorate this with `@InternalApi`?

aajtodd: it's generated with `internal` visibility so I don't think so.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
kggilmer: Personally I don't mind #1 but am fine w/ #2 and also do not like #3.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
ianbotsf: Oh right. Wasn't thinking.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
ianbotsf: I prefer option 2 since it more closely aligns with the usage of the default value then it does to the definition of the symbol.

PR #505, opened 2021-10-20
fix(rt): emulate a real response body more closely to help catch subtle codegen bugs

aajtodd:
## Issue \#

sibling PR: https://github.com/awslabs/aws-sdk-kotlin/pull/358

## Description of changes

Use a streaming single use `HttpBody` variant for protocol tests which will more closely reflect reality.

## Scope

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

### DIFF
--- a/runtime/smithy-test/common/src/aws/smithy/kotlin/runtime/smithy/test/HttpResponseTest.kt
+++ b/runtime/smithy-test/common/src/aws/smithy/kotlin/runtime/smithy/test/HttpResponseTest.kt
@@ -91,10 +91,10 @@ fun <T> httpResponseTest(block: HttpResponseTestBuilder<T>.() -> Unit) = runSusp
                 // emulate a real response stream which typically can only be consumed once
                 // see: https://github.com/awslabs/aws-sdk-kotlin/issues/356
                 object : HttpBody.Streaming() {
-                    private var cnt = 0
+                    private var consumed = false
                     override fun readFrom(): SdkByteReadChannel {
-                        val content = if (cnt > 0) ByteArray(0) else it.encodeToByteArray()
-                        cnt++
+                        val content = if (consumed) ByteArray(0) else it.encodeToByteArray()
+                        consumed = true
                         return SdkByteReadChannel(content)
                     }
                 }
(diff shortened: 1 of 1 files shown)


runtime/smithy-test/common/src/aws/smithy/kotlin/runtime/smithy/test/HttpResponseTest.kt
```diff
@@ -88,7 +88,16 @@ fun <T> httpResponseTest(block: HttpResponseTestBuilder<T>.() -> Unit) = runSusp
             }
 
             val body: HttpBody = testBuilder.expected.body?.let {
-                ByteArrayContent(it.encodeToByteArray())
+                // emulate a real response stream which typically can only be consumed once
+                // see: https://github.com/awslabs/aws-sdk-kotlin/issues/356
+                object : HttpBody.Streaming() {
+                    private var cnt = 0
```
kggilmer, line 94: nit

a boolean would suffice here but it hardly matters in a unit test

runtime/smithy-test/common/src/aws/smithy/kotlin/runtime/smithy/test/HttpResponseTest.kt
```diff
@@ -88,7 +88,16 @@ fun <T> httpResponseTest(block: HttpResponseTestBuilder<T>.() -> Unit) = runSusp
...
+                // see: https://github.com/awslabs/aws-sdk-kotlin/issues/356
+                object : HttpBody.Streaming() {
+                    private var consumed = false
+                    override fun readFrom(): SdkByteReadChannel {
+                        val content = if (consumed) ByteArray(0) else it.encodeToByteArray()
+                        consumed = true
+                        return SdkByteReadChannel(content)
+                    }
+                }
             } ?: HttpBody.Empty
```
ianbotsf, line 101: **Question**: While I recognize that we were missing how consume-once streams were handled before, it seems like we'll now miss how replayable streams and simple byte arrays are handled. Is there value in testing both cases?

aajtodd: Response bodies are _always_ going to be a single use stream. At least starting from the engine. What happens in middleware may change this (as we do with wrapped response).

PR #528, opened 2021-11-08
refactor(rt): add retry logging; rename retry middleware

### LINKED ISSUE #524: Integrate SdkLogMode.LogRetries with retry middleware
The [SdkLogMode.LogRetries](https://github.com/awslabs/smithy-kotlin/blob/dc4e5949342157d0a598f076152d5eab2052e1af/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/client/SdkLogMode.kt#L56) setting is supposed to enable additional retry logging. 

We missed this when implementing retries. Task is to investigate if this setting is even necessary and if it is to consume this from the retry middleware to produce useful debug logging for retries.

aajtodd:
## Issue \#

closes #524 

## Description of changes

* remove `SdkLogMode.LogRetries`. This mode isn't required as retry logging doesn't require an opt-in behavior from the customer since no sensitive information is logged by retries
* add logging around when requests are retried and when a retry policy has decided to terminate and fail
* rename the retry middleware to just `Retry` (none of the other middleware/features have a `Feature` suffix)


runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/Retry.kt
```diff
@@ -11,31 +11,67 @@ import aws.smithy.kotlin.runtime.http.HttpBody
...
             if (req.subject.isRetryable) {
-                strategy.retry(policy) {
+                var attempt = 1
+                val logger = req.context.getLogger("Retry")
+                val wrappedPolicy = PolicyLogger(policy, logger)
+
+                strategy.retry(wrappedPolicy) {
+                    if (attempt > 1) {
+                        logger.debug { "retrying request, attempt $attempt" }
+                    }
```
ianbotsf, line 73: **Comment**: This feels weird. We're duplicating the attempt-counting machinery that already exists inside the retry strategy. Is this because request context isn't available outside of the `intercept` method? And we need request context to get a context-aware logger?

If so, then I wonder if request context could actually be available from `CoroutineContext` or something more global that would enable easier access to things that aren't `intercept` methods.

Logically, it seems like this logging ("retrying request, attempt _n_") should happen regardless of whether we're in a request or not. It feels like it belongs inside the strategy, not the feature.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/Retry.kt
```diff
@@ -11,31 +11,67 @@ import aws.smithy.kotlin.runtime.http.HttpBody
...
+ * @param strategy the [RetryStrategy] to retry failed requests with
+ * @param policy the [RetryPolicy] used to determine when to retry
+ */
+@InternalApi
+class Retry(
+    private val strategy: RetryStrategy,
+    private val policy: RetryPolicy<Any?>
+) : Feature {
 
-class RetryFeature(private val strategy: RetryStrategy, private val policy: RetryPolicy<Any?>) : Feature {
```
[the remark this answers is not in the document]
aajtodd: I don't recall why I changed it off hand. I find it easier to read the signature this way but there was no reason explicitly for the change other than I was probably doing something else and on autopilot.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/Retry.kt
kggilmer: FYI holding off on this PR as it seems the work is still in progress, LMK if this is not the case.

aajtodd: @kggilmer No it's ready, this is a point of discussion for sure but otherwise it's done

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/Retry.kt
```diff
@@ -11,31 +11,67 @@ import aws.smithy.kotlin.runtime.http.HttpBody
...
+    /**
+     * Wrapper around [policy] that logs termination decisions
+     */
+    private class PolicyLogger(
+        private val policy: RetryPolicy<Any?>,
+        private val logger: Logger,
+    ) : RetryPolicy<Any?> {
+        override fun evaluate(result: Result<Any?>): RetryDirective = policy.evaluate(result).also {
+            if (it is RetryDirective.TerminateAndFail) {
+                logger.debug { "request failed with non-retryable error" }
```
kggilmer, line 58: suggestion

emit the error that was non-retryable

kggilmer (approved):
No concerns found outside of the design debate

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/Retry.kt
aajtodd: It should throw with the error already right?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/Retry.kt
ianbotsf: Circling back, I can see some pros in moving logging to the middleware and some other pros in keeping it part of the strategies. I think the difference comes down to style so I won't block on this point.

ianbotsf (approved):

ianbotsf (approved):

PR #531, opened 2021-11-11
chore: ci updates for windows

aajtodd:
## Issue \#

support for https://github.com/awslabs/aws-sdk-kotlin/pull/313

## Description of changes


kggilmer (changes requested):
We should coordinate on the memory property settings.  IMO if general dev workflow is fine with defaults it should be removed (as you have here) and any jobs that need more can provide an override.

kggilmer (commented):
Discussed, will shift memory settings to doc generation script to leave smithy-kotlin more generic.

kggilmer (approved):

ianbotsf (approved):

PR #533, opened 2021-11-12
chore: optimize codegen for size round1

aajtodd:
## Issue \#

https://github.com/awslabs/aws-sdk-kotlin/issues/411

## Description of changes

* remove `suspend` from deserialization (we aren't using it anyway and the benchmarks show that it is too slow so we aren't likely to ever use it)
    * NOTE: The errors deserialization is still implemented with suspend because exception deserializers implement `HttpDeserialize`
* added `inline` to several places where it was missing
    * inspection of generated code revealed that we were often capturing variables in a lambda and thus a backing class was being generated by the compiler
* inlined the middleware registration which removes a lot of unnecessary extra classes being generated
    * this still suffers from the lambda closures being generated all over the place, e.g. `op.install(Feature) { <lambda capture> }`
    * there will be some follow up work to refactor our middleware types to probably just implement `Middleware` directly rather than feature. I don't see how to inline these as designed
    * this also has the advantage of middleware registration/generation having complete access to the internals of a service client including config, private variables, not to mention access to the input type directly (@ianbotsf this would probably allow the predict endpoint middleware to be implemented differently...)
 

These changes show a 1MB reduction in jar size for dynamodb
```
> ls -lsah
total 18128

10056 -rw-r--r--  1    4.9M Nov 12 10:16 dynamodb-0.9.2-alpha-original.jar
 8072 -rw-r--r--   1    3.9M Nov 12 11:01 dynamodb-0.9.2-alpha.jar
```

### DIFF
--- a/runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/SdkObjectDescriptor.kt
+++ b/runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/SdkObjectDescriptor.kt
@@ -7,40 +7,29 @@ package aws.smithy.kotlin.runtime.serde
 /**
  * Metadata container for all fields of an object/class
  */
-class SdkObjectDescriptor private constructor(builder: BuilderImpl) : SdkFieldDescriptor(
+class SdkObjectDescriptor private constructor(builder: Builder) : SdkFieldDescriptor(
     kind = SerialKind.Struct, traits = builder.traits
 ) {
     val fields: List<SdkFieldDescriptor> = builder.fields
 
     companion object {
-        fun build(block: DslBuilder.() -> Unit): SdkObjectDescriptor = BuilderImpl().apply(block).build()
+        inline fun build(block: Builder.() -> Unit): SdkObjectDescriptor = Builder().apply(block).build()
     }
 
-    interface DslBuilder {
-        /**
-         * Declare a field belonging to this object
-         */
-        fun field(field: SdkFieldDescriptor)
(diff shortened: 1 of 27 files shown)


aajtodd:
Codegen diff preview

https://github.com/awslabs/aws-sdk-kotlin/compare/__generated-main...__generated-optimize-codegen

ianbotsf (approved):

kggilmer:
The general idea here is that we want to get this into DP for the size savings?  Or that this would follow DP?

aajtodd:
> The general idea here is that we want to get this into DP for the size savings? Or that this would follow DP?

For the size savings and overall correctness.

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/SdkObjectDescriptor.kt
```diff
@@ -7,40 +7,29 @@ package aws.smithy.kotlin.runtime.serde
...
             fields.add(field)
         }
 
-        override fun trait(trait: FieldTrait) {
+        fun trait(trait: FieldTrait) {
             traits.add(trait)
         }
 
-        override fun build(): SdkObjectDescriptor = SdkObjectDescriptor(this)
+        @PublishedApi
```
kggilmer, line 32: TIL

kggilmer (approved):
Looks like something is failing in CI but LGTM otherwise.

PR #536, opened 2021-11-16
refactor: remove middleware Feature concept

aajtodd:
## Issue \#

related: aws-sdk-kotlin#411

## Description of changes

* Remove the `Feature` concept from middleware and replace it with more explicit middleware types targeted at specific phases. The closest thing to `Feature` that still exists is (currently) called `InlineMiddleware`. Most middleware only operate on a single phase so by providing explicit types for those middleware it's easier to reason about at a glance
    * The only downside to this is there is no longer a unifying concept to middleware installation (hence all the overloads) since each phase has different generics. The only way to unify it would be to lift the generics into the interface/class but this has downsides by making it impossible to instantiate type agnostic middleware (i.e. middleware that doesn't care about the input/output type and only modifies the request). 
* Provide a new type of middleware `ModifyRequestMiddleware` that doesn't have access to the next handler in the chain. This middleware can only mutate the input, it can't operate on the output or change how the next handler in the chain is called. Most (all but 2 or 3) of the middleware we have falls into this category where it only mutates the outgoing `SdkHttpRequest` instance.
* update `ProtocolMiddleware` in codegen to implement installation as necessary for each specific middleware component. This cuts down the generated size as we weren't able to `inline` the `Feature.create` calls and it was causing extra backing classes for the closures in each operation.
* (fix): bug in `closeAndOpenBlock` that was causing an unbalanced `indent()` call and resulting in rightward drift in all future `write()` calls


runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/operation/OperationMiddleware.kt
```diff
@@ -0,0 +1,81 @@
+/*
```
kggilmer, line 1: Nice!

kggilmer (approved):
Really nice to see all the red.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/MutateHeaders.kt
```diff
@@ -6,16 +6,28 @@
...
+) : ModifyRequestMiddleware {
     private val overrides = HeadersBuilder()
     private val additional = HeadersBuilder()
     private val conditionallySet = HeadersBuilder()
 
+    init {
+        override.forEach { (key, value) -> set(key, value) }
+        append.forEach { (key, value) -> append(key, value) }
+        setMissing.forEach { (key, value) -> setIfMissing(key, value) }
+    }
```
ianbotsf, line 29: **Question**: What are `set`, `append`, and `setIfMissing` operating on? How do these k/v pairs get into `overrides`, `additional`, and `conditionallySet` respectively?

ianbotsf (approved):
Loved it then, still love it now.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/MutateHeaders.kt
aajtodd: `set`, `append` and `setIfMissing` are functions on this type. They could just directly modify the appropriate `HeadersBuilder` if that's preferable

ianbotsf: Oh I missed them in the code. I see them now. It's fine as-is.

PR #537, opened 2021-11-16
fix: remove leaky trace logging from XML serde

ianbotsf:
## Issue \#

(none)

## Description of changes

While auditing our logging posture, I noticed that our XML deserializer logged response data (including potentially-sensitive data) at `TRACE` level. That's unnecessary given that it can be enabled by `SdkLogMode.LogResponseWithBody`. Thus, removing it.

### DIFF
--- a/runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
+++ b/runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
@@ -5,7 +5,6 @@
 
 package aws.smithy.kotlin.runtime.serde.xml
 
-import aws.smithy.kotlin.runtime.logging.Logger
 import aws.smithy.kotlin.runtime.serde.*
 
 // Represents aspects of SdkFieldDescriptor that are particular to the Xml format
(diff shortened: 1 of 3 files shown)


runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
```diff
@@ -5,7 +5,6 @@
 
 package aws.smithy.kotlin.runtime.serde.xml
 
-import aws.smithy.kotlin.runtime.logging.Logger
```
aajtodd, line 8: question
if we don't use it at all can we drop the dependency?

aajtodd (approved):

runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
ianbotsf: Yes, sure can. I wish we had a ktlint-style rule for removing unused dependencies like we have for unused imports.

kggilmer (approved):

PR #542, opened 2021-11-22
feat: add debug call logging

ianbotsf:
## Issue \#

Closes aws-sdk-kotlin#430

## Description of changes

Adds an integration/middleware for logging a call summary at DEBUG level.


aajtodd (commented):
I'm not really sure we need a new middleware to do this since the `httpTraceMiddleware` is already installed and would be a perfect place to extend.


I'm also not sure we need this logging at all. If we get a successful response why do we care about the HTTP status code in a debug message or how long it took? If you get an error response we'll get back an exception (which you can get the HTTP status code out of if you want). 

I don't have an issue per-say with this I'm just not sure it's adding value to me.

ianbotsf:
Cancelling this PR in favor of a better one.

PR #553, opened 2021-12-16
paginator design update

kggilmer:
* Update paginator design to favor flow-based aproach over async iterator abstraction

## Issue \#

https://github.com/awslabs/smithy-kotlin/issues/124

## Description of changes

* Update the design to specify Kotlin Flow based approach to pagination


docs/design/paginators.md
```diff
@@ -70,127 +69,113 @@ used in a model.
...
-/**
- * Terminal operator that collects the given iterable and calls [action] for each result
- */
-inline fun <T> SdkAsyncIterable<T>.collect(crossinline action: suspend (T) -> Unit):Unit = TODO()
+Here is an example that follows from the previous example (codegen):
+
+```kotlin
+fun Flow<ListFunctionsResponse>.onFunctionConfiguration(): Flow<FunctionConfiguration> =
+   transform() { response ->
+      response.functions?.forEach {
```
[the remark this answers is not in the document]
aajtodd: Oh I see it's emitting individual functions, well now you have alternative form to use/consider if needed.

docs/design/paginators.md
```diff
@@ -70,127 +69,113 @@ used in a model.
...
+The targeted type will provide iteration to the `item` element for the user automatically.
 
-/**
- * Terminal operator that collects the given iterable and calls [action] for each result
- */
-inline fun <T> SdkAsyncIterable<T>.collect(crossinline action: suspend (T) -> Unit):Unit = TODO()
+Here is an example that follows from the previous example (codegen):
+
+```kotlin
+fun Flow<ListFunctionsResponse>.onFunctionConfiguration(): Flow<FunctionConfiguration> =
```
aajtodd, line 108: I don't love this name. `on` implies event handling to me. We could probably just name it after the name of `items` (and in the case of longer nesting use the leaf name)

```kotlin
fun Flow<ListFunctionsResponse>.functions(): Flow<FunctionConfiguration>
```

docs/design/paginators.md
```diff
@@ -70,127 +69,113 @@ used in a model.
...
 
-Pagination by default will always generate an `SdkAsyncIterable` over the normal operation output type. 
-If `items` is specified then the generated paginator will be “specialized” to have an additional field that iterates 
-only over the member targeted by `items`. The targeted type will be flattened for the user automatically.
 
-An example of this is demonstrated below:
+### Creating a Paginator
 
+Each operation that supports pagination would receive an extension function (based from the service client interface) to create a paginator. 
+The function generated always begins with the prefix `paginate`, helping with the `discoverablity` design consideration, by allowing users to quickly
```
aajtodd, line 120: Previously we had chosen the operation name + `Paginated` also in the name of discoverability, e.g. `listFunctionsPaginated()`. This was chosen such that when you type out the operation in the IDE you can get the suggestion for the paginated version in the same results as the normal operation. 

I can see it both ways though that being able to see everything that can be paginated is more discoverable with this proposal. I'm not sure I have a super strong opinion but I think I kind of favor the prior design more.

docs/design/paginators.md
```diff
@@ -70,127 +69,113 @@ used in a model.
...
-   private var cursor: String? = null
-   private var isFirstPage: Boolean = true
+Using the model in the introduction would produce the following:
 
-   override operator fun iterator(): SdkAsyncIterator<ListFunctionsResponse> = this
+### Examples
 
-   override suspend operator fun hasNext(): Boolean
-           = isFirstPage || (cursor?.isNotEmpty() ?: false)
+#### Usage - Manual Pagination
```
aajtodd, line 129: nit: this header makes no sense anymore and could just be renamed to simply `Usage`. Manual pagination referred to driving a paginator class from the previous design

docs/design/paginators.md
```diff
@@ -70,127 +69,113 @@ used in a model.
...
+#### Usage - Iterating over Modeled Item
 
-### Creating a Paginator
-
-Each operation that supports pagination would receive an extension function to create a paginator. Using the model in the introduction would produce the following:
+Notice that this example is a superset of the previous.  Simply by providing a transform we are able to iterate over the 
+nested element.  A key design feature of this approach is that users may extend the abstraction as needed simply by providing
+their own flow transforms.  
 
+Additionally, the `item` specified in the model may be deeply nested and involve operations and types with many words.  Generating a
```
aajtodd, line 151: I see. Do you have an example of this? It would be helpful to clarify with a real example. I'm also wondering if we just use the leaf member targeted by `items`

docs/design/paginators.md
```diff
@@ -70,127 +69,113 @@ used in a model.
...
+          println("${key}: ${value}")
+      }
 }
 ```
 
-
 ## Alternatives Considered
 
-
 ### Construction ALT 1 - Extension on the operation input
```
aajtodd, line 179: GH won't let me comment but the example for ALT1 is still based on `SdkAsyncIterable` rather than flow.

docs/design/paginators.md
```diff
@@ -70,127 +69,113 @@ used in a model.
...
-   /**
-    * Returns true if more results are expected
-    */
-   suspend operator fun hasNext(): Boolean
+```kotlin
+fun LambdaClient.paginateListFunctions(initialRequest: ListFunctionsRequest): Flow<ListFunctionsResponse> {
+   return flow {
+      var cursor: kotlin.String? = null
+      var isFirstPage: Boolean = true
+      val req = initialRequest.copy {
```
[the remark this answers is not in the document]
kggilmer: I mean..only if you want the pagination party to end..

jk, yes good catch.

docs/design/paginators.md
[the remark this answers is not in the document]
kggilmer: Yes I'm aware of the previous design here.  I opted to go with `transform` over `flatMapConcat` for a few reasons.
1. `flatMapConcat` is annotated as a "flow preview" and as such is unstable
2. The docs of the function say something to the effect "are you sure this is really what you want?" and I wasn't sure.
3. I find the `transform` function to be is simpler and clearly articulating what's going on, and don't see any downsides to it's usage.  

Regarding `flatMapConcat`, are you saying that the Smithy team recommended this function in particular?  If so do you recall the reasoning?

docs/design/paginators.md
kggilmer: I see your point about the overuse of `on`.  what about `with` or `over`?  Regarding the rest of the function name, I think that specifying the type that the customer will collect in the flow is helpful over the member name, but see your point about using the terminal member specified by items.  I could go either way.

docs/design/paginators.md
kggilmer: I moved to the prefix form because it's my understanding that is how waiter codegen will play out.  I think which ever way we go, we should be consistent.  Another minor point for the prefix is that in postfix it's often off the view in the IDE until you select/scroll to a specific function, and that because it's a static string we know that at least some of the rest of the function name will be legible in the IDE.  Part of discovery I guess.  But yeah, if we opt for postfix so be it.

docs/design/paginators.md
kggilmer: From grepping the models:

```
./redshift.2012-12-01.json:                    "items": "DefaultClusterParameters.Parameters",
./configservice.2014-11-12.json:                    "items": "OrganizationConformancePackDetailedStatuses",
./servicequotas.2019-06-24.json:                    "items": "ServiceQuotaIncreaseRequestInTemplateList",
./ec2.2016-11-15.json:                    "items": "LocalGatewayRouteTableVirtualInterfaceGroupAssociations",
```
I removed the part about the nesting, from what I see at most there are two levels of testing.  It's just the long winded "composite nouns" that present readability issues in codegen afaik.

docs/design/paginators.md
kggilmer: I left this as I wasn't sure how much of it was tied to the original design.  I've updated to mention flow.

docs/design/paginators.md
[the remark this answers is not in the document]
aajtodd: > Regarding flatMapConcat, are you saying that the Smithy team recommended this function in particular? If so do you recall the reasoning?

No sorry I misunderstood the code originally and thought we were not flattening the result (the function signature suggests we are though and thats why I commented again). This is fine as is.

docs/design/paginators.md
aajtodd: > I think that specifying the type that the customer will collect in the flow is helpful over the member name,

I could see that. I see it from the member name which is what customers interact with when they aren't paginating. The lambda example may be overly simplified but *if* it's even a little bit representative `functions` is clearer to me than the type itself. Probably good to hear from @ianbotsf or @kiiadi for their gut reaction here as well.

docs/design/paginators.md
aajtodd: Indeed those are some long names even without nesting. Either proposal is going to yield long function names. 

Here is an example from Java v2 for the `LocalGatewayRouteTableVirtualInterfaceGroupAssociations`
* [EC2 Paginator Factory Method](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/ec2/Ec2Client.html#describeLocalGatewayVirtualInterfaceGroupsPaginator-software.amazon.awssdk.services.ec2.model.DescribeLocalGatewayVirtualInterfaceGroupsRequest-)
* [EC2 Paginator Type](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/ec2/paginators/DescribeLocalGatewayRouteTableVirtualInterfaceGroupAssociationsIterable.html#localGatewayRouteTableVirtualInterfaceGroupAssociations--)

And here is the one for Config Service `OrganizationConformancePackDetailedStatuses`
* [Config Paginator Type](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/config/paginators/GetOrganizationConformancePackDetailedStatusIterable.html)

docs/design/paginators.md
efemoney: I know this is anecdotal but as a consumer I also agree with @aajtodd. We already use the *items'* property name to access it, I believe it'll be natural to have the same as an extension function on the flow.

Also works better (?) in the case of DynamoDb queries because I'm not sure what name the initial proposal wouldve generated vs below:
```
fun Flow<QueryResponse>.items(): Flow<List<Map<String, AttributeValue>>?>
```

kggilmer: Hi @efemoney , thanks for your feedback!  Yes in fact we received more feedback internally regarding this point in particular.  I've just updated the design and the current plan is to generate the mapping function to always be named `items()`.  Here is how the transform would look:

```kotlin
fun Flow<ScanResponse>.items(): Flow<Map<String, AttributeValue>>
```

and sample usage:

```kotlin
dynamoDbClient
    .paginateScan(ScanRequest { tableName = "some-table" })
    .items()
    .collect { item -> // Map<String, AttributeValue>
        println(item)
    }
```

Any thoughts on this?

docs/design/paginators.md
kggilmer: As mentioned above, design has been updated.  The proposal is now to generate the string literal `items` as the function name to access the iterating value.

docs/design/paginators.md
aajtodd: To clarify my suggestion was not to always name it `items()` but rather use the member name targeted by the `@paginated` trait `items` field. In this case it's confusing because `DynamoDb::Scan` actually targets a member named [Items](https://github.com/awslabs/aws-sdk-kotlin/blob/v0.9.5-beta/codegen/sdk/aws-models/dynamodb.2012-08-10.json#L7335). 

So to illustrate with a less confusing example, the `DynamoDb::ListTables` paginator has a [target member](https://github.com/awslabs/aws-sdk-kotlin/blob/v0.9.5-beta/codegen/sdk/aws-models/dynamodb.2012-08-10.json#L5170) of `TableNames` so using my suggestion it would result in an extension like:

```kotlin
fun Flow<ListTablesResponse>.tableNames(): Flow<...>
```

I think this makes the most sense as it has the best symmetry between regular (un-paginated) and paginated responses:

e.g.
**Un-paginated**
```kotlin
val resp = lambda.listFunctions()

resp.functions?.forEach { ... }
```

**Paginated**

```kotlin
lambda.listFunctionsPaginated()
    .functions()     // extension function for the member name targeted by `@paginated.items`
    .collect()
```

In the case of something like redshift which targets a nested member [DefaultClusterParameters.Parameters](https://github.com/awslabs/aws-sdk-kotlin/blob/v0.9.5-beta/codegen/sdk/aws-models/redshift.2012-12-01.json#L5589) we would use the leaf member name (so in this case just `parameters()`). This [appears](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/redshift/paginators/DescribeDefaultClusterParametersIterable.html) to be what Java V2 is doing as it turns out.

ianbotsf: I agree with @aajtodd here. Using `items` may be confusing because:
* It doesn't match with non-paginated call patterns (e.g., there is no `items` member on a regular result object)
* It may conflict with service-local semantics of an Item (e.g., DynamoDB Items won't be returned by a paginated `ListTables` paginated call despite having an `items` member)

JordonPhillips: I think `items` (or some other consistent name) works better than using the member name. In my experience*, most people prefer consistency since once you're working with several different services any amount of inconsistency in interfaces becomes a major sore spot.

> It doesn't match with non-paginated call patterns (e.g., there is no `items` member on a regular result object)

I don't think this is a bad thing. A paginated operation *isn't* the a normal call, and shouldn't be easily confused with one.

> It may conflict with service-local semantics of an Item

I think this is an acceptable trade-off, but would something like `.iterator()` work better? I'm not sure how / if that term is used in the rust community, but to my mind it would be a good description of what the function is returning.

*For some context, this is mostly informed by my experience of working on the Python SDK and CLI for several years. We got a *lot* of complaints any time there were inconsistencies.

aajtodd: > I think items (or some other consistent name) works better than using the member name. In my experience*, most people prefer consistency since once you're working with several different services any amount of inconsistency in interfaces becomes a major sore spot.

Agreed that inconsistency is bad but consistency is relative to something. I still think using the member name is the right choice as it provides consistency w.r.t to the singular response. Using the literal identifier `items()` provides consistency w.r.t other services/calls but it isn't descriptive at all as to what it means. Customers would have to look at the function signature and documentation to figure out what is being paginated. This is particularly true if the thing being paginated over is builtin like `String`. 

e.g. The DDB `ListTables` paginator paginates over `TableNames` which is a list of strings. These will be flattened so the resulting signature would look like:

```kotlin
// Returns a flow of table names
fun Flow<ListTablesResponse>.items(): Flow<String>
```

What does that `String` represent without looking at the documentation? I think `tableNames()` would be much clearer.

JordonPhillips: > What does that `String` represent without looking at the documentation? I think `tableNames()` would be much clearer.

It represents the `Tables` from `ListTables`. The member name for the paginated response list is typically stuttered from the operation name.

> consistency w.r.t to the singular response

This is not something we want. The singular response is not the same thing. Confusing the two is already a problem in other languages that have a stronger distinction than this.

aajtodd: > It represents the Tables from ListTables. 

Yes *I* understand what it is, my point was that the name `items()` tells _customers_ nothing about it without looking at the documentation.

> The member name for the paginated response list is typically stuttered from the operation name.

I'm not sure what you mean here exactly. The docs for the [paginated trait](https://awslabs.github.io/smithy/1.0/spec/core/behavior-traits.html#paginated-trait) state that `items` is:

> The path to an output member of the operation that contains the data that is being paginated across many responses. The named output member, if specified, MUST target a list or map.


---------

> This is not something we want. The singular response is not the same thing. Confusing the two is already a problem in other languages that have a stronger distinction than this.

I don't feel like we are having the same conversation here. The `items` targeted by the trait must target an output member name. When I say consistency w.r.t to the singular response that's what I'm referring to. If a customer interacts with the singular response it looks like:

```kotlin
val singular = lambda.listFunctions { }
singular.functions?.forEach { ... }
```

Paginated using the member name:

```kotlin
lambda.listFunctionsPaginated()
     .functions()
     .collect { ... }
```

This feels like the right consistency to me. It's clear at a glance what I'm trying to do and what is being paged over.

kiiadi: why not both?

JordonPhillips: > Yes I understand what it is, my point was that the name `items()` tells customers nothing about it without looking at the documentation.

That method invocation doesn't exist in a void, it must only follow after a statement that give the necessary context.

> >The member name for the paginated response list is typically stuttered from the operation name.

> I'm not sure what you mean here exactly.

Operation outputs must always be structures, so services teams must give some name to the list they're returning. In the majority of cases, the name of the property is a suffix of the name of the operation or something very close to that. E.g.:

Operation name `ListFoos` becomes output structure member `ListFoosResponse.foos` becomes operation invocation `client.listFoos().foos`.

It's stuttered because it's duplicated information, essentially. You can't avoid that in a singular invocation, but there's no need to preserve it if you're iterating. So you don't lose context going from `listFoosPaginated.foos()` to `listFoosPaginated.items()`.

> I don't feel like we are having the same conversation here.

There are three cases.

1. Single operation invocation:

```
val singular = lambda.listFunctions { }
singular.functions?.forEach { ... }
```

2. Paginating:

```
lambda.listFunctionsPaginated()
     .collect { ... }
```

3. Iterating:

```
lambda.listFunctionsPaginated()
     .items()
     .collect { ... }
```

I'm concerned with making all of these as distinct as possible on an operation level while making each case consistent across operations. 2 and 3 looking like each other is not great because the return types are different and the way you have to interact with them is very different. 1 and 3 looking like each other is not good because 1 is not paginated and confusion there can mean that not all items are looked at (if they thought they were paginating but were not) or that too many are looked at (if they didn't think they were paginating but were). These might sound like purely theoretical issues, but I have personally assisted customers with these issues in the past.

aajtodd: Thank you for clarifying.

kggilmer: In the design review we opted to go with the following, and will solicit feedback from customers once an implementation is available for them to test drive.  Either way the codegen changes are minimal.

> use the member name targeted by the @paginated trait items field

docs/design/paginators.md
kggilmer: It was my misunderstanding that waiter functionality would provide extensions from the client, and so no consistency is needed/possible.  In design review the majority was the following:

> operation name + Paginated

ianbotsf (approved):

aajtodd (approved):

PR #554, opened 2021-12-16
feat(rt): allow query params to be set from resolved endpoints

aajtodd:
## Issue \#

support for aws-sdk-kotlin#475

## Description of changes

* Allow resolved endpoints to add query parameters. This allows the ECS credentials provider `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` or `AWS_CONTAINER_CREDENTIALS_FULL_URI` to specify query parameters.
* Also fixes the formatting of combining endpoint paths and operation paths when the operation path is empty or `/`. Previously this would result in `/endpoint-path/` and now it will be just `/endpoint-path` without the trailing slash. 

### DIFF
--- a/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/ResolveEndpoint.kt
+++ b/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/ResolveEndpoint.kt
@@ -39,6 +39,10 @@ fun setRequestEndpoint(req: SdkHttpRequest, endpoint: Endpoint) {
     if (endpoint.uri.path.isNotBlank()) {
         val pathPrefix = endpoint.uri.path.removeSuffix("/")
         val original = req.subject.url.path.removePrefix("/")
-        req.subject.url.path = "$pathPrefix/$original"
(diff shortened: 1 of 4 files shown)


runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/ResolveEndpoint.kt
```diff
@@ -37,8 +37,17 @@ fun setRequestEndpoint(req: SdkHttpRequest, endpoint: Endpoint) {
...
+        val endpointPath = endpoint.uri.path.removeSuffix("/")
+        req.subject.url.path = if (opPath.isNotBlank()) {
+            "$endpointPath/${opPath.removePrefix("/")}"
+        } else {
+            // keep endpoint path as is
+            endpoint.uri.path
+        }
+    }
+
+    if (!endpoint.uri.parameters.isEmpty()) {
```
kggilmer, line 50: nit/style

the predicate is unnecessary, you could just append the list either way it seems.

kggilmer (approved):
LGTM

PR #560, opened 2022-01-04
fix(rt): only retry client errors if metadata allows

### LINKED ISSUE #555: Default retry policy always retries client exceptions
## Describe the bug

The `StandardRetryPolicy` [always](https://github.com/awslabs/smithy-kotlin/blob/v0.7.4-beta/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicy.kt#L53) evaluates `ClientException` (and sub classes) as retryable

## Expected Behavior
Client exceptions are in general probably not retryable since they all stem from the runtime. The errors usually represent some form of configuration error, serde issue, etc. The policy should be updated to at the very least check the `isRetryable` metadata (which is present on [SdkBaseException](https://github.com/awslabs/smithy-kotlin/blob/v0.7.4-beta/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/Exceptions.kt#L54)) and defaults to `false`. This would allow for client exceptions to be retried if they explicitly opt-in.

aajtodd:
## Issue \#

fixes #555 

## Description of changes

Do not retry `ClientException` unless metadata explicitly allows it. In general client exceptions are a class of exception that aren't retryable (serialization/deserialization errors, state errors, configuration errors, etc)

### DIFF
--- a/runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicyTest.kt
+++ b/runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicyTest.kt
@@ -38,7 +38,7 @@ class StandardRetryPolicyTest {
     }
 
     @Test
-    fun testClientSideException() {
+    fun testRetryableClientException() {
         val ex = ServiceException()
         ex.sdkErrorMetadata.attributes[ErrorMetadata.Retryable] = true
         ex.sdkErrorMetadata.attributes[ServiceErrorMetadata.ErrorType] = ServiceException.ErrorType.Client
@@ -47,9 +47,9 @@ class StandardRetryPolicyTest {
     }
 
     @Test
-    fun testClientException() {
+    fun testNonRetryableClientException() {
         val result = test(ClientException())
-        assertEquals(RetryDirective.RetryError(RetryErrorType.ClientSide), result)
+        assertEquals(RetryDirective.TerminateAndFail, result)
     }
 
     @Test
--- a/runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicyTest.kt
+++ b/runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicyTest.kt
@@ -38,10 +38,18 @@ class StandardRetryPolicyTest {
     }
 
     @Test
-    fun testRetryableClientException() {
+    fun testClientSideException() {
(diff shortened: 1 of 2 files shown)


kggilmer (approved):

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicyTest.kt
```diff
@@ -47,9 +47,9 @@ class StandardRetryPolicyTest {
     }
 
     @Test
-    fun testClientException() {
+    fun testNonRetryableClientException() {
         val result = test(ClientException())
-        assertEquals(RetryDirective.RetryError(RetryErrorType.ClientSide), result)
+        assertEquals(RetryDirective.TerminateAndFail, result)
     }
```
ianbotsf, line 53: **Nit**: Missing a new unit test for `ClientException` that is retryable.

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicyTest.kt
```diff
@@ -38,7 +38,7 @@ class StandardRetryPolicyTest {
...
 
     @Test
-    fun testClientSideException() {
+    fun testRetryableClientException() {
```
ianbotsf, line 41: **Question**: Why did this name change? It's not technically testing for a `ClientException` where `isRetryable` is true...it's still testing a `ServiceException` with `errorType == ErrorType.Client`.

ianbotsf (approved):

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicyTest.kt
aajtodd: Oh you're right I made the update here quickly thinking htis was testing a retryable client exception and missed the fact that this is a service side exception with error type of client (confused yet?). I'll update the name to something like `testClientErrorType` or something.

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries/impl/StandardRetryPolicyTest.kt
aajtodd: Yes I see that now, I thought thats what the other function you commented on was doing but I see now it's not.

PR #564, opened 2022-01-08
Fix smithy sdk no default client

kggilmer:
## Issue \#

N/A

## Description of changes

* This is a precursor to [KMP project generation](https://github.com/awslabs/smithy-kotlin/issues/437) in `smithy-kotlin`.  As-is, any generated Smithy client SDK is invalid (does not compile) because the client interface codegens references to a concrete client type that is not generated because no protocol is available.  This has impact in codegen tests where we have to write test code in a specific way in order for the smithy sdk to compile, and also impacts ongoing work to develop and test kmp projects.  It is likely that we'll have some kind of protocol in the future in `smithy-kotlin` at which point we may wish to revert this change, however this is unlikely to happen soon and so I choose this approach so that codegen tests can be cleaned up and kmp work can be unblocked.  Also we'll probably always want to retain the capability to generate a client sdk sans protocol.

## Additional Testing
* Verified that codegen'd smithy sdk compiles successfully
* Verified that paginator codegen tests can use any name for concrete client implementation


smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ServiceGenerator.kt
```diff
@@ -72,7 +73,7 @@ class ServiceGenerator(private val ctx: RenderingContext<ServiceShape>) {
                     SectionServiceCompanionObject,
                     context = mapOf(SectionServiceCompanionObject.ServiceSymbol to serviceSymbol)
                 ) {
-                    renderCompanionObject()
+                    renderCompanionObject(ctx.protocolGenerator != null)
```
aajtodd, line 76: style

`ctx` is a property on the class, no real reason to thread through the parameter

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ServiceGenerator.kt
```diff
@@ -109,16 +110,20 @@ class ServiceGenerator(private val ctx: RenderingContext<ServiceShape>) {
...
+     * determines if the SDK will generate a default client implementation. If there is no [ProtocolGenerator], do not
+     * codegen references to the non-existent default client.
      */
-    private fun renderCompanionObject() {
+    private fun renderCompanionObject(hasProtocolGenerator: Boolean) {
         writer.withBlock("companion object {", "}") {
-            withBlock("operator fun invoke(block: Config.Builder.() -> Unit = {}): ${serviceSymbol.name} {", "}") {
-                write("val config = Config.Builder().apply(block).build()")
-                write("return Default${serviceSymbol.name}(config)")
+            this.callIf(hasProtocolGenerator) {
```
aajtodd, line 119: nit/style does `this` need qualified here?

aajtodd (approved):

ianbotsf (approved):

PR #567, opened 2022-01-11
Chore test cleanup

kggilmer:
## Issue \#

N/A

## Description of changes

* Rework the paginator compile test to use jvm target over kmp for simpler build file
* Update compile test to "Smithy SDK" for consistency


tests/codegen/paginator-tests/build.gradle.kts
```diff
@@ -19,82 +18,56 @@ buildscript {
...
 }
 
-tasks["smithyBuildJar"].enabled = false
+val coroutinesVersion: String by project
+val sdkVersion: String by project
 
-val codegen by configurations.creating
+tasks["smithyBuildJar"].enabled = false
 
+val codegen: Configuration by configurations.creating
```
aajtodd, line 36: you don't need this, it's an empty configuration. The reason it exists in [benchmarks](https://github.com/awslabs/smithy-kotlin/blob/main/tests/benchmarks/serde-benchmarks/build.gradle.kts#L88) is because we need to add a dependency that smithy build task requires to work the way we want.

You should be able to remove all references to it (e.g. L39 and L41)

tests/codegen/paginator-tests/build.gradle.kts
```diff
@@ -19,82 +18,56 @@ buildscript {
...
+}
+
 dependencies {
+    val smithyCliConfig = configurations.maybeCreate("smithyCli")
+    val kotlinVersion: String by project
+    val smithyVersion: String by project
+
+    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutinesVersion")
+
     smithyCliConfig("software.amazon.smithy:smithy-cli:$smithyVersion")
```
aajtodd, line 68: does this work here? 

This might need to happen in a `buildscript` block to be picked up appropriately. You can tell if the output on the commandline says something like `using explicitly configured Smithy CLI` instead of `detected ...`

tests/codegen/paginator-tests/build.gradle.kts
```diff
@@ -19,82 +18,56 @@ buildscript {
...
-            kotlin.srcDir("${project.buildDir}/generated-src/src")
-        }
-    }
+kotlin.sourceSets.getByName("main") {
+    kotlin.srcDir("${project.buildDir}/generated-src/src")
+    kotlin.srcDir("${project.buildDir}/smithyprojections/paginator-tests/paginator-tests/kotlin-codegen")
 }
 
-tasks["smithyBuildJar"].enabled = false
+val coroutinesVersion: String by project
```
aajtodd, line 31: inconsistent property getter location. You have `kotlinVersion` and `smithyVersion` inside the `dependencies` block but these outside. I'm not seeing a reason for the difference.

aajtodd (approved):
Couple minor comments.

ianbotsf (approved):

tests/codegen/paginator-tests/build.gradle.kts
kggilmer: When I remove the code so that references to `codegen` are removed, I am left with:

```
tasks["smithyBuildJar"].enabled = false

val generateSdk = tasks.create<SmithyBuild>("generateSdk") {
    group = "codegen"
    inputs.file(projectDir.resolve("smithy-build.json"))
}

val stageGeneratedSources = tasks.register("stageGeneratedSources") {
    group = "codegen"
    dependsOn(generateSdk)
}
```

When I execute a test, I get the following error:

```
Projection paginator-tests failed: software.amazon.smithy.build.SmithyBuildException: Unable to find a plugin named `kotlin-codegen` in the `paginator-tests` projection. Is this the correct spelling? Are you missing a dependency? Is your classpath configured correctly?
software.amazon.smithy.build.SmithyBuildImpl.applyPlugin(SmithyBuildImpl.java:384)
software.amazon.smithy.build.SmithyBuildImpl.applyProjection(SmithyBuildImpl.java:325)
```

It's not clear to me what I would do other than keep the code as-is.  Suggestions?

aajtodd: Yeah you need a dependency on `project(":smithy-kotlin-codegen")` and then either set `addRuntimeClasspath = true` or `addBuildScriptClasspath = true` in the `generateSdk` task depending on what you choose. It's unclear why this even works in the first place to be honest. 

Currently it looks like you have it defined in the normal [normal](https://github.com/awslabs/smithy-kotlin/blob/91d6d3bee6bef717626686af98d4d95fb3fbad34/tests/codegen/paginator-tests/build.gradle.kts#L69) dependencies block which would correspond to adding `addRuntimeClasspath`. 


Either that or leave `codegen` configuration there and change the dependencies block to look like:
```kotlin
dependencies {
    codegen(project(":smithy-kotlin-codegen"))
}
```

tests/codegen/paginator-tests/build.gradle.kts
kggilmer: Hmm, well it worked locally.  In looking at output when executing a test I see neither of those strings.  I do see the following regardless if `smithyCliConfig` is in`buildscript {}` or not:

```
> Configure project :tests:codegen:paginator-tests
(scanned and found a Smithy CLI version 1.13.1
```

In any case, I've moved this code into the `buildscript` block as suggested as it appears more correct.

tests/codegen/paginator-tests/build.gradle.kts
kggilmer: thanks, that worked :+1:

PR #568, opened 2022-01-11
feat(codegen): allow custom rendering of config properties

aajtodd:
## Issue \#

support for https://github.com/awslabs/aws-sdk-kotlin/pull/498

## Description of changes

* provide a way to render a config property using a custom render implementation

There are alternative options we could consider. Please share any ideas if you've got them.


smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigGenerator.kt
```diff
@@ -140,6 +140,7 @@ class ClientConfigGenerator(
                         prop.propertyType.default
                     )
                 }
+                is ClientConfigPropertyType.Custom -> prop.propertyType.render(prop, ctx.writer)
```
ianbotsf, line 143: **Style**: Since we're already switching on `prop.propertyType`, this and other repeated references to `prop.propertyType` could be cleaned up with a temporary capture:

```kotlin
when (val type = prop.propertyType) {
    is ClientConfigPropertyType.Custom -> type.render(prop, ctx.writer)
}
```

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
```diff
@@ -180,6 +180,11 @@ sealed class ClientConfigPropertyType {
      * @param default the value to assign if the corresponding builder property is null
      */
     data class RequiredWithDefault(val default: String) : ClientConfigPropertyType()
+
+    /**
+     * A configuration property that uses [render] to manually render the actual immutable property
+     */
+    data class Custom(val render: (ClientConfigProperty, KotlinWriter) -> Unit) : ClientConfigPropertyType()
```
ianbotsf, line 187: **Question**: Does anything about this custom type _require_ the property be immutable as stated in the KDoc?

ianbotsf (approved):

kggilmer (approved):

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
aajtodd: Technically no (although all property types are currently rendered as `val`). 

The alternative here to `Custom` would have been to allow existing property types to have an optional `mapping` argument that apples a mapping to the builder.

Issue #495, opened 2021-10-11
HTTP Client Engine SPI/Auto load


aajtodd:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue, please leave a comment

## Describe the Feature

Implement SPI support for `HttpClientEngine`. This would allow generated clients to not necessarily take a hard dependency on a specific HTTP client engine. Instead codegen could generate something like:

```kotlin

val engine = config.engine ?: sdkHttpClient(/* engine loaded automatically or throw if none found */) {
     // configure common engine properties
} 

```

An example of what ktor does to support KMP is here:
* [common factory](https://github.com/ktorio/ktor/blob/master/ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt#L42)
* [jvm spi implementation](https://github.com/ktorio/ktor/blob/master/ktor-client/ktor-client-core/jvm/src/io/ktor/client/HttpClientJvm.kt#L42)
* [native implementation](https://github.com/ktorio/ktor/blob/master/ktor-client/ktor-client-core/posix/src/io/ktor/client/engine/Loader.kt#L18)
    * [example](https://github.com/ktorio/ktor/blob/main/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/Curl.kt#L30-L40) of curl engine registering itself using init hook

## Additional Context

- [ ] I may be able to implement this feature request

aajtodd:
This would also be useful for the runtime to not have to depend on an explicit engine (e.g. `ImdsClient`, credential providers, etc)

ianbotsf:
This path forward is no longer favored. Resolving in favor of a different task to provide an explicit default artifact (#602).

PR #577, opened 2022-01-26
Kotlin MP gradle file generation

kggilmer:
## Issue \#
[](https://github.com/awslabs/smithy-kotlin/issues/437)

## Description of changes

* Add flag to `KotlinSettings` to control JVM or Multiplatform gradle build file generation
* Add facility to `CodeWriter` and `KotlinWriter` to allow template style codegen, inspired by `smithy-rs`
* Add necessary KMP abstractions to runtime library to support pure `common` SDK generation
* Update `GradleGenerator` to generate KMP projects for existing supported variations specified by `KotlinSettings`.
* Minor bug fixes found in testing, cleanup.
* Extract `CodeWriter` logic into separate file/test from `KotlinWriter` logic.

### NOTE

We opted to not take the compile-tests from [the draft](https://github.com/awslabs/smithy-kotlin/pull/574) that verify gradle file generation is valid in favor of a future Gradle-based testing approach.  As such, Gradle file generation correctness lacks test coverage (as was before this PR).


kggilmer:
Related to https://github.com/awslabs/smithy/issues/1066

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/GradleGenerator.kt
```diff
@@ -9,78 +9,202 @@ import software.amazon.smithy.kotlin.codegen.KotlinSettings
...
+    if (!isKmp) {
+        writer.write("implementation(kotlin(\"stdlib\"))")
+    }
+
+    // TODO - can we make kotlin dependencies not specify a version e.g. kotlin("kotlin-test")
+    // TODO - Kotlin MPP setup (pass through KotlinSettings) - maybe separate gradle writers
+    val orderedDependencies = dependencies.sortedWith(compareBy({ it.config }, { it.artifact }))
+    orderedDependencies
+        .filter {
+            if (isSrcScope) !it.config.isTestScope else it.config.isTestScope
```
ianbotsf, line 182: **Nit**: Boolean logic can be simplified:

`.filter { it.config.isTestScope != isSrcScope }`

This highlights the incongruity in these flags. We have one local flag for source scope and one broader flag for test scope. We might be able to clean this up by making the local flag refer to test scope or by changing both to an enum or anything else that unifies their intent.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/KotlinTypes.kt
```diff
@@ -34,9 +34,14 @@ object KotlinTypes {
...
-    val List: Symbol = builtInSymbol("List", "kotlin.collections")
-    val Set: Symbol = builtInSymbol("Set", "kotlin.collections")
-    val Map: Symbol = builtInSymbol("Map", "kotlin.collections")
+    object Collections {
+        val List: Symbol = builtInSymbol("List", "kotlin.collections")
+        val Set: Symbol = builtInSymbol("Set", "kotlin.collections")
+        val Map: Symbol = builtInSymbol("Map", "kotlin.collections")
+    }
+    object Text {
+        val encodeToByeArraySymbol = builtInSymbol("encodeToByteArray", "kotlin.text")
```
ianbotsf, line 43: **Nit**: `Bye` → `Byte`

Also, we don't typically suffix `Symbol` to these definitions. Everything in this file should be symbols so explicitly calling them such is not super-helpful.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolClientGenerator.kt
```diff
@@ -94,7 +94,7 @@ abstract class HttpProtocolClientGenerator(
      * or else you need to override [renderInit] and construct it manually
      */
     protected open val defaultHttpClientEngineSymbol: Symbol = buildSymbol {
-        name = "KtorEngine"
+        name = "DefaultHttpClientEngine"
         namespace(KotlinDependency.HTTP_KTOR_ENGINE)
     }
```
ianbotsf, line 99: **Question**: Is an overridable definition still required here? Or now that we have an `expect` definition can we just move this to **RuntimeTypes.kt** like all our other static symbols?

ianbotsf (approved):
Looks great overall! Minor nits and questions only.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinWriter.kt
```diff
@@ -26,158 +26,17 @@ import software.amazon.smithy.utils.CodeWriter
 import java.util.function.BiFunction
 
 /**
- * Extension function that is more idiomatic Kotlin that is roughly the same purpose as
- * the provided function `openBlock(String textBeforeNewline, String textAfterNewline, Runnable r)`
+ * Provides capability of writing Kotlin source code. An instance of a KotlinWriter corresponds to a file emitted
+ * from codegen.
```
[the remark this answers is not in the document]
aajtodd: I think in spirit yes that is still true or at least that is how I would want people to conceptually think about a writer in most cases. Even when inline writing with `W` it is rendered to the same file as the parent writer.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/KotlinTypes.kt
aajtodd: nit

we don't add `Symbol` suffix elsewhere and could just be `encodeToByteArray`. This is just carryover from relocation I think.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolClientGenerator.kt
aajtodd: Yes for now we still need this. `expect` / `actual` doesn't apply to dependent code. They are resolved at compile time for the given sourceSet. IIRC there is something on the kotlin roadmap to maybe allow that sort of thing where you can say I expect you to provide this but it doesn't exist right now.


Eventually I want this to be removed and rely on #495

runtime/protocol/http-client-engines/http-client-engine-ktor/common/src/aws/smithy/kotlin/runtime/http/engine/ktor/DefaultHttpClientEngine.kt
```diff
@@ -0,0 +1,14 @@
...
+package aws.smithy.kotlin.runtime.http.engine.ktor
+
+import aws.smithy.kotlin.runtime.http.engine.*
+
+/**
+ * Specifies the common default http client of which platform specific [HttpClientEngine]'s actualize
+ *
+ * @param config Provides configuration for the DefaultHttpClientEngine
+ */
+expect class DefaultHttpClientEngine(config: HttpClientEngineConfig = HttpClientEngineConfig.Default) : HttpClientEngineBase
```
aajtodd, line 14: Why rename this? `Default` tells me nothing about the underlying engine. It's based on ktor and even the artifact is still `http-client-engine-ktor`. The fact that it's chosen as the "default" in codegen shouldn't influence this name. In particular because when a customer chooses an explicit engine to override the default, those names should be explicit, e.g.

* `CrtHttpEngine`
* `NettyHttpEngine`
* `KtorHttpEngine`
* `CurlHttpEngine`
* etc
I'm not saying we will support all of those or anything just that as the number of possible engines we do support expands over time they ought to be named explicitly IMHO.

When we get to KMP support for all this I'll probably rework how we wire up the default anyway because I'd like it to be based on some form of #495 rather than needing to be explicitly set in codegen but I don't have enough to go on yet to make a recommendation w.r.t what that will look like eventually.

Perhaps we just `typealias DefaultHttpClientEngine = KtorEngine` if the name is important to everyone else.

runtime/protocol/http-client-engines/http-client-engine-ktor/build.gradle.kts
```diff
@@ -11,6 +11,11 @@ val ktorVersion: String by project
...
 kotlin {
     sourceSets {
+        commonMain {
+            dependencies {
+                api(project(":runtime:protocol:http"))
+            }
+        }
         jvmMain {
             dependencies {
                 api(project(":runtime:protocol:http"))
```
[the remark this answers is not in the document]
aajtodd: it does not

runtime/protocol/http-client-engines/http-client-engine-ktor/jvm/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngine.kt
```diff
@@ -32,9 +32,16 @@ import aws.smithy.kotlin.runtime.http.response.HttpResponse as SdkHttpResponse
 /**
  * JVM [HttpClientEngine] backed by Ktor
  */
-class KtorEngine(
-    private val config: HttpClientEngineConfig = HttpClientEngineConfig.Default
+actual class DefaultHttpClientEngine actual constructor(
+    config: HttpClientEngineConfig
 ) : HttpClientEngineBase("ktor-okhttp") {
```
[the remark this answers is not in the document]
aajtodd: see my other comment, I don't love this name direction

runtime/protocol/http-client-engines/http-client-engine-ktor/jvm/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngine.kt
```diff
@@ -32,9 +32,16 @@ import aws.smithy.kotlin.runtime.http.response.HttpResponse as SdkHttpResponse
...
+
+    private val config: HttpClientEngineConfig
```
aajtodd, line 39: nit

I believe you can just do

```kotlin
actual class DefaultHttpClientEngine actual constructor(
    private val config: Config
)
```

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/CodeWriterExt.kt
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * Convenience function to get a typed value out of the context or throw if the key doesn't exist
+ * or the type is wrong
+ */
+inline fun <reified T> CodeWriter.getContextValue(key: String): T = checkNotNull(getContext(key) as? T) {
+    "Expected `$key` in CodeWriter context"
+}
+
+// Specifies a function that receives a [CodeWriter]
+typealias InlineCodeWriter = CodeWriter.() -> Unit
```
aajtodd, line 115: nit

Still think `Writable` may help connect to the formatter `W` better as far as cognitive overhead

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinWriter.kt
```diff
@@ -383,6 +296,27 @@ class KotlinPropertyFormatter(
     }
 }
 
+// Specifies a function that receives a [KotlinWriter]
+typealias InlineKotlinWriter = KotlinWriter.() -> Unit
+/**
+ * Formatter to enable passing a writing function
+ * @param codeWriterCreator function that creates a new [CodeWriter] instance used to generate output of inline content
+ */
+class InlineKotlinWriterFormatter(private val parent: KotlinWriter) : BiFunction<Any, String, String> {
```
aajtodd, line 305: Why do we need two different versions of this? I think we ought to maybe discourage the use of a raw `CodeWriter` and opt for `KotlinWriter` in our codebase.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/KotlinTypes.kt
```diff
@@ -34,9 +34,14 @@ object KotlinTypes {
     val Double: Symbol = builtInSymbol("Double")
     val Boolean: Symbol = builtInSymbol("Boolean")
 
-    val List: Symbol = builtInSymbol("List", "kotlin.collections")
-    val Set: Symbol = builtInSymbol("Set", "kotlin.collections")
-    val Map: Symbol = builtInSymbol("Map", "kotlin.collections")
+    object Collections {
```
aajtodd, line 37: nice

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/GradleGenerator.kt
```diff
@@ -9,78 +9,202 @@ import software.amazon.smithy.kotlin.codegen.KotlinSettings
...
-                )
+    val pluginsBodyRenderer: InlineCodeWriter = {
+        val pluginName = if (isKmp) "multiplatform" else "jvm"
 
-                writer.write(formatted)
+        write(
+            "kotlin(\"$pluginName\") #W",
+            { w: CodeWriter ->
+                if (isRootModule) {
+                    w.write("""version "#L"""", KOTLIN_COMPILER_VERSION)
```
aajtodd, line 52: could use `w.write("version #S", KOTLIN_COMPILER_VERSION)` to get string escaping

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/GradleGenerator.kt
```diff
@@ -9,78 +9,202 @@ import software.amazon.smithy.kotlin.codegen.KotlinSettings
...
+                    useJUnit()
+                }
+            }
+            sourceSets {
+                val commonMain by getting {
+                    dependencies {
+                        #W
+                    }
+                }
+                val jvmMain by getting
```
aajtodd, line 114: shouldn't be necessary

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/GradleGenerator.kt
```diff
@@ -9,78 +9,202 @@ import software.amazon.smithy.kotlin.codegen.KotlinSettings
...
+        }  
+        #W
+        
+        kotlin {
+            jvm {
+                compilations.all {
+                    kotlinOptions.jvmTarget = "#L"
+                }
+                testRuns["test"].executionTask.configure {
+                    useJUnit()
```
aajtodd, line 105: fix

This should match the JVM version and be configured for junit5

```kotlin
    jvm {
        testRuns["test"].executionTask.configure {
            useJUnitPlatform()
            testLogging {
                events("passed", "skipped", "failed")
                showStandardStreams = true
            }
        }
    }

```

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/GradleGenerator.kt
```diff
@@ -9,78 +9,202 @@ import software.amazon.smithy.kotlin.codegen.KotlinSettings
...
+    writer.write(
+        """
+        plugins {
+            #W
+        }  
+        #W
+        
+        kotlin {
+            jvm {
+                compilations.all {
```
aajtodd, line 101: question

Should this entire block be dependent on whether we generate the root module or not?

In general when not generating the root module we have been sort of just relying on the parent project to take care of specific configuration like this. 

e.g. for non root module we maybe just generate:

```kotlin
plugins {... }

kotlin {
    jvm()
    
    sourceSets { ... }

    ...
}

```

I'm not sure it matters right now but it would align with what we have been doing.

aajtodd (commented):
Looks great overall. Couple minor questions, nits, suggestions, etc

runtime/protocol/http-client-engines/http-client-engine-ktor/common/src/aws/smithy/kotlin/runtime/http/engine/ktor/DefaultHttpClientEngine.kt
ianbotsf: Interesting. What if we don't support some engine on some platform for some random reason (e.g., KTOR on JS). Won't having an `expect` then mean we can't compile for the platform that's missing engine implementations?

aajtodd: It's not a requirement that any particular engine support all platforms or even more than one.

To be clear I'm not necessarily saying we shouldn't use `expect/actual` here in this case. KTOR has implementations for each of the platforms such that we could provide a KMP version of our HTTP engine abstraction using ktor. Which solves the issue Ken was trying to solve here (enabling us to generate a KMP client from smithy (the CRT engine is already KMP so we don't have this problem)).

aajtodd: > Won't having an expect then mean we can't compile for the platform that's missing engine implementations?

`expect` applies to every `target` declared in the build. An artifact need not support _every_ possible KMP target. It will only of course be usable as a dependency for those specific targets.

I think what you're really getting at is if we eventually have an engine that only works on say `linuxX64` "won't that mean that a customer can't use that in common"? The answer is yes. However, if they are providing a custom HTTP engine that isn't the default then that is on them to manage. If they are using KMP they may need to write a little bit of `expect/actual` themselves to handle per/target configuration differences.

Our only requirement is that what we generate is KMP compatible. A specific HTTP engine implementation can't have the same requirement because obviously some engines are only available on some platforms. 

Happy to chat offline about this as well, it's a meaty subject.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinWriter.kt
aajtodd: I see we (I) used a raw `CodeWriter` in the gradle codegen. I don't recall why that was done at this point. Might be OBE though 🤷

runtime/protocol/http-client-engines/http-client-engine-ktor/build.gradle.kts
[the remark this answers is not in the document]
kggilmer: nice catch, fixed in subsequent push

runtime/protocol/http-client-engines/http-client-engine-ktor/common/src/aws/smithy/kotlin/runtime/http/engine/ktor/DefaultHttpClientEngine.kt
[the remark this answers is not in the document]
kggilmer: Kotlin compiler says: `Expected class constructor cannot have a property parameter`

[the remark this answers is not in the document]
ianbotsf: Boo.

Okay, then how about making `val config: HttpClientEngineConfig` part of the `expect` class body? We've seen recently that classes which don't make their immutable inputs available post-hoc can be inconvenient.

kggilmer: Regarding the name change, I renamed it because I didn't find `KtorEngine` to lend itself as well to forward-compat.  I choose `Default` as that seemed closest to the intention of the type...meaning the default HTTP client that the SDK ships with.  By using a generic name, we may have more flexibility in the future if we decide to use a different implementation as default.  In any case, since you seem to feel strongly about the existing name, I'll change it back.

If there are other `expect/actual` concerns in this thread please LMK, it isn't clear to me that there is.

runtime/protocol/http-client-engines/http-client-engine-ktor/jvm/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngine.kt
kggilmer: I wasn't able to get that to compile.

runtime/protocol/http-client-engines/http-client-engine-ktor/common/src/aws/smithy/kotlin/runtime/http/engine/ktor/DefaultHttpClientEngine.kt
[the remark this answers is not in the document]
kggilmer: Sure I can do that.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinWriter.kt
kggilmer: Well `KotlinWriter` does some kotlin-y things (comment, package, imports) that wouldn't make sense in a Gradle file.  As discussed I investigated another approach where we would have a similar type, `GradleWriter` however found  that to be relatively complex for the value and found this approach to be better.

We may at some point wish to have a "CodeWriterPlus" type that provides some of our features on top of the Smithy Java `CodeWriter`, but without tying it specifically to Kotlin language files.  But,  I didn't see particular merit in adding it in this PR.

kggilmer: Regarding why we need two types (for `CodeWriter` and `KotlinWriter`), it comes down to the limitations of `typealias` and generics.  There may be a way of consolidating this once we're able to access the settings of `CodeWriter`.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/CodeWriterExt.kt
kggilmer: This issue is that there are two, one for `CodeWriter` and one for `KotlinWriter`.  Do you prefer `KotlinWritable` and `CodeWritable` to what is here?

runtime/protocol/http-client-engines/http-client-engine-ktor/common/src/aws/smithy/kotlin/runtime/http/engine/ktor/DefaultHttpClientEngine.kt
aajtodd: I think this is worth a discussion offline to align but no other concerns though.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/CodeWriterExt.kt
aajtodd: Meh with two of them I guess I don't care as much. It's fine as is

aajtodd (approved):

Issue #572, opened 2022-01-13
Implement waiters


ianbotsf:
## Describe the Feature

Implement code-generated waiters as described by #125.

PR #578, opened 2022-01-27
feat: implement Jmespath expression parsing for all modeled AWS services

ianbotsf:
## Issue \#

Addresses #572 

## Description of changes

This change implements handling for the remaining types of Jmespath expressions that are currently used in AWS service models. It explicitly does not implement expression types which are today unused (e.g., boolean and expressions, index/slice expressions, multiselect hash expressions, certain types of literals, etc.).

This change overhauls the existing skeleton of `KotlinJmespathExpressionVisitor` and removes all usage of the model. This greatly simplifies the implementation but, since code generator is now unaware of the exact types of shapes it's operating on, there are two main downsides:
* Use of null checks and null-safe operators (e.g., `?.`, `?:`) in places that are technically unnecessary. The compiler issues warnings when building the generated code but it runs as expected.
* The Jmespath `length` function works over strings and collections but in Kotlin, strings have `length` but collections have `size`. To allow the codegen to work independent of type, I added a convenience property for `Collection<T>.length` which delegates to `size`.

### DIFF
--- a/smithy-kotlin-codegen/build.gradle.kts
+++ b/smithy-kotlin-codegen/build.gradle.kts
@@ -53,6 +53,7 @@ val generateSdkRuntimeVersion by tasks.registering {
 // unlike the runtime, smithy-kotlin codegen package is not expected to run on Android...we can target 1.8
 tasks.compileKotlin {
     kotlinOptions.jvmTarget = "1.8"
+    kotlinOptions.freeCompilerArgs += "-opt-in=kotlin.RequiresOptIn"
     dependsOn(generateSdkRuntimeVersion)
 }
 
--- a/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitorTest.kt
+++ b/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitorTest.kt
@@ -18,217 +18,244 @@ import kotlin.test.fail
 class KotlinJmespathExpressionVisitorTest {
     @Test
     fun testAndExpression() {
-        assertUnimplemented("foo && bar")
+        assertUnimplemented(path = "foo && bar")
     }
 
     @Test
     fun testComparatorExpression_Equality() {
         assertVisit(
-            "foo == bar",
-            "comparison",
-            "val foo = it?.foo",
-            "val bar = it?.bar",
-            "val comparison = foo == bar",
+            path = "foo == bar",
+            expectedActualName = "comparison",
+            expectedCodegen = """
+                val foo = it?.foo
+                val bar = it?.bar
+                val comparison = foo == bar
+            """.trimIndent(),
         )
     }
 
     @Test
     fun testComparatorExpression_EqualityWithString() {
         assertVisit(
-            "`\"foo\"` == bar",
-            "comparison",
-            "val string = \"foo\"",
-            "val bar = it?.bar",
-            "val comparison = string == bar?.toString()",
+            path = "`\"foo\"` == bar",
+            expectedActualName = "comparison",
+            expectedCodegen = """
+                val string = "foo"
+                val bar = it?.bar
+                val comparison = string == bar?.toString()
+            """.trimIndent(),
         )
 
         assertVisit(
-            "foo == `\"bar\"`",
-            "comparison",
-            "val foo = it?.foo",
-            "val string = \"bar\"",
-            "val comparison = foo?.toString() == string",
+            path = "foo == `\"bar\"`",
+            expectedActualName = "comparison",
+            expectedCodegen = """
+                val foo = it?.foo
+                val string = "bar"
+                val comparison = foo?.toString() == string
+            """.trimIndent(),
         )
     }
 
     @Test
     fun testComparatorExpression_EqualityWithTwoStrings() {
         assertVisit(
-            "`\"foo\"` != `\"bar\"`",
-            "comparison",
-            "val string = \"foo\"",
-            "val string2 = \"bar\"",
-            "val comparison = string != string2",
+            path = "`\"foo\"` != `\"bar\"`",
+            expectedActualName = "comparison",
+            expectedCodegen = """
+                val string = "foo"
+                val string2 = "bar"
+                val comparison = string != string2
+            """.trimIndent(),
+
         )
     }
 
     @Test
     fun testComparatorExpression_Inequality() {
         assertVisit(
-            "foo <= bar",
-            "comparison",
-            "val foo = it?.foo",
-            "val bar = it?.bar",
-            "val comparison = if (foo == null || bar == null) null else foo <= bar",
+            path = "foo <= bar",
+            expectedActualName = "comparison",
+            expectedCodegen = """
+                val foo = it?.foo
+                val bar = it?.bar
+                val comparison = if (foo == null || bar == null) null else foo <= bar
+            """.trimIndent(),
         )
     }
 
     @Test
     fun testCurrentNodeInFlattenExpression() {
         assertVisit(
-            "foo[]",
-            "fooOrEmpty",
-            "val foo = it?.foo",
-            "val fooOrEmpty = foo ?: listOf()",
+            path = "foo[]",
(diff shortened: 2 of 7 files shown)


smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitorTest.kt
```diff
@@ -5,91 +5,261 @@
...
+        assertUnimplemented("!foo")
+    }
+
+    @Test
+    fun testObjectProjection() {
+        assertVisit(
+            "foo.*.bar",
+            "projection",
+            "val foo = it?.foo",
+            "val fooValues = foo?.values ?: listOf()",
```
aajtodd, line 192: where does `.values` come from?

aajtodd (approved):
Looks good overall. If generating this is substantially easier without worrying about nullability then it seems ok to treat everything as nullable. It will just be an unnecessary check (although possibly elided by the compiler anyway). 

The extra extension `Collections<T>.length` extension seems fine for now as well. I guess if there are other things that require a `length` function in the future we may need to revisit and add back context here.


The only concern I really have is it's a little hard to follow the tests. I can see that we are checking given some JMES path, what is the name of the final generated identifier, and what are all the individual steps in computing that. I guess I'm missing how that end result is used since in some cases the expression is a boolean, in others it's list, or null, etc

smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitorTest.kt
ianbotsf: That comes from `visitObjectProjection`. It refers specifically to the values of a map, which will then be run through whatever mapping step applies to the projection.

smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitorTest.kt
```diff
@@ -5,91 +5,261 @@
...
+    }
+
+    @Test
+    fun testComparatorExpression_Inequality() {
+        assertVisit(
+            "foo <= bar",
+            "comparison",
+            "val foo = it?.foo",
+            "val bar = it?.bar",
+            "val comparison = if (foo == null || bar == null) null else foo <= bar",
```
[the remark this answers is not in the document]
ianbotsf: The return value here is `Boolean?`. There's no final checks anywhere else. Most everything should be safe because we treat all vals as nullable everywhere _except_ in the return of the `filter` call. That's missing a null-safe conclusion so I'll add one.

ianbotsf:
> The only concern I really have is it's a little hard to follow the tests. I can see that we are checking given some JMES path, what is the name of the final generated identifier, and what are all the individual steps in computing that. I guess I'm missing how that end result is used since in some cases the expression is a boolean, in others it's list, or null, etc

Right, the final usage of the result of the expression isn't part of the tests because it's also not part of the unit under test. The job of this class is solely to generate the code that effectively "runs" a Jmespath expression on a request/response and yields the outcome of that expression (which may be a boolean, list, null, string, etc.).

For the waiters use case, the final comparison of that expression to an expected value is codegenned in `AcceptorGenerator`. There's where we take these expression results and evaluate whether they pass the expectations of the matcher.

Is there some comment, test construct, etc. that could help clarify this for future readers?

aajtodd:
> > The only concern I really have is it's a little hard to follow the tests. I can see that we are checking given some JMES path, what is the name of the final generated identifier, and what are all the individual steps in computing that. I guess I'm missing how that end result is used since in some cases the expression is a boolean, in others it's list, or null, etc
> 
> Right, the final usage of the result of the expression isn't part of the tests because it's also not part of the unit under test. The job of this class is solely to generate the code that effectively "runs" a Jmespath expression on a request/response and yields the outcome of that expression (which may be a boolean, list, null, string, etc.).
> 
> For the waiters use case, the final comparison of that expression to an expected value is codegenned in `AcceptorGenerator`. There's where we take these expression results and evaluate whether they pass the expectations of the matcher.
> 
> Is there some comment, test construct, etc. that could help clarify this for future readers?

I see now yeah this was part of my disconnect is where the result was being used, which appears to be [here](https://github.com/awslabs/smithy-kotlin/blob/feat-jmespath-full/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/AcceptorGenerator.kt#L102-L107)

smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitorTest.kt
aajtodd: Ok so this MUST be a map type that the object expression applies to then?

smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitorTest.kt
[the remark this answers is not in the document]
aajtodd: I see now, it was related to my other more general question on the usage of the result of evaluating the expression.

smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitorTest.kt
ianbotsf: Correct, although the codegen has no way of knowing whether that's true. If the target is _not_ a map the generated code will fail to compile.

smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitorTest.kt
```diff
@@ -5,91 +5,261 @@
...
+
+    @Test
+    fun testSliceExpression() {
+        assertUnimplemented("foo[0:1]")
+    }
+
+    @Test
+    fun testSubExpressions() {
+        assertVisit(
+            "foo.bar.baz",
```
kggilmer, line 227: suggestion

supplying the parameter names for `path` and `expectedActualName` when calling `assertVisit()` in unit tests would go a long way in lending more readability in these tests IMO

kggilmer (approved):

smithy-kotlin-codegen/build.gradle.kts
```diff
@@ -53,6 +53,7 @@ val generateSdkRuntimeVersion by tasks.registering {
 // unlike the runtime, smithy-kotlin codegen package is not expected to run on Android...we can target 1.8
 tasks.compileKotlin {
     kotlinOptions.jvmTarget = "1.8"
+    kotlinOptions.freeCompilerArgs += "-opt-in=kotlin.RequiresOptIn"
```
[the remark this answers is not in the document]
ianbotsf: I'm not sure what you mean. I added this specifically because of the new `codegenReq` function which uses (still-experimental) contracts. There should be no other effect on our codegen execution.

smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitorTest.kt
ianbotsf: Excellent idea!

PR #583, opened 2022-02-03
fix: make waiter comparison codegen type-safe

ianbotsf:
## Issue \#

Related to #572 

## Description of changes

This was missed somehow in service codegen testing. Because Smithy's JMESPath parser always interprets literals (e.g., `` `42` ``) as doubles (e.g., `42.0`), these numbers aren't suitable for direct operator comparison with non-double values. Updating to use `compareTo` which allows comparison across types.


aajtodd (approved):
I'm assuming this is required because we don't actually know the type to cast/compare with?

PR #589, opened 2022-02-14
fix: handle stream response exceptions properly in Ktor engine

ianbotsf:
## Issue \#

aws-sdk-kotlin#524

## Description of changes

Fixes a bug with the Ktor engine's waiter. Due to [a bad refactor](https://github.com/awslabs/smithy-kotlin/commit/a6602fa3299a127fce2a3d45d4a465a79c02349b#diff-3e45cb866e3a7a231d35f49b6986ca43fc637a764d6d59ba84223c595f79784c), the implementation is unable to handle exceptions that occur before the waiter begins waiting (e.g., in the case an exception is thrown). Rather than restore the previous implementation based on channels, switched for a simpler and (hopefully) clearer implementation based on mutexes.

Added unit tests that verify waiter behavior. In particular, `testSignalWhenNotWaiting` failed with the prior implementation.


runtime/protocol/http-client-engines/http-client-engine-ktor/jvm/test/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngineTest.kt
```diff
@@ -0,0 +1,46 @@
...
+    @Test
+    fun testSignalWhenNotWaiting() = runBlockingTest {
+        val start = currentTime
+
+        val waiter = Waiter()
+        launch {
+            delay(500.milliseconds)
+            waiter.signal()
+        }
+        delay(1000.milliseconds)
```
aajtodd, line 42: we might consider changing these to something smaller, no need to wait a full second in the test I don't think

aajtodd (approved):
I don't recall the behavior of `kotlinx.coroutines.Mutex` w.r.t same thread. This refactor could be susceptible to incorrect behavior if the mutex is re-entrant (i.e. same thread that locked it is allowed to acquire). 

I don't _think_ that the mutex is re-entrant (especially since there is an explicit re-entrant lock type in `kotlinx.atomicfu` package) but worth maybe double checking.

runtime/protocol/http-client-engines/http-client-engine-ktor/jvm/test/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngineTest.kt
ianbotsf: This is in a `runBlockingTest` so time has no meaning.

ianbotsf:
> I don't recall the behavior of `kotlinx.coroutines.Mutex` w.r.t same thread. This refactor could be susceptible to incorrect behavior if the mutex is re-entrant (i.e. same thread that locked it is allowed to acquire).

According to the [`Mutex` documentation](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-mutex/), it is non-reentrant.

PR #591, opened 2022-02-15
feat: add DSL overloads to paginator methods

ianbotsf:
## Issue \#

(none)

## Description of changes

This change adds DSL overloads to paginator methods. Previously, the only way to call paginators was with explicit request objects:

```kotlin
val client = SomeClient(…)
val initialRequest = ListItemsRequest {
    foo = …
    bar = …
}
val result = client.listItemsPaginated(initialRequest)
```

With this new change, paginators can now be called using DSL overloads that create request objects internally:

```kotlin
val client = SomeClient(…)
val result = client.listItemsPaginated {
    foo = …
    bar = …
}
```

This brings parity with the invocation options available for regular service operations, waiters, config, etc.

A minor change alongside this is an update to the KDoc formatting done in codegen. Previously, all blank lines were removed. Now, only leading, trailing, and consecutive blank lines are removed. This allows KDocs to contain empty lines which denote paragraph breaks.

### DIFF
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
@@ -15,6 +15,7 @@ import software.amazon.smithy.kotlin.codegen.core.KotlinWriter
 import software.amazon.smithy.kotlin.codegen.core.defaultName
 import software.amazon.smithy.kotlin.codegen.core.withBlock
 import software.amazon.smithy.kotlin.codegen.integration.KotlinIntegration
+import software.amazon.smithy.kotlin.codegen.lang.KotlinTypes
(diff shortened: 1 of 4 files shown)


smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
```diff
@@ -155,6 +154,25 @@ class PaginatorGenerator : KotlinIntegration {
...
+
+        writer.write("")
+        writer
+            .dokka("""
+                $docBody
+                @param block A builder block used for DSL-style invocation of the operation
+                $docReturn
+            """.trimIndent())
+            .withBlock(
+                "fun #T.#LPaginated(block: #T.Builder.() -> Unit): #T<#T> =",
```
aajtodd, line 166: `KotlinTypes.Unit` is available as well and we may want that to be fully qualified since there is a model with `Unit` as a shape IIRC

aajtodd (approved):

PR #593, opened 2022-02-23
Bootstrap event streams

aajtodd:
## Issue \#

sibling: https://github.com/awslabs/aws-sdk-kotlin/pull/537

## Description of changes

This is a WIP branch for event stream support.

* fix(rt): ensure `SdkByteBuffer` has capacity when writing primitives
* doc: update event stream design doc with agreed upon error handling/stream representation
* feat: carve out a path in HTTP binding protocol generator for event stream serialization/deserialization
* misc codegen and runtime updates to handle event streams


docs/design/event-streams.md
```diff
@@ -141,15 +142,70 @@ Modeling the event stream as a field of the request or response allows for [init
 to be implemented. If we directly returned or took a `Flow<T>` as the input or output type we would not be able to represent the initial request or response fields when present.
 
 
+#### Event Stream Error Representation
+
+Event stream unions may model exceptions that can appear on the stream. These exceptions are terminal messages that are intended to be surfaced to the client using
+idiomatic error handling mechansims of the target language. This means that in the model the possible (modeled) errors a consumer may see after the stream has started
+are part of the overall union that makes up the possible events.
```
ianbotsf, line 149: **Style**: This sentence is slightly unclear. May I suggest:

"Thus the modeled errors a consumer may see during the stream are part of the overall union that makes up the possible events."

docs/design/event-streams.md
```diff
@@ -141,15 +142,70 @@ Modeling the event stream as a field of the request or response allows for [init
...
+    data class ThrottlingError(val value: ThrottlingError): MovementEvents()
+    object SdkUnknown : MovementEvents()
+}
+```
+
+This is undesirable though since event stream errors are terminal and end the stream. Keeping them in the set of possible events also means it may be easier for consumers to ignore
+errors depending on what events they are looking for (e.g. by having a catch all `else` branch they may inadvertently ignore an error and think the stream completed successfully).
+
+
+Event stream unions will be special cased to filter out variants targeting error shapes. When these errors are emitted by the service on the stream they will be converted
```
ianbotsf, line 185: **Nit**: "special cased" → "special-cased" or "specially cased"

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt
```diff
@@ -178,7 +178,27 @@ class KotlinSymbolProvider(private val model: Model, private val settings: Kotli
...
+            // a top level operation input/output member referencing a streaming union is represented by a Flow<T>
+            buildSymbol {
+                name = "Flow<${targetSymbol.fullName}>"
+                nullable = true
+                reference(targetSymbol, SymbolReference.ContextOption.DECLARE)
+                reference(RuntimeTypes.KotlinxCoroutines.Flow.Flow, SymbolReference.ContextOption.DECLARE)
+            }
+        } else {
+            targetSymbol
+        }
```
ianbotsf, line 201: **Question**: Would `!isOperationInputOrOutput && isEventStream` be an error case here? If so, should we throw a codegen exception?

ianbotsf (approved):

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt
aajtodd: I was going to check for it but it results in an invalid smithy model and is validated by smithy.

PR #595, opened 2022-03-07
refactor: mangle serializers and deserializers

aajtodd:
## Issue \#

N/A

## Description of changes

* Add `identifier.name` to writer context for inline symbol rendering. This removes a small paper cut where we have a symbol rendered for a known symbol name and still have to figure out the name inside the render implementation. Encapsulating this also allows us to remove some of these function names and force consumers to go through `Symbol` (which would allow mangling to work correctly)
* Adds a primitive mangling implementation for generating payload/document serializers and deserializers with only a subset of the target members.
    * This is mostly to support event streams since all other uses of these functions must target a `MemberShape` (and as such the entire target of that member)

### DIFF
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/SerdeExt.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/SerdeExt.kt
@@ -11,6 +11,7 @@ import software.amazon.smithy.codegen.core.SymbolReference
 import software.amazon.smithy.kotlin.codegen.KotlinSettings
 import software.amazon.smithy.kotlin.codegen.core.SymbolRenderer
 import software.amazon.smithy.kotlin.codegen.core.defaultName
+import software.amazon.smithy.kotlin.codegen.core.mangledSuffix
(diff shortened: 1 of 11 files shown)


smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/SerdeExt.kt
```diff
@@ -59,48 +63,71 @@ fun OperationShape.bodyDeserializerName(): String = "deserialize" + StringUtils.
...
  */
-fun OperationShape.bodyDeserializer(settings: KotlinSettings, block: SymbolRenderer): Symbol = buildSymbol {
+fun OperationShape.bodyDeserializer(
+    settings: KotlinSettings,
+    block: SymbolRenderer
+): Symbol = buildSymbol {
     name = bodyDeserializerName()
     namespace = "${settings.pkg.name}.transform"
-    // place body serializer in same file as operation serializer implementaiton
+    // place body serializer in same file as operation serializer implementation
```
ianbotsf, line 72: **Comment**: May as well fix the typo on L46 too.

ianbotsf (approved):

PR #603, opened 2022-03-15
test: bootstrap common http engine test suite

### LINKED ISSUE #552: Provide re-usable test suite for all http engine implementations
Provide a re-usable test artifact for validating required behaviors of all `HttpClientEngine` implementations.

aajtodd:
## Issue \#

closes #552 

## Description of changes

Provide a common HTTP client engine test suite that all supported engines can test against. These tests run against a real local server that is spun up (from Gradle) before the tests run. Each KMP target will have to provide an instance of `AbstractEngineTest` that iterates over supported engines for that target. 

This initial set of tests is based on adapting tests we already have spread out over the Ktor and CRT engine tests. I expect that over time we will expand this test suite as we find more behaviors that we want to ensure are upheld across different engines. 

Possible future expansions:
* TLS / H2 tests
* Proxy tests
Both require quite a bit more thought/work to provide the necessary local test server instances. We may be better off writing TLS tests against e.g. badssl.com (which CRT uses FWIW).

### DIFF
--- a/runtime/protocol/http-client-engines/test-suite/build.gradle.kts
+++ b/runtime/protocol/http-client-engines/test-suite/build.gradle.kts
@@ -1,0 +1,91 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+import java.io.*
+import java.net.*
+
+description = "Common HTTP Client Engine Test Suite"
+extra["moduleName"] = "aws.smithy.kotlin.http.test"
+
+extra["skipPublish"] = true
+
+val coroutinesVersion: String by project
+val ktorVersion: String by project
+
+kotlin {
+    sourceSets {
+        commonMain {
+            dependencies {
+                implementation(project(":runtime:protocol:http"))
+                implementation(project(":runtime:protocol:http-test"))
+                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:$coroutinesVersion")
+            }
+        }
+
+        jvmMain {
+            dependencies {
+                implementation("io.ktor:ktor-server-cio:$ktorVersion")
+
+                implementation(project(":runtime:protocol:http-client-engines:http-client-engine-ktor"))
+            }
+        }
+
+        all {
+            languageSettings.optIn("aws.smithy.kotlin.runtime.util.InternalApi")
+        }
+    }
+}
+
+open class LocalTestServer : DefaultTask() {
+    @Internal
+    var server: Closeable? = null
+        private set
+
+    @Internal
+    lateinit var main: String
+
+    @Internal
+    lateinit var classpath: FileCollection
+
+    @TaskAction
+    fun exec() {
+        try {
+            println("[TestServer] start")
+            val urlClassLoaderSource = classpath.map { file -> file.toURI().toURL() }.toTypedArray()
+            val loader = URLClassLoader(urlClassLoaderSource, ClassLoader.getSystemClassLoader())
+
+            val mainClass = loader.loadClass(main)
+            val main = mainClass.getMethod("startServer")
+            server = main.invoke(null) as Closeable
+            println("[TestServer] started")
+        } catch (cause: Throwable) {
+            println("[TestServer] failed: ${cause.message}")
+            cause.printStackTrace()
+        }
+    }
+}
+
+val osName = System.getProperty("os.name")
+
+val startTestServer = task<LocalTestServer>("startTestServer") {
+    dependsOn(tasks["jvmJar"])
+
+    main = "aws.smithy.kotlin.runtime.http.test.util.TestServerKt"
+    val kotlinCompilation = kotlin.targets.getByName("jvm").compilations["test"]
+    classpath = (kotlinCompilation as org.jetbrains.kotlin.gradle.plugin.KotlinCompilationToRunnableFiles<*>).runtimeDependencyFiles
+}
+
+val testTasks = listOf("allTests", "jvmTest")
+    .forEach {
+        tasks.named(it) {
+            dependsOn(startTestServer)
+        }
+    }
+
+gradle.buildFinished {
+    if (startTestServer.server != null) {
+        startTestServer.server?.close()
+        println("[TestServer] stop")
+    }
+}
--- a/runtime/protocol/http-client-engines/test-suite/build.gradle.kts
+++ b/runtime/protocol/http-client-engines/test-suite/build.gradle.kts
@@ -61,7 +61,14 @@ open class LocalTestServer : DefaultTask() {
             println("[TestServer] started")
         } catch (cause: Throwable) {
             println("[TestServer] failed: ${cause.message}")
-            cause.printStackTrace()
+            throw cause
+        }
+    }
+
+    fun stop() {
+        if (server != null) {
+            server?.close()
+            println("[TestServer] stop")
         }
     }
 }
@@ -84,8 +91,5 @@ val testTasks = listOf("allTests", "jvmTest")
     }
 
 gradle.buildFinished {
-    if (startTestServer.server != null) {
-        startTestServer.server?.close()
-        println("[TestServer] stop")
-    }
+    startTestServer.stop()
 }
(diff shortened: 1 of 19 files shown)


runtime/protocol/http-client-engines/test-suite/jvm/src/aws/smithy/kotlin/runtime/http/test/util/AbstractEngineTestJVM.kt
```diff
@@ -0,0 +1,44 @@
...
+import kotlin.coroutines.CoroutineContext
+import kotlin.time.Duration
+
+actual abstract class AbstractEngineTest actual constructor() {
+    actual fun testEngines(block: EngineTestBuilder.() -> Unit) {
+        engines().forEach { engine ->
+            val client = sdkHttpClient(engine)
+            testWithClient(client, block = block)
+        }
+    }
```
ianbotsf, line 23: **Question**: This implementation like it would be the same for any platform...iterate through the list of available engines, construct a client, and run the test. What other actions could a platform want to take? Would it be better to make `engines()` the `expect`/`actual` method and pull `testEngines` up into common code?

runtime/protocol/http-client-engines/test-suite/jvm/src/aws/smithy/kotlin/runtime/http/test/util/AbstractEngineTestJVM.kt
```diff
@@ -0,0 +1,44 @@
...
+    runBlocking(context) {
+        if (timeout != null) {
+            withTimeout(timeout) {
+                block()
+            }
+        } else {
+            block()
+        }
+    }
+}
```
ianbotsf, line 44: **Question**: This implementation like it would be the same for any platform...why make it platform-specific?

runtime/protocol/http-client-engines/test-suite/common/test/aws/smithy/kotlin/runtime/http/test/AsyncStressTest.kt
```diff
@@ -0,0 +1,41 @@
...
+                url(env.testServer)
+                url.path = "concurrent"
+            }
+
+            val call = client.call(req)
+            assertEquals(HttpStatusCode.OK, call.response.status)
+
+            try {
+                call.response.body.readAll()
+            } finally {
```
ianbotsf, line 36: **Comment**: I recommend adding an assertion for the response body too.

runtime/protocol/http-client-engines/test-suite/common/test/aws/smithy/kotlin/runtime/http/test/UploadTest.kt
```diff
@@ -0,0 +1,83 @@
...
+import kotlinx.coroutines.launch
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class UploadTest : AbstractEngineTest() {
+    @Test
+    fun testUploadIntegrity() = testEngines {
+        // test that what we write the entire contents given to us
+        test { env, client ->
+            val data = ByteArray(16 * 1024 * 1023) { it.toByte() }
```
ianbotsf, line 32: **Question**: Is `1023` intentional here and elsewhere in this file?

runtime/protocol/http-client-engines/test-suite/jvm/src/aws/smithy/kotlin/runtime/http/test/util/TestServer.kt
```diff
@@ -0,0 +1,71 @@
...
+    downloadTests()
+    uploadTests()
+    concurrentTests()
+
+    routing {
+        post("/echo") {
+            val response = call.receiveText()
+            call.respond(response)
+        }
+    }
```
ianbotsf, line 70: **Question**: Unused?

runtime/protocol/http-client-engines/test-suite/jvm/src/aws/smithy/kotlin/runtime/http/test/util/TestServer.kt
```diff
@@ -0,0 +1,71 @@
...
+            testRoutes()
+        }.start()
+
+        servers.add(Closeable { server.stop(0L, 0L, TimeUnit.MILLISECONDS) })
+
+        // ensure server is up and listening before tests run
+        Thread.sleep(500)
+    } catch (ex: Exception) {
+        servers.close()
+    }
```
ianbotsf, line 53: **Question**: Why swallow this exception?

runtime/protocol/http-client-engines/test-suite/jvm/src/aws/smithy/kotlin/runtime/http/test/util/TestServer.kt
```diff
@@ -0,0 +1,71 @@
...
+    private val resources = mutableListOf<Closeable>()
+
+    fun add(resource: Closeable) {
+        resources.add(resource)
+    }
+
+    override fun close() {
+        resources.forEach(Closeable::close)
+    }
+}
```
ianbotsf, line 33: **Question**: Why is a list of resources necessary? Seems like we'll only ever have one.

ianbotsf (approved):
Generally looks great, I'm excited to see this here!

**Question**: Do we have redundant engine- or platform-specific tests now that can be cleaned up? (e.g., I seem to recall a stress test that looks similar to the one now implemented as an engine test)

**Comment**: I find myself wishing we could more closely tie together the test procedure and the server routing. It'd be nice if both were in the test classes/files so that one could quickly see both in a single context. Ideally each test would configure its own server logic as part of test setup but I'm not sure if Ktor allows runtime changes to routing.

runtime/protocol/http-client-engines/test-suite/build.gradle.kts
```diff
@@ -0,0 +1,91 @@
...
+            val loader = URLClassLoader(urlClassLoaderSource, ClassLoader.getSystemClassLoader())
+
+            val mainClass = loader.loadClass(main)
+            val main = mainClass.getMethod("startServer")
+            server = main.invoke(null) as Closeable
+            println("[TestServer] started")
+        } catch (cause: Throwable) {
+            println("[TestServer] failed: ${cause.message}")
+            cause.printStackTrace()
+        }
```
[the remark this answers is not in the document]
aajtodd: Hmm good question. I can rethrow it I suppose. In practice the tests will fail because the server isn't there

runtime/protocol/http-client-engines/test-suite/jvm/src/aws/smithy/kotlin/runtime/http/test/util/AbstractEngineTestJVM.kt
aajtodd: If you look at the ktor equivalent they have an additional parameter to allow ignoring specific engines for a test (they also have specific debugging per/target). If we don't think we'll ever need per/platform behavior in here then we could move it up

runtime/protocol/http-client-engines/test-suite/jvm/src/aws/smithy/kotlin/runtime/http/test/util/AbstractEngineTestJVM.kt
aajtodd: `runBlocking` is not available on all platforms

runtime/protocol/http-client-engines/test-suite/common/test/aws/smithy/kotlin/runtime/http/test/UploadTest.kt
aajtodd: ha no probably a fat finger, good catch

runtime/protocol/http-client-engines/test-suite/build.gradle.kts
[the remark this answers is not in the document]
ianbotsf: If it causes the build to fail more rapidly with a clearer error message then I think it's worth making a change. Seeing every test fail because the server isn't there seems less useful from a debugging standpoint.

runtime/protocol/http-client-engines/test-suite/jvm/src/aws/smithy/kotlin/runtime/http/test/util/AbstractEngineTestJVM.kt
ianbotsf: Seems like ignoring engines for some tests could also be handled by making `engines()` non-final.

aajtodd: Well the other reason would be to allow per/test engine configuration. Right now we just statically create `KtorEngine` with a default `HttpClientEngineConfig`. Which means we'll need to change to a factory pattern for `engines(config: HttpClientConfig)`.

I'll pull it up for now though since I don't see us needing to skip engines yet (since the whole point here is to assert expected behaviors across all supported engines)

PR #606, opened 2022-03-18
feat(rt): introduce opaque KMP default HTTP client engine

### LINKED ISSUE #602: Default HTTP engine artifacts
### Community Note

## Describe the Feature

Provide default HTTP engine artifacts for KMP.

(n/a)

## Proposed Solution

(n/a)

HTTP engine SPI/auto-load as described in #495 

## Additional Context

(n/a)

aajtodd:
## Issue \#

closes #602 

## Description of changes

* Introduce a new artifact `http-client-engine-default` that provides an `HttpClientEngine` implementation that works for KMP. This allows us to not hard code against a specific engine (which would require that engine be KMP compatible). 
* Fix ktor engine handling of query parameter encoding. We had some signing issues with presign tests because the presigner implementation currently encodes the query parameters. Ktor was encoding these a separate time resulting in query parameters like `%252F` instead of just `%2F` when `/` was present in a query value. The fix for this was to turn off any encoding by ktor and use the encoding used by the SDK. 
* Refactor the `KtorEngine` into `common` and make it a utility that wraps any ktor compatible engine rather than being synonymous with a particular engine. See breaking change notes below

# Breaking Change

This PR marks the `KtorEngine` constructor as deprecated since it no longer provides a concrete engine. Instead it now wraps _some_ ktor compatible engine. 

```kotlin
val client = FooClient {
    ...
    httpClientEngine = KtorEngine()
}
```

Users that are currently constructing service clients like the above snippet will no longer compile on update. The default engine is now the Ktor OKHttp engine and so the recommended fix would be to simply not override `httpClientEngine`.


runtime/protocol/http-client-engines/http-client-engine-default/common/src/aws/smithy/kotlin/runtime/http/engine/DefaultHttpEngine.kt
```diff
@@ -0,0 +1,20 @@
...
+ * Factory function to create a new HTTP client engine using the default for the current KMP target
+ */
+fun SdkHttpEngine(config: HttpClientEngineConfig = HttpClientEngineConfig.Default): HttpClientEngine =
+    newDefaultHttpEngine(config)
+
+fun SdkHttpEngine(block: HttpClientEngineConfig.Builder.() -> Unit): HttpClientEngine {
+    val builder = HttpClientEngineConfig.Builder().apply(block)
+    val config = HttpClientEngineConfig(builder)
+    return SdkHttpEngine(config)
+}
```
ianbotsf, line 18: **Comment**: The name `SdkHttpEngine` feels ambiguous and inaccurate to me. Technically the CRT engine is also an "SDK HTTP Engine". It'd be clearer to me if it were `DefaultHttpEngine` or something similar.

runtime/protocol/http-client-engines/http-client-engine-ktor/common/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngine.kt
```diff
@@ -4,15 +4,157 @@
...
+
+    @Suppress("UNUSED_PARAMETER")
+    @Deprecated(
+        message = "KtorEngine was previously synonymous with the OkHttp engine. It has been modified to wrap any " +
+            "Ktor compliant engine. The default engine has been changed from CRT to Ktor/OkHttp. To fix either " +
+            "remove setting `httpClientEngine` explicitly or instantiate a Ktor compliant engine of your own and " +
+            "use KtorEngine to wrap it. This constructor will be removed in a future release before GA.",
+        level = DeprecationLevel.ERROR
+    )
+    constructor(config: HttpClientEngineConfig = HttpClientEngineConfig.Default) : this(DeprecationEngine)
```
ianbotsf, line 47: **Comment**: It seems cleaner to throw an error immediately upon invoking `KtorEngine()` rather than later:

```kotlin
companion object {
    @Suppress("UNUSED_PARAMETER")
    @Deprecated(message = "KtorEngine was previously...", level = DeprecationLevel.ERROR)
    operator fun invoke(config: HttpClientEngineConfig = HttpClientEngineConfig.Default): KtorEngine =
        error("Invoking KtorEngine without an engine is deprecated")
}
```

ianbotsf (approved):

runtime/protocol/http-client-engines/http-client-engine-ktor/common/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngine.kt
aajtodd: It's irrelevant, code doesn't compile if you target this constructor

PR #619, opened 2022-04-04
fix conditional content-type

lucix-aws:
Fix (reinstate) setting of `Content-Type` based on empty body. Protocol tests in aws-sdk-kotlin will be added to improve catching breakage of this in the future.

## Issue \#
https://github.com/awslabs/aws-sdk-kotlin/issues/567

## Description of changes
Allows requests with no body (eg. `s3::CreateBucket`) to sign properly.


smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt
```diff
@@ -274,7 +274,9 @@ abstract class HttpBindingProtocolGenerator : ProtocolGenerator {
         }
 
         resolver.determineRequestContentType(op)?.let { contentType ->
-            writer.write("builder.headers.setMissing(\"Content-Type\", #S)", contentType)
+            writer.openBlock("if (builder.body !is HttpBody.Empty) {", "}") {
+                writer.write("builder.headers.setMissing(\"Content-Type\", #S)", contentType)
+            }
```
ianbotsf, line 279: **Style**: `openBlock` is part of the Java **smithy** interface and it accomplishes the desired task here but we often prefer [our own Kotlin extension methods](https://github.com/awslabs/smithy-kotlin/blob/main/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/CodeWriterExt.kt) such as `withBlock`, which includes the writer as a lambda receiver. This would allow you to slightly simplify this code by omitting the explicit writer on L278:

```kotlin
writer.withBlock("if (builder.body !is HttpBody.Empty) {", "}") {
    write("builder.headers.setMissing(\"Content-Type\", #S)", contentType)
}
```

ianbotsf (approved):
Looks good. One optional suggestion below.

In your investigation, did you happen to verify that **Content-Length** is also being set under the correct conditions?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt
lucix-aws: I included `Content-Length` verification in the new protocol tests, the outcomes for both with and without a body seem to be correct there.

aajtodd: FYI `Content-Length` is a bit more nuanced anyway since each HTTP engine supports how it is set differently. For instance the ktor based engine doesn't allow content-length to be set and instead will set it based on the `HttpBody` length property.

aajtodd (approved):

PR #620, opened 2022-04-05
fix: readAvailable fallback behavior

### LINKED ISSUE #613: Wrapped ByteStreams fail when used as streaming request bodies
## Describe the bug

Streaming request operations (e.g., S3 `PutObject`) require using `ByteStream` instances. Using `ByteStream` instances which return readers that directly wrap Ktor channels works correctly (e.g., `ByteStream.fromFile(...)`). Attempting to wrap those streams, even without adding any additional logic, fails. Specifically, the stream appears to hang and never complete signing.

## Expected Behavior

Wrapped `ByteStream` instances should be usable as streaming request bodies.

## Steps to Reproduce

The following example demonstrates the problem:

```kotlin
val file = File("...")
val underlyingStream = ByteStream.fromFile(file) as ByteStream.ReplayableStream

val wrappedStream = object : ByteStream.ReplayableStream() {
    override fun newReader(): SdkByteReadChannel {
        val underlyingReader = underlyingStream.newReader()
        return object : SdkByteReadChannel by underlyingReader
    }
}

val s3 = S3Client.fromEnvironment { }
s3.putObject {
    bucket = "..."
    key = "..."
    body = wrappedStream
}
```

aajtodd:
## Issue \#

fixes #613

## Description of changes

* Fill in missing behavior with unit test for `readAvailableFallback`
* Add test to HTTP engine test suite for a wrapped stream

### DIFF
--- a/runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannel.kt
+++ b/runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannel.kt
@@ -158,6 +158,7 @@ internal suspend fun SdkByteReadChannel.readAvailableFallback(dest: SdkByteBuffe
     // channel was closed while waiting and no further content was made available
     if (availableForRead == 0 && isClosedForRead) return -1
     val tmp = ByteArray(minOf(availableForRead.toLong(), limit, Int.MAX_VALUE.toLong()).toInt())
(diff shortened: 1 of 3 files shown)


ianbotsf (approved):
Works for streams with explicit content lengths but not those without. The fix for _that_ can proceed independent of this work in #621.

Issue #617, opened 2022-03-31
Implement KMP AwsSigner


ianbotsf:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue, please leave a comment

## Describe the Feature

Related to #616, provide an implementation of an `AwsSigner` that is purely written in Kotlin and avoids the CRT as a dependency. This implies a KMP (including `expect`/`actual` if necessary) implementation of HMAC SHA256.

## Is your Feature Request related to a problem?

n/a

## Proposed Solution

n/a

## Describe alternatives you've considered

n/a

## Additional Context

- [x] I may be able to implement this feature request

PR #632, opened 2022-04-26
refactor!: separate hashing functions into new subproject

ianbotsf:
## Issue \#

Related to #617

## Description of changes

This moves hashing functions into their own dedicated subproject, **:runtime:hashing**, separates out each algorithm into its own file, and adds a few convenience methods for performing hashes over arbitrary algorithms or with suppliers. It also marks all the classes as `@InternalApi` since they're not intended to be used outside of internal business logic.

Because of namespace changes and marking classes as newly `@InternalApi`, this is a **breaking change**.

**Companion PR**: [aws-sdk-kotlin#593](https://github.com/awslabs/aws-sdk-kotlin/pull/593)


runtime/hashing/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+
+description = "Hashing functions for Smithy services generated by smithy-kotlin"
+extra["displayName"] = "Smithy :: Kotlin :: Hashing"
+extra["moduleName"] = "aws.smithy.kotlin.runtime.hashing"
+
+kotlin {
+    sourceSets {
+        commonMain {
+            dependencies {
+                api(project(":runtime:utils"))
```
aajtodd, line 14: nit add a comment about what we are exposing

aajtodd (approved):

runtime/hashing/build.gradle.kts
ianbotsf: Actually, I've realized this can be `implementation` instead of `api`. Changing...

PR #634, opened 2022-05-02
feat: convert html to markdown for improved dokka readability

lucix-aws:
## Issue \#
#136 

## Description of changes
Convert commonmark/html to kdoc/markdown such that formatting of original model docs aren't destroyed in the Kotlin output.

### DIFF
--- a/smithy-kotlin-codegen/build.gradle.kts
+++ b/smithy-kotlin-codegen/build.gradle.kts
@@ -20,6 +20,7 @@ val smithyVersion: String by project
 val kotlinVersion: String by project
 val junitVersion: String by project
 val kotestVersion: String by project
+val jsoupVersion: String by project
 
 dependencies {
     implementation(kotlin("stdlib-jdk8"))
@@ -27,7 +28,7 @@ dependencies {
     api("software.amazon.smithy:smithy-waiters:$smithyVersion")
     implementation("software.amazon.smithy:smithy-aws-traits:$smithyVersion")
     implementation("software.amazon.smithy:smithy-protocol-test-traits:$smithyVersion")
-    implementation("org.jsoup:jsoup:1.14.1")
+    implementation("org.jsoup:jsoup:$jsoupVersion")
 
     // Test dependencies
     // These are not set as test dependencies so they can be shared with other modules
--- a/smithy-kotlin-codegen/build.gradle.kts
+++ b/smithy-kotlin-codegen/build.gradle.kts
@@ -27,6 +27,7 @@ dependencies {
     api("software.amazon.smithy:smithy-waiters:$smithyVersion")
     implementation("software.amazon.smithy:smithy-aws-traits:$smithyVersion")
     implementation("software.amazon.smithy:smithy-protocol-test-traits:$smithyVersion")
+    implementation("org.jsoup:jsoup:1.14.1")
 
     // Test dependencies
     // These are not set as test dependencies so they can be shared with other modules
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinWriter.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinWriter.kt
@@ -194,7 +194,7 @@ class KotlinWriter(
      * Clean/escape any content from the doc that would invalidate the Kotlin output.
      */
     private fun cleanForWriter(doc: String) = doc
-        // Docs can have valid $ characters that shouldn't run through formatters.
+        // Docs can have valid # characters that shouldn't run through formatters.
         .replace("#", "##")
         // Services may have comment string literals embedded in documentation.
         .replace("/*", "&##47;*")
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinWriter.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinWriter.kt
@@ -184,11 +184,21 @@ class KotlinWriter(
     fun dokka(docs: String): KotlinWriter =
         dokka {
             write(
-                formatDocumentation(
-                    sanitizeDocumentation(docs)
+                cleanForWriter(
+                    formatDocumentation(docs)
                 )
             )
         }
+
+    /**
+     * Clean/escape any content from the doc that would invalidate the Kotlin output.
+     */
+    private fun cleanForWriter(doc: String) = doc
+        // Docs can have valid $ characters that shouldn't run through formatters.
+        .replace("#", "##")
+        // Services may have comment string literals embedded in documentation.
+        .replace("/*", "&##47;*")
+        .replace("*/", "*&##47;")
 
     /**
      * Adds appropriate annotations to generated declarations.
@@ -341,36 +351,6 @@ class InlineKotlinWriterFormatter(private val parent: KotlinWriter) : BiFunction
     }
 }
 
-// Most commonly occurring (but not exhaustive) set of HTML tags found in AWS models
-private val commonHtmlTags = setOf(
-    "a",
-    "b",
-    "code",
-    "dd",
-    "dl",
-    "dt",
-    "i",
-    "important",
-    "li",
-    "note",
-    "p",
-    "strong",
-    "ul"
-).map { listOf("<$it>", "</$it>") }.flatten()
-
-// Replace characters in the input documentation to prevent issues in codegen or rendering.
-// NOTE: Currently we look for specific strings of Html tags commonly found in docs
-//       and remove them.  A better solution would be to generally convert from HTML to "pure"
-//       markdown such that formatting is preserved.
-// TODO: https://github.com/awslabs/smithy-kotlin/issues/136
-private fun sanitizeDocumentation(doc: String): String = doc
-    .stripAll(commonHtmlTags)
-    // Docs can have valid $ characters that shouldn't run through formatters.
-    .replace("#", "##")
-    // Services may have comment string literals embedded in documentation.
-    .replace("/*", "&##47;*")
-    .replace("*/", "*&##47;")
-
 // Remove all strings from source string and return the result
 private fun String.stripAll(stripList: List<String>): String {
     var newStr = this
@@ -379,7 +359,6 @@ private fun String.stripAll(stripList: List<String>): String {
     return newStr
 }
 
-// Remove whitespace from the beginning and end of each line of documentation
 // Remove leading, trailing, and consecutive blank lines
 private fun formatDocumentation(doc: String, lineSeparator: String = "\n") =
     doc
@@ -387,7 +366,7 @@ private fun formatDocumentation(doc: String, lineSeparator: String = "\n") =
         .dropWhile { it.isBlank() } // Drop leading blank lines
         .dropLastWhile { it.isBlank() } // Drop trailing blank lines
         .dropConsecutive { it.isBlank() } // Remove consecutive empty lines
-        .joinToString(separator = lineSeparator) { it.trim() } // Trim line
+        .joinToString(separator = lineSeparator)
 
 /**
  * Filters out consecutive items matching the given [predicate].
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
@@ -10,6 +10,7 @@ import org.jsoup.nodes.Document
 import org.jsoup.nodes.Node
 import org.jsoup.nodes.TextNode
 import org.jsoup.select.NodeVisitor
+import software.amazon.smithy.codegen.core.CodegenException
 import software.amazon.smithy.kotlin.codegen.KotlinSettings
 import software.amazon.smithy.kotlin.codegen.integration.KotlinIntegration
 import software.amazon.smithy.model.Model
@@ -222,7 +223,7 @@ private fun String.applyWithin(start: String, end: String, transform: (String) -
 
     val stringToTransform = substring(substringStart, substringEnd)
     if (stringToTransform.indexOf(start) != -1) {
-        throw RuntimeException("string contains nested start delimiter")
+        throw CodegenException("string contains nested start delimiter")
     }
 
     return substring(0, substringStart) + transform(stringToTransform) + end +
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
@@ -11,8 +11,15 @@ import software.amazon.smithy.model.Model
 import software.amazon.smithy.model.traits.DocumentationTrait
 import software.amazon.smithy.model.transform.ModelTransformer
 
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Document
+import org.jsoup.nodes.Node
+import org.jsoup.nodes.TextNode
+import org.jsoup.select.NodeTraversor
+import org.jsoup.select.NodeVisitor
+
 /**
- * Sanitize all instances of [DocumentationTrait]
+ * Sanitize all instances of [DocumentationTrait] and converts them to KDoc-compliant strings.
  */
 class DocumentationPreprocessor : KotlinIntegration {
 
@@ -21,7 +28,13 @@ class DocumentationPreprocessor : KotlinIntegration {
         return transformer.mapTraits(model) { _, trait ->
             when (trait) {
                 is DocumentationTrait -> {
-                    val docs = sanitize(trait.value)
+                    // There's definitely some improperly escaped HTML characters within preformat blocks in existing
+                    // models. Ensure we strip those now, the parser is VERY forgiving and will mistreat any sequences
+                    // of characters that happen to form tags as such.
+                    val sanitizedDoc = trait.value
+                        .applyWithin("<code>", "</code>", String::escapeHtml)
+                        .applyWithin("<pre>", "</pre>", String::escapeHtml)
+                    val docs = toKdoc(sanitizedDoc)
                     DocumentationTrait(docs, trait.sourceLocation)
                 }
                 else -> trait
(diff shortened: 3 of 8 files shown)


smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
```diff
@@ -21,17 +26,226 @@ class DocumentationPreprocessor : KotlinIntegration {
         return transformer.mapTraits(model) { _, trait ->
             when (trait) {
                 is DocumentationTrait -> {
-                    val docs = sanitize(trait.value)
+                    // There's definitely some improperly escaped HTML characters within preformat blocks in existing
+                    // models. Ensure we strip those now, the parser is VERY forgiving and will mistreat any sequences
+                    // of characters that happen to form tags as such.
+                    val sanitizedDoc = trait.value
+                        .applyWithin("<code>", "</code>", String::escapeHtml)
```
aajtodd, line 33: what if there are multiple `<code>` blocks in a single trait? looks like the `applyWithin` will find first occurrence.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
```diff
@@ -21,17 +26,226 @@ class DocumentationPreprocessor : KotlinIntegration {
...
+
+    private class MarkdownRenderer : NodeVisitor {
+        companion object {
+            const val SUBLIST_INDENT = "   "
+            const val PREFORMAT_MARKER = "`"
+            const val BOLD_MARKER = "**"
+            const val ITALIC_MARKER = "*"
+        }
+
+        var text: String = ""
```
aajtodd, line 67: this looks like `StringBuilder` might be better suited

smithy-kotlin-codegen/build.gradle.kts
```diff
@@ -27,6 +27,7 @@ dependencies {
     api("software.amazon.smithy:smithy-waiters:$smithyVersion")
     implementation("software.amazon.smithy:smithy-aws-traits:$smithyVersion")
     implementation("software.amazon.smithy:smithy-protocol-test-traits:$smithyVersion")
+    implementation("org.jsoup:jsoup:1.14.1")
```
aajtodd, line 30: add `jsoupVersion` to `gradle.properites` with the other `codegen` related versions and reference it the way we do other versions here

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
```diff
@@ -21,17 +26,226 @@ class DocumentationPreprocessor : KotlinIntegration {
...
+        .replace("[", "&#91;")
+        .replace("]", "&#93;")
+
+/**
+ * Operates on all substrings that fall within the provided section delimiters. Returns a new string where all
+ * substrings enclosed as specified have been modified according to the provided transform.
+ */
+private fun String.applyWithin(start: String, end: String, transform: (String) -> String): String {
+    val substringStart = this.indexOf(start) + start.length
+    if (substringStart == -1) return this
```
aajtodd, line 237: adding `start.length` messes this check up doesn't it? It would never be `-1` if `start` has a non-zero length.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
aajtodd: Oh nevermind I missed that it was recursive

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
```diff
@@ -21,17 +26,226 @@ class DocumentationPreprocessor : KotlinIntegration {
...
+
+    val substringEnd = this.indexOf(end, substringStart)
+    if (substringEnd == -1) return this
+
+    val stringToTransform = this.substring(substringStart, substringEnd)
+    return this.substring(0, substringStart) + transform(stringToTransform) + end +
```
aajtodd, line 243: This actually might be easier as a [Regex](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-regex/) which has a `replace` function that will replace all occurrences.

aajtodd (commented):
Overall this looks great! Excited to have much better documentation. Couple minor comments/questions.

smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessorTest.kt
```diff
@@ -68,4 +68,493 @@ class DocumentationPreprocessorTest {
...
     }
+
+    @Test
+    fun `it renders paragraphs`() {
+        val input = "<p>this is paragraph</p><div>this too is paragraph</div>"
+        val expected = """
+        this is paragraph
+        
+        this too is paragraph
+
```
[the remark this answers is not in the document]
lucix-aws: There can be a few at the end because the transformer doesn't check whether the document is ending when it terminates a paragraph, list, etc.

The existing `KotlinWriter` implementation will drop trailing whitespace like this.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinWriter.kt
```diff
@@ -184,12 +184,22 @@ class KotlinWriter(
...
                 )
             )
         }
 
+    /**
+     * Clean/escape any content from the doc that would invalidate the Kotlin output.
+     */
+    private fun cleanForWriter(doc: String) = doc
+        // Docs can have valid $ characters that shouldn't run through formatters.
+        .replace("#", "##")
```
ianbotsf, line 198: **Question**: Should this comment say that `"Docs can have valid # characters…"` (instead of `$`)?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
```diff
@@ -21,17 +26,226 @@ class DocumentationPreprocessor : KotlinIntegration {
...
+        override fun tail(node: Node, depth: Int) {
+            when (node.nodeName()) {
+                "p", "div", "dd" -> {
+                    val nextSibling = node.nextSibling()
+                    text += when {
+                        // break to give the upcoming list a new line
+                        (nextSibling != null && nextSibling.isList()) -> "\n"
+                        // if we're inside a list, the outer list item will close out the line for us
+                        (node.hasAncestor(Node::isList)) -> ""
+                        // all other cases: this is a standalone "text block" which should be displayed as its own
```
ianbotsf, line 148: **Nit**: Unnecessary parentheses. Simplified:

```kotlin
when {
    // break to give the upcoming list a new line
    nextSibling != null && nextSibling.isList() -> "\n"

    // if we're inside a list, the outer list item will close out the line for us
    node.hasAncestor(Node::isList) -> ""

    …
```

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
```diff
@@ -21,17 +26,226 @@ class DocumentationPreprocessor : KotlinIntegration {
...
+                }
+                "code", "pre" -> {
+                    text += PREFORMAT_MARKER
+                }
+                "b", "strong" -> {
+                    text += BOLD_MARKER
+                }
+                "i", "em" -> {
+                    text += ITALIC_MARKER
+                }
```
ianbotsf, line 104: **Style**: braces and newlines are not required for single-statement cases (but feel free to keep them when a line is too long or requires an adjacent comment). Example:

```kotlin
"code", "pre" -> text += PREFORMAT_MARKER
"b", "strong" -> text += BOLD_MARKER
"i", "em" -> text += ITALIC_MARKER
```

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
```diff
@@ -21,17 +26,226 @@ class DocumentationPreprocessor : KotlinIntegration {
...
+private fun Node.stripBlankTextNodes() {
+    if (this is TextNode && this.isBlank) {
+        this.remove()
+        return
+    }
+
+    val childNodes = this.childNodes()
+    if (childNodes.isNotEmpty()) {
+        childNodes.forEach(Node::stripBlankTextNodes)
+    }
```
ianbotsf, line 211: **Nit**: Checking if a collection is empty before iterating it is unnecessary. This can be simplified to:

```kotlin
childNodes().forEach(Node::stripBlankTextNodes)
```

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
```diff
@@ -21,17 +26,226 @@ class DocumentationPreprocessor : KotlinIntegration {
...
+}
+
+private fun Node.hasAncestor(predicate: (Node) -> Boolean): Boolean {
+    if (!this.hasParent()) return false
+
+    val parent = this.parent() as Node
+    return predicate(parent) || parent.hasAncestor(predicate)
+}
```
ianbotsf, line 219: **Style**: Checking if the parent node exists (or is not null) can be collapsed into a safe call operation:

```kotlin
private fun Node.hasAncestor(predicate: (Node) -> Boolean) =
    parent()?.let { predicate(it) || it.hasAncestor(predicate) } == true
```

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
```diff
@@ -21,17 +26,226 @@ class DocumentationPreprocessor : KotlinIntegration {
...
+/**
+ * Operates on all substrings that fall within the provided section delimiters. Returns a new string where all
+ * substrings enclosed as specified have been modified according to the provided transform.
+ */
+private fun String.applyWithin(start: String, end: String, transform: (String) -> String): String {
+    val substringStart = this.indexOf(start) + start.length
+    if (substringStart == -1) return this
+
+    val substringEnd = this.indexOf(end, substringStart)
+    if (substringEnd == -1) return this
```
ianbotsf, line 240: **Question**: Is this intended to operate on nested constructs? If so, I think it might be broken because it assumes that `end` will be found before any subsequent `start` instances.

smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessorTest.kt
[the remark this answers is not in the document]
ianbotsf: I would expect the preprocessor to cleanly return markdown that doesn't need to be processed again (i.e., by trimming whitespace).

smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessorTest.kt
```diff
@@ -68,4 +68,493 @@ class DocumentationPreprocessorTest {
...
++ Specify access permissions explicitly with the `x-amz-grant-read`, `x-amz-grant-read-acp`, `x-amz-grant-write-acp`, and `x-amz-grant-full-control` headers. These parameters map to the set of permissions that Amazon S3 supports in an ACL. For more information, see [Access Control List (ACL) Overview](https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html).
+You can use either a canned ACL or specify access permissions explicitly. You cannot do both.
+
+
+
+## Server-Side- Encryption-Specific Request Headers
+You can optionally tell Amazon S3 to encrypt data at rest using server-side encryption. Server-side encryption is for data encryption at rest. Amazon S3 encrypts your data as it writes it to disks in its data centers and decrypts it when you access it. The option you use depends on whether you want to use Amazon Web Services managed encryption keys or provide your own encryption key. 
++ Use encryption keys managed by Amazon S3 or customer managed key stored in Amazon Web Services Key Management Service (Amazon Web Services KMS) – If you want Amazon Web Services to manage the keys used to encrypt data, specify the following headers in the request.
```
ianbotsf, line 480: **Style**: It seems unnecessary blank lines can creep in even before the document ends. You may want to keep track of a flag that indicates whether intervening blank lines are required rather than greedily appending them to the output to avoid piling up too many of them.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
lucix-aws: It definitely isn't built to handle that. That said, within the specific context of what it's accomplishing in this module (ie. sanitizing HTML within preformat) any documentation that came to us like that would be very broken on its own(in a different way than just having preformat blocks with unescaped `<>&` characters).

eg. Consider the block
```
<code>
<shouldEscapeTheseBrackets>
someFunction();<code></code>
<shouldEscapeTheseBracketsToo>
</code>
```

If those inner tags aren't escaped, the structure of the document is basically destroyed past the first closing tag. The existing implementation would read the block to operate on as 
```
<code>
<shouldEscapeTheseBrackets>
someFunction();<code></code>
```
and the HTML we feed into the parser afterward would be garbage.

I think there's a balance to be struck here, given that we're making a concerted effort to clean up what is essentially malformed HTML to begin with. This may all be moot if there ends up being a regex solution.

Assuming no regex, for this particular application, we'd really want to assume that the inner tags SHOULD be rendered with the tags escaped. Therefore we'd ignore any nested delimiters and let that text be mutated as part of the enclosing pair. **So I think the function should definitely be modified to not get tripped up on nested structures, but it should basically ignore them (and even then, you're still gonna get to operate on that text within the nested structure in your transform, you're just not gonna get a standalone callback for it).**

There is also the copout route of just throwing up hands if we encounter a nesting like this, but if we can detect it we might as well handle it sensibly. Curious what you guys think.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
aajtodd: Ya regex may be difficult to get right but something like:

```
<code>(.*?)<\/code>
```
With option to enable `.` to also match newline characters would probably work.

e.g.

```kotlin
val startToken = Regex.escape(start)
val endToken = Regex.escape(end)
val re = Regex("$startToken(.*?)$endToken", setOf(Regex.DOT_MATCHES_ALL))
...
```

Your approach is fine as well though, it's up to you.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
```diff
@@ -21,17 +26,226 @@ class DocumentationPreprocessor : KotlinIntegration {
...
+    val startIndex = indexOf(start)
+    if (startIndex == -1) return this
+
+    val substringStart = indexOf(start) + start.length
+    val substringEnd = indexOf(end, substringStart)
+    if (substringEnd == -1) return this
+
+    val stringToTransform = substring(substringStart, substringEnd)
+    if (stringToTransform.indexOf(start) != -1) {
+        throw RuntimeException("string contains nested start delimiter")
```
aajtodd, line 225: should probably be `CodegenException`

aajtodd (approved):

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
```diff
@@ -231,21 +209,43 @@ private fun TextNode.markdownText() =
...
+    val substringEnd = indexOf(end, substringStart)
     if (substringEnd == -1) return this
 
-    val stringToTransform = this.substring(substringStart, substringEnd)
-    return this.substring(0, substringStart) + transform(stringToTransform) + end +
-        this.substring(substringEnd + end.length).applyWithin(start, end, transform)
+    val stringToTransform = substring(substringStart, substringEnd)
+    if (stringToTransform.indexOf(start) != -1) {
+        throw RuntimeException("string contains nested start delimiter")
+    }
```
ianbotsf, line 227: **Nit**: Let's use a better exception type here like `CodegenException`.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
```diff
@@ -124,10 +124,10 @@ class PaginatorGenerator : KotlinIntegration {
...
             .dokka(
                 """
-                    $docBody
-                    @param initialRequest A [${inputSymbol.name}] to start pagination
-                    $docReturn
-                """.trimIndent()
+                    |$docBody
+                    |@param initialRequest A [${inputSymbol.name}] to start pagination
+                    |$docReturn
+                """.trimMargin()
```
ianbotsf, line 130: **Question**: Why is an explicit margin necessary now?

ianbotsf (commented):
This looks much improved! Still one lingering question below.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
ianbotsf: Looks like you already made this change during the middle of my review. Disregard!

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
lucix-aws: So when you compound multiline strings like this, they don't yield the indent structure you might expect, especially if they've been `trimIndent()`ed

eg. consider the following
```kotlin
val multiline0 = """
  foo
  bar
  baz
""".trimIndent()

val multiline1 = """
  $multiline0
  qux
  qux2
""".trimIndent()
```

The value of `multiline1` is now the literal
```
  foo
bar
baz
  qux
  qux2
```
because you cut the indent on `multiline0` but then reintroduced it on the first line when you interpolate it in. The "common indent" detected by `trimIndent` is now just an empty string, because the 2nd line onward of the inserted value `multiline0` still has no indent, so the 2nd `trimIndent()` has no effect and the output is scuffed.

We were always generating a string like this before for this pagination doc, but it was hidden by the `KotlinWriter` trimming all head/tail whitespace on comments. We can't just blindly strip whitespace anymore because it's sometimes required by markdown (sublists) so some tests caught the change here.

Providing the margin lets you compose multiline strings like this more clearly/reliably.

ianbotsf: Oh right, I wasn't remembering that the inner strings would've already been unindented. This makes sense!

PR #636, opened 2022-05-09
feat: expose Byte.percentEncodeTo for downstream recursion detection

lucix-aws:
## Description of changes
Adjust smithy-kotlin API support for recursion detection (https://github.com/awslabs/aws-sdk-kotlin/pull/602).


runtime/utils/common/src/aws/smithy/kotlin/runtime/util/text/Text.kt
```diff
@@ -139,3 +140,8 @@ private val percentEncodedParam = """%([A-Fa-f0-9]{2})""".toRegex()
 public fun String.urlDecodeComponent(): String = this
     .replace('+', ' ')
     .replace(percentEncodedParam) { it.groupValues[1].toInt(radix = 16).toChar().toString() }
+
+private fun Char.defaultShouldUrlEncode() = when (this) {
+    in 'a'..'z', in 'A'..'Z', in '0'..'9', '-', '_', '.', '~' -> false
```
lucix-aws, line 145: Looking at this again, it should really just be a single boolean expression.

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/text/Text.kt
```diff
@@ -99,7 +99,7 @@ private const val upperHex: String = "0123456789ABCDEF"
 private val upperHexSet = upperHex.toSet()
 
 // $2.1 Percent-Encoding
-private fun Byte.percentEncode(): String = buildString(3) {
+fun Byte.percentEncode(): String = buildString(3) {
```
aajtodd, line 102: This should be marked `@InternalApi` 

We also might consider changing this to two different functions:

```kotlin

@InternalApi
fun Byte.percentEncodeTo(out: Appendable): Unit {
    val code = toInt() and 0xff
    out.append('%')
    out.append(upperHex[code shr 4])
    out.append(upperHex[code and 0x0f])
}

@InternalApi
private fun Byte.percentEncode(): String {
    val byte = this
    return StringBuilder(3)
        .apply { byte.percentEncodeTo(this) }
        .toString()

}
```

This would allow percent encoding to append directly to an existing string builder (e.g. in our existing encode functions)

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/text/Text.kt
```diff
@@ -99,11 +99,20 @@ private const val upperHex: String = "0123456789ABCDEF"
...
+    out.append(upperHex[code shr 4])
+    out.append(upperHex[code and 0x0f])
+}
+
+@InternalApi
+fun Byte.percentEncode(): String {
+    val byte = this
+    return StringBuilder(3)
+        .apply { byte.percentEncodeTo(this) }
+        .toString()
```
ianbotsf, line 115: **Style**: The `apply` call requires aliasing `this` whereas `also` captures the context object as `it` (vs `this`) and can thus simplify the call chain:

```kotlin
fun Byte.percentEncode(): String = StringBuilder(3).also(::percentEncodeTo).toString()
```

ianbotsf (approved):
Minor feedback. Fix and ship!

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/text/Text.kt
aajtodd: Yeah that's better, what Ian said.

aajtodd (approved):
Looks good other than Ians suggestion. Fix and ship

PR #639, opened 2022-05-10
feat: add convenience getters for union members

ianbotsf:
## Issue \#

Closes [aws-sdk-kotlin#393](https://github.com/awslabs/aws-sdk-kotlin/issues/393)

## Description of changes

This change adds extension getters to unions for retrieving member values. This is useful to reduce boilerplate when working with known subtypes at compile time. For example, DDB's `AttributeValue` has subtypes `S`, `N`, `B`, etc., Users that knew the subtype at compile time needed to cast and retrieve the value like so:

```kotlin
val attr: AttributeValue = TODO()
val stringValue = (attr as AttributeValue.S).value
```

With this PR, users can now retrieve that same value with:

```kotlin
val stringValue = attr.S
```

Note that union members are title-cased in our codegen. I opted to use the same capitalization in the extension getters' names for consistency. An alternative implementation may use camel-cased names instead (e.g., `attr.s` instead of `attr.S`).

### DIFF
--- a/.changes/02852edb-3c2d-4f29-a1ad-5f7e1b884268.json
+++ b/.changes/02852edb-3c2d-4f29-a1ad-5f7e1b884268.json
@@ -1,0 +1,8 @@
+{
+    "id": "02852edb-3c2d-4f29-a1ad-5f7e1b884268",
+    "type": "feature",
+    "description": "Add convenience getters for union members",
+    "issues": [
+        "awslabs/aws-sdk-kotlin#393"
+    ]
+}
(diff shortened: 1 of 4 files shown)


ianbotsf (commented):
Based on an offline conversation, we decided that methods (e.g., `UnionType.asFoo()`) are clearer and more idiomatic than uppercased properties (e.g., `UnionType.Foo`). They also leave the door open to nullable variants (e.g., `UnionType.asFooOrNull()`). I'll rework this PR shortly to switch to those.

aajtodd (approved):
I would maybe move the new methods _after_ all the variants are declared. This will make seeing all the variants easier at a glance in generated code vs having to sift through the `as*` methods in-between them.

Otherwise looks good.

PR #642, opened 2022-05-13
feat: usability enhancements for no required inputs

lucix-aws:
## Issue \#
(https://github.com/awslabs/smithy-kotlin/issues/129)

## Description of changes
Add convenience overload for methods with no required inputs, eg.
```kotlin
S3Client.fromEnvironment().use {
  it.listBuckets().buckets?.forEach {
    println(it.name)
  }
}
```

### DIFF
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/ShapeExt.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/ShapeExt.kt
@@ -108,14 +108,17 @@ fun ServiceShape.hasIdempotentTokenMember(model: Model): Boolean {
 /**
  * Return the formatted (Kotlin) function signature for the given operation
  */
-fun OperationIndex.operationSignature(model: Model, symbolProvider: SymbolProvider, op: OperationShape): String {
+fun OperationIndex.operationSignature(
+    model: Model, symbolProvider: SymbolProvider, op: OperationShape, includeOptionalDefault: Boolean = false): String {
     val inputShape = this.getInput(op)
     val outputShape = this.getOutput(op)
     val input = inputShape.map { symbolProvider.toSymbol(it).name }
     val output = outputShape.map { symbolProvider.toSymbol(it).name }
 
     val hasOutputStream = outputShape.map { it.hasStreamingMember(model) }.orElse(false)
-    val inputParam = input.map { "input: $it" }.orElse("")
+    val inputParam = input.map {
+        if (includeOptionalDefault && inputShape.get().isOptional()) "input: $it = $it{}" else "input: $it"
+    }.orElse("")
     val outputParam = output.map { ": $it" }.orElse("")
 
     val operationName = op.defaultName()
@@ -220,3 +223,8 @@ fun UnionShape.filterEventStreamErrors(model: Model): Collection<MemberShape> {
         target.isError
     }
 }
+
+/**
+ * Test if a shape is optional.
+ */
+fun Shape.isOptional(): Boolean = members().none { it.isRequired }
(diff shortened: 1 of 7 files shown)


smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/ShapeExt.kt
```diff
@@ -108,14 +108,21 @@ fun ServiceShape.hasIdempotentTokenMember(model: Model): Boolean {
...
+): String {
     val inputShape = this.getInput(op)
     val outputShape = this.getOutput(op)
     val input = inputShape.map { symbolProvider.toSymbol(it).name }
     val output = outputShape.map { symbolProvider.toSymbol(it).name }
 
     val hasOutputStream = outputShape.map { it.hasStreamingMember(model) }.orElse(false)
-    val inputParam = input.map { "input: $it" }.orElse("")
+    val inputParam = input.map {
+        if (includeOptionalDefault && inputShape.get().isOptional()) "input: $it = $it{}" else "input: $it"
```
ianbotsf, line 124: **Nit**: Needs a space between the type name and the empty braces.

ianbotsf (approved):

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/ShapeExt.kt
```diff
@@ -108,14 +108,21 @@ fun ServiceShape.hasIdempotentTokenMember(model: Model): Boolean {
 /**
  * Return the formatted (Kotlin) function signature for the given operation
  */
-fun OperationIndex.operationSignature(model: Model, symbolProvider: SymbolProvider, op: OperationShape): String {
+fun OperationIndex.operationSignature(
+    model: Model,
+    symbolProvider: SymbolProvider,
+    op: OperationShape,
+    includeOptionalDefault: Boolean = false
```
aajtodd, line 115: question

Why do we need this parameter? It seems like for any operation we should have only one possible signature.

aajtodd (approved):
One question but otherwise looks good.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/ShapeExt.kt
lucix-aws: It has to do with where you're rendering the signature.

Kotlin requires default args to be defined at the superclass / interface level. So for us that's `{Service}Client`. We're not allowed to mirror that same default arg declaration at the override level ie. `Default{Service}Client`. Thus the parameter is introduced here to control that behavior.

I'm not really a diehard fan of flat-out bool args, the alternative in my mind would be to have the "create signature" method return some structure that the caller can then use conditionally (ie. `render()` vs `renderWithDefaults()`), but the effort vs. value added seemed not worth it just to avoid the bool in the one scenario this is used.

aajtodd: Oh right, gotcha.

Boolean params are fine, especially when combined with named arguments at the call site:

```kotlin
opIndex.operationSignature(model, symbolProvider, op, includeOptionalDefault = true)
```

You might update the function documentation though to document that param in particular.

ianbotsf: If you really don't like boolean here (I don't care too much) you could switch to an enum that indicates the use case:

```kotlin
enum ParameterLocation {
    Definition,
    Override,
}
```

PR #643, opened 2022-05-16
fix: don't escape markdown within preformat blocks

lucix-aws:
## Description of changes
Markdown characters inside of preformat (code/pre) HTML shouldn't be escaped.

eg. the square brackets within
`<code> bracket_list = [1,2,3] </code>`
should be rendered as-is in the final markdown.

### DIFF
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
@@ -200,12 +200,19 @@ private fun Node.hasAncestor(predicate: (Node) -> Boolean): Boolean =
 private fun Node.isList() =
     nodeName().let { it == "ul" || it == "ol" }
 
-private fun TextNode.markdownText() =
-    text()
+private fun Node.isPreformat() =
+    nodeName().let { it == "code" || it == "pre" }
+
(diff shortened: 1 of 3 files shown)


aajtodd (approved):

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/DocumentationPreprocessor.kt
```diff
@@ -200,12 +200,19 @@ private fun Node.hasAncestor(predicate: (Node) -> Boolean): Boolean =
...
+private fun TextNode.markdownText(): String {
+    // If we're inside a preformat block, everything is literal, ie. no escapes required.
+    if (hasAncestor(Node::isPreformat)) return text()
+
+    return text()
         // Replace square brackets with escaped equivalents so that they are not rendered as invalid Markdown
         // links.
         .replace("[", "&#91;")
         .replace("]", "&#93;")
+}
```
ianbotsf, line 215: **Style**: Multiple returns? Gross! Could be simplified with a `when` block:

```kotlin
private fun TextNode.markdownText(): String = when {
    hasAncestor(Node::isPreformat)) -> text()

    // Replace square brackets with escaped equivalents so that they are not rendered as invalid Markdown
    // links.
    else -> text().replace("[", "&#91;").replace("]", "&#93;")
}
```

ianbotsf (approved):

Issue #629, opened 2022-04-18
Bind to OkHttp Engine Directly


aajtodd:
Add a new `http-client-engine-okhttp` artifact which binds to OkHTTP engine directly (rather than using `KtorEngine` + `ktor-engine-okhttp`). Replace the default HTTP engine with this new abstraction (the default is already OkHTTP, it's just going through ktor abstraction layer).

This should allow us to access the duplex stream API's which is one of the things still blocking event streams (the other is event stream signing). 

Longer term it looks like OkHTTP 5 may support KMP which would allow us to use the same default engine potentially across all/multiple KMP targets.

FredyH:
Ktor recently updated to version 2.0 which changed a lot of package locations and renamed a bunch of things.  When mixing dependencies between Ktor 1.x and Ktor 2.0 this will lead to linker errors during runtime which currently prevents me from using the Kotlin AWS client in my Ktor 2.0 project.
I think this change would completely bypass the issue while sacrificing no functionality.

aajtodd:
@FredyH Thanks for weighing in.

This unfortunately won't completely solve the issue. We still have a reliance on `ByteReadChannel` from `ktor-io`. This issue is more about being able to get at all the capability that `okhttp` provides. `ktor` exposes a common subset of capability across different implementations and this unfortunately means that sometimes the more advanced/nuanced features are not accessible.  

That particular issue is being tracked in https://github.com/awslabs/aws-sdk-kotlin/issues/480

Feel free to comment there or upvote.

PR #644, opened 2022-05-17
feat(rt): add additional telemetry for HTTP requests

aajtodd:
## Issue \#

N/A

## Description of changes

* Implement retry header SEP which propagates `amz-sdk-invocation-id` header with our existing client side SdkRequestId value.
* Implement `okhttp::EventListener` interface and register it with the engine. This logs every event from the engine at trace level. Currently this pulls out the (new) invocation ID header to propagate sdk request ID into the logs for correlation with the rest of the logging messages related to that operation. When we implement #629 we can use the `tags` functionality of an okhttp request to propagate whatever context we want.
* Refactor `HttpClientEngine` interface to take `ExecutionContext`. 

### DIFF
--- a/runtime/protocol/http-client-engines/http-client-engine-ktor/common/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngine.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-ktor/common/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngine.kt
@@ -59,6 +59,7 @@ class KtorEngine(
 
     override suspend fun roundTrip(request: HttpRequest): HttpCall {
         val callContext = callContext()
+        val sdkRequestId = request.headers["__sdkRequestId"]
 
         val respChannel = Channel<HttpCall>(Channel.RENDEZVOUS)
 
@@ -77,14 +78,15 @@ class KtorEngine(
         }
 
         // wait for the response to be available, the content will be read as a stream
-        logger.trace("waiting on response to be available")
+        logger.trace("[sdkRequestId=$sdkRequestId] waiting on response to be available")
 
         try {
             val resp = respChannel.receive()
-            logger.trace("response is available continuing")
+            logger.trace("[sdkRequestId=$sdkRequestId] response is available continuing")
             return resp
         } catch (ex: Exception) {
-            throw logger.throwing(ex)
+            logger.trace(ex){ "[sdkRequestId=$sdkRequestId] failed to receive response" }
+            throw ex
         }
     }
 
@@ -93,6 +95,7 @@ class KtorEngine(
         sdkRequest: HttpRequest,
         channel: SendChannel<HttpCall>
     ) {
+        val sdkRequestId = sdkRequest.headers["__sdkRequestId"]
         val builder = KtorRequestAdapter(sdkRequest, callContext).toBuilder()
         val waiter = Waiter()
         val reqTime = Instant.now()
@@ -117,14 +120,14 @@ class KtorEngine(
                 body,
             )
 
-            logger.trace("signalling response")
+            logger.trace("[sdkRequestId=$sdkRequestId] signalling response")
             val call = HttpCall(sdkRequest, resp, reqTime, respTime, callContext)
             channel.send(call)
 
-            logger.trace("waiting on body to be consumed")
+            logger.trace("[sdkRequestId=$sdkRequestId] waiting on body to be consumed")
             // wait for the receiving end to finish with the HTTP body
             waiter.wait()
-            logger.trace("request done")
+            logger.trace("[sdkRequestId=$sdkRequestId] request done")
         }
     }
 
--- a/runtime/protocol/http-client-engines/http-client-engine-ktor/common/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngine.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-ktor/common/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngine.kt
@@ -4,6 +4,7 @@
  */
 package aws.smithy.kotlin.runtime.http.engine.ktor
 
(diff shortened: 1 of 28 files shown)


runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/operation/SdkOperationExecution.kt
```diff
@@ -16,11 +16,11 @@ import aws.smithy.kotlin.runtime.http.response.HttpResponse
 import aws.smithy.kotlin.runtime.http.response.complete
 import aws.smithy.kotlin.runtime.http.response.dumpResponse
 import aws.smithy.kotlin.runtime.io.*
-import aws.smithy.kotlin.runtime.io.middleware.MapRequest
 import aws.smithy.kotlin.runtime.io.middleware.Middleware
 import aws.smithy.kotlin.runtime.io.middleware.Phase
 import aws.smithy.kotlin.runtime.logging.Logger
 import aws.smithy.kotlin.runtime.util.InternalApi
+import aws.smithy.kotlin.runtime.util.get
 import kotlin.time.ExperimentalTime
```
ianbotsf, line 24: **Question**: This import is new but I don't see any altered code in this file using it. Is this unused? Or is this replacing some other `get` implementation that was being used before?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/Retry.kt
```diff
@@ -41,6 +53,9 @@ class Retry<O>(
                 // Deep copy the request because later middlewares (e.g., signing) mutate it
                 val requestCopy = request.deepCopy()
 
+                // we don't know max attempts, it comes from the strategy and setting ttl would never be accurate
+                // set attempt which is the only additional metadata we know
+                requestCopy.subject.header(AMZ_SDK_REQUEST_HEADER, "attempt=$attempt")
```
ianbotsf, line 58: **Comment**: We could expose this relatively simply:

```kotlin
interface RetryStrategy {
    val options: RetryOptions
    suspend fun <R> retry(policy: RetryPolicy<R>, block: suspend () -> R): Outcome<R>
}

interface RetryOptions {
    val maxAttempts: Int? = null
}

data class StandardRetryStrategyOptions(override val maxAttempts: Int) : RetryOptions {
    ...
}
```

Keeping the `RetryOptions` member null allows exposing a relatively common retry option without forcing retry strategies to use it internally (even though our _only_ retry strategy _does_ implement it).

What do you think about this?

ianbotsf (approved):

runtime/protocol/http-client-engines/http-client-engine-ktor/common/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngine.kt
```diff
@@ -117,18 +123,18 @@ class KtorEngine(
...
-            logger.trace("request done")
+            reqLogger.trace("request done")
         }
     }
 
-    override fun close() {
+    override fun shutdown() {
         client.close()
         engine.close()
     }
```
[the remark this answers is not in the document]
aajtodd: We shouldn't be overriding `close()`. `HttpClientEngineBase` implements common `close()` logic and waits for all `in-flight` requests to complete and then calls `shutdown()` which is the hook provided to implementations.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/operation/SdkOperationExecution.kt
aajtodd: Hmm yeah I'll take a look, weird if lint didn't catch it.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/Retry.kt
aajtodd: Yeah we could. It is called out in SEP that if the value is unknown or not reliable to omit it. Depends on how important we really think it is to include. My vote is not all that important (I almost left out `AMZ_SDK_REQUEST_HEADER` all together)

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/engine/HttpClientEngine.kt
```diff
@@ -15,20 +15,15 @@ import kotlinx.coroutines.*
...
 interface HttpClientEngine : Closeable, CoroutineScope {
     /**
      * Execute a single HTTP request and return the response.
      * Consumers *MUST* call `HttpCall.complete()` when finished processing the response
      */
-    suspend fun roundTrip(request: HttpRequest): HttpCall
-
-    /**
-     * Execute a single HTTP request and return the response
-     * Consumers *MUST* call `HttpCall.complete()` when finished processing the response
```
lucix-aws, line 29: Is consumer use of `complete()` no longer required?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/Retry.kt
ianbotsf: I'd like to include it if we could. It may provide useful information to services (IDK if any of them make use of it at this point).

lucix-aws (approved):

runtime/protocol/http-client-engines/http-client-engine-ktor/common/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorEngine.kt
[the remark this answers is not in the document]
aajtodd: Unknown if it was causing any issues but it was definitely not written as intended.

ianbotsf (approved):
Looks like a good separation of concerns.

PR #648, opened 2022-05-23
feat: update design of smithy Document type

lucix-aws:
## Description of changes
Update design / implementation for proposed `Document` type.


docs/design/document-type.md
```diff
@@ -14,346 +14,164 @@ such as `struct`, `union`, etc. which are handled by the [Kotlin Smithy SDK](kot
...
 
     /**
-     * Adds given [value] as [SmithyString] to the current [SmithyMap] with [this] as a key
+     * Implemented by structures that can be transformed into a serialized Document.
+     * Must serialize to JSON.
      */
-    infix fun String.to(value: String) {
-        require(content[this] == null) {"Key $this is already registered in builder"}
-        content[this] = value.toDocument()
+    interface Transformable {
```
aajtodd, line 103: I don't think we want to support this right now. We want as minimal surface area as possible. We may eventually support something like this but it would more than likely be either as a compiler plugin OR in one of the `serde` packages. We don't want to tie the type definition to any particular serialization.

docs/design/document-type.md
```diff
@@ -14,346 +14,164 @@ such as `struct`, `union`, etc. which are handled by the [Kotlin Smithy SDK](kot
...
-     * Adds [this] value to the current [SmithyArray] as [SmithyString]
-     */
-    operator fun String.unaryPlus() {
-        content.add(SmithyString(this))
+    companion object {
+        /**
+         * Overloads to create a document from possible values.
+         */
+        operator fun invoke(init: Builder.() -> Unit): Document
+        operator fun invoke(value: kotlin.Number): Document
```
aajtodd, line 40: these extra overloads for types with an explicit variant seem unnecessary to me.

```kotlin
val x = Document(12)
```

vs explicit

```kotlin
val x = Document.Number(12)
```

If we felt strongly about keeping them I'd remove these as `operator invoke` overloads to free "builder" functions outside of the class definition like:

```kotlin
sealed class Document { ... }

/**
 * Construct a document from a [Number]
 */
fun Document(value: kotlin.Number): Document = Document.Number(value)
...
```

In part because we should move anything unrelated to the core type outside of the type definition where possible but mostly because intellij will still suggest these functions to you but then insert things like `Document.invoke{...}` instead of what you want `Document(...)`

docs/design/document-type.md
```diff
@@ -14,346 +14,164 @@ such as `struct`, `union`, etc. which are handled by the [Kotlin Smithy SDK](kot
...
+        operator fun invoke(value: kotlin.String): Document
+        operator fun invoke(value: kotlin.Boolean): Document
+        operator fun invoke(value: Transformable): Document
+
+        /**
+         * Unwraps a Document from a serialized value.
+         */
+        fun fromString(value: kotlin.String): Document
```
aajtodd, line 48: we shouldn't support this here

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/smithy/Document.kt
```diff
@@ -4,161 +4,185 @@
...
+    operator fun get(i: kotlin.String) = asMap()[i]
 
     /**
-     * Returns [index] th element of an array as [SmithyMap] if the element is of that type or null if not.
-     *
-     * @throws IndexOutOfBoundsException if there is no element with given index
+     * DSL builder for a [Map]-based [Document].
      */
-    fun getMap(index: Int) = content[index] as? SmithyMap
+    class Builder internal constructor() {
```
aajtodd, line 150: Not a hill I'm going to die on but I think I liked the document builders separate (especially since we've moved the definition of each variant into the sealed class body making it harder to read)

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/smithy/Document.kt
```diff
@@ -4,161 +4,185 @@
...
+                value.doubleOrNull != null -> Number(value.double)
+                value.isString -> String(value.content)
+                value.booleanOrNull != null -> Boolean(value.boolean)
+                else -> throw IllegalArgumentException("json value $value could not be deserialized")
+            }
+
+        /**
+         * Creates a [List] from the provided values.
+         */
+        fun listOf(vararg values: Any?): Document =
```
aajtodd, line 63: I don't even think we need this.

If the builder has an infix function to a list like:
```kotlin
infix fun kotlin.String.to(value: List<Any?>)
```

then the builder can work like:
```kotlin
Document {
    "myList" to listOf(12, null, true)
}
```

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/smithy/Document.kt
```diff
@@ -4,161 +4,185 @@
...
+        override fun toString() = "null"
+    }
 
-    /**
-     * Returns [index] th element of an array as [SmithyBool] if the element is of that type or null if not.
-     *
-     * @throws IndexOutOfBoundsException if there is no element with given index
-     */
-    fun getBoolean(index: Int) = content[index] as? SmithyBool
+    fun asNumber() = (this as Number).value
```
aajtodd, line 129: all of these should have explicit return types (which reminds me we really need to turn that check on...)

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/smithy/Document.kt
```diff
@@ -4,161 +4,185 @@
...
-     * Returns [index] th element of an array as [SmithyArray] if the element is of that type or null if not.
-     *
-     * @throws IndexOutOfBoundsException if there is no element with given index
-     */
-    fun getArray(index: Int) = content[index] as? SmithyArray
+    val isNull: kotlin.Boolean
+        get() = this == Null
+
+    operator fun get(i: Int) = asList()[i]
+    operator fun get(i: kotlin.String) = asMap()[i]
```
aajtodd, line 145: I don't know how I feel about all of these. We have casts here that can obviously fail. 

This needs further discussion I think.

lucix-aws: Isn't that just kind of a part-in-parcel risk of data being unstructured like this?

We could add the `getOrNull`-style methods as well.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/smithy/Document.kt
lucix-aws: I actually had that at first, but then you can't construct a Document outside of the builder directly as an array eg.

```kotlin
val Document = Document.listOf(1, null, "blah")
```

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/smithy/Document.kt
lucix-aws: This was a question of consistency for me - all of our request objects localize the builder within the class itself.

aajtodd: Our generated code doesn't have the luxury of being hand written though. The choices we make for generated code deal with a lot of other concerns. 

Like I said this isn't a huge deal so if you feel strongly I'm fine with it here.

docs/design/document-type.md
```diff
@@ -14,346 +14,164 @@ such as `struct`, `union`, etc. which are handled by the [Kotlin Smithy SDK](kot
...
-        content.add(SmithyBool(this))
-    }
-
-    /**
-     * Adds [this] value to the current [SmithyArray] as [Document]
-     */
-    operator fun Document.unaryPlus() {
-        content.add(this)
-    }
+    data class Number(val value: kotlin.Number) : Document()
```
ianbotsf, line 59: **Question**: Is accepting `Number` unsafe? It's not a sealed interface so 3P classes can implement it. Moreover, if/when Kotlin adds 1P `BigInteger` and `BigDecimal` implementations, they will likely inherit `Number` just like [the Java equivalents do](https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html).

docs/design/document-type.md
```diff
@@ -14,346 +14,164 @@ such as `struct`, `union`, etc. which are handled by the [Kotlin Smithy SDK](kot
...
-    builder.init()
-    return SmithyArray(builder.content)
-}
-
-/**
- * DSL builder for a [Document] as a [SmithyMap]
- */
-class DocumentBuilder internal constructor() {
-    internal val content: MutableMap<String, Document> = linkedMapOf()
+    fun asNumber() = (this as Number).value
```
ianbotsf, line 69: **Question**: Why would we need a public `asNumber()` converter if we have `asByte()`, `asInt()`, etc. which all return types that extend `Number`?

docs/design/document-type.md
```diff
@@ -14,346 +14,164 @@ such as `struct`, `union`, etc. which are handled by the [Kotlin Smithy SDK](kot
...
+
+        /**
+         * Unwraps a Document from a serialized value.
+         */
+        fun fromString(value: kotlin.String): Document
+        
+        /**
+         * Creates a Document list from the provided values.
+         */
+        fun listOf(vararg values: Any?): Document
```
ianbotsf, line 53: **Comment**: It's a shame we can't provide a narrower parameter list than this. If Kotlin ever supports union types this might be good place to use them.

docs/design/document-type.md
```diff
@@ -14,346 +14,164 @@ such as `struct`, `union`, etc. which are handled by the [Kotlin Smithy SDK](kot
...
+            "quux" to Document.listOf(
+                202L,
+                12,
+                false,
+                true,
+                "blah",
+                null,
+                Document.listOf(1, null, 2)
+            )
         }
```
ianbotsf, line 138: **Question**: This feels verbose. Does it hamper readability too much to use shorter names like `doc` and `arr`?

```kotlin
"baz" to doc {
    "quux" to arr(
        202L,
        12,
        false,
        true,
        "blah",
        null,
        arr(1, null, 2),
    )
}
```

docs/design/document-type.md
aajtodd: It's not a sealed class but it's also [not an open class](https://github.com/JetBrains/kotlin/blob/6a670dc5f38fc73eb01d754d8f7c158ae0176ceb/core/builtins/native/kotlin/Number.kt#L22)

docs/design/document-type.md
lucix-aws: Theoretically we could, though not quite in the same saw - I was looking at kotlinx serialization and this is a sample of their DSL builder for json objects:

```kotlin
val json = buildJsonObject {
    put("booleanKey", true)
    putJsonArray("arrayKey") {
        for (i in 1..10) add(i)
    }
    putJsonObject("objectKey") {
        put("stringKey", "stringValue")
    }
}
```
If we do it this way, all of the builder members could be defined as typed overloads (add number, add string, etc) and I'd still consider it at least somewhat "idiomatic" considering we're following how kotlinx does a thing.

docs/design/document-type.md
lucix-aws: `Document` was to mimic how our codegen request structures work ie.

```kotlin
val request = ListObjectsRequest {
   // ...
   foo = "blah"
}
```

Don't feel strongly about it. kotlinx serialization builder, commented above, doesn't do it that way at all.

docs/design/document-type.md
lucix-aws: But to Ian's point a caller can definitely extend it and shove in some bogus value, as long as they implement all the `toWhateverNumericForm()` methods.

We could instead make the constructor private, and override `invoke` etc. to take the specific subclasses.

aajtodd: What am I missing here? If `Number` can't be extended from I am not seeing the concern.

lucix-aws: It can be though, that's what I was saying--

```kotlin
class Number2: Number() {
    override fun toByte(): Byte {
        TODO("Not yet implemented")
    }

    override fun toChar(): Char {
        TODO("Not yet implemented")
    }

    override fun toDouble(): Double {
        TODO("Not yet implemented")
    }

    override fun toFloat(): Float {
        TODO("Not yet implemented")
    }

    override fun toInt(): Int {
        TODO("Not yet implemented")
    }

    override fun toLong(): Long {
        TODO("Not yet implemented")
    }

    override fun toShort(): Short {
        TODO("Not yet implemented")
    }
}

val invalidDoc = Document.Number(Number2()) // ???
```

aajtodd: oh duh, abstract classes are open. I think I need to take a break for the day.

aajtodd: I definitely see Ian's point. However, I think this is overall a minor concern. Worse case we would fail serialization because we can't convert it into a builtin type.

docs/design/document-type.md
```diff
@@ -1,7 +1,7 @@
 # Document type
 
 * **Type**: Design
-* **Author(s)**: Aaron Todd
+* **Author(s)**: Aaron Todd, Luc Talatinian
 
 ## Work in progress
```
aajtodd, line 6: We can remove this WIP note

docs/design/document-type.md
```diff
@@ -14,346 +14,94 @@ such as `struct`, `union`, etc. which are handled by the [Kotlin Smithy SDK](kot
 
 # Design
 
-The `document` type is an untyped JSON-like value that can take on the following types: null, boolean, string, byte,
-short, integer, long, float, double, an array of these types, or a map of these types where the key is string.
+The `document` type is an untyped JSON-like value that can take on the following types:
+* null
+* boolean
+* string
+* number (byte, short, integer, long, float, double)
```
aajtodd, line 21: We should clarify arbitrary precision number is called out in spec and in scope.

docs/design/document-type.md
```diff
@@ -14,346 +14,94 @@ such as `struct`, `union`, etc. which are handled by the [Kotlin Smithy SDK](kot
...
 This type is best represented as a sum type (sealed class is closest we can get in Kotlin). See the
 [JSON type](https://github.com/Kotlin/kotlinx.serialization/blob/master/runtime/commonMain/src/kotlinx/serialization/json/JsonElement.kt)
-from the kotlinx.serialization lib for an example on which the following is derived. We should provide our own type but
-we may be able to internally deal with serialization by going through the one from the kotlinx.serialization library.
+from the kotlinx.serialization lib for an example on which the following is derived.
 
-````kotlin
-package com.amazonaws.smithy.runtime
+# Document
+Document directly implements the subclasses that wrap all the possible values the type can hold.
```
aajtodd, line 30: The wording here is confusing. `Document` doesn't implement anything. 

Maybe something like:

> The `Document` shape will be represented as a sealed class in Kotlin with each variant (subclass) representing one of the possible values the type can hold.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/smithy/Document.kt
```diff
@@ -5,160 +5,120 @@
...
-            separator = ",",
-            prefix = "{",
-            postfix = "}",
-            transform = { (k, v) -> """"$k":$v""" }
-        )
-}
+/**
+ * Construct a [Document] from a [List].
+ */
+fun Document(value: List<Document>): Document = Document.List(value)
```
aajtodd, line 119: these constructor functions could probably all be `inline`

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/smithy/DocumentBuilder.kt
```diff
@@ -0,0 +1,67 @@
...
+        require(content[this] == null) { "Key $this is already registered in builder" }
+        content[this] = Document(value)
+    }
+
+    infix fun String.to(value: Document?) {
+        require(content[this] == null) { "Key $this is already registered in builder" }
+        content[this] = value ?: Document.Null
+    }
+
+    class ListBuilder internal constructor() {
```
aajtodd, line 30: Question: should we include an infix fun to `List<Document>`?

That would allow someone to do 

```kotlin
buildDocument {
    "foo" to listOf(1, 2, 3).map { Document(it) }
}
```

aajtodd (approved):
Overall looks great. Couple questions/suggestions but nothing blocking.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/smithy/DocumentBuilder.kt
ianbotsf: If we provide that, would it also make sense to provide convenience mappers to specifically-typed lists like `List<Int?>`, `List<Boolean?>`, etc.? I have a gut feeling that _most_ of the time people are dealing with lists, they won't be heterogeneously-typed...they'll be (possibly-nullable) singly-typed lists.

docs/design/document-type.md
```diff
@@ -24,150 +24,80 @@ The `document` type is an untyped JSON-like value that can take on the following
...
-     */
     data class Number(val value: kotlin.Number) : Document()
     data class String(val value: kotlin.String) : Document()
     data class Boolean(val value: kotlin.Boolean) : Document()
-    data class List(val value: kotlin.collections.List<Document>) : Document()
-    data class Map(val value: kotlin.collections.Map<kotlin.String, Document>) : Document()
+    data class List(val value: kotlin.collections.List<Document>) :
+        Document(), kotlin.collections.List<Document> by value
+    data class Map(val value: kotlin.collections.Map<kotlin.String, Document>) :
+        Document(), kotlin.collections.Map<kotlin.String, Document> by value
```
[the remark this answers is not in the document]
aajtodd: You definitely would have to fully qualify it in the actual code or else the type `List` refers to `Document.List` in this scope. Same for all the other types that have a builtin of the same name (which is everything)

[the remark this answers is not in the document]
ianbotsf: Oh of course. 🤦‍♂️ The answer is right in front of me. Nevermind!

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/smithy/DocumentBuilder.kt
lucix-aws: I would say no, given that we're adding `addAll` to the list builder.

ianbotsf (approved):

PR #656, opened 2022-06-03
bind default HTTP engine to okhttp directly

### LINKED ISSUE #629: Bind to OkHttp Engine Directly

aajtodd:
## Issue \#

closes #629 

## Description of changes

The default engine is already based on okhttp but goes through ktor APIs. This PR removes the ktor middleman and binds directly to okhttp. 

There are a few reasons we want to go this route:
1. Support for duplex streaming (required for event streams) isn't exposed in ktor
2. Trailers aren't supported by ktor (although okhttp support for this may be incomplete)
3. We can remove a layer of indirection and buffering (not to mention additional coroutines used by ktor)
4. okhttp has support for custom [dns](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-dns/) resolver
5. okhttp is working on KMP support (which may or may not pan out but it would give us 1 API to consider potentially)

NOTE: Duplex streaming isn't supported in this PR, this is the bare minimum to bind directly with existing functionality.

NOTE: I have grabbed the latest version of okhttp in this PR which is marked as "alpha". The [changelog](https://github.com/square/okhttp/blob/master/CHANGELOG.md) indicates:
> The alpha releases in the 5.0.0 series have production-quality code and an unstable API. We expect to make changes to the APIs introduced in 5.0.0-alpha.X. These releases are safe for production use and 'alpha' strictly signals that we're still experimenting with some new APIs. If you're eager for the fixes or features below, please upgrade.

Since this is all internal I'm inclined to just keep moving forward and take upgrades to this dependency as they come. We can discuss if people disagree with this take.


runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/test/aws/smithy/kotlin/runtime/http/engine/okhttp/ByteChannelRequestBodyTest.kt
```diff
@@ -0,0 +1,121 @@
...
+        val actual = ByteChannelRequestBody(body, callContext)
+
+        assertEquals(body.contentLength, actual.contentLength())
+        assertFalse(actual.isDuplex())
+        assertTrue(actual.isOneShot())
+
+        val buffer = Buffer()
+        actual.writeTo(buffer)
+        val actualSha256 = buffer.sha256().hex()
+        assertEquals(expectedSha256, actualSha256)
```
ianbotsf, line 46: **Question**: Why compare the hashes? Why not just compare the bytes themselves?

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/HttpEngineEventListener.kt
```diff
@@ -2,30 +2,31 @@
...
-
 internal class HttpEngineEventListener(
     private val pool: ConnectionPool
 ) : EventListener() {
     private val logger = Logger.getLogger<HttpEngineEventListener>()
 
     private inline fun traceCall(call: Call, crossinline msg: () -> Any) {
-        val sdkRequestId = call.request().header(AMZ_SDK_INVOCATION_ID_HEADER)
+        val sdkRequestTag = call.request().tag<SdkRequestTag>()
+        val sdkRequestId = sdkRequestTag?.execContext?.getOrNull(HttpOperationContext.SdkRequestId)
```
ianbotsf, line 23: **Question**: What does it mean if any of these are null and we cannot derive the `sdkRequestId`?

ianbotsf (approved):

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/test/aws/smithy/kotlin/runtime/http/engine/okhttp/ByteChannelRequestBodyTest.kt
aajtodd: That would work but this matches the rest of our "integrity" style checks scattered in the runtime tests. It also encapsulates two checks in one, are they the same length and are the contents the same?

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/HttpEngineEventListener.kt
aajtodd: As in what will happen? Nothing, the log statement will have `[sdkRequestId=null]`. In practice this should never happen in an actual generated SDK because we guarantee that element to be present at the time the request is constructed. The reason I've relaxed it to `getOrNull` is that many of the unit tests here and in the integration test suite don't include this additional `ExecutionContext` setup.

lucix-aws (approved):

PR #662, opened 2022-06-15
HTTP Engine Benchmarks

aajtodd:

## Issue \#
N/A

## Description of changes
<!--- Why is this change required? What problem does it solve? -->
* Bootstrap a benchmark suite for HTTP client engines. This is a very rough test suite that we can build from.

I'm currently seeing the following two errors when running the benchmarks:
1. 
```
failed to consume body: failed to read all HttpBody bytes from stream: isClosedForRead: false/true; isClosedForWrite: true/true; availableForRead: 0/0: hello    
```
This only happens on the `KtorEngine` and `OkHttpEngine` which both use the same `SdkByteReadChannel` implementation (backed by ktor) whereas the `CRT` engine uses the custom implementation.

2.
```
stopping server                                                                                                                                                         Jun 15, 2022 11:07:11 AM io.netty.channel.AbstractChannelHandlerContext invokeExceptionCaught                                                                           WARNING: Failed to submit an exceptionCaught() event.                                                                                                                   java.util.concurrent.RejectedExecutionException: event executor terminated                 
        at io.netty.util.concurrent.SingleThreadEventExecutor.reject(SingleThreadEventExecutor.java:932)                                                                        at io.netty.util.concurrent.SingleThreadEventExecutor.offerTask(SingleThreadEventExecutor.java:351)                                                                     at io.netty.util.concurrent.SingleThreadEventExecutor.addTask(SingleThreadEventExecutor.java:344)                                                                       at io.netty.util.concurrent.SingleThreadEventExecutor.execute(SingleThreadEventExecutor.java:834)                                                                       at io.netty.util.concurrent.SingleThreadEventExecutor.execute0(SingleThreadEventExecutor.java:825)    
        at io.netty.util.concurrent.SingleThreadEventExecutor.execute(SingleThreadEventExecutor.java:815)                                                                       at io.netty.channel.AbstractChannelHandlerContext.invokeExceptionCaught(AbstractChannelHandlerContext.java:284)                                                         at io.netty.channel.AbstractChannelHandlerContext.fireExceptionCaught(AbstractChannelHandlerContext.java:273)                                                           at io.netty.channel.ChannelInboundHandlerAdapter.exceptionCaught(ChannelInboundHandlerAdapter.java:143)
        at io.netty.channel.AbstractChannelHandlerContext.invokeExceptionCaught(AbstractChannelHandlerContext.java:302)                                                         at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:264)
        at io.netty.channel.AbstractChannelHandlerContext.access$300(AbstractChannelHandlerContext.java:61)                                                             
        at io.netty.channel.AbstractChannelHandlerContext$4.run(AbstractChannelHandlerContext.java:253)                                                               
        at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:174)                                                                     
        at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:167)                                                                   
        at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasksFrom(SingleThreadEventExecutor.java:426)                                                       
        at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:375)          
        at io.netty.util.concurrent.SingleThreadEventExecutor.confirmShutdown(SingleThreadEventExecutor.java:761)                                                       
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:530)                                                                                               
        at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:995)                
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)                                                                                    
        at io.ktor.server.netty.EventLoopGroupProxy$Companion.create$lambda-1$lambda-0(NettyApplicationEngine.kt:260)                                                           at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)                                                                                at java.base/java.lang.Thread.run(Thread.java:829)                                                                                                                                                                                                                         
```

The second one (2) is likely just not shutting down the server cleanly/correctly but I haven't found the right knob yet. It doesn't happen everytime or on every run and doesn't seem to be affecting anything else. 

The first one (1) is the one that worries me but I haven't been able to reproduce it outside of the benchmark test suite. It is a pre-existing condition. I've updated the `check` to include additional state in the event it does happen in production. We probably want to revisit this API and not limit it to `Int.MAX_VALUE` bytes but rather `Long.MAX_VALUE`. This would necessitate returning a different type other than `ByteArray` though which is [limited](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-byte-array/-init-.html) to `Int.MAX_VALUE`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -136,7 +136,7 @@ dependencies {
 val lintPaths = listOf(
     "smithy-kotlin-codegen/src/**/*.kt",
     "runtime/**/*.kt",
-    "benchmarks/**/jvm/*.kt",
+    "tests/**/**/jvm/*.kt",
 )
 
 tasks.register<JavaExec>("ktlint") {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -136,7 +136,7 @@ dependencies {
 val lintPaths = listOf(
     "smithy-kotlin-codegen/src/**/*.kt",
     "runtime/**/*.kt",
-    "tests/**/**/jvm/*.kt",
+    "tests/**/jvm/**/*.kt",
 )
 
 tasks.register<JavaExec>("ktlint") {
--- a/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
@@ -8,7 +8,6 @@ package aws.smithy.kotlin.runtime.http.engine.okhttp
 import aws.smithy.kotlin.runtime.client.ExecutionContext
 import aws.smithy.kotlin.runtime.http.engine.AlpnId
 import aws.smithy.kotlin.runtime.http.engine.HttpClientEngineBase
-import aws.smithy.kotlin.runtime.http.engine.HttpClientEngineConfig
 import aws.smithy.kotlin.runtime.http.engine.callContext
 import aws.smithy.kotlin.runtime.http.request.HttpRequest
 import aws.smithy.kotlin.runtime.http.response.HttpCall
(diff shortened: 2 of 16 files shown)


runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
```diff
@@ -23,12 +22,49 @@ import kotlin.time.toJavaDuration
...
-    // TODO - expose thread count and use custom dispatcher/ExecutionService
+    public companion object {
+        public operator fun invoke(block: OkHttpEngineConfig.Builder.() -> Unit): OkHttpEngine = OkHttpEngine(
+            OkHttpEngineConfig.Builder().apply(block).build()
+        )
+    }
+
+    private val client = config.buildClient()
+
+    override suspend fun roundTrip(context: ExecutionContext, request: HttpRequest): HttpCall {
```
aajtodd, line 37: none of this changed, the diff is just confusing. I moved the construction of the client on L54 into a method `buildClient()`

tests/benchmarks/http-benchmarks/README.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+Baseline `0.10.3-SNAPSHOT` on EC2 **[m5.4xlarge](https://aws.amazon.com/ec2/instance-types/m5/)** with **Corretto-11.0.15.9.1**:
+
+The download/upload throughput benchmarks are an approximation of how much data in MB/s we are able to process.
+
+```
+jvm summary:
+Benchmark                                      (httpClientName)   Mode  Cnt      Score     Error  Units
+HttpEngineBenchmarks.downloadThroughputNoTls             OkHttp  thrpt    5    129.300 ±   1.796  ops/s
```
aajtodd, line 18: I haven't been able to explain the difference between `OkHttp` download throughput and `Ktor_OkHttp`. Theoretically these should be much closer given they are the same underlying engine but on the m5 instance they are not. Locally when I run the benchmark they are closer together (with `OkHttp` sometimes barely beating `Ktor_OkHttp`). 


Also FYI the benchmark throughput here is a rough estimation of `MB/s` but I can't figure out how to change the Units in `jmh`/`kotlinx-benchmark`.

tests/benchmarks/http-benchmarks/jvm/src/aws/smithy/kotlin/benchmarks/http/HttpEngineBenchmarks.kt
```diff
@@ -0,0 +1,209 @@
...
+
+    /**
+     * Sequential requests raw throughput
+     */
+    @Benchmark
+    fun roundTripSequentialNoTls(blackhole: Blackhole) = runBlocking {
+        val call = httpClient.call(helloRequest)
+        try {
+            val body = call.response.body.readAll()
+            blackhole.consume(body)
```
ianbotsf, line 146: **Comment**: TIL

build.gradle.kts
```diff
@@ -136,7 +136,7 @@ dependencies {
 val lintPaths = listOf(
     "smithy-kotlin-codegen/src/**/*.kt",
     "runtime/**/*.kt",
-    "benchmarks/**/jvm/*.kt",
+    "tests/**/**/jvm/*.kt",
```
ianbotsf, line 139: **Comment**: Looks like this was incorrect before and it's still incorrect now. We don't need an extra `**` before `jvm` (because `**/**` is equivalent to `**`) but we _do_ need a `**` between `jvm` and `*.kt`:

```
"tests/**/jvm/**/*.kt",
```

Changing this picks up several new problems in **HttpEngineBenchmarks** and some existing ones in **AwsSignerBenchmark**.

ianbotsf (approved):
Minor ktlint problems. Fix and ship!

build.gradle.kts
aajtodd: How embarrassing, good catch I'll fix it up.

lucix-aws (approved):

PR #664, opened 2022-06-16
feat!: use kotlin nullability for null documents instead of explicit subclass

lucix-aws:
## Description of changes
Remove `Document.Null` and instead use kotlin nullability to represent null documents.

### DIFF
--- a/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/smithy/DocumentBuilder.kt
+++ b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/smithy/DocumentBuilder.kt
@@ -10,27 +10,27 @@ class DocumentBuilder internal constructor() {
     val content: MutableMap<String, Document?> = linkedMapOf()
 
     infix fun String.to(value: Number?) {
-        require(!content.containsKey(this)) { "Key $this is already registered in builder" }
+        require(this !in content) { "Key $this is already registered in builder" }
         content[this] = if (value != null) Document(value) else null
     }
 
     infix fun String.to(value: String?) {
-        require(!content.containsKey(this)) { "Key $this is already registered in builder" }
+        require(this !in content) { "Key $this is already registered in builder" }
         content[this] = if (value != null) Document(value) else null
     }
 
     infix fun String.to(value: Boolean?) {
-        require(!content.containsKey(this)) { "Key $this is already registered in builder" }
+        require(this !in content) { "Key $this is already registered in builder" }
         content[this] = if (value != null) Document(value) else null
     }
 
     infix fun String.to(value: Document?) {
-        require(!content.containsKey(this)) { "Key $this is already registered in builder" }
+        require(this !in content) { "Key $this is already registered in builder" }
         content[this] = value
     }
 
     infix fun String.to(@Suppress("UNUSED_PARAMETER") value: Nothing?) {
-        require(!content.containsKey(this)) { "Key $this is already registered in builder" }
+        require(this !in content) { "Key $this is already registered in builder" }
         content[this] = null
     }
 
--- a/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/smithy/DocumentBuilder.kt
+++ b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/smithy/DocumentBuilder.kt
@@ -7,46 +7,46 @@ package aws.smithy.kotlin.runtime.smithy
 import kotlin.jvm.JvmName
 
 class DocumentBuilder internal constructor() {
-    val content: MutableMap<String, Document> = linkedMapOf()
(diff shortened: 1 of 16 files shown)


runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/Deserializer.kt
```diff
@@ -198,7 +198,7 @@ interface PrimitiveDeserializer {
      * If the document's value is a list or map, this method will deserialize all elements or fields recursively - the
      * caller need not further inspect the value to attempt to do so manually.
      */
-    fun deserializeDocument(): Document
+    fun deserializeDocument(): Document?
```
aajtodd, line 201: I don't _think_ we need the `?` qualifier. An explicit `null` is the same as not sending it in the payload as far as serialization and deserialization is concerned. JSON for example [won't even report `null` values](https://github.com/awslabs/smithy-kotlin/blob/2547699c13857f5dbd57447ff5e2bdf0308aeefa/runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonDeserializer.kt#L195).

aajtodd: Hmm I guess maybe this is due to the internal nullability of a document? In which case you _could_ make the deserializer interface non-nullable return and delegate intrernally to a function that allows it to be `null`

aajtodd (commented):
Looks good. One question on nullability of the exposed serialize/deserialize interfaces

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/Deserializer.kt
lucix-aws: Can you clarify your objection to it being nullable? It's done that way because of internal Document nullability as you suspected.

It certainly can be split out like you said, but it'll just bulk up and duplicate some code (you'll basically have this deserialize switch twice), and to avoid what I'm not quite sure.

aajtodd: It's different than how the rest of the deserializer interface works. 

We only ever call `deserialize<Type>` when we have found a matching field:


```kotlin
deserializer.deserializeStruct(OBJ_DESCRIPTOR) {
    loop@while (true) {
        when (findNextFieldIndex()) {
            MY_DOCUMENT_MEMBER.index -> builder.myDocumentMember = deserializeDocument()
            ...
        }
    }
}
```

`deserializeDocument()` will only ever be called if the value associated with the key represented by `MY_DOCUMENT_MEMBER` is non-null. e.g.

```json
{
     "myDocumentMember": "foo"
}
```

whereas: 
```json
{
     "myDocumentMember":  null
}
```

will never execute that branch.


It's maybe a minor thing overall I'm just questioning whether we should codify that difference. Internally a document may be null so I would expect _if_ we wanted the top level deserializer function to return a non-null document it could delegate like so:

```
fun deserializeDocument(): Document = checkNotNull(deserializeDocumentImpl()) { "expected non-null document field" }

private fun deserializeDocumentImpl(): Document? { ... what you have now basically }
```

ianbotsf: As long as our existing deserializer logic correctly shortcuts explicitly-null member and absent member (i.e., implicitly-null) then I favor using the same semantics as other methods. That is, `deserializeDocument(): Document` (non-null).

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/smithy/DocumentBuilder.kt
```diff
@@ -7,46 +7,46 @@ package aws.smithy.kotlin.runtime.smithy
 import kotlin.jvm.JvmName
 
 class DocumentBuilder internal constructor() {
-    val content: MutableMap<String, Document> = linkedMapOf()
+    val content: MutableMap<String, Document?> = linkedMapOf()
 
     infix fun String.to(value: Number?) {
-        require(content[this] == null) { "Key $this is already registered in builder" }
-        content[this] = if (value != null) Document(value) else Document.Null
+        require(!content.containsKey(this)) { "Key $this is already registered in builder" }
```
ianbotsf, line 13: **Style**: Kotlin provides the `in` and `!in` operators for `contains` checks. This might be more idiomatic as:

```kotlin
require(this !in content) { "..." }
```

ianbotsf (approved):
These code changes look good but let's also update [the design doc](https://github.com/awslabs/smithy-kotlin/blob/main/docs/design/document-type.md) as part of this push.

ianbotsf (changes requested):
Actually missed one critical thing: since this is a breaking change we must update the version in **gradle.properties**.

aajtodd (approved):

PR #666, opened 2022-06-20
fix: validate that members bound to URI paths are non-null at object construction

### LINKED ISSUE #139: Validate members bound to URI
For required inputs we should consider if we should be validating them or not.

e.g. URI labels are always marked `required` but we don't currently do any validation on them to verify they aren't null

They are inlined directly into the URI without validation:
e.g
```
path = "/2018-10-31/layers/${input.layerName}/versions/${input.versionNumber}/policy"
```

We could be doing this as part of construction of the type (potentially). 

*internal id*: 176997800

ianbotsf:
## Issue \#

Closes #139 

## Description of changes

This change adds validation that URI-bound members are set to non-null values at object construction time. This prevents malformed requests from being sent to services with `null` serialized into URIs.

Companion PR: (link coming soon)

### DIFF
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolUnitTestResponseGenerator.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolUnitTestResponseGenerator.kt
@@ -6,10 +6,12 @@ package software.amazon.smithy.kotlin.codegen.rendering.protocol
 
 import software.amazon.smithy.codegen.core.Symbol
 import software.amazon.smithy.kotlin.codegen.core.*
+import software.amazon.smithy.kotlin.codegen.model.defaultUnboxedValue
 import software.amazon.smithy.kotlin.codegen.model.hasStreamingMember
 import software.amazon.smithy.kotlin.codegen.model.hasTrait
 import software.amazon.smithy.kotlin.codegen.rendering.ShapeValueGenerator
 import software.amazon.smithy.model.shapes.*
+import software.amazon.smithy.model.traits.HttpLabelTrait
 import software.amazon.smithy.model.traits.StreamingTrait
 import software.amazon.smithy.protocoltests.traits.HttpResponseTestCase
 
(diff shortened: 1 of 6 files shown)


smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolUnitTestResponseGenerator.kt
```diff
@@ -80,8 +82,15 @@ open class HttpProtocolUnitTestResponseGenerator protected constructor(builder:
                     val inputShape = model.expectShape(it)
                     val inputSymbol = symbolProvider.toSymbol(inputShape)
 
+                    val labelMembers = inputShape.members().filter { it.hasTrait<HttpLabelTrait>() }
+
                     // invoke the DSL builder for the input type
-                    writer.write("val input = ${inputSymbol.name}{}")
+                    writer.withBlock("val input = #T {", "}", inputSymbol) {
+                        labelMembers.forEach { memberShape ->
```
aajtodd, line 89: Question: I'm assuming this is because some `httpResponseTest` fails without it?

It looks like we set default values for string only, [httpLabel](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httplabel-trait) can target:

> structure > member[trait|required] :test(> :test(string, number, boolean, timestamp)

Wonder if we should nix setting a default value for string and just handle the possible cases inline here?

aajtodd (approved):
Looks good overall, nothing blocking.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolUnitTestResponseGenerator.kt
ianbotsf: We already provided default values for [all number types](https://github.com/awslabs/smithy-kotlin/blob/main/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt#L67-L82) and [boolean](https://github.com/awslabs/smithy-kotlin/blob/main/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt#L104-L105). We didn't previously set defaults for strings or timestamps and so I added one for string to get the tests to pass. I'm not sure what a sensible default would be for timestamp so I left it blank.

We certainly could handle the param values inline in the test generator but I wonder what else we plan to use default values for if not something like this.

ianbotsf: After more discussion we decided that the global `Symbol.defaultValue` wasn't the right place for this, especially since there's not really a valid default for strings or timestamps. Given that the _need_ for a default is local to the test generator, I'll move the _specification_ of those defaults there as well. Reverting changes to **SymbolExt.kt** and **KotlinSymbolProvider.kt**...

PR #670, opened 2022-06-30
feat(rt): add support for HTTP_REQUEST_EVENT signature type

aajtodd:
## Issue \#

aws-sdk-kotlin#543
sibling PR: https://github.com/awslabs/aws-sdk-kotlin/pull/644

## Description of changes

Add support for `HTTP_REQUEST_EVENT` signature type. This works the same as chunk signing and only varies in the how the canonical string is formed. `HTTP_REQUEST_CHUNK` uses the sha-256 of an empty body whereas `HTTP_REQUEST_EVENT` uses the sha-256 of the `:date` event stream header.

NOTE: I have duplicated the logic for the binary representation of this header here to avoid a direct dependency on `aws-event-stream` package from `aws-sdk-kotlin`. It's unclear if there are other scenarios or considerations here ([CRT seems to think so](https://github.com/awslabs/aws-c-auth/blob/f29ffca89dc0b6a1d722ae9bacb230ec9dc765f4/source/signing_config.c#L59-L60)).

### DIFF
--- a/runtime/auth/aws-signing-default/common/test/aws/smithy/kotlin/runtime/auth/awssigning/DefaultSignatureCalculatorTest.kt
+++ b/runtime/auth/aws-signing-default/common/test/aws/smithy/kotlin/runtime/auth/awssigning/DefaultSignatureCalculatorTest.kt
@@ -5,7 +5,9 @@
 package aws.smithy.kotlin.runtime.auth.awssigning
 
 import aws.smithy.kotlin.runtime.auth.awscredentials.Credentials
+import aws.smithy.kotlin.runtime.auth.awssigning.tests.asStaticProvider
 import aws.smithy.kotlin.runtime.auth.awssigning.tests.testCredentialsProvider
+import aws.smithy.kotlin.runtime.hashing.sha256
 import aws.smithy.kotlin.runtime.time.Instant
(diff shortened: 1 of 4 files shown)


lucix-aws (approved):

runtime/auth/aws-signing-default/common/test/aws/smithy/kotlin/runtime/auth/awssigning/DefaultSignatureCalculatorTest.kt
```diff
@@ -80,4 +82,44 @@ class DefaultSignatureCalculatorTest {
...
+            )
+        )
+
+        val epoch = Instant.fromEpochSeconds(123_456_789L, 1234)
+        val prevSignature = "last message sts".encodeToByteArray().sha256().encodeToHex().encodeToByteArray()
+        val chunkBody = "test payload".encodeToByteArray()
+
+        for (test in tests) {
+            val config = AwsSigningConfig {
+                credentialsProvider = Credentials("fake access key", "fake secret key").asStaticProvider()
```
ianbotsf, line 106: **Question**: You don't need these credentials for anything, do you? If not, you can just use `testCredentialsProvider`.

ianbotsf (approved):

PR #671, opened 2022-06-30
feat: add support for smithy document type

lucix-aws:
## Issue \#
(https://github.com/awslabs/smithy-kotlin/issues/123)

## Description of changes
Wire in codegen for `Document` serde.

### DIFF
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
@@ -138,7 +138,7 @@ class ShapeValueGenerator(
     ) : NodeVisitor<Unit> {
 
         override fun objectNode(node: ObjectNode) {
-            if (currShape.type === ShapeType.DOCUMENT) {
+            if (currShape.type == ShapeType.DOCUMENT) {
                 writer
                     .writeInline("#T {\n", RuntimeTypes.Core.Smithy.buildDocument)
                     .indent()
@@ -229,22 +229,16 @@ class ShapeValueGenerator(
         override fun arrayNode(node: ArrayNode) {
             when (currShape.type) {
                 ShapeType.DOCUMENT -> {
-                    writer
-                        .writeInline("#T(\n", RuntimeTypes.Core.Smithy.Document)
-                        .indent()
-                        .writeInline("listOf(\n")
-                        .indent()
-
-                    node.elements.forEach {
-                        generator.writeShapeValueInline(writer, currShape, it)
-                        writer.writeInline(",\n")
+                    writer.withInlineBlock("#T(", ")", RuntimeTypes.Core.Smithy.Document) {
+                        writer.withInlineBlock("listOf(", ")") {
+                            node.elements.forEachIndexed { index, it ->
+                                generator.writeShapeValueInline(writer, currShape, it)
+                                if (index < node.elements.size - 1) {
+                                    writer.writeInline(",\n")
+                                }
+                            }
+                        }
                     }
-
-                    writer
-                        .dedent()
-                        .writeInline(")\n")
-                        .dedent()
-                        .writeInline(")")
                 }
 
                 else -> {
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
@@ -231,11 +231,9 @@ class ShapeValueGenerator(
                 ShapeType.DOCUMENT -> {
                     writer.withInlineBlock("#T(", ")", RuntimeTypes.Core.Smithy.Document) {
                         writer.withInlineBlock("listOf(", ")") {
-                            node.elements.forEachIndexed { index, it ->
+                            node.elements.forEach {
                                 generator.writeShapeValueInline(writer, currShape, it)
-                                if (index < node.elements.size - 1) {
-                                    writer.writeInline(",\n")
-                                }
+                                writer.writeInline(",\n")
                             }
                         }
                     }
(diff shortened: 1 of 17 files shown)


aajtodd (commented):
I wish we could improve the `ShapeValueGenerator` some but the ugliness there predates this PR.

Looks good overall, one suggestion/improvement to look into.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
```diff
@@ -138,6 +138,12 @@ class ShapeValueGenerator(
     ) : NodeVisitor<Unit> {
 
         override fun objectNode(node: ObjectNode) {
+            if (currShape.type === ShapeType.DOCUMENT) {
```
ianbotsf, line 141: **Question**: Why `===`? Won't `==` be sufficient here?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
```diff
@@ -211,14 +227,37 @@ class ShapeValueGenerator(
...
+                    node.elements.forEach {
+                        generator.writeShapeValueInline(writer, currShape, it)
+                        writer.writeInline(",\n")
+                    }
+
+                    writer
+                        .dedent()
+                        .writeInline(")\n")
+                        .dedent()
+                        .writeInline(")")
```
ianbotsf, line 247: **Style**: The manual indenting/dedenting is unnecessary here:

```kotlin
writer.withBlock("#T(", ")", RuntimeTypes.Core.Smithy.Document) {
    withBlock("listOf(", ")") {
        node.elements.forEach {
            generator.writeShapeValueInline(writer, currShape, it)
            writeInline(",\n")
        }
    }
}
```

ianbotsf (changes requested):
**Comment**: I see there's still [an unaddressed TODO in **SerializeStructGeneratorTest.kt**](https://github.com/awslabs/smithy-kotlin/blob/main/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/SerializeStructGeneratorTest.kt#L238) around tests for document type. We should add some tests there or (if they _truly_ aren't necessary) just delete the TODO.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
lucix-aws: You have witnessed firsthand my javascript reflex. I didn't even realize I did this!

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
lucix-aws: So this was deliberate -- I had `writeBlock`ified at first, but it doesn't generate output in the spirit of the method `writeShapeValueInline` - keyword being "inline", you end up with some extra line breaks. The code is still functional but not necessarily reader-friendly and the output isn't truly inline at that point.

It doesn't super matter to me, especially considering that this `ShapeValueGenerator` is effectively only generating for test code, but I suppose it could be otherwise in the future.

We could also expose `withInlineBlock` on the writer, because the manual *denting is indeed nasty.

At the very least, if we do keep it, a comment might be warranted.

ianbotsf: Oh I see, I missed that the last write doesn't want a newline. `withInlineBlock` is a great idea and might lead to a simplification of this new block and probably some existing blocks as well. Up to you if you want to tackle it in this PR. Otherwise, my concern is addressed.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt
```diff
@@ -454,7 +454,10 @@ abstract class HttpBindingProtocolGenerator : ProtocolGenerator {
                 writer.write("builder.body = #T(payload)", RuntimeTypes.Http.ByteArrayContent)
             }
             ShapeType.DOCUMENT -> {
-                // TODO - deal with document members
+                writer
+                    .write("val serializer = #T()", RuntimeTypes.Serde.SerdeJson.JsonSerializer)
+                    .write("serializer.serializeDocument(input.#L)", memberName)
```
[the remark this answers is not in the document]
lucix-aws: @ianbotsf Make sure you take a look at the updates for this.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
```diff
@@ -229,22 +229,16 @@ class ShapeValueGenerator(
...
-                    node.elements.forEach {
-                        generator.writeShapeValueInline(writer, currShape, it)
-                        writer.writeInline(",\n")
+                    writer.withInlineBlock("#T(", ")", RuntimeTypes.Core.Smithy.Document) {
+                        writer.withInlineBlock("listOf(", ")") {
+                            node.elements.forEachIndexed { index, it ->
+                                generator.writeShapeValueInline(writer, currShape, it)
+                                if (index < node.elements.size - 1) {
+                                    writer.writeInline(",\n")
+                                }
```
ianbotsf, line 238: **Style**: There's no reason to omit the final comma. You could simplify this to the original `forEach` loop that always adds the trailer.

ianbotsf (approved):

PR #672, opened 2022-07-01
chore: upgrade Smithy to 1.22

### LINKED ISSUE #599: Upgrade to Smithy 1.18

ianbotsf:
## Issue \#

Closes #599 

## Description of changes

This change upgrades from Smithy 1.17 to 1.22:
* The biggest change between the versions is the deprecation of `set` shapes and their replacement with `list` shapes annotated with the `@uniqueItems` trait. Accordingly, this PR updates several uses of set-based classes from Smithy and adds an automatic translator to replace legacy `set` shapes in model files with new lists.
* Additionally, `CodeWriter` is deprecated and replaced with `AbstractCodeWriter` and a default `SimpleCodeWriter` implementation.

**Companion PR**: [aws-sdk-kotlin#647](https://github.com/awslabs/aws-sdk-kotlin/pull/647)

### DIFF
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
@@ -196,7 +196,7 @@ class KotlinJmespathExpressionVisitor(val writer: KotlinWriter) : ExpressionVisi
         val operandName = expression.expression.accept(this)
         val truthinessName = addTempVar("${operandName}Truthiness", "truthiness($operandName)")
         val notName = "not${operandName.replaceFirstChar(Char::uppercaseChar)}"
-        return addTempVar(notName, "!${truthinessName}")
+        return addTempVar(notName, "!$truthinessName")
     }
 
     override fun visitObjectProjection(expression: ObjectProjectionExpression): String {
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
@@ -73,12 +73,10 @@ class KotlinJmespathExpressionVisitor(val writer: KotlinWriter) : ExpressionVisi
     override fun visitAnd(expression: AndExpression): String {
         writer.addImport(RuntimeTypes.Utils.truthiness)
 
-        val left = expression.left
-        val leftName = left.accept(this)
+        val leftName = expression.left.accept(this)
         val leftTruthinessName = addTempVar("${leftName}Truthiness", "truthiness($leftName)")
 
-        val right = expression.right
-        val rightName = right.accept(this)
+        val rightName = expression.right.accept(this)
 
         return addTempVar("and", "if ($leftTruthinessName) $rightName else $leftName")
     }
@@ -208,12 +206,10 @@ class KotlinJmespathExpressionVisitor(val writer: KotlinWriter) : ExpressionVisi
     override fun visitOr(expression: OrExpression): String {
         writer.addImport(RuntimeTypes.Utils.truthiness)
 
-        val left = expression.left
-        val leftName = left.accept(this)
+        val leftName = expression.left.accept(this)
         val leftTruthinessName = addTempVar("${leftName}Truthiness", "truthiness($leftName)")
 
-        val right = expression.right
-        val rightName = right.accept(this)
+        val rightName = expression.right.accept(this)
 
         return addTempVar("or", "if ($leftTruthinessName) $leftName else $rightName")
     }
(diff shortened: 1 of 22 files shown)


lucix-aws (approved):

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -71,11 +71,20 @@ class KotlinJmespathExpressionVisitor(val writer: KotlinWriter) : ExpressionVisi
     }
 
     override fun visitAnd(expression: AndExpression): String {
-        throw CodegenException("AndExpression is unsupported")
+        writer.addImport(RuntimeTypes.Utils.truthiness)
```
lucix-aws, line 74: No `#T` formatter in the writer?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -71,11 +71,20 @@ class KotlinJmespathExpressionVisitor(val writer: KotlinWriter) : ExpressionVisi
...
+
+        val left = expression.left
```
lucix-aws, line 76: It's a nit for sure, but these `left`/`right` temporaries don't seem warranted.

ianbotsf:
> IMO it's worth considering actually generating full boolean expressions instead of generating a series of temporary variables that cascade into evaluating the value of the expression. This would allow us to take advantage of short-circuiting as these waiter expressions could get more complex.

Interesting idea. So right now we generate something like this for OR:

```kotlin
val foo = it?.foo
val fooTruthiness = truthiness(foo)
val bar = it?.bar
val or = if (fooTruthiness) foo else bar
```

Are you suggesting a more compact version of this that omits some/all of `foo`/`fooTruthiness`/`bar`? Something like:

```kotlin
val or = if (truthiness(it?.foo)) it?.foo else it?.bar
```

If so, it'd potentially short-circuit an access of `it?.bar` (the right-hand operand, basically). Is that the general idea?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
ianbotsf: You're right, these were originally more complicated and can be omitted now.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
ianbotsf: No, the existing `addTempVar` method accepts a literal code block and doesn't accept/pass placeholder args down to the writer. I could expand it but we've already pushed the scope of this PR far more than I'd like. I'd prefer to keep it simple for now.

lucix-aws:
> > IMO it's worth considering actually generating full boolean expressions instead of generating a series of temporary variables that cascade into evaluating the value of the expression. This would allow us to take advantage of short-circuiting as these waiter expressions could get more complex.
> 
> Interesting idea. So right now we generate something like this for OR:
> 
> ```kotlin
> val foo = it?.foo
> val fooTruthiness = truthiness(foo)
> val bar = it?.bar
> val or = if (fooTruthiness) foo else bar
> ```
> 
> Are you suggesting a more compact version of this that omits some/all of `foo`/`fooTruthiness`/`bar`? Something like:
> 
> ```kotlin
> val or = if (truthiness(it?.foo)) it?.foo else it?.bar
> ```
> 
> If so, it'd potentially short-circuit an access of `it?.bar` (the right-hand operand, basically). Is that the general idea?

Correct.

ianbotsf:
> > > IMO it's worth considering actually generating full boolean expressions instead of generating a series of temporary variables that cascade into evaluating the value of the expression. This would allow us to take advantage of short-circuiting as these waiter expressions could get more complex.
> > 
> > Interesting idea. So right now we generate something like this for OR:
> > ```kotlin
> > val foo = it?.foo
> > val fooTruthiness = truthiness(foo)
> > val bar = it?.bar
> > val or = if (fooTruthiness) foo else bar
> > ```
> > 
> > Are you suggesting a more compact version of this that omits some/all of `foo`/`fooTruthiness`/`bar`? Something like:
> > ```kotlin
> > val or = if (truthiness(it?.foo)) it?.foo else it?.bar
> > ```
> > 
> > If so, it'd potentially short-circuit an access of `it?.bar` (the right-hand operand, basically). Is that the general idea?
> 
> Correct.

We discussed skipping intermediate variables altogether when initially implementing waiters. It would definitely have the benefit of making code smaller, less verbose, and possibly easier to read under simple path specifications (which account for a large portion of the waiters).

We ultimately decided to use granular intermediate variables for the following reasons:
* Without them, the code would become quite lengthy and difficult to mentally parse for complex path expressions. (Some waiters have dozens of subexpressions, all of which could be crowded onto a single line. This would be pretty painful to read much less debug.)
* In-process performance isn't meaningfully impacted by the more verbose code. (Most of these temporary variables will be optimized away by the compiler anyway. More importantly, we've just completed a network IO wait and deserialization run which dwarf the time needed to perform these in-process calculations.)
* Keeping each intermediate calculation in a separate variable makes it easier to trace-log the calculations for aid in debugging. (Note that the trace logging itself isn't added—this is merely a future possibility.)

I'm inclined to apply the same logic to these new `if`/`else` expressions for AND and OR. Short-circuiting is a new benefit but doesn't outweigh the downsides to me. In waiter evaluations, all the data we'd be passing through `if`/`else` blocks is already in memory. Accessing it is fast, cheap, and free of side effects. Enhanced readability/debuggability and (future) granular trace-logging capabilities are far better gains in this situation.

PR #674, opened 2022-07-14
Feat dsl overloads as ext methods

ianbotsf:
## Issue \#

(none)

## Description of changes

Move DSL overloads for builder-style invocation of generated operations out of the client interface and into extension methods. This is generally correct since the behavior is a convenience and not a separate function that inheritors need to implement. It has the additional benefit of simplifying unit testing and mocking.

This is a **breaking change** since now these extension methods require a different import than before.

**Companion PR**: [aws-sdk-kotlin#653](https://github.com/awslabs/aws-sdk-kotlin/pull/653)

### DIFF
--- a/.changes/e1fa257d-73b3-4616-8c39-f32956008124.json
+++ b/.changes/e1fa257d-73b3-4616-8c39-f32956008124.json
@@ -1,0 +1,5 @@
+{
+    "id": "e1fa257d-73b3-4616-8c39-f32956008124",
+    "type": "bugfix",
+    "description": "**Breaking**: Move DSL overloads on generated clients to extension methods"
+}
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ServiceGenerator.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ServiceGenerator.kt
@@ -80,16 +80,12 @@ class ServiceGenerator(private val ctx: RenderingContext<ServiceShape>) {
                 renderServiceConfig()
             }
             .call {
-                operations.forEach { op ->
-                    renderOperation(operationsIndex, op)
-                }
+                operations.forEach { renderOperation(operationsIndex, it) }
             }
             .closeBlock("}")
             .write("")
 
-        operations.forEach { op ->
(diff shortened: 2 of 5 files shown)


smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ServiceGenerator.kt
```diff
@@ -86,6 +86,10 @@ class ServiceGenerator(private val ctx: RenderingContext<ServiceShape>) {
             }
             .closeBlock("}")
             .write("")
+
+        operations.forEach { op ->
```
lucix-aws, line 90: Why `op` instead of implicit `it`?

.changes/e1fa257d-73b3-4616-8c39-f32956008124.json
```diff
@@ -0,0 +1,5 @@
+{
+    "id": "e1fa257d-73b3-4616-8c39-f32956008124",
+    "type": "bugfix",
```
lucix-aws, line 3: I question whether this is a bugfix, but I wouldn't say it's exactly a feature either...

lucix-aws (approved):

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ServiceGenerator.kt
ianbotsf: Oh because I copied/pasted from the similar block up on L83. I'll simplify it.

.changes/e1fa257d-73b3-4616-8c39-f32956008124.json
ianbotsf: I also struggled with this and ultimately decided on **bugfix** because I consider the placement of the DSL overloads in the to be a "design bug".

But it's not a perfect fit. Our options are **bugfix**, **feature**, **documentation**, or **miscellaneous** which I generally reserve for chores. Does **miscellaneous** fit better here?

lucix-aws: Perhaps we're missing a "refactor" category?

PR #679, opened 2022-08-01
fix: parse Instant correctly from epoch offset even when exponential notation is used

ianbotsf:
## Issue \#

[aws-sdk-kotlin#665](https://github.com/awslabs/aws-sdk-kotlin/issues/665)

## Description of changes

This change adds support for parsing epoch timestamps into `Instant`s when exponential notation is used (e.g., `1.234567890E9`). The parser will first normalize the exponentiation into its decimal expansion before continuing with the existing parsing logic.


lucix-aws (approved):

lucix-aws:
Changelog?

PR #686, opened 2022-08-12
feat: endpoints v2: stdlib, runtime additions

lucix-aws:
- Implements common standard-library features for endpoint v2 rules.
- Minor additions to the `Endpoint` type for future ruleset evaluation.
- Implements common-Kotlin URL parser and ipv4/v6 validation to support the endpoint stdlib.

Note that we don't implement `getAttr` here due to the way types are structured in kotlin, it's impractical to implement a scalable/maintable catch-all version of it, we will have to generate it directly into the resolver trees.


runtime/utils/common/src/aws/smithy/kotlin/runtime/util/net/Text.kt
```diff
@@ -0,0 +1,155 @@
...
+
+    for (i in 1 until length) {
+        val c = this[i]
+        if (!c.isLetterOrDigit() && c != '-') {
+            return false
+        }
+    }
+
+    return true
+}
```
ianbotsf, line 68: **Style**: Consider:

```kotlin
public fun String.isValidHostname(): Boolean =
    length in 1..63 && this[0].isLetterOrDigit() && drop(1).all { it.isLetterOrDigit() || it == '-' }
```

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/net/Text.kt
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * Validates a string as an IPv6 address according to [RFC 4291 §2.2](https://www.rfc-editor.org/rfc/rfc4291.html#section-2.2).
+ * Does NOT validate for [address prefixes (§2.3)](https://www.rfc-editor.org/rfc/rfc4291.html#section-2.3).
+ *
+ * The address MAY include a scope ID which is validated per [RFC 6874](https://www.rfc-editor.org/rfc/rfc6874).
+ */
+@InternalApi
+public fun String.isIpv6(): Boolean {
+    val components = split("%25") // delimiter and scope id are %-encoded
```
ianbotsf, line 84: **Question**: Why are delimiter and scope ID percent-encoded? Does some code handle that for us or is it some invariant of IPv6 addresses?

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/net/Text.kt
```diff
@@ -0,0 +1,155 @@
...
+    val rightSegments = if (explicitSegmentGroups[1] == "") listOf() else explicitSegmentGroups[1].split(':')
+
+    val totalSegmentCount = if (rightSegments.lastOrNull()?.contains('.') == true) ipv6DualSegmentCount else ipv6StandardSegmentCount
+    val implicitSegmentCount = totalSegmentCount - leftSegments.size - rightSegments.size
+
+    return buildList {
+        addAll(leftSegments)
+        for (i in 0 until implicitSegmentCount) {
+            add("0")
+        }
```
ianbotsf, line 115: **Style**: Consider `repeat(implicitSegmentCount) { add("0") }`

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/net/Text.kt
```diff
@@ -0,0 +1,155 @@
...
+        if (hostEndIndex != -1 && hostEndIndex != length) substring(hostEndIndex + 1).toInt() else null
+    )
+}
+
+/**
+ * Validates a hostname per [RFC 1123](https://www.ietf.org/rfc/rfc1123.txt).
+ */
+@InternalApi
+public fun String.isValidHostname(): Boolean {
+    if (length !in 1..63 || !this[0].isLetterOrDigit()) {
```
ianbotsf, line 56: **Question**: [RFC 1123 says that](https://www.ietf.org/rfc/rfc1123.txt#:~:text=Host%20software%20MUST%20handle%20host%20names%20of%20up%20to%2063%20characters%20and%0A%20%20%20%20%20%20SHOULD%20handle%20host%20names%20of%20up%20to%20255%20characters.):

> Host software MUST handle host names of up to 63 characters and SHOULD handle host names of up to 255 characters.

Why limit to 63 in this case?

ianbotsf (changes requested):

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpUtils.kt
```diff
@@ -175,3 +176,29 @@ internal class OkHttpProxySelector(
...
+            val parsedParameters = uri.query.splitAsQueryParameters()
+            parameters.appendAll(parsedParameters)
+        }
+
+        if (uri.userInfo != null && uri.userInfo.isNotBlank()) {
+            val userInfoParts = uri.userInfo.split(":")
+            val user = userInfoParts[0]
+            val pw = if (userInfoParts.size > 1) userInfoParts[1] else ""
+            userInfo = UserInfo(user, pw)
+        }
```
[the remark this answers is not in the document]
lucix-aws: You may not have noticed that this code was just relocated from elsewhere, but yeah, I can reincorporate the constructer overload I created.

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/net/Text.kt
lucix-aws: Yeah, this is specified in the RFC for scope ID. https://www.rfc-editor.org/rfc/rfc4007

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/net/Text.kt
lucix-aws: Yeah I noticed this. It's explicitly called out to limit to the base 63 in the endpoints2 SEP, but it's worth asking those folks why they don't want to handle the 255, so I'll do that.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/util/Util.kt
```diff
@@ -0,0 +1,74 @@
...
+}
+
+@InternalApi
+public data class Url(
+    public val scheme: String,
+    public val authority: String,
+    public val path: String,
+    public val normalizedPath: String,
+    public val isIp: Boolean,
+)
```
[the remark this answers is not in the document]
lucix-aws: I think it's semantically clearer to have a discrete type here that maps from our SDK's public URL type. That said, extension `val`s would also be great, and probably more idiomatic.

The only downside I can see from that (correct me if I'm wrong) would be that we'd technically start recomputing the extension vals whenever they were evaluated in rulesets.

eg. `isIp` is checked once in the type mapping route. As an extension, which I imagine we'd implement as `public val ... get() = ...` would then perform the check per-access.

I definitely don't want to add these properties into the main type, that would mean we'd start duplicating some information across fields.

runtime/protocol/http/common/test/aws/smithy/kotlin/runtime/http/UrlTest.kt
```diff
@@ -168,10 +171,54 @@ class UrlTest {
...
+    @Test
+    fun itParsesPathQueryFragment() =
+        testSymmetricParse("http://host/path?n=1#fragment")
+
     @Test
     fun itParsesIpv6Hosts() {
         val actual = Url.parse("http://[2001:db8::1]:80")
-        assertEquals("[2001:db8::1]", actual.host)
+        assertEquals("2001:db8::1", actual.host)
     }
```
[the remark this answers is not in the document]
lucix-aws: Yeah, this was incorrect before. The brackets aren't part of the text encoding for IPv6 addresses, they only exist in URLs to isolate the address for the purpose of parsing.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/Url.kt
```diff
@@ -30,14 +31,14 @@ public data class Url(
...
+    public val encodeParameters: Boolean = true,
 ) {
     init {
         require(port in 1..65536) { "port must be in between 1 and 65536" }
     }
 
     public companion object {
-        public fun parse(url: String): Url = platformUrlParse(url)
+        public fun parse(url: String): Url = url.toUrl()
     }
```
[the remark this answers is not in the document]
lucix-aws: `String.toUrl` is package-internal, and it's just split out like this to not have that logic have to be dictated out within this `companion object` block.

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/net/Text.kt
lucix-aws: It's enforced at 63 to be consistent with how Amazon (primarily S3) handles hostnames today.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
```diff
@@ -17,6 +17,10 @@ import aws.smithy.kotlin.runtime.http.Url
 
  * @property isHostnameImmutable Flag indicating that the hostname can be modified by the SDK client.
  *
+ * @property attributes A grab-bag property map of endpoint attributes. The values here are considered unstable.
+ *
+ * @property headers A map of additional headers to be set when making calls against this endpoint.
+ *
```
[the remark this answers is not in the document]
lucix-aws: Yes.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
```diff
@@ -17,6 +17,10 @@ import aws.smithy.kotlin.runtime.http.Url
 
  * @property isHostnameImmutable Flag indicating that the hostname can be modified by the SDK client.
  *
+ * @property attributes A grab-bag property map of endpoint attributes. The values here are considered unstable.
```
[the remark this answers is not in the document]
lucix-aws: Basically (got clarification from SEP crowd), that they are entirely arbitrary and shouldn't have any expectations placed on them (specific values being set, structure, etc.). In the future. I'll clarify this in the doc.

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/net/Text.kt
ianbotsf: Okay, I see where that doc says [`%` is the delimiter between address and zone ID](https://www.rfc-editor.org/rfc/rfc4007#section-5:~:text=%60%25%27%20is%20a%20delimiter%20character%20to%20distinguish%20between%20%3Caddress%3E%20and%0A%20%20%20%20%20%20%3Czone_id%3E.) and that [such a delimiter may need special handling in URIs](https://www.rfc-editor.org/rfc/rfc4007#section-5:~:text=Also%2C%20the%20format%20for%20non%2Dglobal%20addresses%20might%20conflict%20with%20the%20URI%0A%20%20%20syntax%20%5B13%5D%2C%20since%20the%20syntax%20defines%20the%20delimiter%20character%20%28%60%25%27%29%0A%20%20%20as%20the%20escape%20character.%20%20This%20conflict%20would%20require%2C%20for%20example%2C%0A%20%20%20that%20the%20%3Czone_id%3E%20part%20for%20zone%201%20with%20the%20delimiter%20be%20represented%0A%20%20%20as%20%27%25251%27.). What I don't yet understand is what mechanism will have already percent-encoded the string?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/util/Util.kt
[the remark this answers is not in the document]
ianbotsf: How often would we be accessing these fields? Does precomputing them save us a significant amount of cycles?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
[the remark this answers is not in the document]
ianbotsf: How would a user construct an endpoint (e.g., in their own custom endpoint provider) and fill in these values?

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/net/Text.kt
lucix-aws: Ok you had me tripped up until I realized there are TWO RFCs for this behavior. I linked 4007. We're parsing zone IDs that appear in URIs which means we want to pay attention to https://datatracker.ietf.org/doc/html/rfc6874. The latter is what my code references, I inadvertently discovered and linked it when I went to respond to this question.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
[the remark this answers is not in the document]
lucix-aws: Both it and headers are specified within results for rule evaluation.

So, for example, a ruleset might say "if parameter x is an arn, resolve to foo.com. otherwise, resolve to foobar.com, send this header, and throw this value into the property bag". The SDK implementation has to know what values to look for and how to handle them. An example of something in the AWS-specific property bag is information about signing scope.

We'll generate the code for the rulesets, but someone could obviously write their own custom one and programmatically dictate that an endpoint with some arbitrary parameters be returned in whatever case.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/util/Util.kt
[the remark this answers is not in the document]
lucix-aws: Per-request, since endpoints are now to be resolved in that way.

But, since I have ADT-ified the IP address per another suggestion above, the "repeated" check would now just be an `is` branch, instead of the address parsing we had before, so I'm more inclined to make them extensions at this point.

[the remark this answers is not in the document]
lucix-aws: I ended up not going the extension route - I noticed there would have been a conflict (the SEP specifies a scheme property, and so do we). Thought about it some more and it seems to make the most sense to keep these values discrete, considering their narrow use and similarity to our own SDK type.

runtime/protocol/http/common/test/aws/smithy/kotlin/runtime/http/UrlTest.kt
```diff
@@ -168,10 +201,60 @@ class UrlTest {
         }
     }
 
+    @Test
+    fun itParsesMinimum() =
```
aajtodd, line 205: If we are set on rolling our own URL parsing then we likely need a much more expansive test suite.

For reference here is okhttp's URL parsing tests: https://github.com/square/okhttp/blob/master/okhttp/src/jvmTest/java/okhttp3/HttpUrlTest.kt

We were relying on the JVM stdlib before so we could get away with a small set of sanity tests. Rolling our own may be the right thing to do long term for KMP but it's also something we may want to punt on until we need to pay the cost. 

My 2 cents would be to use the stdlib for now still if we can and lump this into the cost of delivering KMP rather than the cost of delivering endpoints 2.0

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/util/Util.kt
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+package aws.smithy.kotlin.runtime.http.endpoints.util
```
aajtodd, line 5: perhaps some of this should live in `../endpoints/internal` package rather than `util` package?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/util/Util.kt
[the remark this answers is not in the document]
aajtodd: I'm not seeing a compelling reason to have a separate `Url` type here (and certainly not one named the same thing)

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
```diff
@@ -27,9 +23,52 @@ import aws.smithy.kotlin.runtime.http.Url
...
+
+    /**
+     * Flag indicating whether the hostname can be modified by the SDK client.
+     */
+    public val isHostnameImmutable: Boolean
+
+    /**
+     * A map of additional HTTP headers to be set when making calls against this endpoint.
+     */
+    public val headers: Map<String, List<String>>
```
aajtodd, line 41: question: Should we re-use our `Headers` type here or is it worthwhile to have this be a `Map`?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
```diff
@@ -27,9 +23,52 @@ import aws.smithy.kotlin.runtime.http.Url
...
+
+    /**
+     * A grab-bag property map of endpoint attributes. The values here are only set when the endpoint returned from
+     * evaluating a ruleset.
+     */
+    @InternalApi
+    public val attributes: Map<String, Any>
```
aajtodd, line 48: We have a typed property bag [Attributes](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/utils/common/src/aws/smithy/kotlin/runtime/util/Attributes.kt) already. Is there any reason to not leverage that here?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
```diff
@@ -27,9 +23,52 @@ import aws.smithy.kotlin.runtime.http.Url
...
+    /**
+     * The base URL endpoint clients will use to make API calls to e.g. "api.myservice.com".
+     * NOTE: Only `scheme`, `port`, `host` `path`, and `parameters` are valid. Other URL elements are ignored.
+     */
+    public val uri: Url
+
+    /**
+     * Flag indicating whether the hostname can be modified by the SDK client.
+     */
+    public val isHostnameImmutable: Boolean
```
aajtodd, line 36: Could/should this just be an attribute? (We can always have it as an extension)

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
```diff
@@ -27,9 +23,52 @@ import aws.smithy.kotlin.runtime.http.Url
...
+    ) {
+        this.uri = uri
+        this.isHostnameImmutable = isHostnameImmutable
+        this.headers = headers
+        this.attributes = mapOf()
+    }
+
     public constructor(uri: String) : this(Url.parse(uri))
+    @InternalApi
+    public constructor(
```
aajtodd, line 63: why so many constructors? Seems like a single constructor that defaults hostnameImmutable/headers/attributes would suffice

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/util/Util.kt
```diff
@@ -0,0 +1,74 @@
...
+
+@InternalApi
+public fun isValidHostLabel(value: String, allowSubdomains: Boolean): Boolean =
+    if (allowSubdomains) value.isValidHostname() else value.split('.').all(String::isValidHostname)
+
+@InternalApi
+public fun uriEncode(value: String): String = value.urlEncodeComponent(formUrlEncode = false)
+
+@InternalApi
+public fun booleanEquals(i: Boolean, j: Boolean): Boolean = i == j
```
aajtodd, line 39: where is this used? Could probably be `inline` (in fact several of these could be)

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/util/Util.kt
```diff
@@ -0,0 +1,74 @@
...
+
+import aws.smithy.kotlin.runtime.util.InternalApi
+import aws.smithy.kotlin.runtime.util.net.Host
+import aws.smithy.kotlin.runtime.util.net.isIpv6
+import aws.smithy.kotlin.runtime.util.net.isValidHostname
+import aws.smithy.kotlin.runtime.util.text.ensureSuffix
+import aws.smithy.kotlin.runtime.util.text.urlEncodeComponent
+
+@InternalApi
+public fun isSet(value: Any?): Boolean = value != null
```
aajtodd, line 15: I'm assuming these are endpoint stdlib functions we are providing? 

If so I'd suggest adding some documentation to that effect so that we can at least mentally grok this belongs to an SEP. I'd also suggest  moving these out of a generic `Util` and into something like `Runtime` or `Functions` or whatever makes sense to match SEP verbiage.

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/net/Host.kt
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+package aws.smithy.kotlin.runtime.util.net
+
+/**
+ * A [Host] represents a parsed internet host. This may be an internet address (IPv4, IPv6) or a domain name.
+ */
+public sealed class Host {
```
aajtodd, line 10: I'm not 100% certain I agree an ADT is better here but I can understand the desire. Curious if there was discussion on this and the tradeoffs?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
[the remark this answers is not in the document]
aajtodd: I was going to comment on the other PR but I was wondering if the addition of a property bag means we can get rid of `AwsEndpoint` and `AwsEndpointResolver`? IIRC the only difference on the endpoint type is the credential scope. It seems worthwhile to see if there is a path to simplify and have `smithy-kotlin` endpoint and endpoint resolver be able to handle AWS use case.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
```diff
@@ -31,11 +23,52 @@ import aws.smithy.kotlin.runtime.http.Url
  * is expected to be mutable and the client cannot modify the endpoint correctly, the operation
  * will likely fail.
  */
-public data class Endpoint(
-    public val uri: Url,
-    public val isHostnameImmutable: Boolean = false,
-    public val attributes: Map<String, Any> = mapOf(),
-    public val headers: Map<String, List<String>> = mapOf(),
-) {
+public class Endpoint {
```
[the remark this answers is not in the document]
aajtodd: Yeah seems like data class would still be preferrable unless I'm missing something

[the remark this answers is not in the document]
lucix-aws: Yeah - fixed this.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
lucix-aws: The typing here was dictated by the SEP, but it's probably acceptable (and preferred) to deviate from strictness in the interest of being more idiomatic, assuming that functionally they're the same (which they are).

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
lucix-aws: Same comment as in `headers`.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
lucix-aws: This wasn't a new field, it just looked that way because I changed the Endpoint to non-data class, that's since been reverted.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
lucix-aws: Fixed by reverting to ADT.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/util/Util.kt
lucix-aws: I completely agree, but apparently the compiler doesn't like it:
`Expected performance impact from inlining is insignificant. Inlining works best for functions with parameters of functional types`.

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/net/Host.kt
lucix-aws: Yeah, so basically, the semantics I originally had was you'd have a string and then you could call `isIpv4` / `isIpv6` to figure out what it was.

The ADT was Ian's suggestion further up, which I went with because I'm a strong proponent of assigning explicitness to things where we can (as opposed to having a contextless string that you have to sniff the format to determine what it is).

The only _real_ downside I see is it's a bit more effort to work with the value (in that you have to wrap/unwrap from/to string).

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
aajtodd: I think I'm ok either way TBH, was just bringing up the question. If the SEP dictates a map type there may be good reason to not deviate then. Not sure

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/util/Util.kt
aajtodd: My next question was going to be do some of these _need_ to be actual functions or can they be dealt with in codegen directly?

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/net/Host.kt
aajtodd: Ok. Yeah I don't have a super strong opinion either way. I suppose since we have a constructor for a `Url` from string directly and `Host.parse()` that those should be sufficient.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
aajtodd: Yeah in this case is there any reason to not use `Attributes` directly though? It's effectively the same thing except it's a typed property bag rather than untyped. I care less about headers than I do this I think 🤔

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
aajtodd: Right yeah no what I was suggesting/wondering is if we have this new `attributes` field (especially if it's our typed property bag) we could just make this an attribute rather than an explicit field. 

e.g.



```kotlin
data class Endpoint(val url: Url, val attributes: Attributes = ...)

val Endpoint.isHostNameImmutable: Boolean
    get() = attributes[EndpointImmutable] ?: false

```

It's fine as is too I was just wondering how much we can shove into the property bag since that's effectively what this is, an attribute of the endpoint.

ianbotsf (approved):

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/QueryParameters.kt
```diff
@@ -49,6 +49,8 @@ public fun Map<String, String>.toQueryParameters(): QueryParameters {
 
 private class QueryParametersImpl(values: Map<String, List<String>> = emptyMap()) : QueryParameters, StringValuesMapImpl(true, values) {
     override fun toString(): String = "QueryParameters ${entries()}"
+
+    override fun equals(other: Any?): Boolean = other is QueryParametersImpl && entries() == other.entries()
```
ianbotsf, line 53: **Question**: Why require `other` to be `QueryParametersImpl` vs `QueryParameters`? This means that the following code will fail but believe we'd want it to succeed:

```kotlin
val expected = QueryParameters.Empty
val actual = QueryParameters { }
assertEquals(expected, actual)
```

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/functions/Functions.kt
```diff
@@ -2,7 +2,10 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0.
  */
-package aws.smithy.kotlin.runtime.http.endpoints.util
+/**
+ * This package implements common standard library functions used by endpoint resolvers.
+ */
+package aws.smithy.kotlin.runtime.http.endpoints.functions
```
ianbotsf, line 8: **Comment**: I don't believe this actually applies KDoc to the package. See https://kotlinlang.org/docs/kotlin-doc.html#module-and-package-documentation for more detail. I suggest you either:
* Move it to a separate markdown file if the intended audience is users
* Change this to a non-doc comment if the intended audience is only SDK code authors
* Delete it if this package doesn't warrant its own documentation

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/Attributes.kt
```diff
@@ -123,4 +123,7 @@ private class AttributesImpl : Attributes {
 
     override val keys: Set<AttributeKey<*>>
         get() = map.keys
+
+    override fun equals(other: Any?): Boolean =
+        other is AttributesImpl && map == other.map
```
ianbotsf, line 128: **Question**: Same question above about checking specific implementations vs the interface.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
```diff
@@ -27,9 +35,18 @@ import aws.smithy.kotlin.runtime.http.Url
  * is expected to be mutable and the client cannot modify the endpoint correctly, the operation
  * will likely fail.
  */
-public data class Endpoint(
+public data class Endpoint @InternalApi constructor(
```
aajtodd, line 38: why is this constructor internal and why do we still need the constructor on L47?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/functions/Functions.kt
```diff
@@ -0,0 +1,62 @@
...
+        scheme = sdkUrl.scheme.protocolName,
+        authority,
+        path = sdkUrl.path,
+        normalizedPath = sdkUrl.path.ensureSuffix("/"),
+        isIp = sdkUrl.host is Host.IPv4 || sdkUrl.host is Host.IPv6,
+    )
+}
+
+@InternalApi
+public data class Url(
```
aajtodd, line 56: Why do we need this type as opposed to re-using our existing `Url` type? If there is a reason is there a better name for this thing that makes it less ambiguous?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/QueryParameters.kt
```diff
@@ -49,6 +49,8 @@ public fun Map<String, String>.toQueryParameters(): QueryParameters {
 
 private class QueryParametersImpl(values: Map<String, List<String>> = emptyMap()) : QueryParameters, StringValuesMapImpl(true, values) {
     override fun toString(): String = "QueryParameters ${entries()}"
+
+    override fun equals(other: Any?): Boolean = other is QueryParameters && entries() == other.entries()
```
aajtodd, line 53: why are we defining equality for this? I'm not sure this does what we expect since query params can use an internal case insensitive map that _does not_ keep the uses the original given key when returning the entry set.

runtime/utils/common/src/aws/smithy/kotlin/runtime/util/Attributes.kt
```diff
@@ -123,4 +123,12 @@ private class AttributesImpl : Attributes {
 
     override val keys: Set<AttributeKey<*>>
         get() = map.keys
+
+    override fun equals(other: Any?): Boolean =
```
aajtodd, line 127: same question, why define equality for this? What does it mean to have equality on a property bag of arbitrary values (that may not have meaningful equality defined).

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/QueryParameters.kt
lucix-aws: It was added to support the `assertEquals` checks in the new parser test (to check equality on a full URL). The alternative would be to inspect individual members in those tests (or adjust the equality check if we want it and feel this implementation isn't adequate).

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/Endpoint.kt
lucix-aws: Callers (people writing endpoint resolvers of their own) aren't allowed to directly manipulate the properties field. This is SEP-dictated. There's been a LOT of discussion around this field internally, we can discuss further.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/util/StringValuesMap.kt
```diff
@@ -81,6 +81,22 @@ internal open class StringValuesMapImpl(
     override fun contains(name: String, value: String): Boolean = getAll(name)?.contains(value) ?: false
 
     override fun isEmpty(): Boolean = values.isEmpty()
+
+    override fun equals(other: Any?): Boolean =
+        other is StringValuesMap && names().size == other.names().size &&
+            names().all {
```
aajtodd, line 87: I think this can be simplified to:

```kotlin
other is StringValuesMap && 
     names().size == other.names().size && 
     names().all { getAll(it) == other.getAll(it) }
```

if it's a case insensitive map then the [get implementation](https://github.com/awslabs/smithy-kotlin/blob/8d17f0666d9bec313d60838e8e00bdef950f7a13/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/util/CaseInsensitiveMap.kt#L33) takes care of that

aajtodd (approved):

Issue #677, opened 2022-07-22
Tracing framework


ianbotsf:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue, please leave a comment

## Describe the Feature

Replace the existing logging functionality with a full-featured tracing mechanism, which should allow for hierarchical trace spans, swappable probe (or sink) implementations, and provide for metrics in addition to logging.

PR #688, opened 2022-08-16
Initial tracing framework

ianbotsf:
## Issue \#

Initial part of #677 

## Description of changes

This change adds support for a tracing framework including:
* Adding the **runtime/tracing/*** projects classes (🌟 start reviewing here)
* Enabling context receivers in our build configurations
* Adding new client config for trace probe and client name (I expect 👈 to be controversial)
* Adding a required `TraceSpan` context to `SdkHttpOperation.roundTrip`. There are still _a lot_ more places to thread context through.
  * This requires updates to a lot of tests. I wanted to create a new `runTestWithTraceSpan` test method (basically `runTest` with an automatic no-op trace span) but [KT-53551](https://youtrack.jetbrains.com/issue/KT-53551) stymied me on that front. For now, every test has been modified to explicit set a no-op trace span.
* Codegenning the creation of a root trace span and request-based child span creation

Companion PR: [aws-sdk-kotlin#680](https://github.com/awslabs/aws-sdk-kotlin/pull/680)


lucix-aws (approved):

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/Url.kt
```diff
@@ -12,7 +12,7 @@ import aws.smithy.kotlin.runtime.util.text.encodeUrlPath
  * Represents an immutable URL of the form: `[scheme:][//[userinfo@]host][/]path[?query][#fragment]`
  *
  * @property scheme The wire protocol (e.g. http, https, ws, wss, etc)
- * @property host hostname
+ * @property host hostnameAwsProtocolModelDeclaration
```
aajtodd, line 15: ?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/operation/SdkHttpOperation.kt
```diff
@@ -26,9 +27,12 @@ public class SdkHttpOperation<I, O>(
     internal val serializer: HttpSerialize<I>,
     internal val deserializer: HttpDeserialize<O>,
 ) {
+    /**
+     * The SDK-side request ID for this operation.
+     */
+    public val sdkRequestId: String = Uuid.random().toString()
```
aajtodd, line 33: I think you can get this off the context already or define an extension to pull it out of the context without needing to also make it a property

runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/DefaultTracer.kt
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.tracing
+
+import aws.smithy.kotlin.runtime.util.InternalApi
+
+@InternalApi
+public class DefaultTraceSpan(
```
aajtodd, line 10: Does this need to be `public` or is it just an internal detail of `DefaultTracer`?

runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/TraceSpan.kt
```diff
@@ -0,0 +1,129 @@
...
+     * @param events The events to record.
+     */
+    public fun postEvents(events: Iterable<TraceEvent>)
+}
+
+/**
+ * Records a single new event that has occurred within the logical context of this span.
+ * @param event The event to record.
+ */
+public fun TraceSpan.postEvent(event: TraceEvent): Unit = postEvents(listOf(event))
```
aajtodd, line 45: I'm guessing this is going to be our primary use case (single event). Is there any concern/thought given to the overhead of list allocation on every event?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/GradleGenerator.kt
```diff
@@ -184,15 +186,24 @@ fun renderJvmGradleBuild(
...
         annotationRenderer,
         { w: GradleWriter -> if (isRootModule) w.write("explicitApi()") },
+        { w: GradleWriter -> if (isRootModule) renderCompilerFlags(w) },
     )
 }
 
+private fun renderCompilerFlags(writer: GradleWriter) = with(writer) {
+    write("")
+    withBlock("tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>{", "}") {
+        write("""kotlinOptions.freeCompilerArgs += "-Xcontext-receivers"""")
```
aajtodd, line 203: how confident are we in context receivers stabilizing?

smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigGeneratorTest.kt
```diff
@@ -77,6 +88,11 @@ public class Config private constructor(builder: Builder): HttpClientConfig, Ide
          * debug purposes.
          */
         public var sdkLogMode: SdkLogMode = SdkLogMode.Default
+        /**
+         * The probe that receives tracing events such as logging messages and metrics. This probe can be used
+         * to send tracing events to other frameworks outside the SDK. By default, a no-op probe is selected.
+         */
+        public var traceProbe: TraceProbe? = null
```
aajtodd, line 95: should we be taking a probe here or should we be taking a `TraceSpan` and defaulting it to a `DefaultTracer` with `NoOpTraceProbe`? 

If we leave it as is (which I think is a reasonable path) it means we are always the root span which means the SDK clients can belong to some other parent span. This is probably ok since we don't really expect customers to instrument their code with our tracing abstractions but worth some thought.

docs/design/tracing.md
```diff
@@ -0,0 +1,150 @@
...
+Trace events can be one of three types:
+* `Message`: Typically, a free-form text message used for logging
+* `Count`: The numerical measurement of some value (e.g., results returned, bytes written, etc.)
+* `Timespan`: The temporal measurement of some occurrence (e.g., time elapsed, latency, etc.)
+
+Probes are free to handle these different types of events however they see fit (e.g., they may log some messages,
+aggregate some metrics, ignore some events, etc.).
+
+Event data values (i.e., message text, count values, and timespan durations) are provided as lambdas rather than with
+direct values. This allows probe implementations to skip calculating them in the even they would otherwise be discarded
```
aajtodd, line 108: "in the **event**"

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolClientGenerator.kt
```diff
@@ -208,16 +212,17 @@ abstract class HttpProtocolClientGenerator(
...
-            writer
-                .addImport(RuntimeTypes.Http.Operation.execute)
-                .write("return op.#T(client, #L, block)", RuntimeTypes.Http.Operation.execute, inputVariableName)
-        } else {
-            writer.addImport(RuntimeTypes.Http.Operation.roundTrip)
-            if (outputShape.isPresent) {
-                writer.write("return op.#T(client, #L)", RuntimeTypes.Http.Operation.roundTrip, inputVariableName)
-            } else {
-                writer.write("op.#T(client, #L)", RuntimeTypes.Http.Operation.roundTrip, inputVariableName)
+        writer.withBlock("return rootTraceSpan.child(op.sdkRequestId).#T { traceSpan ->", "}", RuntimeTypes.IO.use) {
```
aajtodd, line 215: I wonder if the child span should include the operation name and the request id (e.g. such that the resulting log is like `DynamoDb/PutItem/<request-id>` 🤔

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/operation/SdkHttpOperation.kt
```diff
@@ -59,6 +63,7 @@ public class SdkHttpOperation<I, O>(
 /**
  * Round trip an operation using the given [HttpHandler]
  */
+context(TraceSpan)
```
aajtodd, line 66: I'm guessing this is the most contentious decision. What I'm wondering is how this actually plays out. It seems it will propagate `context(TraceSpan)` all over the place. Of course passing it down explicitly is roughly the same result. 

The alternative (which I'm not saying is better just thinking out loud) is to thread it into `ExecutionContext` which is already [a placeholder for per operation metadata](https://github.com/awslabs/smithy-kotlin/blob/cfa0fd3a30b4c50b75485786f043d4e2ad803f55/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/client/ExecutionContext.kt#L9-L12). 

I think the way to think about which route may be better is to answer the question "how deep are nested child spans going to be"? If *most* child spans are created off of the operation specific span then `ExecutionContext` becomes more attractive since we already thread it through so many places (e.g. every middleware, http engine, etc). If we have mostly deeply nested trace spans then creating child spans off the operation span doesn't help and putting into execution context makes less sense.

I'd like to explore further why we think context receivers is the right choice and what the difference between approaches looks like (which is sort of hard to do in this PR since we create no child spans and can't see any output yet)

aajtodd (commented):
Overall this looks really good, nice work!

runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/TraceSpan.kt
ianbotsf: Giving some thought to this, yes there would be some additional overhead of list allocation. I believe that's justified by two factors:

* The receiving side of this (`TraceProbe`) also specifies a list-typed consuming method: `fun postEvents(span: TraceSpan, events: Iterable<TraceEvent>)`. If we offered a single-item `TraceSpan.postEvent` method, we'd need to support a similar single-event `TraceProbe.postEvent` method in the consumer. This would force implementors to write two methods for handling events, which seems like poor interface design. It's important to be able to support the batch case so that implementors can take advantage of scaling efficiencies in their downstream systems/logic so it also seems wrong to eliminate the list-typed overload.
* The [implementation of `listOf(element: T)`](https://github.com/JetBrains/kotlin/blob/ea836fd46a1fef07d77c96f9d7e8d7807f793453/libraries/stdlib/jvm/src/kotlin/collections/CollectionsJVM.kt#L20) on JVM delegates down to [`Collections.singletonList`](https://github.com/openjdk/jdk/blob/jdk8-b120/jdk/src/share/classes/java/util/Collections.java#L4847-L4898), which utilizes an optimized list class with very little overhead (compared to mutable list implementations). I believe the cost here is low.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolClientGenerator.kt
ianbotsf: Spans are context boundaries which have a meaningful beginning and end. For the entire client, the beginning and end are clear: the lifetime of the client itself, which begins at instantiation and ends when `close` is called. For requests, the beginning and end align with the lifetime of the request and its various middleware. These boundaries are important (especially the end) because probe implementations may collect events _until_ the end of a span before calculating any additional metadata or emitting them downstream.

For a child span that's defined by operation name, what are the beginning and end conditions? It seems like they could only be coterminous with an individual request or (more likely) the client itself. Neither case seems to add incremental benefit for event batching.

aajtodd: Yes I suppose `sdkRequestId` and `OperationName` have the same lifetime. I was thinking of this mostly through a debugging lens. Being able to only emit events for a specific operation would be easier with the operation name whereas request ID is dynamic per invocation obviously. It would be harder to filter spans of interest (you either get all of a specific client or all of a specific request ID, nothing in-between).

ianbotsf: I suppose it's possible the span IDs for requests could combine the operation name and a unique request ID:

Example spans:
* `s3`
  * `listObjects:89e1cbef-00f5-4005-9863-9d18aff6ac52`
    * `attempt 0`
      * ...

That said, I worry we're conflating the logical scope/lifestime of spans with additional information that may be useful for debugging, binning, or filtering. I wonder if we're missing a design concept of additional metadata (or what [CloudWatch calls "dimensions"](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html#Dimension)) which could be better suited to, for example, probes filtering events to specific operation names.

ianbotsf:
Closing in favor of #707.

PR #695, opened 2022-08-31
misc: add design document for endpoints2 implementation

lucix-aws:

### DIFF
--- a/docs/design/endpoints.md
+++ b/docs/design/endpoints.md
@@ -249,7 +249,8 @@ this value from the client config.
 
 ## Codegen
 ### Package "endpoints"
-The provider parameters and interface will generate to a new top-level `endpoints` package.
+The provider parameters and interface will generate to a new top-level `endpoints` package underneath each service
+package (e.g., `aws.sdk.kotlin.services.s3.endpoints`).
 
 ### Provider parameters
 
@@ -343,19 +344,19 @@ public open class EndpointResolverMiddlewareGenerator(
 This will generate the following two-phase middleware:
 ```kotlin
 internal class OperationResolveEndpointMiddleware(
-  private val config: SmithyService.Config,
+    private val config: SmithyService.Config,
 ) : InlineMiddleware<OperationRequest, OperationResponse> {
-  private var endpoint: Endpoint
+    private var endpoint: Endpoint
 
-  override fun install(op: SdkHttpOperation<GetResourceRequest, GetResourceResponse>) {
-      op.execution.initialize.intercept { req, next ->
-          // bind provider parameters and resolve endpoint, saving the value for the mutate phase
-      }
+    override fun install(op: SdkHttpOperation<GetResourceRequest, GetResourceResponse>) {
+        op.execution.initialize.intercept { req, next ->
+            // bind provider parameters and resolve endpoint, saving the value for the mutate phase
+        }
 
-      op.execution.mutate.intercept { req, next ->
-          // set the request endpoint
-      }
-  }
+        op.execution.mutate.intercept { req, next ->
+            // set the request endpoint
+        }
+    }
 }
 ```
 
--- a/docs/design/endpoints.md
+++ b/docs/design/endpoints.md
@@ -253,15 +253,12 @@ The provider parameters and interface will generate to a new top-level `endpoint
 
 ### Provider parameters
 
-Provider parameters will be generated as a data class, since backwards compatability of parameter values is guaranteed (
-field order MUST be preserved, and obsolete values are marked deprecated and not removed).
+Provider parameters will be generated as a data class, since backwards compatability of parameter values is guaranteed
+(field order MUST be preserved, and obsolete values are marked deprecated and not removed).
 
 The presence of required values is checked at construction, with an error thrown if any are missing.
 
-The generation includes DSL builder support for several reasons:
-1. caller convenience - SDK users are able to implement and call their own endpoint providers
-2. doing so provides greater flexibility for codegen for middleware - the binding of a value can be expressed as an
-   arbitrary set of statements, rather than being restricted to a single expression in a constructor argument list
+The generation includes DSL builder support to simplify construction.
 ```kotlin
 public data class ServiceEndpointParams(
     // ...args
@@ -287,7 +284,7 @@ authors as needed.
 ```kotlin
 public open class EndpointProviderGenerator(
     public val writer: KotlinWriter,
-    public val ruleSet: EndpointRuleset,
+    public val rules: EndpointRuleset,
     // ...
 ) {
     /**
@@ -296,9 +293,9 @@ public open class EndpointProviderGenerator(
     public fun render()
 
     /**
-     * Render a standard library function referenced in a rule set.
+     * Provide a symbol for a standard library function referenced in a rule set.
      */
-    public open fun renderFunction(func: String, args: List<FunctionArg>)
+    public open fun resolveFunction(func: String): Symbol
 }
 ```
 
@@ -317,7 +314,7 @@ have unique middlewares generated to bind those specific values.
 public open class EndpointResolverMiddlewareGenerator(
     public val writer: KotlinWriter,
     public val operation: OperationShape,
-    public val ruleSet: EndpointRuleset,
+    public val rules: EndpointRuleset,
     // will likely need some additional params such as service shape etc.
 ) {
     /**
@@ -327,16 +324,6 @@ public open class EndpointResolverMiddlewareGenerator(
      * Client/static/general context parameters are inspected here to determine how to bind values.
      */
     public fun render()
-
-    /**
-     * Implement this API to generate additional parameters to be passed to the middleware.
-     * By default, the middleware takes two parameters:
-     * 1. service client config
-     * 2. operation's request object
-     * 
-     * The implementer is responsible for code generation of the middleware installation.
-     */
-    public open fun getMiddlewareParams(): List<Symbol>
     
     /**
      * SDK authors extend this method to handle built-ins.
@@ -353,17 +340,22 @@ public open class EndpointResolverMiddlewareGenerator(
 }
 ```
 
-This will generate middleware with the following signature:
+This will generate the following two-phase middleware:
 ```kotlin
-public class OperationMiddleware(
-    private val config: SmithyService.Config, // has the endpoint provider
-    private val request: OperationRequest,
-) : ModifyRequestMiddleware {
-    override suspend fun modifyRequest(req: SdkHttpRequest): SdkHttpRequest {
-        /**
-         * Source provider parameters, resolve endpoint, set it on the request.
-         */
-    }
+internal class OperationResolveEndpointMiddleware(
+  private val config: SmithyService.Config,
+) : InlineMiddleware<OperationRequest, OperationResponse> {
+  private var endpoint: Endpoint
+
+  override fun install(op: SdkHttpOperation<GetResourceRequest, GetResourceResponse>) {
+      op.execution.initialize.intercept { req, next ->
+          // bind provider parameters and resolve endpoint, saving the value for the mutate phase
+      }
+
+      op.execution.mutate.intercept { req, next ->
+          // set the request endpoint
+      }
+  }
 }
 ```
 
@@ -475,19 +467,25 @@ internal class DefaultSmithyServiceEndpointProvider: SmithyServiceEndpointProvid
 
 internal class GetResourceResolveEndpointMiddleware(
     private val config: SmithyService.Config,
-    private val request: GetResourceRequest,
-) : ModifyRequestMiddleware {
-    override suspend fun modifyRequest(req: SdkHttpRequest): SdkHttpRequest {
-        val params = SmithyServiceEndpointParams {
-            resourceId = request.resource
+) : InlineMiddleware<GetResourceRequest, GetResourceResponse> {
+    private var endpoint: Endpoint
+
+    override fun install(op: SdkHttpOperation<GetResourceRequest, GetResourceResponse>) {
+        op.execution.initialize.intercept { req, next ->
+            val params = SmithyServiceEndpointParams {
+              resourceId = req.subject.resource
+            }
+            val endpoint = config.provider.resolveEndpoint(params)
+            next.call(req)
         }
-        val endpoint = config.provider.resolveEndpoint(params)
-            
-        setRequestEndpoint(req, endpoint)
         
-        // further action can be taken here as part of renderPostResolution...
-        
-        return req
+        op.execution.mutate.intercept { req, next -> ...
+            setRequestEndpoint(req, endpoint)
+
+            // further action can be taken here as part of renderPostResolution...
+
+            next.call(req)
+        }
     }
 }
 ```
--- a/docs/design/endpoints.md
+++ b/docs/design/endpoints.md
@@ -1,0 +1,292 @@
+# Endpoints
+
+* **Type**: Design
+* **Author(s)**: Luc Talatinian
+
+This document discusses the design approach to modern endpoint resolution (Endpoints v2.0) in the Kotlin SDK.
+
+## Overview
+The Smithy IDL exposes a "rules language", which allows a model author to define the programmatic steps that must be
+taken in order to determine the endpoint that calls should be made to. This language is provided inline with the model
+as the `smithy.rules#endpointRuleSet` service trait. The schema for this trait will be described as part of Smithy's
+extended documentation in the future.
+
+The smithy-kotlin repository implements the core code-generation for endpoint providers based on rule sets defined this
+way, as well as the middleware that sources the parameters for the resolver and calls it on each request.
+
+## Core Concepts
+### Endpoint
+The `Endpoint` is the result of calling an endpoint provider implementation at runtime to determine what endpoint to use
+when making service calls. The type is public and shared across all service clients, and all provider implementations
+MUST return values of this type. All values of the endpoint are provided by the ruleset.
+
+The (annotated) type is provided here:
+
+```kotlin
+public data class Endpoint @InternalApi constructor(
+    /**
+     * The actual URL against which service operations are called.
+     */
+    public val uri: Url,
+
+    /**
+     * Any additional HTTP headers to send on calls to this endpoint.
+     */
+    public val headers: Headers = Headers.Empty,
+
+    /**
+     * SDK-specific attributes tied to this endpoint.
+     * 
+     * The values within the set of attributes are unstable. Additionally, custom provider implementations cannot set
+     * these values as part of endpoints they resolve to (hence the internal annotation). They can only be specified
+     * in the Smithy ruleset and returned as part of generated code therein.
+     * 
+     * SDK authors are encouraged to define strongly-typed extensions for values relevant to their SDK's operation.
+     */
+    @InternalApi
+    public val attributes: Attributes = Attributes(),
+)
+```
+
+### Endpoint provider
+
+Providers house the core logic for determining the endpoint to use **on a per-request basis**. Providers are generated
+as both a service-specific interface (to be tied to that service's specific provider parameters type and allow for
+custom implementations) and a default provider that implements the logic defined in the model.
+
+```kotlin
+public fun interface EndpointProvider {
+    public suspend fun resolveEndpoint(params: ServiceEndpointParams): Endpoint
+}
+```
+Client config will include a field for the resolver unique to that service, and SDK operations will use the generated
+default unless specified otherwise by the caller.
+
+### Provider library
+The endpoint rulesets as provided by Smithy require some "standard-library"-type functions that support or drive the
+evaluation of the rules.
+
+Some of these are trivial (eg. "is a value non-null") and are generated inline as part of the
+ruleset evaluation. The more complex functions (eg. URL parsing) are hosted in the runtime at
+`package aws.smithy.kotlin.runtime.http.endpoints.functions`. Runtime-hosted standard library functions are annotated as
+internal and should only be called by code-generated endpoint providers.
+
+### Provider parameters
+Each service client has a specific set of parameters passed to invocations of its endpoint provider. Endpoint resolution
+is a pre-serialization operation, and the rules language specifies several ways in which parameters are sourced.
+
+While provider parameters are defined per-service, the sources of the values may vary per-operation based on context
+parameters defined in the model.
+
+#### `clientContextParams`
+Service-level annotation. Defines an additional configuration field to be added to the service's client config. Bound to
+a field on the provider parameters at resolution time.
+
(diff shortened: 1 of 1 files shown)


docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+      "conditions": [],
+      "endpoint": {
+        "url": "https://global.api"
+      }
+    }
+  ]
+}
+```
+*Note*: The signature of the substring method is `substring(s: String, start: Int, end: Int, reverseIndices: false): String`.
+The argv property in the corresponding function call describes how to source the arguments (or input constants directly
```
ianbotsf, line 238: **Style**: "describes how to source" → "provides"

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+
+public class GetResourceResolveEndpointMiddleware(
+    private val config: SmithyService.Config,
+    private val request: GetResourceRequest,
+    private val provider: EndpointProvider,
+) : ModifyRequestMiddleware {
+    override suspend fun modifyRequest(req: SdkHttpRequest): SdkHttpRequest {
+        val params = SmithyServiceEndpointParams {
+            resourceId = request.resource
+        }
```
ianbotsf, line 282: **Comment**: Your above definition of `SmithyServiceEndpointParams` was a data class with a public constructor and no mention of a builder whereas here it's invoked using a builder lambda. The former is preferable if we can get away with it (e.g., backward compatibility isn't a concern).

ianbotsf (changes requested):

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+Client config will include a field for the resolver unique to that service, and SDK operations will use the generated
+default unless specified otherwise by the caller.
+
+### Provider library
+The endpoint rulesets as provided by Smithy require some "standard-library"-type functions that support or drive the
+evaluation of the rules.
+
+Some of these are trivial (eg. "is a value non-null") and are generated inline as part of the
+ruleset evaluation. The more complex functions (eg. URL parsing) are hosted in the runtime at
+`package aws.smithy.kotlin.runtime.http.endpoints.functions`. Runtime-hosted standard library functions are annotated as
```
aajtodd, line 71: Tangent tot his design, something I've wondered is whether we should pull endpoints out of `http` into it's own module (or core depending). The only hiccup here (and likely why I placed it here in the first place) is that the type relies on `Url`. Something to think about whether we want to tie endpoints to http though.

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+) {
+    /**
+     * Main codegen routine - walk the rules tree and generate the provider.
+     */
+    public fun render()
+
+    /**
+     * Resolve a standard library function referenced in a ruleset.
+     */
+    public open fun resolveFunction(func: String)
```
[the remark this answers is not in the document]
aajtodd: I was wondering if this should be `Symbol` (likely)

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+
+Given a sample operation `GetResource`, with a single operation input `resource` bound to the `ResourceId` provider
+parameter via context traits, this ruleset would roughly generate the following code:
+
+```kotlin
+public data class SmithyServiceEndpointParams(
+    public val resourceId: String?,
+) {
+    init {
+        require(resourceId != null) { "param resourceId is required" }
```
aajtodd, line 249: if it's required why is it nullable in the signature?

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+```
+*Note*: The signature of the substring method is `substring(s: String, start: Int, end: Int, reverseIndices: false): String`.
+The argv property in the corresponding function call describes how to source the arguments (or input constants directly
+in the case of the indices).
+
+Given a sample operation `GetResource`, with a single operation input `resource` bound to the `ResourceId` provider
+parameter via context traits, this ruleset would roughly generate the following code:
+
+```kotlin
+public data class SmithyServiceEndpointParams(
```
aajtodd, line 245: to Ian's point below if this is public can we guarantee ordering of parameters will be consistent over time? Otherwise this will break backwards compat one day. 

Perhaps this should follow the builder pattern for operation input/outputs?

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+    public val resourceId: String?,
+) {
+    init {
+        require(resourceId != null) { "param resourceId is required" }
+    }
+}
+
+public fun interface SmithyServiceEndpointProvider {
+    open suspend fun resolveEndpoint(params: SmithyServiceEndpointParams): Endpoint
```
aajtodd, line 254: nit: it's an interface, `open` is implied

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+            // use a special endpoint for government resources
+            return Endpoint(
+                Url.parse("https://gov.api"),
+            )
+        }
+
+        // fallback to global endpoint        
+        return Endpoint(
+            Url.parse("https://global.api"),
+        )
```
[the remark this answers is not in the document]
aajtodd: or something like:

```kotlin
val url = when (resourceIdPrefix) {
     "gov." -> "https://gov.api"
     else -> "https://global.api"
}
return Endpoint(Url.parse(url))
```

This is of course an implementation detail we can go over later when we actually have the full scope of rules to deal with.

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+    }
+}
+
+public class GetResourceResolveEndpointMiddleware(
+    private val config: SmithyService.Config,
+    private val request: GetResourceRequest,
+    private val provider: EndpointProvider,
```
aajtodd, line 277: wouldn't this be the client specific endpoint provider `SmithyServiceEndpointProvider`? (since it takes specific parameters)

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+        }
+
+        // fallback to global endpoint        
+        return Endpoint(
+            Url.parse("https://global.api"),
+        )
+    }
+}
+
+public class GetResourceResolveEndpointMiddleware(
```
aajtodd, line 274: this would be `internal`.

What package do each of these go to? I would assume this would go in the `internal` package?

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+```
+
+### Middleware
+Endpoint resolution, whether generated or written by the caller, is performed per-request as a pre-operation middleware.
+At runtime, endpoint resolution is comprised of two principal tasks:
+1. construct a set of provider parameters, sourcing the values as specified by the model (context params)
+2. using the sourced parameters, call the provider to resolve to an endpoint, pointing the request to it
+
+Since endpoint resolution is performed on a per-operation basis, and the sourcing of provider parameters can vary
+per-operation, middleware must now be generated per-operation (this differs from the original implementation, where the
```
aajtodd, line 133: Question: If an operation only uses service level context could there be a shared middleware for those operations?

This is a bit of premature optimization but we have seen how every additional class/lambda/etc adds up to overall artifact size. I would hazard a guess that most operations are not going to do anything fancy (since most don't now) and that generating a singular implementation that is shared across them would save a bunch on space/compilation/etc for larger models like DDB, EC2, S3, etc. It also would be backwards compatible because if a model ever added per/operation endpoint traits it would just generate a new internal middleware.

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+    init {
+        require(resourceId != null) { "param resourceId is required" }
+    }
+}
+
+public fun interface SmithyServiceEndpointProvider {
+    open suspend fun resolveEndpoint(params: SmithyServiceEndpointParams): Endpoint
+}
+
+public class DefaultSmithyServiceEndpointProvider: SmithyServiceEndpointProvider {
```
aajtodd, line 257: Should probable be `internal` (and live in the `internal` package) as it is now.

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+    public suspend fun resolveEndpoint(params: SmithyServiceEndpointParams): Endpoint {
+        val resourceIdPrefix = substring(params.resourceId, 0, 5, false)
```
aajtodd, line 259: Question: Are rules really going to be inlined like this? There isn't an internal data structure generated and algorithm for resolution (like there is now where we generate a partition table and have a concrete implementation in the runtime that handles resolution)? How large do we expect some rule sets to get?

docs/design/endpoints.md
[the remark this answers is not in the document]
lucix-aws: It's going to be rendering _something_ - either a symbol, or some inline logic.

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+
+### Middleware
+Endpoint resolution, whether generated or written by the caller, is performed per-request as a pre-operation middleware.
+At runtime, endpoint resolution is comprised of two principal tasks:
+1. construct a set of provider parameters, sourcing the values as specified by the model (context params)
+2. using the sourced parameters, call the provider to resolve to an endpoint, pointing the request to it
+
+Since endpoint resolution is performed on a per-operation basis, and the sourcing of provider parameters can vary
+per-operation, middleware must now be generated per-operation (this differs from the original implementation, where the
+endpoint resolver middleware was static across operations and its implementation was part of the runtime).
```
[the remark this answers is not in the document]
lucix-aws: There's no real way around it - the Endpoints2 spec introduces this level of variety.

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
+# Endpoints
+
+* **Type**: Design
+* **Author(s)**: Luc Talatinian
+
+This document discusses the design approach to modern endpoint resolution (Endpoints v2.0) in the Kotlin SDK.
+
+## Overview
+The Smithy IDL exposes a "rules language", which allows a model author to define the programmatic steps that must be
```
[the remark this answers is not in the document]
lucix-aws: Sadly no, even though it's technically released in smithy now. I'm told it'll get public documentation eventually.

docs/design/endpoints.md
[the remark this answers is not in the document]
lucix-aws: Ok, seeing another comment by Aaron further down on this same thing - there _should_ be a world in which we can generate a default middleware for operations that don't have additional parameter bindings.

docs/design/endpoints.md
lucix-aws: It _will_ generate a builder for convenience, especially since callers will be able to make these themselves (it should also simplify the codegen). I'll make sure this doc reflects that.

I'm 99% sure we would enforce struct member ordering in params like we would in any other struct, but I'll check.

docs/design/endpoints.md
lucix-aws: Mostly to follow convention w.r.t how we do path-bound request values now (nullable, but we error on construct)

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+     * with those parameters, and modify the request (and handle failure to resolve if that occurs).
+     * 
+     * Client/static/general context parameters are inspected here to determine where to source values.
+     */
+    public fun render()
+    
+    /**
+     * SDK authors extend this method to handle builtins.
+     */
+    public open fun resolveBuiltin(value: String)
```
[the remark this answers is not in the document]
lucix-aws: It's the same issue as in `resolveFunction` where we basically have to delegate to the caller to let them handle however they need to.

These APIs should probably give a little more context about the expectations made of writer calls, to that effect.

docs/design/endpoints.md
lucix-aws: You might have to speak more to what you're imagining but I think we're far better off codegenning directly, given that we have to evaluate arbitrary boolean expressions, assign intermediates, etc - it's just a task suited to codegen.

As far as how large, we have an s3 model with a ruleset available today. That is probably a pretty good indication of the maximum complexity we'd expect from an AWS ruleset.

docs/design/endpoints.md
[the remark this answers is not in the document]
lucix-aws: A more accurate smithy-esque term would be "bound" "bind" etc. SEP uses this verbiage. I'll update.

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+    {
+      "documentation": "use a special endpoint for government resources",
+      "type": "endpoint",
+      "conditions": [
+        {
+          "fn": "substring",
+          "argv": [
+            {"ref": "ResourceId"},
+            0,
+            5,
```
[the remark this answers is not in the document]
lucix-aws: The end index is exclusive.

docs/design/endpoints.md
```diff
@@ -0,0 +1,292 @@
...
+  "parameters": {
+    "ResourceId": {
+      "type": "string",
+      "required": "true"
+    }
+  },
+  "rules": [
+    {
+      "documentation": "use a special endpoint for government resources",
+      "type": "endpoint",
```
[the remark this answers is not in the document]
lucix-aws: The other type would be "error" - the fallback is just missing the type field, will fix.

docs/design/endpoints.md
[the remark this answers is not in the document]
lucix-aws: Yeah, I had thought about this, unfortunately we don't really have a good way to "label" the symbols - it would just have to be
```kotlin
// documentation field for 1st
object Endpoint0 : Endpoint ( ... )
// documentation field for 2nd
object Endpoint1 : Endpoint ( ...)
// etc.
```

I think exposing them publicly is IDEAL, though - especially since if people write their own resolvers, they might still want to use some of those endpoints.

I think it's worth asking if we can have the spec label them some way such that we can give them meaningful variable names.

docs/design/endpoints.md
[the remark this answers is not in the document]
ianbotsf: Oh if it's rendering then the verb should probably reflect that.

docs/design/endpoints.md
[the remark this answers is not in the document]
aajtodd: I disagree on exposing them unless that is something every SDK is expected to do. Anything we make public is a commitment to backwards compatibility. I'd rather retain flexibility to break the internal implementation as needed.

docs/design/endpoints.md
aajtodd: That is modeling constraint that may or may not apply here (I can't say for sure which one without more context probably)

docs/design/endpoints.md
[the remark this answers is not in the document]
ianbotsf: Right and a substring from 0-4 (inclusive) would be 5 characters long:

```
"gov.foo"
 0123456
```

docs/design/endpoints.md
```diff
@@ -63,52 +63,231 @@ Client config will include a field for the resolver unique to that service, and
...
+
+### Provider parameters
+
+Provider parameters will be generated as a data class, since backwards compatability of parameter values is guaranteed (
+field order MUST be preserved, and obsolete values are marked deprecated and not removed).
+
+The presence of required values is checked at construction, with an error thrown if any are missing.
+
+The generation includes DSL builder support for several reasons:
+1. caller convenience - SDK users are able to implement and call their own endpoint providers
```
aajtodd, line 262: > and call their own endpoint providers

Not sure I understand this part. Why would calling an endpoint provider be an argument for builders?

docs/design/endpoints.md
```diff
@@ -63,52 +63,231 @@ Client config will include a field for the resolver unique to that service, and
...
-implemented to perform to walk the rules tree and generate the core logic. This base class can be extended on by SDK
+Codegen for resolvers works much like any other extendable generated construct in smithy-kotlin: a base generator is
+implemented to walk the rules tree and generate the core logic. This base class can be extended by SDK
 authors as needed.
 
 ```kotlin
 public open class EndpointProviderGenerator(
     public val writer: KotlinWriter,
-    public val ruleset: EndpointRuleset,
+    public val ruleSet: EndpointRuleset,
```
aajtodd, line 290: nit: simply `rules` would probably work here as well

docs/design/endpoints.md
```diff
@@ -117,41 +296,53 @@ public open class EndpointProviderGenerator(
...
 ### Middleware
-Endpoint resolution, whether generated or written by the caller, is performed per-request as a pre-operation middleware.
-At runtime, endpoint resolution is comprised of two principal tasks:
-1. construct a set of provider parameters, sourcing the values as specified by the model (context params)
-2. using the sourced parameters, call the provider to resolve to an endpoint, pointing the request to it
 
-Since endpoint resolution is performed on a per-operation basis, and the sourcing of provider parameters can vary
-per-operation, middleware must now be generated per-operation (this differs from the original implementation, where the
-endpoint resolver middleware was static across operations and its implementation was part of the runtime).
+Endpoint resolution, whether generated or written by the caller, is performed per-request as a pre-serialization
```
aajtodd, line 307: if it's generating an instance of `ModifyRequestMiddleware` then it is post serialization

docs/design/endpoints.md
```diff
@@ -271,16 +471,17 @@ public class DefaultSmithyServiceEndpointProvider: SmithyServiceEndpointProvider
     }
 }
 
-public class GetResourceResolveEndpointMiddleware(
+// package *.services.smithyservice.internal
+
+internal class GetResourceResolveEndpointMiddleware(
     private val config: SmithyService.Config,
     private val request: GetResourceRequest,
```
aajtodd, line 478: Something we need to be careful of here now that I'm thinking about it is that the [initialize](https://github.com/awslabs/smithy-kotlin/blob/cfa0fd3a30b4c50b75485786f043d4e2ad803f55/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/operation/SdkOperationExecution.kt#L44) middleware phase is allowed to act on the operation input (possibly modifying it pre-serialization). We wouldn't be able to just capture the operation input at the time middleware is registered as it could change.

e.g.

```kotlin
suspend fun fooOperation(input: FooInput): FooOutput {
     val op = SdkHttpOperation<FooInput, FooOutput>.build { ... }
     
     op.install(Middleware1())
     ...
     op.install(MiddlewareN())

      // example inline middleware to demonstrate this case, same thing applies to non-inlined middleware
      op.execution.initialize.intercept { req, next ->
          val modifiedFooInput = req.subject.copy { foobar = "baz" }
          val modifiedReq = req.copy(subject = modifiedFooInput)
          next.call(modifiedReq)
      }

     // captures the original input not the one modified by the middleware that will be serialized
     op.install(FooServiceEndpointMiddleware(config, input))

}
```

Now one way around this would be to have the service endpoint middleware implement both `initialize` and `modify` phases. If it's the last `initialize` middleware to run it can capture the finalized operation input that is used during the `modify` phase.

docs/design/endpoints.md
```diff
@@ -117,41 +296,53 @@ public open class EndpointProviderGenerator(
     public fun render()
 
     /**
-     * Resolve a standard library function referenced in a ruleset.
+     * Render a standard library function referenced in a rule set.
      */
-    public open fun resolveFunction(func: String)
+    public open fun renderFunction(func: String, args: List<FunctionArg>)
```
aajtodd, line 301: If these are standard library functions then I would think `resolveFunction(name: String): Symbol` would be appropriate right? The generator obviously knows the parameters already since they are a function argument so simply resolving to a symbol would be enough.

e.g.

```kotlin
// hypothetical codegen path in render()

val xyzStdlibFn = resolveFunction("xyz")
writer.write("val xyzResult = #T(param1, param2)", xyzStdlibFn)

File: docs/design/endpoints.md [hunk 24]
```diff
@@ -117,41 +296,53 @@ public open class EndpointProviderGenerator(
...
+
+    /**
+     * Implement this API to generate additional parameters to be passed to the middleware.
+     * By default, the middleware takes two parameters:
+     * 1. service client config
+     * 2. operation's request object
+     * 
+     * The implementer is responsible for code generation of the middleware installation.
+     */
+    public open fun getMiddlewareParams(): List<Symbol>
```
aajtodd, line 339: question: Why is this necessary? Aren't params bound to client context (config), static, or operation input?

docs/design/endpoints.md
```diff
@@ -117,41 +296,53 @@ public open class EndpointProviderGenerator(
...
     
     /**
-     * SDK authors extend this method to handle builtins.
+     * SDK authors extend this method to handle built-ins.
+     * Called within the builder lambda for the parameters object.
      */
-    public open fun resolveBuiltin(value: String)
+    public open fun renderBuiltin(name: String)
```
aajtodd, line 345: Same question as above around `resolveFunction`, perhaps I'm just not seeing the bigger picture here yet...

docs/design/endpoints.md
```diff
@@ -63,52 +63,228 @@ Client config will include a field for the resolver unique to that service, and
...
+  }
+}
+```
+
+The base middleware generator cannot render the binding of this value. The AWS SDK must implement custom logic to bind
+this value from the client config.
 
 ## Codegen
+### Package "endpoints"
+The provider parameters and interface will generate to a new top-level `endpoints` package.
```
ianbotsf, line 252: **Correctness**: I assume this should be "an `endpoints` package underneath each service package (e.g., `aws.sdk.kotlin.services.s3.endpoints`)", right?

docs/design/endpoints.md
```diff
@@ -162,31 +340,35 @@ public open class EndpointResolverMiddlewareGenerator(
...
+  override fun install(op: SdkHttpOperation<GetResourceRequest, GetResourceResponse>) {
+      op.execution.initialize.intercept { req, next ->
+          // bind provider parameters and resolve endpoint, saving the value for the mutate phase
+      }
+
+      op.execution.mutate.intercept { req, next ->
+          // set the request endpoint
+      }
+  }
 }
```
ianbotsf, line 359: **Style**: Indentation inconsistency

docs/design/endpoints.md
```diff
@@ -63,52 +63,228 @@ Client config will include a field for the resolver unique to that service, and
...
+@clientContextParams(
+    optInBeta: {type: "boolean", documentation: "opt in to the beta environment for this service"}
+)
+service BasicService {
+    // ...
+}
+```
+
+The client config for this service will be generated with an `optInBeta` field, the value of which will be bound for the
+corresponding endpoint parameter field at resolution.
```
ianbotsf, line 127: **Comment**: These new examples and expanded explanations are great. Thanks!

ianbotsf (approved):

aajtodd (approved):

PR #699, opened 2022-09-08
feat: endpoints2: codegen for endpoint params

lucix-aws:
## Description of changes
Implement code generation of `EndpointParameters` for endpoint resolution v2.


smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,147 @@
...
+            it.deprecated.getOrNull(),
+        )
+    }
+
+    fun render() {
+        renderDocumentation()
+        writer.withBlock("public data class EndpointParameters(", ")") {
+            params.forEach {
+                renderConstructorParam(it)
+            }
```
ianbotsf, line 42: **Style**: Consider `params.forEach(::renderConstructorParam)`

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,147 @@
...
+            params.forEach {
+                write("@property #L #L", it.name, it.documentation)
+            }
+        }
+    }
+
+    private fun renderConstructorParam(param: KotlinEndpointParameter) {
+        if (param.deprecated != null) {
+            param.deprecated.writeKotlinAnnotation(writer)
+        }
```
ianbotsf, line 67: **Style**: Consider `param.deprecated?.run { writeKotlinAnnotation(writer) }`

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,147 @@
...
+                    it.deprecated.writeKotlinAnnotation(writer)
+                }
+                write("public var #L: #P = #L", it.name, it.type, it.defaultLiteral)
+                write("")
+            }
+            withBlock("public fun build(): EndpointParameters {", "}") {
+                withBlock("return EndpointParameters(", ")") {
+                    params.forEach {
+                        if (it.deprecated != null) {
+                            writeInline("@Suppress(#S)", "DEPRECATION")
```
ianbotsf, line 105: **Nit**: No need for the string template here.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,147 @@
...
+        writer.withBlock("public companion object {", "}") {
+            write("public operator fun invoke(block: Builder.() -> Unit): EndpointParameters = Builder().apply(block).build()")
+        }
+    }
+
+    private fun renderInit() {
+        writer.withBlock("init {", "}") {
+            params.forEach {
+                if (it.isRequired) {
+                    write("""require(#1L != null) { "endpoint provider parameter #1L is required" }""", it.name)
```
ianbotsf, line 81: **Nit**: `require` → `requireNotNull`

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,147 @@
...
+    asString().replaceFirstChar(Char::lowercase)
+
+private fun ParameterType.toSymbol(): Symbol =
+    buildSymbol {
+        namespace = "kotlin"
+        name = when (this@toSymbol) {
+            ParameterType.STRING -> "String"
+            ParameterType.BOOLEAN -> "Boolean"
+        }
+    }
```
ianbotsf, line 137: **Nit**: We already have [pre-built symbols in `KotlinTypes`](https://github.com/awslabs/smithy-kotlin/blob/main/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/KotlinTypes.kt). This could be simplified:

```kotlin
private fun ParameterType.toSymbol(): Symbol = when (this) {
    ParameterType.STRING -> KotlinTypes.String
    ParameterType.BOOLEAN -> KotlinTypes.Boolean
}
```

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,147 @@
...
+
+private fun Value.toKotlinLiteral(): String =
+    when (this) {
+        is Value.Str -> expectString().doubleQuote()
+        is Value.Bool -> if (expectBool()) "true" else "false"
+        else -> throw IllegalArgumentException("unrecognized parameter value type")
+    }
+
+private fun Parameter.Deprecated.writeKotlinAnnotation(writer: KotlinWriter) =
+    writer.write("@Deprecated(#S)", message ?: "")
```
ianbotsf, line 147: **Comment**: Let's add a basic deprecation string in the event of a null or blank message on the parameter itself. Something like "This field is deprecated and no longer recommended for use."

ianbotsf (changes requested):

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,205 @@
...
+            renderToString()
+            write("")
+            renderCopy()
+            write("")
+            renderBuilder()
+        }
+    }
+
+    private fun renderFields() {
+        params.forEach {
```
ianbotsf, line 66: **Question**: In `StructureGenerator`, we sort members before rendering them—not only at field declaration but in `equals`, `hashCode`, etc. This generally makes classes with a lot of members easier to navigate and parse. Should we consider doing the same here?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+
+    private fun renderInit() {
+        writer.withBlock("init {", "}") {
+            params.forEach {
+                if (it.isRequired) {
+                    write("""requireNotNull(#1L) { "endpoint provider parameter #1L is required" }""", it.name)
+                }
+            }
```
ianbotsf, line 91: **Question**: In `StructureGenerator`, we do this at the field declaration site (e.g., `public val foo: String = requireNotNull(builder.foo) { "..." }`). Is there any reason to favor using an explicit `init` block for this?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,205 @@
...
+            write("return true")
+        }
+    }
+
+    private fun renderHashCode() {
+        writer.withBlock("public override fun hashCode(): Int {", "}") {
+            if (params.isEmpty()) {
+                write("return 0")
+                return@withBlock
+            }
```
ianbotsf, line 112: **Question**: In `StructureGenerator`, we use `this::class.hashCode()` as the zero-param value which seems like it would provide better hash bucketing in situations where multiple classes are mixed. Are there drawbacks to doing the same here?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,205 @@
...
+            params.forEachIndexed { index, it ->
+                ensureSuppressDeprecation(it)
+                write(
+                    if (index < params.size - 1) "append(\"#1L=$#1L,\")" else "append(\"#1L=$#1L\")",
+                    it.name,
+                )
+            }
+            write("append(\")\")")
+        }
+    }
```
ianbotsf, line 136: **Question**: I'm guessing "no" but can any of these parameters be "sensitive" such that we'd want to suppress their output?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,205 @@
...
+
+    private fun renderToString() {
+        writer.withBlock("public override fun toString(): String = buildString {", "}") {
+            write("append(\"#L(\")", CLASS_NAME)
+            params.forEachIndexed { index, it ->
+                ensureSuppressDeprecation(it)
+                write(
+                    if (index < params.size - 1) "append(\"#1L=$#1L,\")" else "append(\"#1L=$#1L\")",
```
ianbotsf, line 130: **Style**: Consider pulling the comma out to a dedicated `val` and `write` argument instead of switching directly in the template argument:

```kotlin
val separator = if (index < sortedMembers.size - 1) "," else ")"
...
write("""append("#1L=$#1L#2L")""", it.name, separator)
```

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,205 @@
...
+            write("")
+            renderCopy()
+            write("")
+            renderBuilder()
+        }
+    }
+
+    private fun renderFields() {
+        params.forEach {
+            writer.ensureSuppressDeprecation(it)
```
ianbotsf, line 67: **Question**: Why suppress the deprecation warning on the field itself (i.e., the same place we declare that it _is_ deprecated)?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,205 @@
...
+
+    private fun renderCopy() {
+        writer.withBlock("public fun copy(block: Builder.() -> Unit = {}): #L {", "}", CLASS_NAME) {
+            withBlock("return Builder().apply {", "}") {
+                params.forEach {
+                    ensureSuppressDeprecation(it)
+                    write("#1L = this@#2L.#1L", it.name, CLASS_NAME)
+                }
+            }
+            write(".apply(block).build()")
```
ianbotsf, line 146: **Nit**: There's no need for two `apply` blocks. Just invoke `build` inside the first `apply` after setting all the parameter values:

```kotlin
withBlock("return Builder().apply {", "}") {
    params.forEach {
        ensureSuppressDeprecation(it)
        write("#1L = this@#2L.#1L", it.name, CLASS_NAME)
    }
    write("block()")
}
```

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
lucix-aws: It's suppressing the deprecation on the assignment from the corresponding builder field.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
lucix-aws: The spec doesn't provide for marking a field as such but it's a good callout. I'll bring it up w/ SEP authors.

lucix-aws: Especially since theoretically any field marked sensitive in smithy could be bound to a parameter via `@contextParam`.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+}
+
+// kotlin-mapped representation of an endpoint parameter
+private data class KotlinEndpointParameter(
+    val name: String,
+    val type: Symbol,
+    val isRequired: Boolean,
+    // All endpoint params are nullable. Standard symbol rendering suppresses the default value for boxed types, but
```
aajtodd, line 166: Sort of, this could be changed. You can also register a new property formatter on the writer if warranted. What you have is fine though, just clarifying this isn't set in stone or how it must work or anything

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
aajtodd: Agreed. Also if we do keep an `init{...}` block for whatever reason I'd consider moving it right after field declaration (before the companion object)

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+
+    private fun renderEquals() {
+        writer.withBlock("public override fun equals(other: Any?): Boolean {", "}") {
+            write("if (this === other) return true")
+            write("if (other !is #L) return false", CLASS_NAME)
+            params.forEach {
+                ensureSuppressDeprecation(it)
+                write("if (this.#1L != other.#1L) return false", it.name)
```
aajtodd, line 101: remind me again, what are the possible types a parameter can be? I'm assuming we don't need consider `ByteArray` here for instance

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
aajtodd: Yeah unless there is good reason I wouldn't deviate

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,205 @@
...
+    private val writer: KotlinWriter,
+    rules: EndpointRuleset,
+) {
+    private val params: List<KotlinEndpointParameter> = rules.parameters.toList().map {
+        KotlinEndpointParameter(
+            it.name.toKotlin(),
+            it.type.toSymbol(),
+            it.isRequired,
+            it.defaultValue.getOrNull()?.toKotlinLiteral() ?: "null",
+            it.documentation.getOrNull(),
```
aajtodd, line 38: Question: Do we run the ruleset documentation through our sanitization? Is that just free because it's part of the model or do we need to explicitly handle that?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -0,0 +1,205 @@
...
+                    if (index < params.size - 1) "append(\"#1L=$#1L,\")" else "append(\"#1L=$#1L\")",
+                    it.name,
+                )
+            }
+            write("append(\")\")")
+        }
+    }
+
+    private fun renderCopy() {
+        writer.withBlock("public fun copy(block: Builder.() -> Unit = {}): #L {", "}", CLASS_NAME) {
```
aajtodd, line 139: this is fine, if you wanted to make it match what we do for structs though we handle it slightly different. We add a constructor to the `Builder` itself and delegate to that.

Random example from lexv2 model
```kotlin

    public inline fun copy(block: Builder.() -> kotlin.Unit = {}): aws.sdk.kotlin.services.lexruntimev2.model.ConfidenceScore = Builder(this).apply(block).build()



    public class Builder {
        @PublishedApi
        internal constructor(x: aws.sdk.kotlin.services.lexruntimev2.model.ConfidenceScore) : this() {
            this.score = x.score
        }
     }
```

aajtodd (approved):
Couple questions/suggestions but nothing blocking from me.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
lucix-aws: `String` and `Boolean` only in the current spec

ianbotsf (approved):

PR #701, opened 2022-09-12
Fix crt stream leak via tracking transfer position

ianbotsf:
## Issue \#

(none)

## Description of changes

Fix the CRT streaming delay bug by tracking the current transfer position.

Mutually exclusive with #700 and #703.


ianbotsf:
We've decided to abandon this approach in favor of #700.

PR #702, opened 2022-09-13
fix: properly check if a member can be nullable

sugmanue:
## Description of changes
Change the way we compute the nullability of members. Before we were only looking at the box trait in the target type and ignoring the trait on the member itself. See for instance the test `can read box trait from member` which has the following model:
```
        namespace com.test
        structure MyStruct {
           @box
           foo: MyFoo
        }
        long MyFoo
```
The code was incorrectly assuming that the foo member was not nullable. This change accounts for this by using the `NullableIndex` which correctly computes whether or not the member or the target type are nullable. 

**NOTE** This is a draft pull request as it depends on the pending release of smithy 1.25.0 scheduled for tomorrow, but sending along to get a head start on the review of the changes.

### DIFF
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt
@@ -9,6 +9,7 @@ import software.amazon.smithy.kotlin.codegen.KotlinSettings
 import software.amazon.smithy.kotlin.codegen.lang.kotlinReservedWords
 import software.amazon.smithy.kotlin.codegen.model.*
 import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.knowledge.NullableIndex
 import software.amazon.smithy.model.shapes.*
 import software.amazon.smithy.model.traits.SparseTrait
 import software.amazon.smithy.model.traits.StreamingTrait
@@ -26,6 +27,7 @@ class KotlinSymbolProvider(private val model: Model, private val settings: Kotli
     private val service = model.expectShape<ServiceShape>(settings.service)
     private val logger = Logger.getLogger(javaClass.name)
     private val escaper: ReservedWordSymbolProvider.Escaper
+    private val nullableIndex = NullableIndex(model)
 
     // model depth; some shapes use `toSymbol()` internally as they convert (e.g.) member shapes to symbols, this tracks
     // how deep in the model we have recursed
@@ -168,8 +170,11 @@ class KotlinSymbolProvider(private val model: Model, private val settings: Kotli
         val targetShape =
             model.getShape(shape.target).orElseThrow { CodegenException("Shape not found: ${shape.target}") }
 
-        val targetSymbol = toSymbol(targetShape)
-
+        val targetSymbol = if (nullableIndex.isMemberNullable(shape, NullableIndex.CheckMode.CLIENT_ZERO_VALUE_V1)) {
+            toSymbol(targetShape).toBuilder().boxed().build()
+        } else {
+            toSymbol(targetShape)
+        }
         // figure out if we are referencing an event stream or not.
         // NOTE: unlike blob streams we actually re-use the target (union) shape which is why we can't do this
         // when visiting a unionShape() like we can for blobShape()
@@ -249,9 +254,7 @@ class KotlinSymbolProvider(private val model: Model, private val settings: Kotli
         val builder = Symbol.builder()
             .putProperty(SymbolProperty.SHAPE_KEY, shape)
             .name(typeName)
-
-        val explicitlyBoxed = shape?.hasTrait<@Suppress("DEPRECATION") software.amazon.smithy.model.traits.BoxTrait>() ?: false
-        if (explicitlyBoxed || boxed) {
+        if (boxed) {
             builder.boxed()
         }
         return builder
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
@@ -9,10 +9,8 @@ import software.amazon.smithy.codegen.core.Symbol
 import software.amazon.smithy.kotlin.codegen.core.*
 import software.amazon.smithy.kotlin.codegen.lang.KotlinTypes
 import software.amazon.smithy.kotlin.codegen.model.*
-import software.amazon.smithy.model.shapes.BlobShape
-import software.amazon.smithy.model.shapes.MemberShape
-import software.amazon.smithy.model.shapes.ShapeType
-import software.amazon.smithy.model.shapes.StructureShape
+import software.amazon.smithy.model.knowledge.NullableIndex
+import software.amazon.smithy.model.shapes.*
 import software.amazon.smithy.model.traits.ErrorTrait
 import software.amazon.smithy.model.traits.HttpLabelTrait
 import software.amazon.smithy.model.traits.RetryableTrait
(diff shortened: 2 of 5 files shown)


smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt
```diff
@@ -39,6 +41,7 @@ class KotlinSymbolProvider(private val model: Model, private val settings: Kotli
             // only escape words when the symbol has a definition file to prevent escaping intentional references to built-in shapes
             .escapePredicate { _, symbol -> symbol.definitionFile.isNotEmpty() }
             .buildEscaper()
+        nullableIndex = NullableIndex(model)
```
ianbotsf, line 44: **Nit**: There's no need to set this in the `init` block. It can be set directly after the field definition on L26:

```kotlin
private val nullableIndex = NullableIndex(model)
```

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt
```diff
@@ -168,7 +171,10 @@ class KotlinSymbolProvider(private val model: Model, private val settings: Kotli
         val targetShape =
             model.getShape(shape.target).orElseThrow { CodegenException("Shape not found: ${shape.target}") }
 
-        val targetSymbol = toSymbol(targetShape)
+        var targetSymbol = toSymbol(targetShape)
+        if (nullableIndex.isMemberNullable(shape, NullableIndex.CheckMode.CLIENT_ZERO_VALUE_V1)) {
+            targetSymbol = targetSymbol.toBuilder().boxed().build()
+        }
```
ianbotsf, line 177: **Nit**: There's no need to make `targetSymbol` mutable:

```kotlin
val targetSymbol = if (nullableIndex.isMemberNullable(shape, NullableIndex.CheckMode.CLIENT_ZERO_VALUE_V1)) {
    toSymbol(targetShape).toBuilder().boxed().build()
} else {
    toSymbol(targetShape)
}
```

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
```diff
@@ -30,6 +31,7 @@ class StructureGenerator(
     private val symbolProvider = ctx.symbolProvider
     private val model = ctx.model
     private val symbol = ctx.symbolProvider.toSymbol(ctx.shape)
+    private val nullableIndex = NullableIndex(ctx.model)
```
ianbotsf, line 34: **Suggestion**: We shouldn't need to reinstantiate the `NullabilityIndex` when it's already present in the `symbolProvider`. I suggest encapsulating the logic to determine if a member is nullable into `KotlinSymbolProvider` and then call that method from this class.

smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/core/SymbolProviderTest.kt
```diff
@@ -97,6 +98,41 @@ class SymbolProviderTest {
...
+        """.prependNamespaceAndService().toSmithyModel()
+
+        val provider: SymbolProvider = KotlinCodegenPlugin.createSymbolProvider(model)
+        val member = model.expectShape<MemberShape>("com.test#MyStruct\$foo")
+        val memberSymbol = provider.toSymbol(member)
+        assertEquals("kotlin", memberSymbol.namespace)
+        assertEquals("null", memberSymbol.defaultValue())
+        assertTrue(memberSymbol.isBoxed)
+    }
     @Test
```
ianbotsf, line 136: **Nit**: Missing newline between methods.

ianbotsf (commented):
Cannot fully review until Smithy 1.25 is release, this PR is refreshed, and the required checks pass. Leaving some intermediate feedback in the meantime...

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt
sugmanue: Make sense, I will change it accordingly.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt
sugmanue: Make sense, I will change it accordingly.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
aajtodd: In most places though I think we rely on the generic `SymbolProvider` rather than the actual `KotlinSymbolProvider` implementation

aajtodd (commented):
As Ian said I'll wait for approval until all the checks can run.

This looks ok for now as far as fixing the specific member trait issue but I am wondering if we want to move away from the "boxed" terminology that is pervasive in the current code. ([example](https://github.com/awslabs/smithy-kotlin/blob/v0.12.5/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/SymbolExt.kt#L43-L74)). 

I'm assuming we will have follow on PR(s) to handle updating how default values are handled?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
sugmanue: I pinged Ian about this yesterday, this change will snowball into a ton of type changes across several files, I don't feel like it's a net-win to be honest, but I already have the changes so I might as well post them for you guys to decide.

ianbotsf: I'm fine with replacing `SymbolProvider` with `KotlinSymbolProvider` as you've posted...despite the number of changes it incurs, I think it's the right change to make.

One alternative worth discussing is retaining the previous type dependencies and just adding an extension method for `isMemberNullable` onto `SymbolProvider`:

```kotlin
fun SymbolProvider.isMemberNullable(member: MemberShape): Boolean = when (this) {
    is KotlinSymbolProvider -> isMemberNullable(member)
    else -> throw NotImplementedError("Cannot determine nullability of $member via $this")
}
```

sugmanue: Synced up offline with Ian, we will keep the `NullableIndex` here for the time being.

aajtodd (approved):
Missing changelog entry. Please fix before merge: https://github.com/awslabs/smithy-kotlin/blob/main/CONTRIBUTING.md#changelog

ianbotsf (approved):

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt
```diff
@@ -53,6 +55,10 @@ class KotlinSymbolProvider(private val model: Model, private val settings: Kotli
         }
     }
 
+    fun isMemberNullable(shape: MemberShape): Boolean {
+        return nullableIndex.isMemberNullable(shape, NullableIndex.CheckMode.CLIENT_ZERO_VALUE_V1_NO_INPUT)
+    }
+
```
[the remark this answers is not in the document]
ianbotsf: Ignore this, it's old feedback that wasn't submitted in time for the approval/merge.

PR #710, opened 2022-09-30
bugfix: correct nullability semantics around generation and usage of struct members

lucix-aws:
## Description of changes

- Fix the nullability index check mode in the symbol provider (to prevent member signature breakage)
- Use a consistent symbol reference for future null checks, instead of re-checking with the index, when generating references to members

### DIFF
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
@@ -148,7 +148,7 @@ class StructureGenerator(
     // Return the appropriate hashCode fragment based on ShapeID of member target.
     private fun selectHashFunctionForShape(member: MemberShape): String {
         val targetShape = model.expectShape(member.target)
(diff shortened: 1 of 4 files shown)


smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
```diff
@@ -148,7 +148,7 @@ class StructureGenerator(
     // Return the appropriate hashCode fragment based on ShapeID of member target.
     private fun selectHashFunctionForShape(member: MemberShape): String {
         val targetShape = model.expectShape(member.target)
-        val isNullable = nullableIndex.isMemberNullable(member, NullableIndex.CheckMode.CLIENT_ZERO_VALUE_V1_NO_INPUT)
+        val isNullable = memberNameSymbolIndex[member]!!.second.isBoxed
```
sugmanue, line 149: I think we can remove the `nullableIndex` after this change.

ianbotsf (changes requested):
These changes look correct but we didn't catch this problem until a model change failed compilation. I think we're missing some unit tests that verify how `default` is codegenned, handled, etc.

ianbotsf (approved):

sugmanue (approved):

PR #719, opened 2022-10-14
feat(rt): add `MIN`/`MAX` values to `Instant`

lauzadis:
Add `Instant.MIN` and `Instant.MAX` which represent the minimum and maximum possible values of an Instant.

## Issue \#
N/A

## Description of changes
This is going to be used by a twin PR in aws-sdk-kotlin to set the expiration time of Credentials when no expiration data is received. When no expiration data is received, we should treat the Credentials as non-expiring. `Instant.MAX` will be used to ensure these Credentials are not refreshed at any point in their use.

### DIFF
--- a/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/time/InstantJVM.kt
+++ b/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/time/InstantJVM.kt
@@ -129,12 +129,12 @@ public actual class Instant(internal val value: jtInstant) : Comparable<Instant>
         /**
          * Create an [Instant] with the minimum possible value
          */
-        public val MIN: Instant = Instant(jtInstant.MIN)
+        public actual val MIN_VALUE: Instant = Instant(jtInstant.MIN)
 
         /**
          * Create an [Instant] with the maximum possible value
          */
-        public val MAX: Instant = Instant(jtInstant.MAX)
+        public actual val MAX_VALUE: Instant = Instant(jtInstant.MAX)
     }
 }
 
--- a/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/time/InstantJVM.kt
+++ b/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/time/InstantJVM.kt
@@ -125,6 +125,16 @@ public actual class Instant(internal val value: jtInstant) : Comparable<Instant>
          * Create an [Instant] from the current system time
          */
         public actual fun now(): Instant = Instant(jtInstant.now())
+
+        /**
+         * Create an [Instant] with the minimum possible value
+         */
+        public val MIN: Instant = Instant(jtInstant.MIN)
+
+        /**
+         * Create an [Instant] with the maximum possible value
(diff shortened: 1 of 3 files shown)


runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/time/InstantJVM.kt
```diff
@@ -125,6 +125,16 @@ public actual class Instant(internal val value: jtInstant) : Comparable<Instant>
...
+
+        /**
+         * Create an [Instant] with the minimum possible value
+         */
+        public val MIN: Instant = Instant(jtInstant.MIN)
+
+        /**
+         * Create an [Instant] with the maximum possible value
+         */
+        public val MAX: Instant = Instant(jtInstant.MAX)
```
ianbotsf, line 137: **Correctness**: I believe these should be `actual`s with corresponding `expect`s in **Instant.kt**.

ianbotsf (changes requested):

aajtodd (commented):
> When no expiration data is received, we should treat the Credentials as non-expiring.

Question/Concern: If no expiration is received why would we set it? It's a [nullable](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/auth/aws-credentials/common/src/aws/smithy/kotlin/runtime/auth/awscredentials/Credentials.kt#L18) field. The lack of a value would be treated as no expiration would it not?

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/time/InstantJVM.kt
lauzadis: Yeah you're right, I completely missed the `actual` and `expects`, I'll update it. And yeah, they should have `_VALUE` for consistency. I think .MAX and .MIN look nicer but we should follow the standard set by Kotlin and Java.

lauzadis:
> > When no expiration data is received, we should treat the Credentials as non-expiring.
> 
> Question/Concern: If no expiration is received why would we set it? It's a [nullable](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/auth/aws-credentials/common/src/aws/smithy/kotlin/runtime/auth/awscredentials/Credentials.kt#L18) field. The lack of a value would be treated as no expiration would it not?

Currently if we receive a null expiration field, we will [set the expiration to `clock.now() + expireCredentialsAfter`](https://github.com/awslabs/aws-sdk-kotlin/blob/main/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CachedCredentialsProvider.kt#L63). But, when we receive a null expiration field in the process credentials, we are supposed to treat them as non-expiring credentials and never refresh them (aka set expiration to `Instant.MAX_VALUE`)

This may be a better discussion on my PR in aws-sdk-kotlin, which will be posted soon

ianbotsf (approved):
Approving since these changes in isolation look good to me now, although I am curious about the outcome of the discussion about how these will be used in credentials.

aajtodd (approved):

PR #721, opened 2022-10-19
feat(codegen): add `order` parameter to ClientConfigProperty to be used for ordering configuration dependencies

lauzadis:
Adds an `order` Integer parameter to ClientConfigProperty which is used to order configuration dependencies.

## Issue \#
[awslabs/aws-sdk-kotlin/711](https://github.com/awslabs/aws-sdk-kotlin/issues/711)

## Description of changes
This was added in order to support passing of certain client configuration properties into other client configuration properties. For example, client configurations can specify an HttpEngine to be used for all requests, which must be passed into the DefaultChainCredentialsProvider which is also a client configuration. This introduces a dependency between client configuration properties which I addressed with the `order` parameter.

When rendering the client configuration properties, we will sort them by `order` (new behavior) and then alphabetically by property name (as it's done today).  This will ensure all client configuration property dependencies are satisfied before we try to access them.

### DIFF
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
@@ -70,6 +70,11 @@ class ClientConfigProperty private constructor(builder: Builder) {
     val additionalImports: List<Symbol> = builder.additionalImports
 
     /**
+     * The priority order of rendering the property. Used to manage dependencies between configuration properties.
+     */
+    val order: Int = builder.order
+
+    /**
      * Flag indicating if this property stems from some base class and needs an override modifier when rendered
      */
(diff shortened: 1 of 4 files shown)


lucix-aws (approved):

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
```diff
@@ -237,6 +244,7 @@ object KotlinClientRuntimeConfigProperty {
             NOTE: The caller is responsible for managing the lifetime of the engine when set. The SDK
             client will not close it when the client is closed.
             """.trimIndent()
+            order = -1
```
aajtodd, line 247: Placing it at `-1` means nothing can be inserted in-between this property and something else at `0`. I'd add a bit of room for growth here (e.g. `-100`).

aajtodd (approved):
Fix and ship

PR #724, opened 2022-10-27
feat(rt): add CRC32C algorithm

lauzadis:
Adds the CRC32C algorithm

## Issue \#
Contributes to the completion of flexible checksums: https://github.com/awslabs/aws-sdk-kotlin/issues/727

## Description of changes
This is a hashing algorithm we must support as part of the flexible checksums specification. It's exactly the same as CRC32, but uses a different polynomial. We get CRC32 from [java.util.zip](https://docs.oracle.com/javase/8/docs/api/java/util/zip/CRC32.html), but they did not begin to support CRC32C until Java 9. We want to support Java 8 at a minimum, so we must implement it ourselves. The Java SDK also [implements it themselves](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/checksums/factory/SdkCrc32C.java).

Validated using https://crccalc.com/.


runtime/hashing/jvm/src/aws/smithy/kotlin/runtime/hashing/Crc32cJVM.kt
```diff
@@ -0,0 +1,664 @@
...
+ *
+ * Copied from Hadoop 2.3.6:
+ * https://git-wip-us.apache.org/repos/asf?p=hadoop.git;a=blob_plain;
+ * f=hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/PureJavaCrc32C.java;
+ * hb=2120de588b92b9f22b1cc4188761d6a8c61aa778
+ * <p>
+ * This class is Not ThreadSafe
+ *
+ */
+private class CRC32CImpl : Checksum {
```
aajtodd, line 66: Is there any reason this can't live in common? 

The only JVM specific thing is inheriting `Checksum` which we don't actually need do we? 

If it lives in common we only have to implement it once. It's possible that on native we may end up using CRT bindings so you could keep the `expect` / `actual` to allow each platform to choose what is best but moving this to common would allow each platform to use this pure kotlin version.

runtime/hashing/common/test/aws/smithy/kotlin/runtime/hashing/Crc32cTest.kt
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.hashing
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class Crc32cTest {
```
aajtodd, line 10: Are there any other tests to add here? Maybe take a look at the [CRT tests](https://github.com/awslabs/aws-checksums/blob/main/tests/crc_test.c) to see if there are other test cases we may want to add to this test suite

runtime/hashing/jvm/src/aws/smithy/kotlin/runtime/hashing/Crc32cJVM.kt
```diff
@@ -0,0 +1,664 @@
...
+public actual class Crc32c : Crc32cBase() {
+    private val md = CRC32CImpl()
+
+    override fun update(input: ByteArray, offset: Int, length: Int): Unit = md.update(input, offset, length)
+
+    override fun digest(): ByteArray {
+        val x = digestValue()
+        reset()
+        return byteArrayOf(
+            ((x shr 24) and 0xffu).toByte(),
```
aajtodd, line 20: Why are we changing endianess here? I'm not sure we actually consume this anywhere but I'm curious where this is stemming from

lauzadis: I initially copied this method from Crc32, which [says that the .digest() byte array should be in big-endian form](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/Crc32.kt#L18)

big-endian means that the most significant byte should be placed in the array first, and the rest of the bytes should be placed in order of decreasing significance. we should use `shr` to accomplish that behavior.

___
example:

The CRC32C value of "foobar" is 224353407 as an unsigned integer, which is `0000 1101 0101 1111 0101 1100 0111 1111` in binary. With `0000 1101 0101 1111 0101 1100 0111 1111 shl 24 and 0xffu`, the first element in the byte array becomes 0, but it should be the top byte `0000 1101`, which is 13.

aajtodd: Right I wasn't questioning the implementation, you are correct that your change results in big-endian. I was just questioning _why_ we were changing the endianess of the return value. Is there a spec somewhere dictating how the CRC value should be encoded or are we just arbitrarily making a decision here where either big/little endian would be fine and you were just updating to reflect the comment?

lauzadis: I wasn't considering any spec when making the change, just updating to reflect the comment. Both little-endian or big-endian would work, but the original configuration with `shl` was neither because we were truncating the top 24 bits

runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/Crc32c.kt
```diff
@@ -0,0 +1,676 @@
...
+@InternalApi
+public expect class Crc32c() : Crc32cBase
+
+/**
+ * Compute the CRC32C hash of the current [ByteArray]
+ */
+@InternalApi
+public fun ByteArray.crc32c(): UInt = Crc32c().apply { update(this@crc32c) }.digestValue()
+
+public actual class Crc32c : Crc32cBase() {
```
aajtodd, line 32: Either
(1) remove `expect`/`actual` all together (only common should have `expect`, `actual` is used by each platform to provide a real/actual implementation of the expected thing) and simplify the type hierarchy here
OR
(2) move `CRC32CImpl` to it's own file making it `internal` and move this `actual` to `Crc32cJVM.kt` in the JVM sourceset

runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/Crc32c.kt
```diff
@@ -0,0 +1,676 @@
...
+    private val md = CRC32CImpl()
+
+    override fun update(input: ByteArray, offset: Int, length: Int): Unit = md.update(input, offset, length)
+
+    override fun digest(): ByteArray {
```
aajtodd, line 37: This doesn't change for each `actual`, we can push this to `Crc32cBase` (same thing with `Crc32` actually)

runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/Crc32c.kt
lauzadis: I think I misinterpreted your previous comment about keeping expect / actual in case we want to use CRT bindings for Kotlin Native. For now I've removed the `expect` / `actual` altogether

runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/Crc32c.kt
lauzadis: Would it be smart to simplify further and combine Crc32cBase and Crc32Base?

aajtodd: Yeah I think so, good catch

aajtodd (approved):

PR #731, opened 2022-11-11
feat(rt): implement `aws-chunked` content encoding

lauzadis:
Implements the `aws-chunked` content encoding.

- https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html
- https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming-trailers.html

## Issue \#
https://github.com/awslabs/aws-sdk-kotlin/issues/747

## Description of changes
This change is required to achieve feature parity with other SDKs. It adds the `aws-chunked` content encoding, which can be used to upload data in chunks.

It works by wrapping an underlying `SdkByteReadChannel`. When reads are performed, the AwsChunked class will read from the underlying channel and transform the content to aws-chunked.

 
 
### DIFF
--- a/runtime/auth/aws-signing-common/build.gradle.kts
+++ b/runtime/auth/aws-signing-common/build.gradle.kts
@@ -7,6 +7,8 @@ description = "Common types for AWS signing"
 extra["displayName"] = "Smithy :: Kotlin :: AWS Signing Common"
 extra["moduleName"] = "aws.smithy.kotlin.runtime.auth.signing.awssigning"
 
+val coroutinesVersion: String by project
+
 kotlin {
     sourceSets {
         commonMain {
@@ -14,6 +16,14 @@ kotlin {
                 api(project(":runtime:auth:aws-credentials"))
                 api(project(":runtime:protocol:http"))
                 implementation(project(":runtime:logging"))
+            }
+        }
+
+        commonTest {
+            dependencies {
+                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:$coroutinesVersion")
+                implementation(project(":runtime:auth:aws-signing-tests"))
+                implementation(project(":runtime:auth:aws-signing-default"))
             }
         }
 
--- a/runtime/auth/aws-signing-common/build.gradle.kts
+++ b/runtime/auth/aws-signing-common/build.gradle.kts
@@ -7,8 +7,6 @@ description = "Common types for AWS signing"
 extra["displayName"] = "Smithy :: Kotlin :: AWS Signing Common"
 extra["moduleName"] = "aws.smithy.kotlin.runtime.auth.signing.awssigning"
 
-val coroutinesVersion: String by project
-
 kotlin {
     sourceSets {
         commonMain {
@@ -16,14 +14,6 @@ kotlin {
                 api(project(":runtime:auth:aws-credentials"))
                 api(project(":runtime:protocol:http"))
                 implementation(project(":runtime:logging"))
-            }
-        }
-
-        commonTest {
-            dependencies {
-                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:$coroutinesVersion")
-                implementation(project(":runtime:auth:aws-signing-tests"))
-                implementation(project(":runtime:auth:aws-signing-default"))
             }
         }
 
--- a/runtime/auth/aws-signing-common/build.gradle.kts
+++ b/runtime/auth/aws-signing-common/build.gradle.kts
@@ -18,13 +18,6 @@ kotlin {
                 implementation(project(":runtime:logging"))
             }
         }
-        commonTest {
-            dependencies {
-                implementation(project(":runtime:auth:aws-signing-default"))
-                implementation(project(":runtime:auth:aws-signing-tests"))
-                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:$coroutinesVersion")
-            }
-        }
 
         commonTest {
             dependencies {
--- a/runtime/auth/aws-signing-common/build.gradle.kts
+++ b/runtime/auth/aws-signing-common/build.gradle.kts
@@ -18,6 +18,13 @@ kotlin {
                 implementation(project(":runtime:logging"))
             }
         }
+        commonTest {
+            dependencies {
+                implementation(project(":runtime:auth:aws-signing-default"))
+                implementation(project(":runtime:auth:aws-signing-tests"))
+                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:$coroutinesVersion")
+            }
+        }
 
         commonTest {
             dependencies {
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunked.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunked.kt
@@ -192,7 +192,7 @@ internal abstract class AbstractAwsChunked(
         }.encodeToByteArray()
 
         chunkOffset = 0
-        return chunkHeader + chunkBody + "\r\n".encodeToByteArray()
+        return chunkHeader + chunkBody
     }
 
     /**
@@ -206,15 +206,19 @@ internal abstract class AbstractAwsChunked(
      * @return a [ByteArray] containing the trailing headers in aws-chunked encoding, ready to send on the wire
      */
     suspend fun getTrailingHeadersChunk(trailingHeaders: Headers): ByteArray {
-
-        var trailerBody = trailingHeaders.entries().joinToString { entry ->
-            "${entry.key}:${entry.value.joinToString(",")}\r\n"
-        }.encodeToByteArray()
+        var trailerBody = trailingHeaders.entries().map {
+            entry -> buildString {
+                append(entry.key)
+                append(":")
+                append(entry.value.joinToString(","))
+                append("\r\n")
+            }.encodeToByteArray()
+        }.reduce { acc, bytes -> acc + bytes }
 
         val trailerSignature = signer.signChunkTrailer(trailerBody, previousSignature, signingConfig).signature
         previousSignature = trailerSignature
 
-        trailerBody += "x-amz-trailer-signature:$trailerSignature\r\n".encodeToByteArray()
+        trailerBody += "x-amz-trailer-signature:${trailerSignature.decodeToString()}\r\n".encodeToByteArray()
 
         chunkOffset = 0
         return trailerBody
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunked.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunked.kt
@@ -13,7 +13,7 @@ internal abstract class AbstractAwsChunked(
     private val signer: AwsSigner,
     private val signingConfig: AwsSigningConfig,
     private var previousSignature: ByteArray,
-    private var trailingHeaders: Headers = Headers.Empty
+    private var trailingHeaders: Headers = Headers.Empty,
 ) : SdkByteReadChannel by chan {
     companion object {
         const val CHUNK_SIZE_BYTES: Int = 65536
@@ -154,7 +154,7 @@ internal abstract class AbstractAwsChunked(
                 }
             }
 
-            return chunk + "\r\n".encodeToByteArray()  // final CRLF to signal end of chunked transaction
+            return chunk + "\r\n".encodeToByteArray() // final CRLF to signal end of chunked transaction
         }
 
         val chunkBody = chan.readRemaining(CHUNK_SIZE_BYTES)
@@ -207,7 +207,8 @@ internal abstract class AbstractAwsChunked(
      */
     suspend fun getTrailingHeadersChunk(trailingHeaders: Headers): ByteArray {
         var trailerBody = trailingHeaders.entries().map {
-            entry -> buildString {
+                entry ->
+            buildString {
                 append(entry.key)
                 append(":")
                 append(entry.value.joinToString(","))
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunked.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunked.kt
@@ -139,7 +139,7 @@ internal abstract class AbstractAwsChunked(
             append(";")
             append("chunk-signature=")
             appendLine(chunkSignature)
-        }.toByteArray()
+        }.encodeToByteArray()
 
         chunkOffset = 0
         return chunkHeader + chunkBody
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunked.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunked.kt
@@ -5,6 +5,7 @@
 
 package aws.smithy.kotlin.runtime.auth.awssigning
 
+import aws.smithy.kotlin.runtime.http.Headers
 import aws.smithy.kotlin.runtime.io.SdkByteReadChannel
 
 internal abstract class AbstractAwsChunked(
@@ -12,6 +13,7 @@ internal abstract class AbstractAwsChunked(
     private val signer: AwsSigner,
     private val signingConfig: AwsSigningConfig,
     private var previousSignature: ByteArray,
+    private var trailingHeaders: Headers = Headers.Empty
 ) : SdkByteReadChannel by chan {
     companion object {
         const val CHUNK_SIZE_BYTES: Int = 65536
@@ -22,6 +24,7 @@ internal abstract class AbstractAwsChunked(
 
     var chunk: ByteArray? = null
     var chunkOffset: Int = 0
+    var hasLastChunkBeenSent: Boolean = false
 
     /**
      * Returns all the bytes remaining in the underlying data source, up to [limit].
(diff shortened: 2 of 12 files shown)


runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunked.kt
```diff
@@ -0,0 +1,227 @@
...
+     */
+    override suspend fun readRemaining(limit: Int): ByteArray {
+        if (chunk == null || chunkOffset >= chunk!!.size) { chunk = getNextChunk() }
+
+        var bytesWritten = 0
+        var bytes = byteArrayOf()
+        while (bytesWritten != limit) {
+            val numBytesToWrite: Int = minOf(limit - bytesWritten, chunk!!.size - chunkOffset)
+
+            bytes += chunk!!.slice(chunkOffset until chunkOffset + numBytesToWrite)
```
aajtodd, line 42: comment: this is not very efficient, can we do better than this?

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunked.kt
```diff
@@ -0,0 +1,227 @@
...
+
+        if (chunk == null || chunkOffset >= chunk!!.size) { chunk = getNextChunk() }
+
+        var bytesWritten = 0
+        while (bytesWritten != length) {
+            if (chunk == null) { throw RuntimeException("Invalid read: unable to fully read $length bytes. missing ${length - bytesWritten} bytes.") }
+
+            val numBytesToWrite: Int = minOf(length, chunk!!.size - chunkOffset)
+
+            val bytes = chunk!!.slice(chunkOffset until chunkOffset + numBytesToWrite).toByteArray()
```
aajtodd, line 79: Comment: Note you could be using `ByteArray.sliceArray` which will give you back a `ByteArray` directly (no conversion to/from `List<Byte>`. A better solution though is to just directly call `chunk.copyInto(sink, ...)` without the extra allocation. [copyInto](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/copy-into.html) accepts the destination offset as well as the source start/end parameters

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunked.kt
```diff
@@ -0,0 +1,227 @@
...
+            chunkOffset += numBytesToWrite
+
+            if (chunkOffset >= chunk!!.size) {
+                chunk = getNextChunk()
+            }
+        }
+    }
+
+    /**
+     * Writes [length] bytes to [sink], starting [offset] bytes from the beginning.
```
aajtodd, line 93: nit: This writes up-to `length` bytes

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunked.kt
```diff
@@ -0,0 +1,227 @@
...
+     * @param offset the number of bytes in [sink] to skip before beginning to write
+     * @param length the number of bytes to write to [sink]
+     * @throws IllegalArgumentException when illegal [offset] and [length] arguments are passed
+     * @throws RuntimeException when the source data is exhausted before [length] bytes are written to [sink]
+     */
+    override suspend fun readFully(sink: ByteArray, offset: Int, length: Int) {
+        require(offset >= 0) { "Invalid read: offset must be positive:  $offset" }
+        require(offset + length <= sink.size) { "Invalid read: offset + length should be less than the destination size: $offset + $length < ${sink.size}" }
+
+        if (chunk == null || chunkOffset >= chunk!!.size) { chunk = getNextChunk() }
```
aajtodd, line 71: Rather than assigning chunk and having this check in each read function you could just move this all into `getNextChunk` which can deal with assigning the `chunk` based on the state of chunk/chunkOffset. Perhaps rename it to `ensureNextChunk` or something. Just an idea, this isn't wrong or anything though

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigner.kt
```diff
@@ -31,4 +31,17 @@ public interface AwsSigner {
...
+
+    /**
+     * Signs a chunked payload's trailer according to the supplied signing configuration
+     * @param trailingHeaders The canonicalized trailing headers
+     * @param finalChunkSignature The signature of the final payload chunk
+     * @param config The signing configuration
+     * @return The signing result, which should be appended as a trailing header itself, named `x-amz-trailer-signature`
+     */
+    public suspend fun signChunkTrailer(
+        trailingHeaders: ByteArray,
```
aajtodd, line 43: Question/correctness: Should this be required to be the canonicalized version of the headers? We don't force that anywhere else in signing and it seems like an internal concern of signing. Should we be taking a `Headers` instance instead and leave canonicalization to the implementation(s)?

runtime/auth/aws-signing-crt/common/src/aws/smithy/kotlin/runtime/auth/awssigning/crt/CrtAwsSigner.kt
```diff
@@ -47,6 +47,14 @@ public object CrtAwsSigner : AwsSigner {
         val crtResult = CrtSigner.signChunk(chunkBody, prevSignature, crtConfig)
         return AwsSigningResult(Unit, crtResult.signature)
     }
+
+    override suspend fun signChunkTrailer(
+        trailingHeaders: ByteArray,
+        finalChunkSignature: ByteArray,
+        config: AwsSigningConfig,
+    ): AwsSigningResult<Unit> {
+        TODO("Not yet implemented")
```
aajtodd, line 56: I think we can implement this already. We may need to wire up everything in `aws-crt-kotlin` to `aws-crt-java`. We should ensure that whatever interface we define we can actually implement...

runtime/auth/aws-signing-common/build.gradle.kts
```diff
@@ -17,6 +19,14 @@ kotlin {
             }
         }
 
+        commonTest {
+            dependencies {
+                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:$coroutinesVersion")
+                implementation(project(":runtime:auth:aws-signing-tests"))
+                implementation(project(":runtime:auth:aws-signing-default"))
```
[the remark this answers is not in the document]
lauzadis: I had to add the `aws-signing-default` dependency to use [`DefaultAwsSigner`](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/auth/aws-signing-default/common/src/aws/smithy/kotlin/runtime/auth/awssigning/DefaultAwsSigner.kt#L13) in my test cases. Similarly the `aws-signing-tests` dependency was so I could use [`testSigningConfig`](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/auth/aws-signing-tests/common/src/aws/smithy/kotlin/runtime/auth/awssigning/tests/SigningTestUtil.kt#L11) from there. I can remove the aws-signing-tests dependency since it's a pretty simple construction, but I figured I should use the existing util variable rather than duplicate it. But I'm not sure how to remove the dependency on aws-signing-default while also continuing to use the `DefaultAwsSigner`.

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigner.kt
lauzadis: We could leave canonicalizing the headers to the implementations, but there are already some cases of [receiving canonicalized requests in the signer](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/auth/aws-signing-default/common/src/aws/smithy/kotlin/runtime/auth/awssigning/SignatureCalculator.kt#L52), so I was following that.

I figured whenever the request gets canonicalized, it could also canonicalize the trailing headers to prevent duplicating the canonicalization code.

I think it would be better to [reuse that code](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/auth/aws-signing-default/common/src/aws/smithy/kotlin/runtime/auth/awssigning/Canonicalizer.kt#L105) rather than rewriting it in each `signChunkTrailer` implementation.

runtime/auth/aws-signing-crt/common/src/aws/smithy/kotlin/runtime/auth/awssigning/crt/CrtAwsSigner.kt
lauzadis: Yeah I believe I could implement this right now. I took a look at the [aws-crt-java function and it takes a List\<Header\> as input](https://github.com/awslabs/aws-crt-java/blob/main/src/main/java/software/amazon/awssdk/crt/auth/signing/AwsSigner.java#L116), so I may need to do some deserialization / serialization, but otherwise it should be fine.

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunked.kt
lauzadis: optimized this by allocating a ByteArray with `limit` size (as that's the absolute maximum we can expect to read), and truncating it before returning

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunked.kt
lauzadis: good idea, I made a function called `ensureValidChunk` which will return a boolean representing if the chunk is valid for reading or not.

it will first check if the existing chunk is valid, and if not, try to load the next chunk.

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsChunked.kt
```diff
@@ -0,0 +1,28 @@
...
+ * This class wraps an SdkByteReadChannel. When reads are performed on this class, it will read the wrapped data
+ * and return it in aws-chunked content encoding.
+ * @see <a href="https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html">SigV4 Streaming</a>
+ * @param chan the underlying [SdkByteReadChannel] which will have its data encoded in aws-chunked format
+ * @param signer the signer to use to sign chunks and (optionally) chunk trailer
+ * @param signingConfig the config to use for signing
+ * @param previousSignature the previous signature to use for signing. in most cases, this should be the seed signature
+ * @param trailingHeaders the optional trailing headers to include in the final chunk
+ */
+internal expect class AwsChunked internal constructor(
```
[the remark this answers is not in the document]
lauzadis: Good point, I went with `AwsChunkedByteReadChannel`

runtime/auth/aws-signing-crt/common/src/aws/smithy/kotlin/runtime/auth/awssigning/crt/CrtAwsSigner.kt
lauzadis: Implemented in the latest commits here and opened a new PR on the aws-crt-kotlin repo: https://github.com/awslabs/aws-crt-kotlin/pull/60

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigner.kt
lauzadis: I ended up adding canonicalization code where it's needed rather than assuming the input is canonicalized. All we need to do for canonicalization is sort alphabetically and lowercase the header names

runtime/auth/aws-signing-common/build.gradle.kts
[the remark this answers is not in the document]
aajtodd: What if we just move the tests for `AwsChunkedByteReadChannel` into `aws-signing-tests`? This would remove the pseudo circular dependency (`aws-signing-default` depends on `aws-signing-common` for the commonMain sourceSet) and allow us to cover both default and crt based signers. 

I realize it's slightly odd to not have the test live with the source but I think in this case it may be the best option.

Thoughts?

runtime/auth/aws-signing-common/common/test/aws/smithy/kotlin/runtime/auth/awssigning/AwsChunkedByteReadChannelTest.kt
```diff
@@ -0,0 +1,913 @@
...
+        val dataLengthBytes = CHUNK_SIZE_BYTES
+        val data = ByteArray(dataLengthBytes) { 0x7A.toByte() }
+        val chan = SdkByteReadChannel(data)
+        val previousSignature: ByteArray = byteArrayOf()
+        val awsChunked = AwsChunkedByteReadChannel(chan, testSigner, testSigningConfig, previousSignature)
+
+        // read the exact chunk length
+        val bytes = awsChunked.readRemaining(
+            CHUNK_SIZE_BYTES + CHUNK_SIZE_BYTES.toString(16).length + 1 +
+                "chunk-signature=".length + 64 + 4 + 1 + 1 + "chunk-signature=".length + 64 + 4 + 2,
```
aajtodd, line 94: This seems like a bunch of magical numbers...can we get some comments in here or define some constants to make the test code clearer? (this goes for several tests, not going to call them all out - the tests are good but they kind of assume an intimate knowledge that hinders overall test understanding/maintenance)

runtime/auth/aws-signing-common/common/test/aws/smithy/kotlin/runtime/auth/awssigning/AwsChunkedByteReadChannelTest.kt
```diff
@@ -0,0 +1,913 @@
...
+
+    @Test
+    fun testReadRemainingExcessiveBytes() = runTest {
+        val dataLengthBytes = CHUNK_SIZE_BYTES
+        val data = ByteArray(dataLengthBytes) { 0x7A.toByte() }
+        val chan = SdkByteReadChannel(data)
+        val previousSignature: ByteArray = byteArrayOf()
+        val awsChunked = AwsChunkedByteReadChannel(chan, testSigner, testSigningConfig, previousSignature)
+
+        val numBytesToRead = 131072 // read double of the available chunk data
```
aajtodd, line 148: since it's a limit why not just make it `Int.MAX_VALUE` or something? The number as written feels oddly specific which doesn't add to understanding the test

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunkedByteReadChannel.kt
```diff
@@ -0,0 +1,204 @@
...
+
+        val chunkHeader = buildString {
+            append(chunkBody.size.toString(16))
+            append(";")
+            append("chunk-signature=")
+            append(chunkSignature.decodeToString())
+            append("\r\n")
+        }.encodeToByteArray()
+
+        return chunkHeader + chunkBody
```
aajtodd, line 174: Comment/no-op: This is fine for now because we are going to be updating this code soon with a refactor of `SdkByteReadChannel` but in general we want to be very cognizant of performance characteristics we use/introduce. Each `plus` operator on a `ByteArray` is an implicit copy behind the scenes. One optimization I would probably consider here is reading to a buffer with large enough capacity to handle the header + body and setting a limit for the chunk. This would avoid excessive copying and read directly into the underlying buffer. 

I call this out not to fix in this PR (again because we have some upcoming expected churn anyway) but as something to pay attention to in general and be aware of.

aajtodd (commented):
Overall looks great

runtime/auth/aws-signing-common/common/test/aws/smithy/kotlin/runtime/auth/awssigning/AwsChunkedByteReadChannelTest.kt
lauzadis: This is `CHUNK_SIZE_BYTES * 2` -- but I can make that clearer along with the magic numbers in other test cases. `Int.MAX_VALUE` works everywhere except `readRemaining` because we allocate a buffer to hold `limit `bytes and the VM runs out of memory.

```
java.lang.OutOfMemoryError: Requested array size exceeds VM limit
```

I'll change it everywhere except the `testReadRemainingExcessiveBytes` case.

aajtodd: Oh I missed that in the code, we really probably shouldn't be allocating the `limit` bytes as we may be way under that in most cases (not to mention it is the [default](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannel.kt#L42)).  

Given we don't ever call `readRemaining()` in any HTTP engines (and never would) and that this will be refactored soon I'm not hugely concerned but I would probably take a second look at this implementation and see if we couldn't do better (e.g. a small optimization would be testing if we are already closed and taking the min(availableForRead, limit)).

runtime/auth/aws-signing-common/build.gradle.kts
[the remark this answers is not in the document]
lauzadis: It is a good idea, but complicates things because `AwsChunkedByteReadChannel` is `internal` to `aws-signing-common` and can't be accessed from `aws-signing-tests`. I can change the visibility, but I think internal is the best choice here.

I was looking into Kotlin's behavior around `internal` and saw it is a group of files compiled together, is there any way to group these two modules together during compilation to allow access in aws-signing-tests?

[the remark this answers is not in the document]
aajtodd: As far as we are concerned `internal` and `public` with `@InternalApi` are nearly the same from a support/maintenance policy (obviously `internal` is better since it is not accessible at all whereas the latter can be accessed with compiler flags). 

I think I'd just mark it as `public` and annotate it with `@InternalApi` but happy to see what others think.

aajtodd (approved):

PR #732, opened 2022-11-12
feat: endpoints2: endpoint provider middleware and final integration

lucix-aws:
## Description of changes
Implement middleware and endpoint parameter binding generators to call endpoint providers per-operation at runtime.

### DIFF
--- a/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/EndpointProvider.kt
+++ b/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/EndpointProvider.kt
@@ -5,19 +5,14 @@
 
 package aws.smithy.kotlin.runtime.http.endpoints
 
-import aws.smithy.kotlin.runtime.util.InternalApi
-
 /**
  * Resolves endpoints for a given service client.
- *
- * This is a static version of the otherwise-generated endpoint resolver interface and is intended for use by internal
- * clients within the runtime.
  */
-@InternalApi
-public fun interface EndpointResolver {
+public fun interface EndpointProvider<T> {
     /**
      * Resolve the endpoint to make requests to
+     * @param params The input context for the resolver function
      * @return an [Endpoint] that can be used to connect to the service
      */
-    public suspend fun resolve(): Endpoint
+    public suspend fun resolveEndpoint(params: T): Endpoint
 }
--- a/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/EndpointResolver.kt
+++ b/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/EndpointResolver.kt
@@ -5,9 +5,15 @@
 
 package aws.smithy.kotlin.runtime.http.endpoints
 
+import aws.smithy.kotlin.runtime.util.InternalApi
+
 /**
- * Resolves endpoints for a given service client
+ * Resolves endpoints for a given service client.
+ *
+ * This is a static version of the otherwise-generated endpoint resolver interface and is intended for use by internal
+ * clients within the runtime.
  */
+@InternalApi
 public fun interface EndpointResolver {
     /**
      * Resolve the endpoint to make requests to
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/RulesEngineExt.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/RulesEngineExt.kt
@@ -31,9 +31,7 @@ fun ParameterType.toSymbol(): Symbol =
         ParameterType.STRING -> KotlinTypes.String
         ParameterType.BOOLEAN -> KotlinTypes.Boolean
     }
-        .toBuilder()
-        .boxed()
-        .build()
+        .asNullable()
 
 /**
  * Get the writable literal for a rules engine value.
@@ -41,6 +39,6 @@ fun ParameterType.toSymbol(): Symbol =
 fun Value.toLiteral(): String =
     when (this) {
         is Value.String -> expectString().doubleQuote()
-        is Value.Bool -> if (expectBool()) "true" else "false"
+        is Value.Bool -> expectBool().toString()
         else -> throw IllegalArgumentException("unrecognized parameter value type")
     }
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/RulesEngineExt.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/RulesEngineExt.kt
@@ -1,0 +1,46 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package software.amazon.smithy.kotlin.codegen.model
+
+import software.amazon.smithy.codegen.core.Symbol
+import software.amazon.smithy.kotlin.codegen.lang.KotlinTypes
+import software.amazon.smithy.kotlin.codegen.utils.doubleQuote
+import software.amazon.smithy.kotlin.codegen.utils.toCamelCase
+import software.amazon.smithy.rulesengine.language.eval.Value
+import software.amazon.smithy.rulesengine.language.syntax.Identifier
+import software.amazon.smithy.rulesengine.language.syntax.parameters.Parameter
+import software.amazon.smithy.rulesengine.language.syntax.parameters.ParameterType
+
+/**
+ * Derive the kotlin variable name for an identifier.
+ */
+fun Identifier.defaultName(): String = asString().toCamelCase()
+
+/**
+ * Derive the kotlin variable name for an endpoint parameter.
+ */
+fun Parameter.defaultName(): String = name.defaultName()
+
+/**
+ * Get the symbol for an endpoint parameter type. All endpoint parameter members are nullable.
+ */
+fun ParameterType.toSymbol(): Symbol =
+    when (this) {
+        ParameterType.STRING -> KotlinTypes.String
+        ParameterType.BOOLEAN -> KotlinTypes.Boolean
+    }
+        .toBuilder()
+        .boxed()
+        .build()
+
+/**
+ * Get the writable literal for a rules engine value.
+ */
+fun Value.toLiteral(): String =
+    when (this) {
+        is Value.String -> expectString().doubleQuote()
+        is Value.Bool -> if (expectBool()) "true" else "false"
+        else -> throw IllegalArgumentException("unrecognized parameter value type")
+    }
--- a/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
+++ b/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
@@ -8,17 +8,10 @@ import software.amazon.smithy.codegen.core.Symbol
 import software.amazon.smithy.kotlin.codegen.KotlinSettings
 import software.amazon.smithy.kotlin.codegen.core.KotlinWriter
 import software.amazon.smithy.kotlin.codegen.core.withBlock
-import software.amazon.smithy.kotlin.codegen.lang.KotlinTypes
-import software.amazon.smithy.kotlin.codegen.model.boxed
-import software.amazon.smithy.kotlin.codegen.model.buildSymbol
-import software.amazon.smithy.kotlin.codegen.utils.doubleQuote
+import software.amazon.smithy.kotlin.codegen.model.*
 import software.amazon.smithy.kotlin.codegen.utils.getOrNull
-import software.amazon.smithy.kotlin.codegen.utils.toCamelCase
 import software.amazon.smithy.rulesengine.language.EndpointRuleSet
-import software.amazon.smithy.rulesengine.language.eval.Value
-import software.amazon.smithy.rulesengine.language.syntax.Identifier
 import software.amazon.smithy.rulesengine.language.syntax.parameters.Parameter
-import software.amazon.smithy.rulesengine.language.syntax.parameters.ParameterType
 
 private const val DEFAULT_DEPRECATED_MESSAGE =
     "This field is deprecated and no longer recommended for use."
@@ -28,7 +21,7 @@ private const val DEFAULT_DEPRECATED_MESSAGE =
  */
 class EndpointParametersGenerator(
     private val writer: KotlinWriter,
-    rules: EndpointRuleSet,
+    rules: EndpointRuleSet?,
 ) {
     companion object {
         const val CLASS_NAME = "EndpointParameters"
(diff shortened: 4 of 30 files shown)


runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/EndpointResolver.kt
```diff
@@ -5,9 +5,15 @@
...
+
 /**
- * Resolves endpoints for a given service client
+ * Resolves endpoints for a given service client.
+ *
+ * This is a static version of the otherwise-generated endpoint resolver interface and is intended for use by internal
+ * clients within the runtime.
  */
+@InternalApi
 public fun interface EndpointResolver {
```
ianbotsf, line 17: **Question**: Is there no supported way to specify a custom endpoint resolver anymore?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/EndpointParametersGenerator.kt
```diff
@@ -38,6 +38,8 @@ class EndpointParametersGenerator(
                 name = CLASS_NAME
                 namespace = "${settings.pkg.name}.endpoints"
             }
+
+        fun getParamKotlinName(param: Parameter): String = param.name.toKotlin()
     }
 
     private val params: List<KotlinEndpointParameter> = rules.parameters.toList()
```
aajtodd, line 45: GH won't let me comment on the unchanged line but we generate a companion object for `EndpointParameters` like this:

```kotlin
    public companion object {
        public operator fun invoke(block: Builder.() -> Unit): EndpointParameters = Builder().apply(block).build()
    }
```

This function should probably be marked `inline` to avoid unnecessary backing lambdas everywhere we have to construct endpoint parameters in generated code (contributes to bloated artifact size)

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt
```diff
@@ -293,7 +293,7 @@ abstract class HttpBindingProtocolGenerator : ProtocolGenerator {
         val pathBindings = requestBindings.filter { it.location == HttpBinding.Location.LABEL }
 
         if (pathBindings.isNotEmpty()) {
-            writer.openBlock("val pathSegments = listOf(", ")") {
+            writer.openBlock("val pathSegments = listOf<String>(", ")") {
```
ianbotsf, line 296: **Question**: Why was this change necessary?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/ResolveEndpointMiddlewareGenerator.kt
```diff
@@ -0,0 +1,151 @@
...
+import software.amazon.smithy.rulesengine.traits.ClientContextParamsTrait
+import software.amazon.smithy.rulesengine.traits.ContextParamTrait
+import software.amazon.smithy.rulesengine.traits.StaticContextParamsTrait
+
+/**
+ * Generates resolver middleware for a specific endpoint provider signature.
+ * The binding of parameter values is generated separately in [EndpointParameterBindingGenerator], and that generated
+ * implementation is plugged in here.
+ */
+class ResolveEndpointMiddlewareGenerator(
```
aajtodd, line 31: Rather than comment on each individual location I'm going to make one long comment on the resulting generated code.

Taken from S3 it seems we are generating the following middleware and middleware registration:

Current middleware being generated (per service):
```kotlin
public class ResolveEndpointMiddleware<I, O>(
    private val endpointProvider: EndpointProvider,
    private val buildParams: EndpointParameters.Builder.(input: I) -> Unit,
) : InlineMiddleware<I, O> {
    private lateinit var endpoint: Endpoint

    public override fun install(op: SdkHttpOperation<I, O>) {
        op.execution.initialize.intercept { req, next ->
            val params = EndpointParameters { buildParams(req.subject) }
            endpoint = endpointProvider.resolveEndpoint(params)
            next.call(req)
        }

        op.execution.mutate.intercept { req, next ->
            setResolvedEndpoint(req, endpoint)
            next.call(req)
        }
    }
}
```

Example current operation setup/registration:
```kotlin
    override suspend fun abortMultipartUpload(input: AbortMultipartUploadRequest): AbortMultipartUploadResponse {
        ...

        op.install(ResolveEndpointMiddleware<AbortMultipartUploadRequest, AbortMultipartUploadResponse>(config.endpointProvider) { input ->
            bindAwsBuiltins(config)
            bucket = input.bucket
        })
      
        ...
```

-----------------

Couple things stand out to me right away.

The middleware takes a `buildParams` lambda which is passed `req.subject` to bind the input shape members to the endpoint parameters (if there are any). This isn't really necessary though as we have access to the operation input when creating the middleware (and is what you are using already anyway, e.g. `input.bucket` rather than the parameter from the lambda). This relationship is a bit confusing and seems like it's there only because we want to do it in middleware.

I'm almost wondering (and this is just an idea I'm throwing out there I don't even know if I like it yet) if we shouldn't just build the endpoint parameters in the operation setup before middleware registration:

```kotlin
val endpointParameters = EndpointParameters {
    bindBuiltins(config)
    bucket = input.bucket
    ...
}



op.install(ResolveEndpointMiddleware(config.endpointProvider, endpointParameters))

```

`ResolveEndpointMiddleware` could then just be a normal `ModifyRequestMiddleware`

--------

We could update the `EndpointResolver` and `ResolveEndpoint` middleware types to be re-usable such that we don't have to generate a separate resolver type per/client but instead use generics:

something like:
```kotlin
public class ResolveEndpoint<T> {
    private val resolver: EndpointResolver<T>,
    private val parameters: T
): ModiifyRequestMiddleware {
    override suspend fun modifyRequest(req: SdkHttpRequest): SdkHttpRequest {
        val endpoint = resolver.resolve(parameters)
        setResolvedEndpoint(req, endpoint)
        ...
    }   
}
```

Resolver w/generics
```kotlin
public fun EndpointResolver<T> {
    public suspend fun resolve(parameters: T): Endpoint
}
```

The generated client config would then be:

```kotlin
public class Config ... {
     ...
     // specialized to whatever the EndpointParameters type is that IS generated per/service
     public val endpointProvider: EndpointProvider<EndpointParameters> = builder.endpointProvider ?: DefaultEndpointProvider()

}
```

Whether we do this with generics or keep per/service generated resolver type isn't super important it's just one idea to try and re-use as much as possible and not need custom codegen for AWS vs generic smithy models.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt
lucix-aws: So this is a side effect of the bucket path preprocessing that now happens on the s3 model.

If you remove both the path segment and the httpLabel trait, our serializer generator renders the bucket to be serialized as part of the payload, which is wrong.

Since the path serializer references off of the httpTrait path itself, I just left the httpLabel trait alone and when we go to render the path serializer it simply ignores the orphaned httpLabel members.

A side effect of that is the list of paths generated here can now be empty, hence the need to explicitly add the String to the declaration.

There may be a better way to express "do not serialize this" in smithy that I'm not aware of.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/ResolveEndpointMiddlewareGenerator.kt
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /**
+     * Render the assignment of an endpoint param value within a builder. The SEP dictates a specific source order to
+     * use when determining how to bind:
+     * 1. staticContextParams (from operation shape)
+     * 2. contextParam (from member of operation input shape)
+     * 3. clientContextParams (from service shape)
+     * 4. builtin binding
+     * 5. builtin default
```
aajtodd, line 125: Do we have any instances of (5)? I'm assuming `bindAwsBuiltins` is type (4) and why we render it _before_ delegating to `EndpointParameterBindingGenerator` in `aws-sdk-kotlin`?

lucix-aws: Correct, and (5) is essentially fulfilled by the config field defaults.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/EndpointProvider.kt
```diff
@@ -5,19 +5,14 @@
...
-
 /**
  * Resolves endpoints for a given service client.
- *
- * This is a static version of the otherwise-generated endpoint resolver interface and is intended for use by internal
- * clients within the runtime.
  */
-@InternalApi
-public fun interface EndpointResolver {
+public fun interface EndpointProvider<T> {
```
ianbotsf, line 11: **Question**: Is there a semantic difference between a "provider" and a "resolver"? This interface's name changed but its KDoc and single method name still refer to "resolution" rather than "providing".

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/RulesEngineExt.kt
```diff
@@ -0,0 +1,46 @@
...
+ * Get the symbol for an endpoint parameter type. All endpoint parameter members are nullable.
+ */
+fun ParameterType.toSymbol(): Symbol =
+    when (this) {
+        ParameterType.STRING -> KotlinTypes.String
+        ParameterType.BOOLEAN -> KotlinTypes.Boolean
+    }
+        .toBuilder()
+        .boxed()
+        .build()
```
ianbotsf, line 36: **Comment**: `.toBuilder().boxed().build()` is a thing we do in a few other places now. Feels like this is officially convenience-method-ready.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/RulesEngineExt.kt
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Get the writable literal for a rules engine value.
+ */
+fun Value.toLiteral(): String =
+    when (this) {
+        is Value.String -> expectString().doubleQuote()
+        is Value.Bool -> if (expectBool()) "true" else "false"
```
ianbotsf, line 44: **Nit**: `expectBool().toString()`

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/ResolveEndpointMiddlewareGenerator.kt
```diff
@@ -31,6 +31,7 @@ import software.amazon.smithy.rulesengine.traits.StaticContextParamsTrait
 class ResolveEndpointMiddlewareGenerator(
     private val ctx: ProtocolGenerator.GenerationContext,
     private val writer: KotlinWriter,
+    private val renderPostResolution: () -> Unit = {},
 ) {
```
ianbotsf, line 35: **Question**: Where is a non-default argument passed for `renderPostResolution`? I couldn't find anywhere in this PR or the **aws-sdk-kotlin** one...

ianbotsf (approved):
Nice work! Nothing blocking, just questions and low-priority suggestions.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/endpoints/EndpointProvider.kt
aajtodd: I don't think there is, I preferred resolver with a single `resolve` method but this matches the SEP/ref arch terminology I believe.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/ResolveEndpointMiddlewareGenerator.kt
```diff
@@ -0,0 +1,153 @@
...
+        writer.closeAndOpenBlock(") : #T<I, O> {", RuntimeTypes.Http.Operation.InlineMiddleware)
+        renderClassMembers()
+        writer.write("")
+        renderInstall()
+        writer.closeBlock("}")
+    }
+
+    private fun renderConstructorParams() {
+        writer.write("private val endpointProvider: #T,", EndpointProviderGenerator.getSymbol(ctx.settings))
+        writer.write("private val buildParams: #T.Builder.(input: I) -> Unit,", EndpointParametersGenerator.getSymbol(ctx.settings))
```
aajtodd, line 58: IDK why this irks me so much but it does. I don't have a better suggestion though so it's fine for now.


Part of my distaste for this is that it results in backing lambdas for each operation, e.g.

```
> javap DefaultS3Client\$deleteBucket\$2                  
Warning: File ./DefaultS3Client$deleteBucket$2.class does not contain class DefaultS3Client$deleteBucket$2
Compiled from "DefaultS3Client.kt"
final class aws.sdk.kotlin.services.s3.DefaultS3Client$deleteBucket$2 extends kotlin.jvm.internal.Lambda implements kotlin.jvm.functions.Function2<aws.sdk.kotlin.services.s3.endpoints.EndpointParameters$Builder, aws.sdk.kotlin.services.s3.model.DeleteBucketRequest, kotlin.Unit> {
  final aws.sdk.kotlin.services.s3.DefaultS3Client this$0;
  aws.sdk.kotlin.services.s3.DefaultS3Client$deleteBucket$2(aws.sdk.kotlin.services.s3.DefaultS3Client);
  public final void invoke(aws.sdk.kotlin.services.s3.endpoints.EndpointParameters$Builder, aws.sdk.kotlin.services.s3.model.DeleteBucketRequest);
  public java.lang.Object invoke(java.lang.Object, java.lang.Object);
}
```

The alternative isn't much better though as it would require us to generate a middleware per/operation anyway

aajtodd (approved):

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/ResolveEndpointMiddlewareGenerator.kt
lucix-aws: The default implementation on the `ProtocolGenerator` method does

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/ResolveEndpointMiddlewareGenerator.kt
lucix-aws: Yeah, I had the per-operation middleware before refactoring to this, which was much worse IMO.

PR #733, opened 2022-11-15
refactor(rt)!: simplify SdkByteChannel interface

aajtodd:
## Issue \#

N/A

## Description of changes

Refactor `SdkByteChannel` to be based on the new `SdkBuffer` type. Moving data between buffers is cheap (especially in bulk) so this type decouples a producer and consumer with minimal overhead. Ktor based channels did some clever things internally but in many common cases it still required an explicit copy for each read/write method. Having a single read/write method to implement makes layering/wrapping a channel much easier than before (similar to how easy it is to layer okio Source/Sink types).

* Removes ktor based channel dependency
* Updates `SdkByteReadChannel/SdkByteWriteChannel` to have a single `read/write` method based on `SdkBuffer`. This makes it effectively an async `Source/Sink` or an async `Pipe` depending on how you want to look at it. 
* Adds benchmarks comparing the new channel to ktor based channel. 
    * NOTE: These benchmarks are not really an apples-to-apples comparison and thus are not really fair. They are however an idiomatic usage of the channel comparison and should be looked at through that lens. 


runtime/io/common/src/aws/smithy/kotlin/runtime/io/RealSdkByteChannel.kt
```diff
@@ -0,0 +1,212 @@
...
+
+        if (autoFlush || availableForWrite == 0) {
+            flush()
+        }
+    }
+
+    private fun afterRead(size: Int) {
+        capacity.completeRead(size)
+        slot.resume()
+    }
```
ianbotsf, line 174: **Suggestion**: If there are things which _must_ occur after reading/writing, I'd consider making them part of the `reading`/`writing` methods.

runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteChannel.kt
```diff
@@ -6,11 +6,12 @@
 package aws.smithy.kotlin.runtime.io
 
 import aws.smithy.kotlin.runtime.util.InternalApi
-import io.ktor.utils.io.ByteChannel as KtorByteChannel
-import io.ktor.utils.io.ByteReadChannel as KtorByteReadChannel
+
+public const val DEFAULT_BYTE_CHANNEL_MAX_BUFFER_SIZE: Int = 8192 * 2
```
ianbotsf, line 10: **Style**: Why write this as `8192 * 2` rather than `16384` or even `16 * 1024`?

runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteChannel.kt
```diff
@@ -20,23 +21,26 @@ import io.ktor.utils.io.ByteReadChannel as KtorByteReadChannel
...
- * Creates a channel for reading from the given byte array.
+ * Creates a channel for reading with the contents of the given byte array.
  */
 @InternalApi
 public fun SdkByteReadChannel(
     content: ByteArray,
     offset: Int = 0,
     length: Int = content.size - offset,
-): SdkByteReadChannel = KtorByteReadChannel(content, offset, length).toSdkChannel()
+): SdkByteReadChannel = RealSdkByteChannel(content, offset, length).apply { close() }
```
ianbotsf, line 46: **Question**: Would closing the channel be better done in the `ByteArray`-accepting constructor? Or do we anticipate a future use case for creating a read channel with an initial byte array and then adding more later?

runtime/io/common/src/aws/smithy/kotlin/runtime/io/internal/ChannelCapacity.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    fun completeWrite(size: Int) {
+        _availableForWrite.minusAssign(size)
+        _pendingToFlush.plusAssign(size)
+    }
+
+    fun completeRead(size: Int) {
+        _availableForRead.minusAssign(size)
+        _availableForWrite.plusAssign(size)
+    }
```
ianbotsf, line 65: **Question**: Sequential atomics are not atomic across the logical operation. Thus it's possible for one value to be updated and other not yet when an intervening read accesses one of them. Is that acceptable in this case?

runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteChannel.kt
aajtodd: Do you have a preference lol? The reason it's written this way was in my head I was thinking in terms of okio segment lengths (8k), which is an internal detail but should result in a decent default).

ianbotsf: Sure that makes sense as a good default and we can even leave it as `8192 * 2`. Let's just add a comment about where 8k is coming from so that it's clearer for future readers.

runtime/io/common/src/aws/smithy/kotlin/runtime/io/internal/ChannelCapacity.kt
aajtodd: > Sequential atomics are not atomic across the logical operation
Correct and yes I believe this is fine. Channels are defined as single reader/single writer. They are invalid for multiple concurrent read or multiple concurrent write calls (and we have checks/tests to the best of our ability to ensure this). 

Concurrent read/write (single reader/writer) should be fine here given that read/write capacity can only go up down based on progress from the other.

runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteChannel.kt
aajtodd: I'm not sure why I put it here, I see no reason it can't be moved to the constructor at first glance though

ianbotsf (approved):
Verifying correct usage of sequential atomics was the main thing predicating my approval. The rest are optional.

runtime/io/common/src/aws/smithy/kotlin/runtime/io/RealSdkByteChannel.kt
aajtodd: I don't recall if there was a specific reason to not include it there but I'll take a look and comment or move things around as appropriate

lauzadis (approved):
Looks good! Thanks for including benchmarking code

runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteChannel.kt
```diff
@@ -20,23 +21,26 @@ import io.ktor.utils.io.ByteReadChannel as KtorByteReadChannel
...
     override fun close() {
-        (this as SdkByteWriteChannel).close()
+        close(null)
     }
 }
 
 /**
  * Create a buffered channel for asynchronous reading and writing of bytes
+ * @param autoFlush Flag indicating if the channel should auto flush after every write, see [SdkByteWriteChannel.autoFlush]
  */
```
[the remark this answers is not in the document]
aajtodd: In this case they are pretty well explained by the kdoc comments but for consistency I have added the `@param` and `@return` annotations to the doc comments

lucix-aws (approved):

PR #734, opened 2022-11-15
fix(rt): handle exceptions in background okhttp response body coroutine

aajtodd:
## Issue \#

fixes https://github.com/awslabs/aws-sdk-kotlin/issues/753

## Description of changes

Wrap response body handling with `runCatching` to prevent the exception from reaching the default [CoroutineExceptionHandler](https://kotlinlang.org/docs/exception-handling.html#coroutineexceptionhandler) (which on JVM will print a stacktrace but on Android it will crash the app). Exception is propagated through the channel which was the original intent.


lauzadis (approved):

ianbotsf (approved):
These changes look correct but there are no new/updated tests. Would it be possible to write a test that fails before the change and succeeds after it?

PR #707, opened 2022-09-27
Initial tracing framework via ExecutionContext

ianbotsf:
## Issue \#

#677 

## Description of changes

This change adds support for tracing via `ExecutionContext`. Pay special attention to the interaction between `Logger` and `TraceSpan` and how the concept of a contextualized logger has been removed. I think this will be unnecessary going forward but I'd love to hear feedback to the contrary.

Publishing in **Draft** while we still debate alternatives.

**Companion PR:** [aws-sdk-kotlin#708](https://github.com/awslabs/aws-sdk-kotlin/pull/708)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #737, opened 2022-11-16
feat: add tracing framework

### LINKED ISSUE #677: Tracing framework

ianbotsf:
## Issue \#

Closes #677 

## Description of changes

This PR adds a new tracing framework which centralizes the handling of log messages and metrics into a single system and provides easy integration points for sending events to downstream systems like **kotlin-logging**. See **docs/design/tracing.md** for more details.

**Note**: This is the final, mergeable version of #722, including changes from the rejected approaches in #688 (context receivers) and #707 (ExecutionContext).

**Companion PR**: [aws-sdk-kotlin#756](https://github.com/awslabs/aws-sdk-kotlin/pull/756)


docs/design/tracing.md
```diff
@@ -0,0 +1,148 @@
...
+
+The `TraceSpan` interface is specified as:
+
+```kotlin
+interface TraceSpan : Closeable {
+    val id: String
+    val parent: TraceSpan?
+
+    fun child(id: String): TraceSpan
+    fun postEvents(events: Iterable<TraceEvent>)
```
aajtodd, line 62: comment: I know we've discussed this before but I was reviewing the tracing section of the ref architecture work and it includes both `emitEvent(event: TraceEvent)` and `emitEvents(events: Iterable<TraceEvent>)`. I'd argue we should remove the plural version and only accept the singular `postEvent(event: TraceEvent)`. This is based on (1) we have yet to have an actual need for the plural when emitting events; I don't expect this will change. (2) The reason for taking plural as explained in the ref arch is performance. That's fine _but_ it shouldn't really be a concern of a `TraceSpan`. It is more appropriate for trace probes to deal with batching/queuing of events (if necessary). I'm rather convinced now that it doesn't belong in this API (or at least we should support both) and I will be recommending the update to the architecture doc as well.

docs/design/tracing.md
```diff
@@ -0,0 +1,148 @@
...
+
+SDKs will typically not bundle many implementations of `TraceProbe` themselves. Common probe implementations may be
+available as separate libraries or from third-party sources. Users may implement probes themselves to bridge SDK events
+to whatever downstream systems they desire.
+
+The `TraceProbe` interface is defined as:
+
+```kotlin
+interface TraceProbe {
+    fun postEvents(span: TraceSpan, events: Iterable<TraceEvent>)
```
aajtodd, line 125: Same concern around singular vs plural. Forcing this into the API feels wrong to me (I'm not going to die on this hill but I feel compelled to bring it up again)

docs/design/tracing.md
```diff
@@ -0,0 +1,148 @@
...
+direct values. This allows probe implementations to skip calculating them in the event they would otherwise be discarded
+(e.g., for events emitted at a level ignored by the probe).
+
+## Trace Probe
+
+A `TraceProbe` is a sink for receiving events from spans. They will typically form a bridge between the SDK's events and
+downstream libraries/frameworks/services which can handle the events. Examples of such downstream systems include Log4j,
+CloudWatch, local files on disk, the console, etc.
+
+SDKs will typically not bundle many implementations of `TraceProbe` themselves. Common probe implementations may be
```
aajtodd, line 117: comment: The ref arch does discuss providing a "multi" sink trace probe that can be used by customers to wire up several sinks as a destination for trace events. This would be a useful runtime component to provide in the future (not part of this PR)

runtime/tracing/tracing-core/build.gradle.kts
```diff
@@ -0,0 +1,36 @@
...
+                // Necessary for Closeable
+                api(project(":runtime:io"))
+
+                // Necessary for InternalApi
+                api(project(":runtime:utils"))
+
+                // Necessary for Logger
+                api(project(":runtime:logging"))
+
+                implementation("io.github.microutils:kotlin-logging:$kotlinLoggingVersion")
```
aajtodd, line 28: question: Why is this necessary? Isn't this hidden as an implementation detail of `:runtime:logging`?

runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/TraceEvent.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * The time at which the event occurred or was recorded.
+     */
+    public val timestamp: Instant,
+
+    /**
+     * The ID of the thread on which this event occurred or was recorded.
+     */
+    public val threadId: String,
```
aajtodd, line 84: question: Should we track coroutine name if it's set in a coroutine context? Which would probably be a good use of open telemetry's concept of attributes...(which I also will be recommending ref arch adopt as part of the tracing API).

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
```diff
@@ -310,5 +312,33 @@ object KotlinClientRuntimeConfigProperty {
...
+
+        // TODO support a nice DSL for this so that callers don't have to be aware of `DefaultTracer` if they don't want
+        // to, they can just call `SomeClient { tracer { clientName = "Foo" } }` in the simple case.
+        Tracer = ClientConfigProperty {
+            symbol = RuntimeTypes.Tracing.Core.Tracer
+            baseClass = tracingClientConfigSymbol
+            documentation = """
+                The tracer that is responsible for creating trace spans and wiring them up to a tracing backend (e.g.,
+                a trace probe). By default, this will create a standard tracer that uses the service name for the root
+                trace span and delegates to a kotlin-logging trace probe (i.e.,
```
aajtodd, line 329: nit: I don't know that I'd mention `kotlin-logging` specifically here so much as just mention it will be logged.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
```diff
@@ -310,5 +312,33 @@ object KotlinClientRuntimeConfigProperty {
...
+                `DefaultTracer(KotlinLoggingTraceProbe, "<service-name>")`).
+            """.trimIndent()
+            propertyType = ClientConfigPropertyType.Custom { prop, writer ->
+                val serviceName = writer.getContext("service.name")?.toString()
+                    ?: throw CodegenException("The service.name context must be set for client config generation")
+                writer.write(
+                    """override val #1L: Tracer = builder.#1L ?: #2T(#3T, "$serviceName-")""",
```
aajtodd, line 336: question: Why is this "`$serviceName-"`? (trailing `-`)

runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/KotlinLoggingTraceProbe.kt
```diff
@@ -0,0 +1,44 @@
...
+package aws.smithy.kotlin.runtime.tracing
+
+import mu.KLogger
+import mu.KotlinLogging
+
+/**
+ * A [TraceProbe] that logs events to the [kotlin-logging](https://github.com/MicroUtils/kotlin-logging) library.
+ * kotlin-logging requires configuration (e.g., of Slf4j on JVM) before events sent to the probe will appear in logs.
+ */
+public object KotlinLoggingTraceProbe : TraceProbe {
```
aajtodd, line 14: correctness: `public`?

smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolClientGeneratorTest.kt
```diff
@@ -145,7 +159,10 @@ class HttpProtocolClientGeneratorTest {
             }
         }
         op.install(MockMiddleware(configurationField1 = "testing"))
-        return op.execute(client, input, block)
+        val rootSpan = config.tracer.createRootSpan("GetFooStreamingOutput-${'$'}{op.context.sdkRequestId}")
+        return coroutineContext.withRootTraceSpan(rootSpan) {
```
aajtodd, line 163: nit: Still would have loved to see some sugar for `withRootTraceSpan` and `withSpan` 😛 

```kotlin
inline suspend fun<T> withSpan(block: () -> T): T = coroutineContext.withSpan(...)
```

such that call sites can just do:

```
suspend fun foo() = withSpan("foo") {...}
```

This is a super minor style thing at the end of the day

aajtodd (approved):
Nothing blocking. Looks great!

runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/TraceSpanExt.kt
```diff
@@ -0,0 +1,219 @@
...
+ * @param ex An optional exception which explains the message
+ * @param content A lambda which provides the content of the message. This content does not need to include any data
+ * from the exception (if any), which may be concatenated later based on probe behavior.
+ */
+public fun TraceSpan.log(level: EventLevel, sourceComponent: String, ex: Throwable? = null, content: () -> Any?) {
+    val event = TraceEvent(
+        level,
+        sourceComponent,
+        Instant.now(),
+        "thread-id", // TODO
```
lauzadis, line 38: **question:** what needs to be done here? is it possible to fetch the thread ID at this point, or will it come in a future PR?

lauzadis (approved):

docs/design/tracing.md
ianbotsf: Discussed offline and came to the conclusion that singular APIs will be better for the average case than plural (and cleaner than having both). Will change!

docs/design/tracing.md
ianbotsf: As above, will change!

docs/design/tracing.md
ianbotsf: Yeah and I see that as pretty trivial to write. Should be easy to add in the future.

runtime/tracing/tracing-core/build.gradle.kts
ianbotsf: No, the `KotlinLoggingTraceProbe` accesses `mu.KLogger` and `mu.KotlinLogging` directly and doesn't go through **:runtime:logging**'s `KotlinLoggingAdapter`. We could _make_ it do that but it seems unnecessary given that we [plan to remove `KotlinLoggingAdapter`](https://github.com/awslabs/aws-sdk-kotlin/issues/757) shortly. At that time we'll probably move (or remove) `Logger` too and then **:runtime:logging** itself will be removed.

runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/TraceEvent.kt
ianbotsf: That sounds like a good idea in theory. In practice, how would we get the coroutine name into this field? At the call site as an additional argument to something like `TraceSpanLogger.info`? That sounds boilerplatey. And getting it _from inside_ the logger/span itself would be tricky because none of those methods are `suspend` and thus don't know about the active coroutine context...

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
ianbotsf: "Logging" is now an overloaded term because the tracing framework itself "logs" things too. Saying that **kotlin-logging** is involved helps disambiguate at which level the logging will occur and also provides a clue for what to configure to adjust the logging output.

Is the concern about tying to a particular implementation? Providing unnecessary/confusing detail? Other?

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
ianbotsf: The argument here is a root prefix—a string which will be prepended to all root-level trace span IDs. If there were no trailing `-` then the service name would blend together with the rest of the span ID (e.g., `DynamoDbPutItem` vs `DynamoDb-PutItem`).

[You suggested in the **aws-sdk-kotlin** PR](https://github.com/awslabs/aws-sdk-kotlin/pull/756#discussion_r1025642640) that the hyphen may be better handled as an implementation detail of the tracer and I agree. I'll modify this to remove the hyphen in the argument and add it in the tracer implementation.

runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/KotlinLoggingTraceProbe.kt
ianbotsf: `public` was intentional. If users want to customize the tracer used in their clients they'll still need access to this probe:

```kotlin
val s3 = S3Client {
    tracer = DefaultTracer(KotlinLoggingTraceProbe, "CustomRootPrefix-")
}
```

Users may also want to use `KotlinLoggingTraceProbe` in a multiplexing or splitting trace probe. I think `public` is correct in this case.

runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/TraceSpanExt.kt
ianbotsf: That's a great question and I don't yet have a great answer. We have a reference architecture which specifies that "thread ID" should be part of trace events but no concrete plan to get it (or use it) yet. [Aaron mentioned above](https://github.com/awslabs/smithy-kotlin/pull/737#discussion_r1025629428) that we might consider putting the coroutine name in this field but I don't yet know how we'd get that in practice.

runtime/tracing/tracing-core/build.gradle.kts
aajtodd: We should get this into the backlog and track as part of GA

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
aajtodd: Mostly exposing what amounts to an implementation detail that we should (are) free to change at will.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
aajtodd: Ok I see I missed that on first review

runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/KotlinLoggingTraceProbe.kt
aajtodd: I would probably just call it a `LoggingTraceProbe` then. You could mention that for JVM it is an SLF4J compatible logger but I wouldn't mention kotlin-logging specifically anywhere.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigProperty.kt
ianbotsf: Okay, will change.

lucix-aws (approved):

PR #743, opened 2022-11-21
refactor(rt)!: allow SdkSource variant for ByteStream and HttpBody

aajtodd:
## Issue \#

N/A

## Description of changes

Refactor `ByteStream` and `HttpBody` to allow an `SdkSource` variant. The primary effect of this is:
1. Http engine implementations (and customers if they are unwrapping `ByteStream` and not using a convenience function) have to deal with a new variant
    * `SdkSource` is a blocking interface which should be fine on JVM/Native but not on JS. We haven't fully thought this through but I expect the answer will be that we shuffle everything on JS through the channel variant and error if we see anything else. We won't be able to make this a compile time error unfortunately without knee capping `ByteStream`/`HttpBody` in common source sets. I chose not to go this route, I'm fine with the experience on JS being slightly less than ideal if you carve your own path outside the one we will define
2. File based input doesn't have to proxy itself through a channel, removing an extra layer of buffering/overhead 

I've also cleaned up `ByteStream` and `HttpBody` to move some of the less common parameters around like `isOneShot` flag, etc. The reason for this is we now have effectively two streaming input variants and duplicating these flags in specific variants of a sealed class felt wrong. It forced engines to unwrap the body to make decisions. 

### DIFF
--- a/runtime/io/common/src/aws/smithy/kotlin/runtime/io/internal/JobChannel.kt
+++ b/runtime/io/common/src/aws/smithy/kotlin/runtime/io/internal/JobChannel.kt
@@ -42,8 +42,4 @@ public class JobChannel(
         }
         return delegate.close(cause)
     }
-
-    override fun close() {
-        delegate.close()
-    }
 }
--- a/runtime/io/common/src/aws/smithy/kotlin/runtime/io/internal/JobChannel.kt
+++ b/runtime/io/common/src/aws/smithy/kotlin/runtime/io/internal/JobChannel.kt
@@ -1,0 +1,49 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package aws.smithy.kotlin.runtime.io.internal
+
+import aws.smithy.kotlin.runtime.io.SdkByteChannel
+import aws.smithy.kotlin.runtime.util.InternalApi
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.Job
+
+/**
+ * Ties the lifetime of the channel and a Job together. Channel failures (closed with exception) will
+ * cause the underlying [Job] to fail with the channel exception.
+ *
+ * This is useful for launching coroutines with the soul purpose of reading/writing data to the channel where
+ * the coroutine should only continue if the channel hasn't failed.
+ */
+@InternalApi
+public class JobChannel(
+    private val delegate: SdkByteChannel = SdkByteChannel(true),
+) : SdkByteChannel by delegate {
+    internal var job: Job? = null
+
+    public fun attachJob(job: Job) {
+        if (isClosedForRead) {
+            job.cancel(CancellationException("channel was already closed", delegate.closedCause))
+            return
+        }
+        this.job = job
+    }
+
+    override fun cancel(cause: Throwable?): Boolean {
+        job?.cancel(CancellationException("channel was cancelled", cause))
+        return delegate.cancel(cause)
+    }
+
+    override fun close(cause: Throwable?): Boolean {
+        if (cause != null) {
+            job?.cancel(CancellationException("channel was closed with cause", cause))
+        }
+        return delegate.close(cause)
+    }
+
+    override fun close() {
+        delegate.close()
+    }
+}
--- a/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/test/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpResponseTest.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/test/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpResponseTest.kt
@@ -98,7 +98,7 @@ class OkHttpResponseTest {
         yield()
 
         val body = sdkResponse.body
-        assertIs<HttpBody.Streaming>(body)
+        assertIs<HttpBody.ChannelContent>(body)
         val ch = body.readFrom()
         assertFalse(ch.isClosedForWrite)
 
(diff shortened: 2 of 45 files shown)


runtime/io/common/src/aws/smithy/kotlin/runtime/io/internal/JobChannel.kt
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Ties the lifetime of the channel and a Job together. Channel failures (closed with exception) will
+ * cause the underlying [Job] to fail with the channel exception.
+ *
+ * This is useful for launching coroutines with the soul purpose of reading/writing data to the channel where
+ * the coroutine should only continue if the channel hasn't failed.
+ */
+@InternalApi
+public class JobChannel(
```
ianbotsf, line 21: **Question**: This only propagates lifecycle events from the channel to the job, not vice versa. Is that important? Would it be possible/correct to also propagate lifecycle from the job to the channel in this class?

runtime/io/common/src/aws/smithy/kotlin/runtime/io/internal/JobChannel.kt
```diff
@@ -0,0 +1,49 @@
...
+    override fun close(cause: Throwable?): Boolean {
+        if (cause != null) {
+            job?.cancel(CancellationException("channel was closed with cause", cause))
+        }
+        return delegate.close(cause)
+    }
+
+    override fun close() {
+        delegate.close()
+    }
```
ianbotsf, line 48: **Question**: Is this necessary? Won't `SdkByteChannel by delegate` already implement this correctly?

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/test/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpResponseTest.kt
```diff
@@ -71,114 +67,4 @@ class OkHttpResponseTest {
 
         assertContentEquals(content, actualBody)
     }
-
-    @Test
-    fun testCallContextCancelsWriter() = runTest {
-        // call context cancelled should result in no children left
```
ianbotsf, line 77: **Question**: This test is no longer necessary because everything's synchronous now?

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/test/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpResponseTest.kt
```diff
@@ -71,114 +67,4 @@ class OkHttpResponseTest {
...
-        assertEquals(1, callContext.job.children.toList().size)
-        callJob.cancel()
-        callJob.complete()
-        callJob.join()
-        assertEquals(0, callContext.job.children.toList().size)
-        assertTrue(ch.isClosedForWrite)
-    }
-
-    @Test
-    fun testSourceReadError(): Unit = runBlocking {
```
ianbotsf, line 114: **Question**: Why is this test no longer necessary?

runtime/io/common/src/aws/smithy/kotlin/runtime/io/internal/JobChannel.kt
aajtodd: It depends on the coroutine but I suppose we might be able to. Example [use](https://github.com/awslabs/smithy-kotlin/blob/323dc3d690e34b80538cb88b96eeeefc609f4027/runtime/io/jvm/src/aws/smithy/kotlin/runtime/io/SdkSourceJVM.kt#L36) where we catch all exceptions and forward to the channel. 

An alternative would be to wire up completion handlers:

```kotlin
job.invokeOnCompletation { cause ->
    ch.cancel(cause)
}
```

I don't necessarily see an issue with this (although given the way backround threads/coroutines work in Android I don't know that we want exceptions to bubble up to coroutine exception handlers anyway). Let me think on this more, we can always add it later if we need it. We don't have a lot of use cases for this yet.

runtime/io/common/src/aws/smithy/kotlin/runtime/io/internal/JobChannel.kt
aajtodd: Good point, it _was_ necessary at one point in time but after I removed `Closeable` from `SdkByteReadChannel` it removed the ambiguity that necessitated this I think.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/test/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpResponseTest.kt
aajtodd: Yeah we aren't launching any background coroutines so there is no cancellation to worry about here. Callers will be reading the body directly rather than proxied via channel

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/test/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpResponseTest.kt
aajtodd: This test was exercising that any exception in the background coroutine resulted in the exception being propagated through the channel (and then later also testing that the exception didn't bubble up which you added recently). 

Neither is necessary now as callers will be reading directly from the underlying OkHttp body. Exceptions will be thrown transparently and there is no cancellation/propagation involved anymore.

lauzadis (approved):

lucix-aws (approved):

PR #746, opened 2022-11-22
feat(rt): use `aws-chunked` 

lauzadis:
This PR enables the usage of the `aws-chunked` content encoding, which was implemented in a [previous PR](https://github.com/awslabs/smithy-kotlin/pull/731). 

## Issue \#

N/A

## Description of changes

This change is required because currently the SDK defaults to unsigned payloads when a body is not replayable. With aws-chunked, we are able to sign the payload even if the body is not replayable.   


runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunkedByteReadChannel.kt
```diff
@@ -203,4 +204,48 @@ public abstract class AbstractAwsChunkedByteReadChannel(
...
+        algorithm = config.algorithm
+        signatureType = AwsSignatureType.HTTP_REQUEST_CHUNK // signature is for a chunk
+        useDoubleUriEncode = config.useDoubleUriEncode
+        normalizeUriPath = config.normalizeUriPath
+        omitSessionToken = config.omitSessionToken
+        hashSpecification = HashSpecification.CalculateFromPayload // calculate the hash from the chunk payload
+        signedBodyHeader = config.signedBodyHeader
+        credentialsProvider = config.credentialsProvider
+        expiresAfter = config.expiresAfter
+    }
```
ianbotsf, line 228: **Nit**: Could be shortened with `toBuilder()`:

```kotlin
private fun getChunkSigningConfig(config: AwsSigningConfig): AwsSigningConfig = config
    .toBuilder()
    .apply {
        signatureType = AwsSignatureType.HTTP_REQUEST_CHUNK // signature is for a chunk
        hashSpecification = HashSpecification.CalculateFromPayload // calculate the hash from the chunk payload
    }
    .build()
```

Applies to `getTrailingHeadersSigningConfig` as well.

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunkedByteReadChannel.kt
```diff
@@ -203,4 +204,48 @@ public abstract class AbstractAwsChunkedByteReadChannel(
...
         return trailerBody
     }
+
+    /**
+     * Given a signing config, make a deep copy, changing the signatureType and hashSpecification configuration values
+     * to specify chunk signing.
+     * @param config the original signing config
+     * @return an [AwsSigningConfig] which can be used by a signer to sign chunks
+     */
+    private fun getChunkSigningConfig(config: AwsSigningConfig): AwsSigningConfig = AwsSigningConfig {
```
ianbotsf, line 214: **Style**: Consider an extension method (e.g., `AwsSigningConfig.toChunkSigningConfig`). Applies to `getTrailingHeadersSigningConfig` as well.

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AbstractAwsChunkedByteReadChannel.kt
```diff
@@ -187,7 +187,8 @@ public abstract class AbstractAwsChunkedByteReadChannel(
      * @return a [ByteArray] containing the trailing headers in aws-chunked encoding, ready to send on the wire
      */
     private suspend fun getTrailingHeadersChunk(trailingHeaders: Headers): ByteArray {
-        val trailerSignature = signer.signChunkTrailer(trailingHeaders, previousSignature, signingConfig).signature
+//        signingConfig.signatureType = AwsSignatureType.HTTP_REQUEST_TRAILING_HEADERS
```
ianbotsf, line 190: **Nit**: Remove commented-out code.

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt
```diff
@@ -29,6 +29,8 @@ public class AwsSigningMiddleware(private val config: Config) : ModifyRequestMid
             requireNotNull(config.signer) { "A signer must be specified for the middleware" }
             return AwsSigningMiddleware(config)
         }
+
+        public var AWS_CHUNKED_THRESHOLD: Int = CHUNK_SIZE_BYTES * 16
```
ianbotsf, line 33: **Correctness**: `var` → `const val`.

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt
```diff
@@ -147,17 +149,28 @@ public class AwsSigningMiddleware(private val config: Config) : ModifyRequestMid
                 contextHashSpecification != null -> contextHashSpecification
                 config.isUnsignedPayload -> HashSpecification.UnsignedPayload
                 body is HttpBody.Empty -> HashSpecification.EmptyBody
-                body is HttpBody.Streaming && !body.isReplayable -> {
-                    coroutineContext.warn<AwsSigningMiddleware> {
-                        "unable to compute hash for unbounded stream; defaulting to unsigned payload"
+                body is HttpBody.Streaming && (!body.isReplayable || body.isEligibleForAwsChunkedStreaming()) -> {
```
ianbotsf, line 152: **Nit**: Seems like the `isReplayable` check could be moved into `isEligibleForAwsChunkedStreaming`.

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt
```diff
@@ -190,3 +212,8 @@ private fun HttpRequestBuilder.update(signedRequest: HttpRequest) {
         }
     }
 }
+
+/**
+ * @return a boolean representing if this HttpBody is large enough to send via aws-chunked
+ */
+private fun HttpBody.isEligibleForAwsChunkedStreaming(): Boolean = (this.contentLength ?: 0) > AwsSigningMiddleware.AWS_CHUNKED_THRESHOLD
```
ianbotsf, line 219: **Style**: This seems like a natural use case for extension `val` (rather than extension `fun`).

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt
```diff
@@ -147,17 +149,28 @@ public class AwsSigningMiddleware(private val config: Config) : ModifyRequestMid
...
 
+        when (signingConfig.hashSpecification) {
+            // Add aws-chunked headers
+            is HashSpecification.StreamingAws4HmacSha256Payload, is HashSpecification.StreamingAws4HmacSha256PayloadWithTrailers -> {
+                req.subject.headers.append("Content-Encoding", "aws-chunked")
+                req.subject.headers.append("Transfer-Encoding", "chunked")
+                req.subject.headers.append("X-Amz-Decoded-Content-Length", (body.contentLength ?: 0).toString())
+            }
+            else -> {}
+        }
```
ianbotsf, line 172: **Question**: It seems odd to do this outside of the signer/canonicalizer. Why is this the right place to mutate the request for chunked encoding?

runtime/auth/aws-signing-tests/common/src/aws/smithy/kotlin/runtime/auth/awssigning/tests/MiddlewareSigningTestBase.kt
```diff
@@ -28,7 +28,9 @@ import kotlin.test.assertEquals
 @Suppress("HttpUrlsUsage")
 @OptIn(ExperimentalCoroutinesApi::class)
 public abstract class MiddlewareSigningTestBase : HasSigner {
-    private fun buildOperation(streaming: Boolean = false, replayable: Boolean = true): SdkHttpOperation<Unit, HttpResponse> = SdkHttpOperation.build {
+    private var readFromChannel: SdkByteReadChannel? = null
```
ianbotsf, line 31: **Suggestion**: Storing the read channel as a side-effect of `buildOperation` works but it seems messy. Consider returning `Pair<SdkHttpOperation<Unit, HttpResponse>, SdkByteReadChannel?>` from `buildOperation` instead. Tests which need it can get it as `val (op, channel) = buildOperation(...)`. Tests which don't can either use `val (op, _) = buildOperation(...)` or use a convenience function (e.g., `buildOperation` which calls `buildOperationWithChannel`).

ianbotsf (changes requested):

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt
lauzadis: I agree, it seems out of place when comparing to the canonicalizer. I just finished played around with it and don't think appending these headers should happen in the canonicalizer, because only the default signer uses it -- the CRT signer canonicalizes requests internally.

Do you think creating a new middleware here makes sense? Something that will check if the body should be uploaded with `aws-chunked`, and append the appropriate headers if so. 

Also, what do you think about adding a new middleware for lines 183-191 below, where the request body is wrapped in an `AwsChunkedByteReadChannel`? It also seems a little out of place here in the signing middleware. One challenge with a new middleware would be getting access to the signing config -- it would have to be added to the execution context here and fetched in the new middleware.

aajtodd: This isn't part of canonicalization though so it doesn't belong there. It's part of `aws-chunked` content encoding. We make a decision to encode it a certain way and need to update the body to reflect that decision. 

I don't think it belongs in a separate middleware (nor the other lines below).

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt
```diff
@@ -29,6 +29,8 @@ public class AwsSigningMiddleware(private val config: Config) : ModifyRequestMid
             requireNotNull(config.signer) { "A signer must be specified for the middleware" }
             return AwsSigningMiddleware(config)
         }
+
+        public const val AWS_CHUNKED_THRESHOLD: Int = CHUNK_SIZE_BYTES * 16
```
aajtodd, line 33: fix: Documentation

question: Should this be `public` or `internal`? (if public should it be `@InternalApi`?)

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt
```diff
@@ -190,3 +212,9 @@ private fun HttpRequestBuilder.update(signedRequest: HttpRequest) {
         }
     }
 }
+
+/**
+ * @return a boolean representing if this HttpBody is eligible to send via aws-chunked content encoding
+ */
+private val HttpBody.isEligibleForAwsChunkedStreaming: Boolean
+    get() = !(this as HttpBody.Streaming).isReplayable || (this.contentLength ?: 0) > AwsSigningMiddleware.AWS_CHUNKED_THRESHOLD
```
aajtodd, line 220: nit: This `as` is only safe because we use it after checking it's `HttpBody.Streaming` on line 152, we could just remove the check on line 152 and move the logic completely in here:

```kotlin
private val HttpBody.isEligibleForAwsChunkedStreaming: Boolean
    get() = (this is HttpBody.Streaming) && (!isReplayable) || (contentLength ?: 0) > AwsSigningMiddleware.AWS_CHUNKED_THRESHOLD)
```

I also think this should probably only apply to streaming bodies, for large in-memory bodies we should just calculate the hash from the payload (the updated logic here would handle that)

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt
```diff
@@ -147,17 +149,28 @@ public class AwsSigningMiddleware(private val config: Config) : ModifyRequestMid
...
                 else -> HashSpecification.CalculateFromPayload
             }
         }
 
+        when (signingConfig.hashSpecification) {
+            // Add aws-chunked headers
+            is HashSpecification.StreamingAws4HmacSha256Payload, is HashSpecification.StreamingAws4HmacSha256PayloadWithTrailers -> {
+                req.subject.headers.append("Content-Encoding", "aws-chunked")
+                req.subject.headers.append("Transfer-Encoding", "chunked")
+                req.subject.headers.append("X-Amz-Decoded-Content-Length", (body.contentLength ?: 0).toString())
```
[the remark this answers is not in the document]
aajtodd: Is it required by `aws-chunked`? If so I'd make it part of the `isEligible` check and send it unsigned if it's not replayable and not eligible.

If it's not required then I'm not sure what should be sent.

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt
```diff
@@ -147,17 +149,28 @@ public class AwsSigningMiddleware(private val config: Config) : ModifyRequestMid
                 contextHashSpecification != null -> contextHashSpecification
                 config.isUnsignedPayload -> HashSpecification.UnsignedPayload
                 body is HttpBody.Empty -> HashSpecification.EmptyBody
-                body is HttpBody.Streaming && !body.isReplayable -> {
-                    coroutineContext.warn<AwsSigningMiddleware> {
-                        "unable to compute hash for unbounded stream; defaulting to unsigned payload"
+                body is HttpBody.Streaming && body.isEligibleForAwsChunkedStreaming -> {
```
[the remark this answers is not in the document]
lauzadis: Yeah, I think that would be a good idea. Maybe I can add a field to the middleware configuration? (similar to `config.isUnsignedPayload`)

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt
lauzadis: Currently it needs to be `internal` to be accessible in some test cases. Since the class AwsSigningMiddleware is `@InternalApi`, does its companion object need to be marked, or is it inherited?

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt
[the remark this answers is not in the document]
aajtodd: No lets leave it right now, I think interceptors will give us the escape hatch (as I talked myself into above)

ianbotsf (approved):

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HttpBody.kt
```diff
@@ -132,6 +132,20 @@ public fun SdkByteReadChannel.toHttpBody(contentLength: Long? = null): HttpBody
     }
 }
 
+/**
+ * Convert an [SdkSource] to an [HttpBody]
+ * @param contentLength the total content length of the channel if known
```
aajtodd, line 137: fix: remove channel/ch naming from this function since it doesn't use a channel

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/internal/AwsChunkedUtil.kt
```diff
@@ -28,3 +34,37 @@ internal fun SdkBuffer.writeTrailers(
...
+    headers.setMissing("Transfer-Encoding", "chunked")
+    headers.setMissing("X-Amz-Decoded-Content-Length", body.contentLength!!.toString())
+}
+
+internal expect fun HttpRequestBuilder.internalSetAwsChunkedBody(signer: AwsSigner, signingConfig: AwsSigningConfig, signature: ByteArray)
+
+/**
+ * Update the HTTP body to use aws-chunked content encoding
+ */
+internal fun HttpRequestBuilder.setAwsChunkedBody(signer: AwsSigner, signingConfig: AwsSigningConfig, signature: ByteArray) {
```
aajtodd, line 68: question: why the redirect to `internalSetAwsChunkedBody`? why not just make this function `expect/actual` instead?

aajtodd (approved):
Looks good! Couple questions and minor fixes requested. No need for second approval though.

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/internal/AwsChunkedUtil.kt
lauzadis: discussed offline, fixed the NoSuchMethodError with a clean build >:)

PR #751, opened 2022-11-28
refactor(rt)!: update SDK I/O abstractions

aajtodd:
## Issue \#

This PR is the final PR to main that is a combination of several smaller PRs already reviewed:

* https://github.com/awslabs/smithy-kotlin/pull/733
* https://github.com/awslabs/smithy-kotlin/pull/736
* https://github.com/awslabs/smithy-kotlin/pull/743
* https://github.com/awslabs/smithy-kotlin/pull/749

## Description of changes

The individual PRs above have all the details but the highlights are copied below:

-----
1. Refactor `SdkByteChannel` to be based on the new `SdkBuffer` type. Moving data between buffers is cheap (especially in bulk) so this type decouples a producer and consumer with minimal overhead. Ktor based channels did some clever things internally but in many common cases it still required an explicit copy for each read/write method. Having a single read/write method to implement makes layering/wrapping a channel much easier than before (similar to how easy it is to layer okio Source/Sink types).

    * Removes ktor based channel dependency
    * Updates `SdkByteReadChannel/SdkByteWriteChannel` to have a single `read/write` method based on `SdkBuffer`. This makes it effectively an async `Source/Sink` or an async `Pipe` depending on how you want to look at it. 

2. Refactor `ByteStream` and `HttpBody` to allow an `SdkSource` variant. The primary effect of this is:
    1. Http engine implementations (and customers if they are unwrapping `ByteStream` and not using a convenience function) have to deal with a new variant
        * `SdkSource` is a blocking interface which should be fine on JVM/Native but not on JS. We haven't fully thought this through but I expect the answer will be that we shuffle everything on JS through the channel variant and error if we see anything else. We won't be able to make this a compile time error unfortunately without knee capping `ByteStream`/`HttpBody` in common source sets. I chose not to go this route, I'm fine with the experience on JS being slightly less than ideal if you carve your own path outside the one we will define
    2. File based input doesn't have to proxy itself through a channel, removing an extra layer of buffering/overhead 


runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannel.kt
```diff
@@ -97,3 +100,34 @@ public suspend fun SdkByteReadChannel.readToBuffer(): SdkBuffer {
     readRemaining(buffer)
     return buffer
 }
+
+/**
+ * Read all bytes from this channel into [sink]. Returns the total number of bytes written.
+ */
+public suspend fun SdkByteReadChannel.readAll(sink: SdkSink): Long = withContext(SdkDispatchers.IO) {
+    val bufferedSink = if (sink is SdkBuffer) sink else sink.buffer()
```
ianbotsf, line 108: **Question**: Doesn't `SdkSink.buffer()` already no-op when `this` is buffered?

runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannel.kt
```diff
@@ -97,3 +100,34 @@ public suspend fun SdkByteReadChannel.readToBuffer(): SdkBuffer {
...
+    var totalWritten = 0L
+    while (true) {
+        val rc = read(bufferedSink.buffer, DEFAULT_BYTE_CHANNEL_MAX_BUFFER_SIZE.toLong())
+        if (rc == -1L) break
+        totalWritten += rc
+        bufferedSink.emit()
+    }
+    bufferedSink.emit()
```
ianbotsf, line 116: **Question**: Remind me why we `emit` inside the loop and again after the loop? We didn't read anything between the last loop `emit` and exiting the loop, right?

ianbotsf (approved):

lauzadis (approved):

runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannel.kt
aajtodd: I don't recall of top of my head. I want to say there is some state you can get into where you have some data leftover from the last read but it's not obvious from looking at the code 🤔 . I can try taking it out and see what breaks. (It's not going to hurt anything though to leave it in)

runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannel.kt
aajtodd: Yes, probably wasn't true at the time this was added (or was missed in rebasing). I can remove it

lauzadis (approved):

PR #756, opened 2022-12-02
Add HashingSource and HashingSink

lauzadis:
Adds HashingSource and HashingSink, which hash their data as they are being read and written, respectively.

## Description of changes

This change is required to generalize the CrcSource to support any hashing function. It is planned to be used in the flexible checksums implementation.

### Note on digestValue() 
Crc32 and Crc32C have an additional function called `digestValue()` which returns an unsigned int. To keep this function but also provide generic support for any hashing function, the HashingSource and HashingSink provide access to the underlying hash function as a member variable called `hash`.

With this, the user can cast the hash function to Crc32 and call the `digestValue()` function.

**Example**
```kotlin
var hs = HashingSource(Crc32(), ByteArray(1024) { 0xf }.source())  // CRC32 hashing source with 1024 bytes
val sink = SdkBuffer()

hs.read(sink, 1024)

val digestValue = (hs.hash as Crc32).digestValue()
```

An alternative design would be to create a class like `CrcHashingSource` which implements `HashingSource` and also provides `digestValue()`, but I thought the first option was better.   


runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashingSink.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import aws.smithy.kotlin.runtime.io.SdkSink
+import aws.smithy.kotlin.runtime.io.internal.SdkSinkObserver
+
+/**
+ * A sink which hashes data as it is being written to
+ * @param hash The [HashFunction] to hash data with
+ * @param sink the [SdkSink] to hash
+ */
+public class HashingSink(public val hash: HashFunction, sink: SdkSink) : HashFunction by hash, SdkSinkObserver(sink) {
```
ianbotsf, line 16: **Correctness**: Why implement `HashFunction`? Is there a use case for calling `update`/`reset` on a sink? Will we ever pass a `HashingSink` as an argument to `ByteArray.hash`?

runtime/hashing/build.gradle.kts
```diff
@@ -12,6 +12,7 @@ kotlin {
         commonMain {
             dependencies {
                 implementation(project(":runtime:utils"))
+                implementation(project(":runtime:io"))
```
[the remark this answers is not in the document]
ianbotsf: Yes, I believe they'd be better off in **io**.

runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashingSink.kt
```diff
@@ -0,0 +1,20 @@
...
+    override fun observe(data: ByteArray, offset: Int, length: Int) {
+        update(data, offset, length)
+    }
+}
```
ianbotsf, line 20: **Correctness**: This largely duplicates [the `private` class in **Canonicalizer.kt**](https://github.com/awslabs/smithy-kotlin/blob/855a77a3ea1e9a4b7f2fe005e297de13c7559732/runtime/auth/aws-signing-default/common/src/aws/smithy/kotlin/runtime/auth/awssigning/Canonicalizer.kt#L182-L189). We should remove that and update the surrounding code to use this.

runtime/hashing/common/test/aws/smithy/kotlin/runtime/hashing/HashingSinkTest.kt
```diff
@@ -0,0 +1,81 @@
...
+
+        val hashingSink = HashingSink(Crc32(), SdkSink.blackhole())
+
+        val expectedHash = Crc32()
+
+        assertEquals(expectedHash.digest().decodeToString(), hashingSink.digest().decodeToString())
+        hashingSink.write(buffer, buffer.size)
+        expectedHash.update(byteArray)
+        assertEquals(expectedHash.digest().decodeToString(), hashingSink.digest().decodeToString())
+    }
```
ianbotsf, line 45: **Question**: Are these two tests identical?

runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashingSink.kt
ianbotsf, line 16: **Comment**: Making the underlying `hash` publicly visible is inadvisable since that means its state can be mutated by object holders. We should probably make it private instead.

runtime/hashing/common/test/aws/smithy/kotlin/runtime/hashing/HashingSourceTest.kt
```diff
@@ -0,0 +1,85 @@
...
+        val sink = SdkBuffer()
+
+        val expectedHash = Crc32()
+        assertEquals(expectedHash.digest().decodeToString(), hashingSource.digest().decodeToString())
+
+        hashingSource.read(sink, 1024)
+        expectedHash.update(byteArray)
+
+        assertEquals(expectedHash.digest().decodeToString(), hashingSource.digest().decodeToString())
+    }
```
ianbotsf, line 47: **Question**: Identical?

ianbotsf (changes requested):

runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashingSink.kt
ianbotsf: Or use a `HashSupplier` instead.

runtime/hashing/common/test/aws/smithy/kotlin/runtime/hashing/HashingSinkTest.kt
lauzadis: They weren't initially, but I refactored and they ended up being basically the same, just using a different hash function. I will remove the duplication

runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashingSink.kt
lauzadis: there's no particular reason, and you're right, there are no strong use cases for directly calling update / reset on this sink, so I'll remove this implementation

runtime/hashing/common/test/aws/smithy/kotlin/runtime/hashing/HashingSinkTest.kt
ianbotsf: Oh my eyes missed the `Crc32` vs `Crc32c`. That's great to test both, maybe just refactor the test itself into a helper method that accepts a hash implementation.

runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashingSink.kt
lauzadis: The reason I made this public was for the `digestValue()` use case for CRC hash functions. I'm not sure how to nicely provide this function if the underlying hash is private, but I understand the problem with mutation by object holders.

From what I see, HashSupplier wouldn't be great here because it returns a new HashFunction each invocation, while we want to use the same HashFunction object to get the hash value of the entire underlying Source/Sink.

For `digestValue()`, I could do this:
```kotlin
public class HashingSource(private val hash: HashFunction, source: SdkSource) : SdkSourceObserver(source) {
    override fun observe(data: ByteArray, offset: Int, length: Int) {
        hash.update(data, offset, length)
    }

    public fun digestValue() : UInt? = (hash as Crc32?)?.digestValue()
}
```

But it seems too special-cased for a generic HashingSource

aajtodd: I'd probably make these `InternalApi`

aajtodd: If anything I would create a `CrcSource/Sink` instead of trying to force them through `HashingSource/Sink`. It's going to look nearly identical but it won't be awkwardly special casing the generic hashing source/sink

runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashingSink.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import aws.smithy.kotlin.runtime.io.SdkSink
+import aws.smithy.kotlin.runtime.io.internal.SdkSinkObserver
+
+/**
+ * A sink which hashes data as it is being written to
+ * @param hash The [HashFunction] to hash data with
+ * @param sink the [SdkSink] to hash
+ */
+public class HashingSink(private val hash: HashFunction, sink: SdkSink) : SdkSinkObserver(sink) {
```
aajtodd, line 16: `@InternalApi`

runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashingSink.kt
```diff
@@ -0,0 +1,26 @@
...
+    override fun observe(data: ByteArray, offset: Int, length: Int) {
+        hash.update(data, offset, length)
+    }
+
+    /**
+     * Provides the digest value as an unsigned integer for the CRC32 function family.
+     * @return unsigned integer representing the value of the digest, if the [HashFunction] is [Crc32] or [Crc32c], and null otherwise.
+     */
+    public fun digestValue() : UInt? = (hash as Crc32?)?.digestValue()
```
aajtodd, line 25: I think either special case OR move this to an `internal` package as an extension function (would require making `hash` `internal` rather than `private`

aajtodd (approved):
Fix and ship

runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashingSink.kt
ianbotsf: Yes, a single unsigned-int digest value is very specific to CRC32. We shouldn't include it in anything that utilizes generic hash functionality. How will this be used eventually outside of things like tests?

lauzadis: It's being used for event steams today. I'm adding it only to change event streams' [CrcSource implementation](https://github.com/awslabs/aws-sdk-kotlin/blob/c099049829d6a45307173383c50bf40f0095825d/aws-runtime/protocols/aws-event-stream/common/src/aws/sdk/kotlin/runtime/protocol/eventstream/CrcUtil.kt#L18) to this new one.

ianbotsf: Why does the event streams code need to treat the CRC as a `UInt`? Could it just as easily use byte arrays?

lauzadis: It's possible. Event streams sends the expected checksum value as `UInt`, so we need to either
1. compare using digestValue()'s `UInt` (how it's done today)
2. convert the given expected checksum from `UInt` to `ByteArray`, and compare using `digest()` instead

I didn't want to make too many changes in event streams besides replacing the CrcSource/Sink that was used there, which is why I went with option 1.

ianbotsf: Discussed offline, we decided to use just `digest()`/`ByteArray` everywhere for hash functions _including_ CRC32. We'll remove specialized `digestValue()` members for the source/sink.

aajtodd (approved):
Fix and ship

PR #757, opened 2022-12-05
fix(rt): allow root spans to inherit their parent from the context when not already set

aajtodd:
## Issue \#

fixes https://github.com/awslabs/aws-sdk-kotlin/issues/789

## Description of changes

Allow a root span to be created when an existing span is present in the current coroutine context as long as the root span has no parent. An existing span and a root span with a parent is still not allowed because it would represent an illegal state (two different span lineages). 

### DIFF
--- a/runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/CoroutineContextUtils.kt
+++ b/runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/CoroutineContextUtils.kt
@@ -46,17 +46,30 @@ public suspend inline fun <R> CoroutineContext.withChildTraceSpan(
 }
 
 @InternalApi
+public class WrappedRootSpan(
+    private val rootSpan: TraceSpan,
+    override val parent: TraceSpan,
+) : TraceSpan by rootSpan
+
+@InternalApi
 public suspend inline fun <R> CoroutineContext.withRootTraceSpan(
     rootSpan: TraceSpan,
     crossinline block: suspend CoroutineScope.() -> R,
 ): R =
     try {
         val existingSpan = get(TraceSpanContextElement)?.traceSpan
-        check(existingSpan == null || existingSpan == rootSpan.parent) {
+        check(existingSpan == null || rootSpan.parent == null || existingSpan == rootSpan.parent) {
             "This method may only be called when no current span exists or the new span is a child of the active span"
         }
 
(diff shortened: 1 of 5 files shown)


runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/CoroutineContextUtils.kt
```diff
@@ -45,18 +45,31 @@ public suspend inline fun <R> CoroutineContext.withChildTraceSpan(
     }
 }
 
+@InternalApi
+public class WrappedRootSpan(
```
ianbotsf, line 49: **Question**: Why `@InternalApi`/`public`? Seems like this class is only necessary in this file and could be `private`.

runtime/tracing/tracing-core/common/test/aws/smithy/kotlin/runtime/tracing/TraceSpanTest.kt
```diff
@@ -0,0 +1,82 @@
...
+import io.kotest.matchers.string.shouldContain
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+import kotlin.test.assertNull
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class TraceSpanTest {
```
ianbotsf, line 17: **Nit**: This is really testing the `CoroutineContext` utils, not the `TraceSpan` interface. We should probably rename the file/class.

ianbotsf (approved):
Minor comments, fix and ship.

runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/CoroutineContextUtils.kt
aajtodd: it's used in an `inline` function so it can't be private. Options are:

1. Public with internal annotation
2. `internal` with `@PublishedApi`
I have no preference

aajtodd: (`@PublishedApi` affects binary compatibility still according to docs)

PR #760, opened 2022-12-09
fix(rt): add special-casing for CrtHttpEngine `aws-chunked` requests

lauzadis:
CRT requires `aws-chunked` requests to use a specific API instead of a normal streaming request body. This PR adds a special-case to the CRT engine, which will use this new API instead of sending an `aws-chunked` body.

## Issue \#

https://github.com/awslabs/smithy-kotlin/issues/759

## Description of changes

Currently `aws-chunked` requests do not work when using the CrtHttpEngine.

### DIFF
--- a/runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
@@ -10,7 +10,6 @@ import aws.sdk.kotlin.crt.io.*
 import aws.smithy.kotlin.runtime.ClientException
 import aws.smithy.kotlin.runtime.client.ExecutionContext
 import aws.smithy.kotlin.runtime.crt.SdkDefaultIO
-import aws.smithy.kotlin.runtime.http.HttpBody
 import aws.smithy.kotlin.runtime.http.engine.HttpClientEngine
 import aws.smithy.kotlin.runtime.http.engine.HttpClientEngineBase
 import aws.smithy.kotlin.runtime.http.engine.ProxyConfig
@@ -18,10 +17,7 @@ import aws.smithy.kotlin.runtime.http.engine.callContext
 import aws.smithy.kotlin.runtime.http.operation.getLogger
 import aws.smithy.kotlin.runtime.http.request.HttpRequest
 import aws.smithy.kotlin.runtime.http.response.HttpCall
-import aws.smithy.kotlin.runtime.io.SdkBuffer
-import aws.smithy.kotlin.runtime.io.buffer
 import aws.smithy.kotlin.runtime.io.internal.SdkDispatchers
-import aws.smithy.kotlin.runtime.io.readToByteArray
 import aws.smithy.kotlin.runtime.logging.Logger
 import aws.smithy.kotlin.runtime.time.Instant
 import kotlinx.coroutines.job
@@ -142,34 +138,5 @@ public class CrtHttpEngine(public val config: CrtHttpEngineConfig) : HttpClientE
             }.build()
             HttpClientConnectionManager(connOpts)
         }
-    }
-}
-
-internal suspend fun HttpStream.sendChunkedBody(body: HttpBody) {
-    when (body) {
-        is HttpBody.SourceContent -> {
-            val source = body.readFrom()
-            val bufferedSource = source.buffer()
-
-            while (!bufferedSource.exhausted()) {
-                bufferedSource.request(CHUNK_BUFFER_SIZE)
-                this.writeChunk(bufferedSource.buffer.readByteArray(), isFinalChunk = bufferedSource.exhausted())
-            }
-        }
-        is HttpBody.ChannelContent -> {
-            val chan = body.readFrom()
-            var buffer = SdkBuffer()
-            val nextBuffer = SdkBuffer()
-
-            while (!chan.isClosedForRead) {
-                chan.read(buffer, CHUNK_BUFFER_SIZE)
-
-                val isFinalChunk = chan.read(nextBuffer, CHUNK_BUFFER_SIZE) == -1L
-
-                this.writeChunk(buffer.readToByteArray(), isFinalChunk)
-                if (isFinalChunk) break else buffer = nextBuffer
-            }
-        }
-        else -> {}
     }
 }
--- a/runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
@@ -19,7 +19,7 @@ import aws.smithy.kotlin.runtime.http.operation.getLogger
 import aws.smithy.kotlin.runtime.http.request.HttpRequest
 import aws.smithy.kotlin.runtime.http.response.HttpCall
 import aws.smithy.kotlin.runtime.io.SdkBuffer
-import aws.smithy.kotlin.runtime.io.readToByteArray
+import aws.smithy.kotlin.runtime.io.buffer
 import aws.smithy.kotlin.runtime.logging.Logger
 import aws.smithy.kotlin.runtime.time.Instant
 import kotlinx.coroutines.job
@@ -28,6 +28,7 @@ import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withTimeoutOrNull
 
 internal const val DEFAULT_WINDOW_SIZE_BYTES: Int = 16 * 1024
+internal const val CHUNK_BUFFER_SIZE: Long = 64 * 1024
 
 /**
  * [HttpClientEngine] based on the AWS Common Runtime HTTP client
@@ -104,23 +105,14 @@ public class CrtHttpEngine(public val config: CrtHttpEngineConfig) : HttpClientE
         stream.activate()
 
         if (request.isChunked) {
-            val bytesToRead = 65536L
-
             when (request.body) {
                 is HttpBody.SourceContent -> {
                     val source = (request.body as HttpBody.SourceContent).readFrom()
-                    val nextBuffer = SdkBuffer()
-                    var buffer = SdkBuffer()
+                    val bufferedSource = source.buffer()
 
-                    while (true) {
-                        source.read(buffer, bytesToRead)
-
-                        // read another set of bytes into a second buffer -- this is how we can "peek" to decide if this is the final chunk or not
-                        val isFinalChunk = source.read(nextBuffer, bytesToRead) == -1L
-
-                        stream.writeChunk(buffer.readToByteArray(), isFinalChunk)
-
-                        if (isFinalChunk) break else buffer = nextBuffer
+                    while (!bufferedSource.exhausted()) {
+                        bufferedSource.request(CHUNK_BUFFER_SIZE)
+                        stream.writeChunk(bufferedSource.buffer.readByteArray(), isFinalChunk = bufferedSource.exhausted())
                     }
                 }
                 is HttpBody.ChannelContent -> {
@@ -128,7 +120,7 @@ public class CrtHttpEngine(public val config: CrtHttpEngineConfig) : HttpClientE
                     val buffer = SdkBuffer()
 
                     while (!chan.isClosedForRead) {
-                        val bytesRead = chan.read(buffer, bytesToRead)
+                        val bytesRead = chan.read(buffer, CHUNK_BUFFER_SIZE)
                         check(bytesRead != -1L) { "unexpected exhaustion of channel" }
                         stream.writeChunk(buffer.readByteArray(bytesRead), isFinalChunk = chan.isClosedForRead)
                     }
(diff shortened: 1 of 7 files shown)


runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
```diff
@@ -100,6 +103,40 @@ public class CrtHttpEngine(public val config: CrtHttpEngineConfig) : HttpClientE
...
+                    while (true) {
+                        source.read(buffer, bytesToRead)
+
+                        // read another set of bytes into a second buffer -- this is how we can "peek" to decide if this is the final chunk or not
+                        val isFinalChunk = source.read(nextBuffer, bytesToRead) == -1L
+
+                        stream.writeChunk(buffer.readToByteArray(), isFinalChunk)
+
+                        if (isFinalChunk) break else buffer = nextBuffer
+                    }
```
ianbotsf, line 124: **Question**: Am I reading this correctly that the first `writeChunk` will get 64K, subsequent `writeChunk`s will get 128K (read, swap, read again), and the final `writeChunk` will get whatever's left? If so, does the initial chunk's smaller size matter?

runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/RequestUtil.kt
```diff
@@ -34,17 +34,18 @@ internal val HttpRequest.uri: Uri
...
+        body is HttpBody.Bytes -> HttpRequestBodyStream.fromByteArray(body.bytes())
+        body is HttpBody.ChannelContent -> ReadChannelBodyStream(body.readFrom(), callContext)
+        body is HttpBody.SourceContent -> {
             val source = body.readFrom()
             callContext.job.invokeOnCompletion {
                 source.close()
             }
             SdkSourceBodyStream(source)
         }
+        else -> null
```
ianbotsf, line 48: **Question**: What could `else` be if not one of the above cases? Should the stream be `null` or should we actually throw an error?

lauzadis: Probably throw an error. I wish there was a better way to format this `when` block, ideally it would enumerate all the `HttpBody` types and nothing else, but I needed access to the `HttpRequest` to call `isChunked`

runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
lauzadis: That's right, and the sizes are all arbitrary. We should select a size that we are comfortable loading into memory all-at-once, and I think 128K is fine.

runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
```diff
@@ -100,6 +103,40 @@ public class CrtHttpEngine(public val config: CrtHttpEngineConfig) : HttpClientE
         val stream = conn.makeRequest(engineRequest, respHandler)
         stream.activate()
 
+        if (request.isChunked) {
+            val bytesToRead = 65536L
```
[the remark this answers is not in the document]
lauzadis: There is no "right" number here, it is just the size of the buffer used to move data from the `aws-chunked` body to the CRT engine. I can clarify that and make it a `private const val`.

runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
```diff
@@ -100,6 +103,40 @@ public class CrtHttpEngine(public val config: CrtHttpEngineConfig) : HttpClientE
...
+            when (request.body) {
+                is HttpBody.SourceContent -> {
+                    val source = (request.body as HttpBody.SourceContent).readFrom()
+                    val nextBuffer = SdkBuffer()
+                    var buffer = SdkBuffer()
+
+                    while (true) {
+                        source.read(buffer, bytesToRead)
+
+                        // read another set of bytes into a second buffer -- this is how we can "peek" to decide if this is the final chunk or not
```
aajtodd, line 118: **fix**: This double buffering logic is unnecessary. 

Instead you can structure this like:

```kotlin
val source = request.body.readFrom()
val bufferedSource = source.buffer()

while(true) {
    bufferedSource.request(bytesToRead)
    
    bufferedSource.buffer.readByteArray()
     ...
    val isFinalChunk = bufferedSource.exhausted()
    ...
}

```

runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/RequestUtil.kt
aajtodd: this should throw an error on the `else` clause now

aajtodd (changes requested):

runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
```diff
@@ -100,6 +103,40 @@ public class CrtHttpEngine(public val config: CrtHttpEngineConfig) : HttpClientE
...
+                        val isFinalChunk = source.read(nextBuffer, bytesToRead) == -1L
+
+                        stream.writeChunk(buffer.readToByteArray(), isFinalChunk)
+
+                        if (isFinalChunk) break else buffer = nextBuffer
+                    }
+                }
+                is HttpBody.ChannelContent -> {
+                    val chan = (request.body as HttpBody.ChannelContent).readFrom()
```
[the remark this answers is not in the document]
lauzadis: I agree, but I tried it earlier and got this warning:

```
Smart cast to 'HttpBody.ChannelContent' is impossible, because 'request.body' is a public API property declared in different module
```

runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/RequestUtil.kt
lauzadis: I refactored to get rid of the dangling `else`:

```kotlin
    val bodyStream = if (isChunked) null else when (body) {
        is HttpBody.Empty -> null
        ...
    }
```

runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
```diff
@@ -100,6 +103,40 @@ public class CrtHttpEngine(public val config: CrtHttpEngineConfig) : HttpClientE
...
+                    val buffer = SdkBuffer()
+
+                    while (!chan.isClosedForRead) {
+                        val bytesRead = chan.read(buffer, bytesToRead)
+                        check(bytesRead != -1L) { "unexpected exhaustion of channel" }
```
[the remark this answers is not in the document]
lauzadis: We can't send a 0-byte chunk as the final chunk, it has to have some data. This is because CRT does a check on the size of the chunk and skips processing if it's 0.

I see the issue, I will refactor it

runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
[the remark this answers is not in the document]
ianbotsf: The compiler _can't_ know this is safe because `body` is a `val` which, while always _read-only_, may be _non-constant_.

runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
```diff
@@ -168,3 +144,32 @@ public class CrtHttpEngine(public val config: CrtHttpEngineConfig) : HttpClientE
...
+            while (!chan.isClosedForRead) {
+                chan.read(buffer, CHUNK_BUFFER_SIZE)
+
+                val isFinalChunk = chan.read(nextBuffer, CHUNK_BUFFER_SIZE) == -1L
+
+                this.writeChunk(buffer.readToByteArray(), isFinalChunk)
+                if (isFinalChunk) break else buffer = nextBuffer
+            }
+        }
+        else -> {}
```
ianbotsf, line 173: **Nit**: `else -> error("sendChunkedBody shouldn't be called for non-chunked body types")` or something like that.

runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
```diff
@@ -168,3 +144,32 @@ public class CrtHttpEngine(public val config: CrtHttpEngineConfig) : HttpClientE
...
+                this.writeChunk(bufferedSource.buffer.readByteArray(), isFinalChunk = bufferedSource.exhausted())
+            }
+        }
+        is HttpBody.ChannelContent -> {
+            val chan = body.readFrom()
+            var buffer = SdkBuffer()
+            val nextBuffer = SdkBuffer()
+
+            while (!chan.isClosedForRead) {
+                chan.read(buffer, CHUNK_BUFFER_SIZE)
```
ianbotsf, line 165: **Question**: Is it possible on the initial iteration for `isClosedForRead` to return `false` but then the first `read` to return `-1` as a result of suspension?

aajtodd: Yes its possible, although wouldn't that be an empty body? We wouldn't be able to handle that if I understood CRT requirements on sending empty chunks this morning.

runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/RequestUtil.kt
```diff
@@ -61,3 +61,11 @@ internal fun HttpRequest.toCrtRequest(callContext: CoroutineContext): aws.sdk.ko
...
     return aws.sdk.kotlin.crt.http.HttpRequest(method.name, url.encodedPath, crtHeaders.build(), bodyStream)
 }
+
+/**
+ * @return whether this HttpRequest is a chunked request.
+ * Specifically, this means return `true` if a request contains a `Transfer-Encoding` header with the value `chunked`,
+ * and the body is either [HttpBody.SourceContent] or [HttpBody.ChannelContent].
+ */
+internal val HttpRequest.isChunked: Boolean get() = (this.body is HttpBody.SourceContent || this.body is HttpBody.ChannelContent) &&
+    this.headers.getAll("Transfer-Encoding")?.contains("chunked") == true
```
aajtodd, line 71: nit: `headers.contains("Transfer-Encoding", "chunked")` already checks if any header contains the value

nit 2: You can generally leave off qualifying `this` on extension functions most of the time.

runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
```diff
@@ -131,3 +144,32 @@ public class CrtHttpEngine(public val config: CrtHttpEngineConfig) : HttpClientE
         }
     }
 }
+
+internal suspend fun HttpStream.sendChunkedBody(body: HttpBody) {
```
aajtodd, line 148: nit: This extension doesn't feel like it belongs in this file, maybe move to `RequestUtil.kt` or something.

runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
lauzadis: Yeah it is possible. It's not necessarily an empty body, it could be a nonempty body that got closed prematurely for some reason. CRT wouldn't be able to handle it either way though. Should I try to handle this case and throw an exception or just let it blow up? It seems like a very unlikely thing to happen.

ianbotsf: I advocate for additional handling of this case. _When_ we have to debug a failure here, it'll be nice to have a clear indication what happened rather than a cryptic CRT exception.

lauzadis: Alright, I added some logic to handle this case where the initial chunk hasn't been sent but the channel is closed. Thanks!

ianbotsf (approved):

PR #761, opened 2022-12-12
refactor(rt): make signing an explicit execution phase

aajtodd:
## Issue \#

https://github.com/awslabs/smithy-kotlin/issues/122

## Description of changes

This is part of refactoring for interceptors.

Refactors signing middleware to be an explicit phase in executing a request rather than just registered off the `finalize` phase of `SdkOperationExecution`. In a future PR I will refactor `SdkOperationExecution` to remove `finalize` all together (since it's primary purpose was mainly for signing). 

* Introduces a new generic `HttpSigner` interface that can be used by any authentication scheme (e.g. sigv4, bearer token, anonymous, etc)
* Refactors `AwsSigningMiddleware` to implement `HttpSigner` rather than `ModifyRequestMiddleware`
* Add `signer` property to `SdkHttpOperation` which codegen explicitly sets instead of installing a middleware
    * NOTE: As the fixme comment in the PR indicates this is temporary. The identity/auth changes that will be coming will remove registration of signing from operation directly

-----

Questions

1. I've left some of the package names as is (e.g. `middleware`). I'm not sure what we want to name these now. e.g. `AwsHttpSigner` is currently in the `middleware` subpackage (because previously it was `AwsSigningMiddleware`). Technically the consumer of these types is a middleware and they aren't meant for consumption except by the runtime. 
    1. Leave as is
    2. Move to an `http` subpackage of signing 
    3. Move to the `internal` package of signing


lucix-aws (approved):

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/auth/HttpSigner.kt
```diff
@@ -0,0 +1,30 @@
...
+         */
+        public val NONE: HttpSigner = object : HttpSigner {
+            override suspend fun sign(context: ExecutionContext, request: HttpRequestBuilder) { }
+        }
+    }
+
+    /**
+     * Sign the provided HTTP request (e.g. add AWS SigV4 headers, Bearer token header, etc)
+     */
+    public suspend fun sign(context: ExecutionContext, request: HttpRequestBuilder)
```
ianbotsf, line 29: **Question**: We generally have to make a copy of a request before signing it since signing happens _per-attempt_ and thus we can't modify the original. Since we're adding this interface and revising `AwsSigningMiddleware` I have to wonder: does the copying belong in the `HttpSigner` interface?

```interface
interface HttpSigner {
    // May return a signed copy or, in rare cases, the request arg unmodified (a no-op signature)
    suspend fun sign(context: ExecutionContext, request: HttpRequest): HttpRequest
}
```

ianbotsf (commented):
> Questions
> 
> 1. I've left some of the package names as is (e.g. `middleware`). I'm not sure what we want to name these now. e.g. `AwsHttpSigner` is currently in the `middleware` subpackage (because previously it was `AwsSigningMiddleware`). Technically the consumer of these types is a middleware and they aren't meant for consumption except by the runtime. 
>     1. Leave as is
>     2. Move to an `http` subpackage of signing 
>     3. Move to the `internal` package of signing

I'd say move it to the base package of **aws-signing-common**. The things in the **internal** package are implementation details of the `AwsSigner` (i.e., truly `internal`) while this is an implementation of an interface defined in another project (i.e., `@InteralAPI public`).

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/auth/HttpSigner.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import aws.smithy.kotlin.runtime.client.ExecutionContext
+import aws.smithy.kotlin.runtime.http.request.HttpRequestBuilder
+import aws.smithy.kotlin.runtime.util.InternalApi
+
+/**
+ * Represents a component capable of signing an HTTP request
+ */
+@InternalApi
+public interface HttpSigner {
```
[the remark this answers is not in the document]
aajtodd: `AwsSigner` is tied to AWS APIs (sigv4). `HttpSigner` would be usable for [non-AWS auth schemes ](https://smithy.io/2.0/spec/authentication-traits.html)(e.g. bearer token, API key, etc)

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsHttpSigner.kt
```diff
@@ -11,26 +11,27 @@ import aws.smithy.kotlin.runtime.auth.awssigning.internal.isEligibleForAwsChunke
...
+import aws.smithy.kotlin.runtime.http.auth.HttpSigner
 import aws.smithy.kotlin.runtime.http.request.HttpRequest
 import aws.smithy.kotlin.runtime.http.request.HttpRequestBuilder
 import aws.smithy.kotlin.runtime.util.InternalApi
 import aws.smithy.kotlin.runtime.util.get
 import kotlin.time.Duration
 
 /**
- * HTTP request pipeline middleware that signs outgoing requests
+ * AWS SigV4 [HttpSigner] that signs outgoing requests using the given [config]
```
lauzadis, line 24: Should we note that this also supports AWS SigV4a if using the CRT signer?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/auth/HttpSigner.kt
```diff
@@ -0,0 +1,30 @@
...
+    public companion object {
+        /**
+         * A no-op signer that does nothing with the request
+         */
+        public val NONE: HttpSigner = object : HttpSigner {
```
lauzadis, line 21: nit: `NONE` or `None`? would `NoOp` be a clearer name?

if this is just temporary, please ignore

lauzadis (approved):

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/auth/HttpSigner.kt
aajtodd: renamed to `Anonymous` which aligns with the ref arch terminology

lauzadis (approved):

runtime/auth/aws-signing-tests/jvm/src/aws/smithy/kotlin/runtime/auth/awssigning/tests/SigningSuiteTestBaseJVM.kt
```diff
@@ -7,7 +7,7 @@ package aws.smithy.kotlin.runtime.auth.awssigning.tests
 import aws.smithy.kotlin.runtime.auth.awscredentials.Credentials
 import aws.smithy.kotlin.runtime.auth.awscredentials.CredentialsProvider
 import aws.smithy.kotlin.runtime.auth.awssigning.*
-import aws.smithy.kotlin.runtime.auth.awssigning.middleware.AwsHttpSigner
+import aws.smithy.kotlin.runtime.auth.awssigning.AwsHttpSigner
```
ianbotsf: **Nit**: Unnecessary named import after star import.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/auth/HttpSigner.kt
```diff
@@ -18,7 +18,7 @@ public interface HttpSigner {
         /**
          * A no-op signer that does nothing with the request
          */
-        public val NONE: HttpSigner = object : HttpSigner {
+        public val Anonymous: HttpSigner = object : HttpSigner {
             override suspend fun sign(context: ExecutionContext, request: HttpRequestBuilder) { }
         }
```
ianbotsf: **Suggestion**: Consider updating the KDoc as well here.

runtime/auth/aws-signing-tests/common/src/aws/smithy/kotlin/runtime/auth/awssigning/tests/BasicSigningTestBase.kt
```diff
@@ -190,4 +189,42 @@ public abstract class BasicSigningTestBase : HasSigner {
         val finalChunkResult = signer.signChunk(ByteArray(0), prevSignature, chunkedSigningConfig)
         assertEquals(EXPECTED_FINAL_CHUNK_SIGNATURE, finalChunkResult.signature.decodeToString())
     }
+
+    @Test
+    public fun testSigningCopiesInput(): TestResult = runTest {
+        // sanity test the signer doesn't mutate the input and instead copies to a new request
```
ianbotsf, line 195: **Comment**: Very nice.

ianbotsf (approved):

runtime/auth/aws-signing-tests/jvm/src/aws/smithy/kotlin/runtime/auth/awssigning/tests/SigningSuiteTestBaseJVM.kt
aajtodd: surprised ktlint didn't catch that

PR #762, opened 2022-12-13
refator: make retries explicit and default behavior of every HTTP operation

aajtodd:
## Issue \#

https://github.com/awslabs/smithy-kotlin/issues/122

## Description of changes

* **refactor**: Renamed `Retry` -> `RetryMiddleware` and made `internal` and no longer `open`
* **refactor**: Refactored `SdkOperationExecution` to include retries by default. Codegen no longer registers a retry middleware, instead it configures the retry strategy and policy which is then consumed by the retry middleware that is configured for you.
    * A positive side effect of this is that the retry middleware doesn't require `RetryPolicy<Any?>`, it can use the generics of `SdkOperationExecution` to get a more specific type. Our existing retry policies (`StandardRetryPolicy` and `AwsStandardRetryPolicy`) implement `RetryPolicy<Any?>` which is API compatible with any generic supplied and can still be used as is
* **refactor**: Removed the `finalize` phase from `SdkOperationExecution`. This phase primarily existed for signing and retries (although retries actually existed in the `mutate` phase oddly enough, not sure if that was correct 🤷 ). This phase was only used in a handful of customizations/middleware and all have been updated to use different/new phases.
* **refactor**: Introduced `onEachAttempt` phase to `SdkOperationExecution` which acts like `mutate` but will run per/attempt. 
* **(unrelated fix)**: There was an issue with CRT signer found where the test would not exit. This was because the signer implementation picked chunked signing but still created a signable body. For channel bodies this sets up a background watchdog coroutine which prevented the test from exiting. I went ahead and fixed this because I needed to verify all tests pass with retries always being present by default now. 
    * The test had a newly introduced `buildOperationWithChannel` function which was not actually respecting the behavior of replayable streams. 


runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/operation/SdkOperationExecution.kt
```diff
@@ -29,39 +34,104 @@ import aws.smithy.kotlin.runtime.io.middleware.decorate as decorateHandler
...
+ * until the actual HTTP client engine call. The response then is returned up the call stack and traverses the phases
+ * in reverse.
+ *
+ * Phases enclosed in brackets `<>` are implicit phases that are always present and cannot be intercepted directly
+ * (only one is allowed to exist). These are usually configured directly when the operation is built.
+ *
+ * ### Default Behaviors
+ *
+ * By default, every operation is:
+ * * Retried by default using the configured [retryStrategy] and [retryPolicy].
```
ianbotsf, line 79: **Question**: Redundant "by default" in the second line?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/operation/SdkOperationExecution.kt
```diff
@@ -29,39 +34,104 @@ import aws.smithy.kotlin.runtime.io.middleware.decorate as decorateHandler
...
+
+    /**
+     * The retry strategy to use. Defaults to [StandardRetryStrategy]
+     */
+    public var retryStrategy: RetryStrategy = StandardRetryStrategy()
+
+    /**
+     * The retry policy to use. Defaults to [StandardRetryPolicy]
+     */
+    public var retryPolicy: RetryPolicy<Response> = StandardRetryPolicy.Default
```
ianbotsf, line 134: **Question**: Do these deserve `// FIXME` comments as well? Or are `public var`s the likely-permanent way middleware/interceptors will modify retry config?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/operation/SdkOperationExecution.kt
```diff
@@ -29,39 +34,104 @@ import aws.smithy.kotlin.runtime.io.middleware.decorate as decorateHandler
...
+    public var signer: HttpSigner = HttpSigner.Anonymous
+
+    /**
+     * The retry strategy to use. Defaults to [StandardRetryStrategy]
+     */
+    public var retryStrategy: RetryStrategy = StandardRetryStrategy()
+
+    /**
+     * The retry policy to use. Defaults to [StandardRetryPolicy]
+     */
```
ianbotsf, line 133: **Nit**: `StandardRetryPolicy` → `StandardRetryPolicy.Default`

runtime/protocol/http/common/test/aws/smithy/kotlin/runtime/http/operation/SdkOperationExecutionTest.kt
```diff
@@ -0,0 +1,103 @@
...
+            assertTrue(req.subject.headers.contains("mutate-header"))
+
+            assertFalse(req.subject.headers.contains("receive-header"))
+            req.subject.headers.append("receive-header", "receive")
+            next.call(req)
+        }
+
+        val client = sdkHttpClient(mockEngine)
+        op.roundTrip(client, Unit)
+        val expectedOrder = listOf("initialize", "mutate", "attempt", "receive", "attempt", "receive")
```
ianbotsf, line 100: **Question**: Is it worth adding verification of the reverse-order call as well?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/policy/RetryPolicy.kt
```diff
@@ -6,13 +6,15 @@
...
+ * A policy that evaluates a [Result] from a retry attempt and indicates the action a
+ * [aws.smithy.kotlin.runtime.retries.RetryStrategy] should take next.
  */
 public interface RetryPolicy<in R> {
     /**
      * Evaluate the given retry attempt.
      * @param result The [Result] of the retry attempt.
-     * @return A [RetryDirective] indicating what action the [RetryStrategy] should take next.
+     * @return A [RetryDirective] indicating what action the [aws.smithy.kotlin.runtime.retries.RetryStrategy] should
+     * take next.
```
ianbotsf, line 17: **Style**: Import `RetryStrategy` instead of fully qualifying it.

ianbotsf (approved):

runtime/protocol/http/common/test/aws/smithy/kotlin/runtime/http/operation/SdkOperationExecutionTest.kt
aajtodd: We could. I thought about it and then decided not to since it's a call stack...

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/operation/SdkOperationExecution.kt
aajtodd: I had planned on these being more permanent. Middleware and interceptors can't really modify these. Plugins likely will be able to but not sure what that looks like. 

Open to discussion though.

lucix-aws (approved):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/policy/RetryPolicy.kt
aajtodd: `RetryStrategy` isn't used by the code anywhere though so fully qualifying it makes sense here (no unused imports)

PR #766, opened 2022-12-15
fix(rt): CRC32C incorrect result on inputs longer than 7 bytes

lauzadis:
During flexible checksums testing, a bug was discovered in the CRC32c hash function when trying to hash more than 7 bytes at a time. The elements were being cast `toByte()`, truncating information and leading to incorrect output. 

## Issue \#

N/A

## Description of changes

This change ensures that CRC32C outputs the correct value when `update()` is called with a `length` greater than 7.

### DIFF
--- a/runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/Crc32c.kt
+++ b/runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/Crc32c.kt
@@ -101,7 +101,7 @@ private class CRC32CImpl {
         crc = localCrc
     }
 
-    fun getValue(): Int = crc.inv() and 0xffffffff.toInt()
+    fun getValue(): Int = crc.inv()
 
     fun update(b: Int) {
         crc = crc ushr 8 xor T[(T8_0_start + ((crc xor b) and 0xff))].toInt()
--- a/runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/Crc32c.kt
+++ b/runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/Crc32c.kt
@@ -6,7 +6,6 @@ package aws.smithy.kotlin.runtime.hashing
 
 import aws.smithy.kotlin.runtime.util.InternalApi
 import kotlin.experimental.and
-import kotlin.experimental.xor
 
 /**
  * Compute the CRC32C hash of the current [ByteArray]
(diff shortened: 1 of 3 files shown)


runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/Crc32c.kt
```diff
@@ -64,48 +63,48 @@ public class Crc32c : Crc32Base() {
...
+            localCrc = (localCrc ushr 8 xor T[T8_0_start + (localCrc xor b[off].toInt() and 0xff)].toInt())
             off += 1
         }
 
-        // Publish localCrc out to object
         crc = localCrc
     }
 
-    fun getValue(): Long = crc.inv() and 0xffffffffL
+    fun getValue(): Int = crc.inv() and 0xffffffff.toInt()
```
ianbotsf, line 104: **Question**: Is `and 0xffffffff.toInt()` really necessary? That's ANDing with 32 positive bits which, for a 32-bit integer, seems redundant.

runtime/hashing/common/test/aws/smithy/kotlin/runtime/hashing/Crc32cTest.kt
```diff
@@ -60,4 +60,24 @@ class Crc32cTest {
...
+
+    @Test
+    fun testLargeInput() {
+        val crc = Crc32c()
+        val input = ByteArray(1024) { 0 }
+        crc.update(input, 0, input.size)
+
+        val bytes = crc.digest()
+        assertEquals("7q7efA==", bytes.encodeBase64String())
+    }
```
ianbotsf, line 82: **Question**: How were these new test values derived? Is there a wider suite of known-correct test cases from which we could draw?

runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/Crc32c.kt
lauzadis: You're right, it's not necessary since this is not a Long anymore

aajtodd (approved):

runtime/hashing/common/test/aws/smithy/kotlin/runtime/hashing/Crc32cTest.kt
lauzadis: These were derived with [the crc32c Python package]( https://pypi.org/project/crc32c/). I took a look and it seems there's no widely-accepted suite of test cases. I found some in [RFC 3720](https://www.rfc-editor.org/rfc/rfc3720#appendix-B.4) but overall it's not focused on CRC, just in the appendix.

ianbotsf (approved):

PR #770, opened 2022-12-19
misc: add design doc for per-op config

lucix-aws:

### DIFF
--- a/docs/design/per-op-config.md
+++ b/docs/design/per-op-config.md
@@ -1,19 +1,23 @@
-# per-op config
+# Per-operation configuration
+
 * **Type**: Design
-* **Author(s)**: AWS SDK for Kotlin team
+* **Author(s)**: Luc Talatinian, Ian Smith Botsford, Aaron Todd, Matas Lauzadis
 
-# abstract
-The smithy-kotlin project generates clients with the ability to perform operation- or block-scoped config overrides.
-This functionality is exposed by supporting a `withConfig` method on clients, which takes a config builder and returns a
-new "subclient" which performs operations using the modified config. We also outline several proposed design
-alternatives that were rejected.
+# Abstract
 
-# design
-All code samples in this document assume the following basic client interface. The client has an arbitrary number of operations,
-the declarations of which are omitted.
-This particular client has two config
-fields: a logging mode and an HTTP client to be used to make requests. Note that HTTP clients (`HttpClientEngine`) in
-the smithy-kotlin runtime implement `Closeable`.
+Service clients are configured at creation time and use the same configuration for every operation for the lifetime of
+the client. It is desirable to enable overriding a subset of client configuration for one or more operations (e.g. using
+different HTTP settings, logging mode, etc.). This document covers the design and implementation of
+the API for overriding service client configuration on a per-operation basis, and outlines several rejected design
+alternatives.
+
+# Design
+
+All code samples in this document assume the following basic client interface. The client has an arbitrary number of
+operations, the declarations of which are omitted.
+
+This particular client has two config fields: a logging mode and an HTTP client to be used to make requests. Note that
+HTTP clients (`HttpClientEngine`) in the smithy-kotlin runtime implement `Closeable`.
 
 ```kotlin
 interface ServiceClient : SdkClient {
@@ -38,23 +42,21 @@ interface ServiceClient : SdkClient {
 class DefaultServiceClient(config: ServiceClient.Config) : ServiceClient { /* ... */ }
 ```
 
-Client interfaces receive a new `withConfig` extension, which takes a config builder and returns a "child" client with
+Client interfaces have a `withConfig` extension, which takes a config builder and returns a new client with
 any fields specified in the builder applied to its config.
 
 ```kotlin
 fun ServiceClient.withConfig(block: ServiceClient.Config.Builder.() -> Unit): ServiceClient { /* ... */ }
 ```
 
-In practice, to perform one or more operations with a modified config in-place, it is most effective to make your call
-within a `use()` block on the returned client to have the `close()` handled for you. It's still possible to pass the
-subclient to an external background routine, and it may even outlive the parent client (see below).
+The client returned from `withConfig` exists independently of the original. It has its own lifetime and should be closed
+when no longer needed. Any runtime-managed closeable resources will be kept alive until all clients using them have been
+closed.
 
-The `withConfig` method is _more efficient_ than creating multiple clients if you are using runtime-managed `Closeable`
-resources such as the default HTTP engine. Users of these defaults are tracked across clients as they are created, and
-the underlying `Closeable` is only fully closed when the last remaining client using it is closed.
-
-For user-owned `Closeable`s, nothing changes - clients do not handle closing these for you today, and any child clients
-created by `withConfig` won't either.
+The `withConfig` method is _more efficient_ than creating multiple clients from scratch (either directly from your
+own `Config` or via `fromEnvironment`) if you are using runtime-managed `Closeable` resources such as the default HTTP
+engine. Instances which use these defaults are tracked across clients as they are created, and the
+underlying `Closeable` is only fully closed when the last remaining client using it is closed.
 
 Sample usage:
 ```kotlin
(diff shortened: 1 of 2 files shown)


docs/design/per-op-config.md
```diff
@@ -0,0 +1,179 @@
...
+```kotlin
+fun ServiceClient.withConfig(block: ServiceClient.Config.Builder.() -> Unit): ServiceClient { /* ... */ }
+```
+
+In practice, to perform one or more operations with a modified config in-place, it is most effective to make your call
+within a `use()` block on the returned client to have the `close()` handled for you. It's still possible to pass the
+subclient to an external background routine, and it may even outlive the parent client (see below).
+
+The `withConfig` method is _more efficient_ than creating multiple clients if you are using runtime-managed `Closeable`
+resources such as the default HTTP engine. Users of these defaults are tracked across clients as they are created, and
```
ianbotsf, line 53: **Comment**: `withConfig` _does_ create multiple clients so perhaps it would be better to contrast with "creating clients from scratch using new config or `fromEnvironment`"?

docs/design/per-op-config.md
```diff
@@ -0,0 +1,179 @@
...
+the underlying `Closeable` is only fully closed when the last remaining client using it is closed.
+
+For user-owned `Closeable`s, nothing changes - clients do not handle closing these for you today, and any child clients
+created by `withConfig` won't either.
```
ianbotsf, line 57: **Comment**: These sentences on resource lifetimes don't feel clear to me.

docs/design/per-op-config.md
```diff
@@ -0,0 +1,179 @@
...
+        })
+    }
+    
+    // application continues...
+}
+```
+
+Some observations about the above sample:
+* Since no explicit HTTP engine is passed to construction of the root client, the SDK creates its own "managed" default.
+* The first use of `withConfig` and `use` demonstrates how to make basic one-off calls. The defaulted HTTP engine is
```
ianbotsf, line 88: **Nit**: defaulted → default

docs/design/per-op-config.md
```diff
@@ -0,0 +1,179 @@
...
+  inherited by the child. When the service client is `closed()` as part of the `use { }` call, the runtime recognizes
```
ianbotsf, line 89: **Nit**: `closed()` isn't a real method name...this should just be "closed" in normal text (without quotes).

docs/design/per-op-config.md
```diff
@@ -0,0 +1,179 @@
...
+  that others are still holding onto its HTTP engine and so it isn't fully closed.
+* The client passed to `doBackgroundRoutine` inherits the default HTTP engine as well. Both the parent and subclient can
+  be closed at any time in either order - the managed HTTP engine will only be reaped when both clients using it are
+  closed.
```
ianbotsf, line 93: **Nit**: reaped → closed

docs/design/per-op-config.md
```diff
@@ -0,0 +1,179 @@
...
+
+Rejected because:
+* Adding our own named field `withConfig` to a structure generated based on a model we don't own introduces a collision
+  risk. There's nothing preventing a model author from adding their own "withConfig" field to a shape.
+* Affords less flexibility. A caller may with to structure a series of operations as a function which accepts an
+  arbitrarily configured client - this route removes the possibility for that separation, forcing the caller to hardcode
+  config overrides in specific method calls.
+* Prevents us from supporting "shape-only" (generate only client interface and input/output shapes, no implementation)
+  codegen in the future, since the `withConfig` property requires additional mechanisms to be generated into the structs
+  in order to be retrieved by the implemented operation.
```
ianbotsf, line 141: **Nit**: "Prevents us from" → "Complicates". There are paths forward here that still allow shape-only codegen but they require complicating our generators.

docs/design/per-op-config.md
ianbotsf, line 53: **Comment**: "Users" generally means the people who write code which uses our SDK. It might be clearer to say "Instances which use".

docs/design/per-op-config.md
```diff
@@ -0,0 +1,179 @@
...
+
+Client interfaces receive a new `withConfig` extension, which takes a config builder and returns a "child" client with
+any fields specified in the builder applied to its config.
+
+```kotlin
+fun ServiceClient.withConfig(block: ServiceClient.Config.Builder.() -> Unit): ServiceClient { /* ... */ }
+```
+
+In practice, to perform one or more operations with a modified config in-place, it is most effective to make your call
+within a `use()` block on the returned client to have the `close()` handled for you. It's still possible to pass the
```
ianbotsf, line 49: **Style**: Avoid talking to "you", "yourself", or describing things as "yours" in design documents. The audience includes more than end users.

ianbotsf (changes requested):

docs/design/per-op-config.md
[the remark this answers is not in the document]
aajtodd: Agreed. Instead of in " in practice" I would just spell out the semantics here.

e.g. something like

> The client returned from `withConfig` is independent from the original client. It has it's own lifetime and should be closed when no longer needed. Any closeable resources will be kept alive until both clients are finished with them.

docs/design/per-op-config.md
aajtodd, line 89: suggestion: "inherited" -> "shared"? (also remove any notion of child relationships)

docs/design/per-op-config.md
```diff
@@ -0,0 +1,179 @@
...
+            withConfig {
+                logMode = SdkLogMode.LogRequestWithBody
+            }
+        }
+    }
+}
+```
+
+Rejected because:
+* Adding our own named field `withConfig` to a structure generated based on a model we don't own introduces a collision
```
aajtodd, line 134: nit: Remove use of "we"

e.g.

> Adding custom un-modeled fields to (Smithy) structures introduces the possibility of collision with future model updates

docs/design/per-op-config.md
```diff
@@ -0,0 +1,179 @@
...
+  risk. There's nothing preventing a model author from adding their own "withConfig" field to a shape.
+* Affords less flexibility. A caller may with to structure a series of operations as a function which accepts an
```
aajtodd, line 136: fix: "A caller may with to" -> "A caller may want to" (I'm assuming that's what you meant anyway)

docs/design/per-op-config.md
```diff
@@ -0,0 +1,179 @@
...
+  arbitrarily configured client - this route removes the possibility for that separation, forcing the caller to hardcode
```
aajtodd, line 137: suggestion: "forcing the caller to specify config overrides in every method call"

docs/design/per-op-config.md
aajtodd: I agree with your statement but I think it's even more than that though.

Generating request or response structures with additional client level configuration makes it almost impossible to re-use "shape only" codegen.

e.g. Let's say we have wanted to support shape only codegen and generate just the request/response structures (no serde, no client interface). Theoretically it should be possible then to depend on that package and use those types as the inputs to a service client operation. Tying them together this way would make them incompatible though.

docs/design/per-op-config.md
```diff
@@ -0,0 +1,179 @@
...
+  
+    client.withConfig({
+        logMode = SdkLogMode.LogRequestWithBody + SdkLogMode.LogResponseWithBody
+    }) {
+        client.operationA { /* ... */ }
+    }
+}
+```
+
+The final accepted design ultimately evolved from this. We argue that there's no reason to restrict the user to making
```
aajtodd, line 170: fix: Remove use of "we" in this document

e.g. Something like:

> This alternative was rejected in-favor of the proposed design which doesn't force a limited scope for configuration override. The same effect can be achieved with the `.use{}` extension. The use of two lambda arguments was also deemed non-idiomatic.

aajtodd (changes requested):

lauzadis (commented):
I agree with all the comments left so far and had trouble finding anything else to comment on. Specifically I agree removing references to "we" (I too learned this in the initial flexible checksums design) and addressing the general formatting fixes will help a lot. 

 Looking forward to reviewing the second revision!

aajtodd (approved):

lauzadis (approved):
Looks good!

docs/design/per-op-config.md
```diff
@@ -38,23 +42,21 @@ interface ServiceClient : SdkClient {
...
 ```kotlin
 fun ServiceClient.withConfig(block: ServiceClient.Config.Builder.() -> Unit): ServiceClient { /* ... */ }
 ```
 
-In practice, to perform one or more operations with a modified config in-place, it is most effective to make your call
-within a `use()` block on the returned client to have the `close()` handled for you. It's still possible to pass the
-subclient to an external background routine, and it may even outlive the parent client (see below).
+The client returned from `withConfig` exists independently of the original. It has its own lifetime and should be closed
+when no longer needed. Any runtime-managed closeable resources will be kept alive until all clients using them have been
+closed.
```
ianbotsf, line 54: **Comment**: Very nice.

docs/design/per-op-config.md
```diff
@@ -38,23 +42,21 @@ interface ServiceClient : SdkClient {
...
 
-The `withConfig` method is _more efficient_ than creating multiple clients if you are using runtime-managed `Closeable`
-resources such as the default HTTP engine. Users of these defaults are tracked across clients as they are created, and
-the underlying `Closeable` is only fully closed when the last remaining client using it is closed.
-
-For user-owned `Closeable`s, nothing changes - clients do not handle closing these for you today, and any child clients
-created by `withConfig` won't either.
+The `withConfig` method is _more efficient_ than creating multiple clients from scratch (either directly from your
+own `Config` or via `fromEnvironment`) if you are using runtime-managed `Closeable` resources such as the default HTTP
```
ianbotsf, line 57: **Nit**: Avoid "your". Consider "directly from explicit `Config` or via `fromEnvironment`".

docs/design/per-op-config.md
```diff
@@ -38,23 +42,21 @@ interface ServiceClient : SdkClient {
...
+engine. Instances which use these defaults are tracked across clients as they are created, and the
```
ianbotsf, line 58: **Nit**: Avoid "you". Consider "when using runtime-managed".

docs/design/per-op-config.md
```diff
@@ -85,21 +87,71 @@ suspend fun main() {
 
 Some observations about the above sample:
 * Since no explicit HTTP engine is passed to construction of the root client, the SDK creates its own "managed" default.
```
ianbotsf, line 89: **Nit**: Avoid "root". Consider "initial client".

docs/design/per-op-config.md
```diff
@@ -85,21 +87,71 @@ suspend fun main() {
...
-* The first use of `withConfig` and `use` demonstrates how to make basic one-off calls. The defaulted HTTP engine is
-  inherited by the child. When the service client is `closed()` as part of the `use { }` call, the runtime recognizes
+* The first use of `withConfig` and `use` demonstrates how to make basic one-off calls. The default HTTP engine is
+  shared with the new client. When the service client is closed as part of the `use { }` call, the runtime recognizes
   that others are still holding onto its HTTP engine and so it isn't fully closed.
-* The client passed to `doBackgroundRoutine` inherits the default HTTP engine as well. Both the parent and subclient can
-  be closed at any time in either order - the managed HTTP engine will only be reaped when both clients using it are
-  closed.
+* The client passed to `doBackgroundRoutine` shares the default HTTP engine as well. Both clients be closed at any time
+  in either order - the managed HTTP engine will only be closed when both clients using it are closed.
```
ianbotsf, line 94: **Style**: Separate independent clauses with an em dash (—) instead of a hyphen (-). No spaces surrounding the punctuation are necessary.

docs/design/per-op-config.md
```diff
@@ -131,18 +183,20 @@ suspend fun main() {
...
 
 Rejected because:
-* Adding our own named field `withConfig` to a structure generated based on a model we don't own introduces a collision
-  risk. There's nothing preventing a model author from adding their own "withConfig" field to a shape.
-* Affords less flexibility. A caller may with to structure a series of operations as a function which accepts an
+
+* Adding custom un-modeled fields to (Smithy) structures introduces the possibility of collision with future model
+  updates.
+* Affords less flexibility. A caller may wish to structure a series of operations as a function which accepts an
   arbitrarily configured client - this route removes the possibility for that separation, forcing the caller to hardcode
```
ianbotsf, line 190: **Style**: Em dash.

docs/design/per-op-config.md
```diff
@@ -85,21 +87,71 @@ suspend fun main() {
...
+## Alt1: optional config on operation calls
+
+Operation methods have the ability to directly override config. There are a number of ways to implement this alternative:
+```kotlin
+interface ServiceClient : SdkClient {
+    // alt1.1: defaulted param
+    suspend fun operationA(input: OperationARequest, config: Config? = null): OperationAResponse
+
+    // alt2.1: Trailing lambda builder
+    suspend fun operationB(input: OperationBRequest, block: Config.Builder.() -> Unit = {}): OperationBResponse
```
ianbotsf, line 109: **Nit**: `operationB` should be "alt1.2", not "alt2.1".

docs/design/per-op-config.md
```diff
@@ -85,21 +87,71 @@ suspend fun main() {
...
+    val client = ServiceClient { /* ... */ }
+
+    val requestA = OperationARequest { /* ... */ }
+    client.operationA(requestA)
+    client.operationA(requestA, client.config.copy {
+        logMode = SdkLogMode.LogRequestWithBody
+    })
+
+    val requestB = OperationBRequest { /* ... */ }
+    client.operationB(operationB)
```
ianbotsf, line 132: **Nit**: `client.operationB(requestB)`

docs/design/per-op-config.md
```diff
@@ -85,21 +87,71 @@ suspend fun main() {
...
+        logMode = SdkLogMode.LogResponseWithBody
+    }
+}
+```
+
+Rejected because:
+
+* Both introduce operation signature bloat.
+* Neither really play into to the adopted builder-based overloads that the SDK supports. Mixed usage of the two may
+  adversely impact an author's ability to produce readable and maintainable code.
```
ianbotsf, line 146: **Style**: "really play into" is confusing here. Consider "mix well with".

docs/design/per-op-config.md
```diff
@@ -167,13 +221,14 @@ suspend fun main() {
 }
 ```
 
-The final accepted design ultimately evolved from this. We argue that there's no reason to restrict the user to making
-calls within this forced scope - we can allow the caller to save and pass around the returned client as long as we're
-responsible with SDK-managed closeable resources that are shared throughout.
+This alternative was rejected in favor of the proposed design since it accomplishes the same thing while expanding the
+scope of use. the SDK can allow the caller to save and pass around the returned client as long as the runtime is
+responsible with managed closeable resources that are shared throughout. The use of two lambda arguments was also deemed
```
ianbotsf, line 226: **Nit**: "the SDK" → "The SDK".

docs/design/per-op-config.md
```diff
@@ -167,13 +221,14 @@ suspend fun main() {
...
+non-idiomatic.
 
 Additionally, the pattern enforced by this API (client only lives within `useBlock`) can still be expressed in the
 accepted design simply by making the `use { }` call on the client yourself, as demonstrated in its example usage.
```
ianbotsf, line 230: **Nit**: Avoid "yourself". Consider "itself" (changes the meaning of the phrase slightly but accurately).

ianbotsf (approved):
Looking pretty good. Down to minor nits and tweaks.

PR #772, opened 2022-12-29
feat: implement flexible checksums customization

lauzadis:
## Description of changes

### DIFF
--- a/runtime/auth/aws-signing-tests/common/src/aws/smithy/kotlin/runtime/auth/awssigning/tests/AwsChunkedTestBase.kt
+++ b/runtime/auth/aws-signing-tests/common/src/aws/smithy/kotlin/runtime/auth/awssigning/tests/AwsChunkedTestBase.kt
@@ -7,9 +7,11 @@ package aws.smithy.kotlin.runtime.auth.awssigning.tests
 
 import aws.smithy.kotlin.runtime.auth.awssigning.*
 import aws.smithy.kotlin.runtime.auth.awssigning.internal.CHUNK_SIZE_BYTES
-import aws.smithy.kotlin.runtime.http.Headers
+import aws.smithy.kotlin.runtime.http.LazyHeaders
+import aws.smithy.kotlin.runtime.http.LazyHeaders.Companion.toHeaders
 import aws.smithy.kotlin.runtime.io.*
 import aws.smithy.kotlin.runtime.time.Instant
+import aws.smithy.kotlin.runtime.util.asyncLazy
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.test.TestResult
 import kotlinx.coroutines.test.runTest
@@ -41,7 +43,7 @@ fun interface AwsChunkedReaderFactory {
         signer: AwsSigner,
         signingConfig: AwsSigningConfig,
         previousSignature: ByteArray,
-        trailingHeaders: Headers,
+        trailingHeaders: LazyHeaders,
     ): AwsChunkedTestReader
 }
 
@@ -50,7 +52,7 @@ fun AwsChunkedReaderFactory.create(
     signer: AwsSigner,
     signingConfig: AwsSigningConfig,
     previousSignature: ByteArray,
-): AwsChunkedTestReader = create(data, signer, signingConfig, previousSignature, Headers.Empty)
+): AwsChunkedTestReader = create(data, signer, signingConfig, previousSignature, LazyHeaders.Empty)
 
 @OptIn(ExperimentalCoroutinesApi::class)
 abstract class AwsChunkedTestBase(
@@ -111,7 +113,7 @@ abstract class AwsChunkedTestBase(
      * Calculates the `aws-chunked` encoded trailing header length
      * Used to calculate how many bytes should be read for all the trailing headers to be consumed
      */
-    fun getTrailingHeadersLength(trailingHeaders: Headers) = trailingHeaders.entries().map {
+    suspend fun getTrailingHeadersLength(trailingHeaders: LazyHeaders) = trailingHeaders.toHeaders().entries().map {
             entry ->
         buildString {
             append(entry.key)
@@ -365,9 +367,9 @@ abstract class AwsChunkedTestBase(
         val data = ByteArray(dataLengthBytes) { 0x7A.toByte() }
         var previousSignature: ByteArray = byteArrayOf()
 
-        val trailingHeaders = Headers {
-            append("x-amz-checksum-crc32", "AAAAAA==")
-            append("x-amz-arbitrary-header-with-value", "UMM")
+        val trailingHeaders = LazyHeaders {
+            append("x-amz-checksum-crc32", asyncLazy { "AAAAAA==" })
+            append("x-amz-arbitrary-header-with-value", asyncLazy { "UMM" })
         }
 
         val trailingHeadersLength = getTrailingHeadersLength(trailingHeaders)
@@ -404,7 +406,7 @@ abstract class AwsChunkedTestBase(
         assertEquals(chunkSizes[0], CHUNK_SIZE_BYTES)
         assertEquals(chunkSizes[1], 0)
 
-        val expectedTrailerSignature = signer.signChunkTrailer(trailingHeaders, previousSignature, testTrailingHeadersSigningConfig).signature
+        val expectedTrailerSignature = signer.signChunkTrailer(trailingHeaders.toHeaders(), previousSignature, testTrailingHeadersSigningConfig).signature
         val trailerSignature = getChunkTrailerSignature(bytesAsString)
         assertEquals(expectedTrailerSignature.decodeToString(), trailerSignature)
     }
--- a/runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashFunction.kt
+++ b/runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashFunction.kt
@@ -57,3 +57,18 @@ public typealias HashSupplier = () -> HashFunction
  */
 @InternalApi
 public fun ByteArray.hash(hashSupplier: HashSupplier): ByteArray = hash(hashSupplier(), this)
+
+@InternalApi
+/**
+ * Return the [HashFunction] which is represented by the current [String], or null if none match.
+ */
+public fun String.toHashFunction(): HashFunction? {
+    return when (this.lowercase()) {
+        "crc32" -> Crc32()
+        "crc32c" -> Crc32c()
+        "sha1" -> Sha1()
+        "sha256" -> Sha256()
+        "md5" -> Md5()
+        else -> null
+    }
+}
--- a/runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashFunction.kt
+++ b/runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashFunction.kt
@@ -60,7 +60,7 @@ public fun ByteArray.hash(hashSupplier: HashSupplier): ByteArray = hash(hashSupp
 
 @InternalApi
 /**
- * Return the [HashFunction] which is represented by the current [String], or null if none match.
+ * Return the [HashFunction] which is represented by this string, or null if none match.
  */
 public fun String.toHashFunction(): HashFunction? {
     return when (this.lowercase()) {
--- a/runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashFunction.kt
+++ b/runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashFunction.kt
@@ -62,13 +62,11 @@ public fun ByteArray.hash(hashSupplier: HashSupplier): ByteArray = hash(hashSupp
 /**
  * Return the [HashFunction] which is represented by this string, or null if none match.
  */
-public fun String.toHashFunction(): HashFunction {
-    return when (this.lowercase()) {
-        "crc32" -> Crc32()
-        "crc32c" -> Crc32c()
-        "sha1" -> Sha1()
-        "sha256" -> Sha256()
-        "md5" -> Md5()
-        else -> throw RuntimeException("$this is not a supported hash function")
-    }
+public fun String.toHashFunction(): HashFunction? = when (this.lowercase()) {
+    "crc32" -> Crc32()
+    "crc32c" -> Crc32c()
+    "sha1" -> Sha1()
+    "sha256" -> Sha256()
+    "md5" -> Md5()
+    else -> null
 }
--- a/runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashFunction.kt
+++ b/runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashFunction.kt
@@ -62,13 +62,13 @@ public fun ByteArray.hash(hashSupplier: HashSupplier): ByteArray = hash(hashSupp
 /**
  * Return the [HashFunction] which is represented by this string, or null if none match.
  */
-public fun String.toHashFunction(): HashFunction? {
+public fun String.toHashFunction(): HashFunction {
     return when (this.lowercase()) {
         "crc32" -> Crc32()
         "crc32c" -> Crc32c()
         "sha1" -> Sha1()
         "sha256" -> Sha256()
         "md5" -> Md5()
-        else -> null
+        else -> throw RuntimeException("$this is not a supported hash function")
     }
 }
--- a/runtime/io/common/src/aws/smithy/kotlin/runtime/io/HashingByteReadChannel.kt
+++ b/runtime/io/common/src/aws/smithy/kotlin/runtime/io/HashingByteReadChannel.kt
@@ -1,0 +1,36 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package aws.smithy.kotlin.runtime.io
+
+import aws.smithy.kotlin.runtime.hashing.HashFunction
+import aws.smithy.kotlin.runtime.io.internal.commonPeek
+import aws.smithy.kotlin.runtime.util.InternalApi
+
+/**
+ * A channel which hashes data as it is being read
+ * @param hash The [HashFunction] to hash data with
+ * @param chan the [SdkByteReadChannel] to hash
+ */
+@InternalApi
+public class HashingByteReadChannel(private val hash: HashFunction, private val chan: SdkByteReadChannel) : SdkByteReadChannel by chan {
+    public override suspend fun read(sink: SdkBuffer, limit: Long): Long {
+        val buffer = SdkBuffer()
+
+        val rc = chan.read(buffer, limit)
+        if (rc == -1L) { return rc } // got 0 bytes, no need to hash anything
+
+        val dataToHash = buffer.commonPeek().readByteArray()
+        hash.update(dataToHash)
+
+        return buffer.read(sink, rc)
+    }
+
+    /**
+     * Provides the digest as a ByteArray
+     * @return a ByteArray representing the contents of the hash
+     */
+    public fun digest(): ByteArray = hash.digest()
+}
--- a/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsRequestMiddleware.kt
+++ b/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsRequestMiddleware.kt
@@ -1,0 +1,158 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package aws.smithy.kotlin.runtime.http.middleware
+
+import aws.smithy.kotlin.runtime.ClientException
+import aws.smithy.kotlin.runtime.hashing.*
+import aws.smithy.kotlin.runtime.http.*
+import aws.smithy.kotlin.runtime.http.operation.*
+import aws.smithy.kotlin.runtime.http.request.HttpRequestBuilder
+import aws.smithy.kotlin.runtime.http.request.header
+import aws.smithy.kotlin.runtime.io.*
+import aws.smithy.kotlin.runtime.util.*
+import kotlin.coroutines.coroutineContext
+
+/**
+ * Mutate a request to enable flexible checksums.
+ *
+ * If the checksum will be sent as a header, calculate the checksum.
+ *
+ * Otherwise, if it will be sent as a trailing header, calculate the checksum as asynchronously as the body is streamed.
+ * In this case, a [LazyAsyncValue] will be added to the execution context which allows the trailing checksum to be sent
+ * after the entire body has been streamed.
+ *
+ * @param checksumAlgorithmName the name of the algorithm used to calculate the checksum
+ */
+@InternalApi
+public class FlexibleChecksumsRequestMiddleware(private val checksumAlgorithmName: String) : ModifyRequestMiddleware {
+    public companion object {
+        public val TrailingHeaders: AttributeKey<LazyHeadersBuilder> = AttributeKey("TrailingHeaders")
+    }
+
+    public override suspend fun modifyRequest(req: SdkHttpRequest): SdkHttpRequest {
+        val logger = coroutineContext.getLogger<FlexibleChecksumsRequestMiddleware>()
+
+        val headerName = "x-amz-checksum-${checksumAlgorithmName.lowercase()}"
+        logger.debug { "Resolved header name: $headerName" }
+
+        // remove all checksum headers except for $headerName
+        // this handles the case where a user inputs a precalculated checksum, but it doesn't match the input checksum algorithm
+        req.subject.headers.removeAllChecksumHeadersExcept(headerName)
+
+        val checksumAlgorithm = checksumAlgorithmName.toHashFunction() ?: throw ClientException("Could not parse checksum algorithm $checksumAlgorithmName")
+        logger.debug { "Resolved checksum algorithm: $checksumAlgorithm" }
+
+        if (!checksumAlgorithm.isSupported) {
+            throw ClientException("Checksum algorithm $checksumAlgorithmName is not supported for flexible checksums")
+        }
+
+        if (req.subject.body.isEligibleForAwsChunkedStreaming) {
+            logger.debug { "Sending checksum as a trailing header" }
+            req.subject.header("x-amz-trailer", headerName)
+
+            val lazyChecksum: LazyAsyncValue<String> = if (req.subject.headers[headerName] != null) {
+                logger.debug { "User supplied a checksum, skipping asynchronous calculation" }
+
+                val checksum = req.subject.headers[headerName]!!
+                req.subject.headers.remove(headerName) // remove the checksum header because it will be sent as a trailing header
+                asyncLazy { checksum }
+            } else {
+                logger.debug { "Calculating checksum asynchronously" }
+
+                req.subject.setHashingBody(checksumAlgorithm)
+                val body = req.subject.body
+                asyncLazy { body.checksum }
+            }
+
+            // add the lazy checksum to the execution context
+            val trailingHeadersBuilder: LazyHeadersBuilder = req.context.getOrNull(TrailingHeaders) ?: LazyHeadersBuilder()
+            trailingHeadersBuilder.append(headerName, lazyChecksum)
+            req.context[TrailingHeaders] = trailingHeadersBuilder
+        } else if (req.subject.headers[headerName] == null) {
+            logger.debug { "Calculating checksum" }
+
+            val checksum = req.subject.body.calculateChecksum(checksumAlgorithm)
+            req.subject.header(headerName, checksum)
+
+            logger.debug { "Calculated checksum: $checksum" }
+        }
+
+        return req
+    }
+
+    // FIXME this duplicates the logic from aws-signing-common, but can't import from there due to circular import.
+    private val HttpBody.isEligibleForAwsChunkedStreaming: Boolean
+        get() = (this is HttpBody.SourceContent || this is HttpBody.ChannelContent) && contentLength != null &&
+            (isOneShot || contentLength!! > 65536 * 16)
+
+    /**
+     * Wrap the HttpRequestBuilder's body with a hashing body, which will observe the data as it's being consumed and maintain
+     * an internal hash.
+     * @param hashFunction the hash function to apply to this body
+     * @throws ClientException when the HttpBody is not supported for hashing bodies
+     */
+    private fun HttpRequestBuilder.setHashingBody(hashFunction: HashFunction) {
+        body = when (body) {
+            is HttpBody.SourceContent ->
+                HashingSource(
+                    hashFunction,
+                    (body as HttpBody.SourceContent).readFrom(),
+                ).toHttpBody(body.contentLength)
+
+            is HttpBody.ChannelContent -> HashingByteReadChannel(
+                hashFunction,
+                (body as HttpBody.ChannelContent).readFrom(),
+            ).toHttpBody(body.contentLength)
+
+            else -> throw ClientException("HttpBody type is not supported")
+        }
+    }
+
+    /**
+     * @return if the [HashFunction] is supported by flexible checksums
+     */
+    private val HashFunction.isSupported: Boolean get() = when (this) {
+        is Crc32, is Crc32c, is Sha256, is Sha1 -> true
+        else -> false
+    }
+
+    /**
+     * Gets the Base64 encoded checksum of an HttpBody
+     * To use this, the HttpBody's underlying data source *must* be either a [HashingSource] or [HashingByteReadChannel],
+     * which means the HttpBody must also be either an [HttpBody.SourceContent] or [HttpBody.ChannelContent]. An exception
+     * will be thrown otherwise.
+     * @return the Base64 encoded checksum of an HttpBody
+     */
+    public val HttpBody.checksum: String get() = when (this) {
+        is HttpBody.SourceContent -> { (readFrom() as HashingSource).digest().encodeBase64String() }
+        is HttpBody.ChannelContent -> { (readFrom() as HashingByteReadChannel).digest().encodeBase64String() }
+        else -> throw ClientException("HttpBody type is not supported")
+    }
+
+    /**
+     * Calculate the checksum synchronously. This will suspend until the entire body has been hashed.
+     * @param hash the [HashFunction] used to calculate the checksum
+     * @return the checksum of the body as a Base64 encoded string
+     */
+    private suspend fun HttpBody.calculateChecksum(hash: HashFunction): String = when (this) {
+        is HttpBody.SourceContent -> readFrom().readToByteArray().hash(hash).encodeBase64String()
+        is HttpBody.ChannelContent -> readFrom().readToBuffer().readToByteArray().hash(hash).encodeBase64String()
+        is HttpBody.Bytes -> bytes().hash(hash).encodeBase64String()
+        is HttpBody.Empty -> throw RuntimeException("can't calculate checksum of an empty body")
+    }
+
+    /**
+     * Removes all checksum headers except [headerName]
+     * @param headerName the checksum header name to keep
+     */
+    private fun HeadersBuilder.removeAllChecksumHeadersExcept(headerName: String) {
+        names().forEach { name ->
+            if (name.startsWith("x-amz-checksum-") && name != headerName) {
+                remove(name)
+            }
+        }
+    }
+}
--- a/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsResponseMiddleware.kt
+++ b/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsResponseMiddleware.kt
@@ -1,0 +1,115 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package aws.smithy.kotlin.runtime.http.middleware
+
+import aws.smithy.kotlin.runtime.ClientException
+import aws.smithy.kotlin.runtime.hashing.HashFunction
+import aws.smithy.kotlin.runtime.hashing.toHashFunction
+import aws.smithy.kotlin.runtime.http.HttpBody
+import aws.smithy.kotlin.runtime.http.isSuccess
+import aws.smithy.kotlin.runtime.http.operation.*
+import aws.smithy.kotlin.runtime.http.response.HttpCall
+import aws.smithy.kotlin.runtime.http.toHttpBody
+import aws.smithy.kotlin.runtime.io.*
+import aws.smithy.kotlin.runtime.util.*
+import kotlin.coroutines.coroutineContext
+
+// The priority to validate response checksums, if multiple are present
+internal val CHECKSUM_HEADER_VALIDATION_PRIORITY_LIST: List<String> = listOf(
+    "x-amz-checksum-crc32c",
+    "x-amz-checksum-crc32",
+    "x-amz-checksum-sha1",
+    "x-amz-checksum-sha256",
+)
+
+internal val MULTIPART_CHECKSUM_HEADER_REGEX = Regex("x-amz-checksum-[a-zA-Z0-9]+-[0-9]+")
+
+/**
+ * Validate a flexible checksums response.
+ *
+ * Wraps the response in a hashing body, calculating the checksum as the response is streamed to the user.
+ * The checksum is validated after the user has consumed the entire body.
+ * Users can see which checksum was validated by referencing the `ResponseChecksumValidated` execution context variable.
+ */
+@InternalApi
+public class FlexibleChecksumsResponseMiddleware : ReceiveMiddleware {
+
+    public companion object {
+        public val ResponseChecksum: AttributeKey<LazyAsyncValue<String>> = AttributeKey("ResponseChecksum")
+        public val ExpectedResponseChecksum: AttributeKey<String> = AttributeKey("ExpectedResponseChecksum")
+
+        public val ChecksumHeaderValidated: AttributeKey<String> = AttributeKey("ChecksumHeaderValidated")
+    }
+
+    public override suspend fun <H : Handler<SdkHttpRequest, HttpCall>> handle(request: SdkHttpRequest, next: H): HttpCall {
+        val logger = coroutineContext.getLogger<FlexibleChecksumsResponseMiddleware>()
+
+        var call = next.call(request)
+        if (!call.response.status.isSuccess()) {
+            logger.error { "Call not successful" }
+            return call
+        }
+
+        val checksumHeader = CHECKSUM_HEADER_VALIDATION_PRIORITY_LIST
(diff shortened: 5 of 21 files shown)


runtime/protocol/http/common/test/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsRequestMiddlewareTest.kt
```diff
@@ -0,0 +1,129 @@
...
+            op.roundTrip(client, Unit)
+        }
+    }
+
+    @Test
+    fun itRemovesChecksumHeadersForAwsChunked() = runTest {
+        val req = HttpRequestBuilder().apply {
+            body = object : HttpBody.SourceContent() {
+                override val contentLength: Long = 1024 * 1024 * 128
+                override fun readFrom(): SdkSource = "a".repeat(1024 * 1024 * 128).toByteArray().source()
```
ianbotsf, line 109: **Nit**: `1024 * 1024 * 128` → `contentLength`

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolClientGenerator.kt
```diff
@@ -282,8 +280,33 @@ abstract class HttpProtocolClientGenerator(
...
+            ?.members()
+            ?.firstOrNull { it.memberName == httpChecksumTrait?.requestAlgorithmMember?.getOrNull() }
+
+        if (hasTrait<HttpChecksumRequiredTrait>() || httpChecksumTrait?.isRequestChecksumRequired == true) {
+            writer.addImport(RuntimeTypes.Http.Middlware.Md5ChecksumMiddleware)
+
+            checksumAlgorithm?.let {
+                writer.withBlock("if (input.#L == null) {", "}", checksumAlgorithm.defaultName()) {
+                    writer.write("op.install(#T())", RuntimeTypes.Http.Middlware.Md5ChecksumMiddleware)
+                }
```
ianbotsf, line 306: **Question**: This installs the MD5 checksum middleware if the input has no checksum algorithm _at invocation time_? Isn't it possible for the checksum algorithm to be set later by an interceptor? Would we still want the MD5 checksum middleware to run in that case?

ianbotsf (changes requested):

runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashFunction.kt
```diff
@@ -57,3 +57,16 @@ public typealias HashSupplier = () -> HashFunction
...
+@InternalApi
+/**
+ * Return the [HashFunction] which is represented by this string, or null if none match.
+ */
+public fun String.toHashFunction(): HashFunction? = when (this.lowercase()) {
+    "crc32" -> Crc32()
+    "crc32c" -> Crc32c()
+    "sha1" -> Sha1()
+    "sha256" -> Sha256()
+    "md5" -> Md5()
```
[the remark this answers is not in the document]
lauzadis: The flexible checksums specification uses this ABNF for algorithm names:
```
algorithms = 1*(Upper-alpha / DIGIT)
lower-alpha = %x61-7A
```

So at least for flexible checksums, we will not have to deal with alternate representations. I think this is a good concise representation of algorithm names and we could use it for our other work as well.

runtime/io/common/src/aws/smithy/kotlin/runtime/io/CompletingByteReadChannel.kt
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * A channel which uses the digest of an underlying [HashingByteReadChannel] to complete a [CompletableDeferred]
+ * @param deferredChecksum The [CompletableDeferred] to be completed when the underlying channel is exhausted
+ * @param hashingChannel the [HashingByteReadChannel] which will be digested and used to complete [deferredChecksum]
+ */
+@InternalApi
+public class CompletingByteReadChannel(
+    private val deferredChecksum: CompletableDeferred<String>,
+    private val hashingChannel: HashingByteReadChannel,
+) : SdkByteReadChannel by hashingChannel {
```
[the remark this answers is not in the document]
lauzadis: I agree, I thought about how to make it more generic. The `CompletingByteReadChannel` could take a function as input which would be executed to obtain the value after the underlying body is exhausted.

```kotlin
public class CompletingByteReadChannel<T>(
    private val deferred: CompletableDeferred<T>,
    private val channel: SdkByteReadChannel,
    private val block: suspend () -> (T),
) : SdkByteReadChannel by channel {
        public override suspend fun read(sink: SdkBuffer, limit: Long): Long =
        channel.read(sink, limit).also {
            if (channel.isClosedForRead) {
                deferredChecksum.complete(block())
            }
        }
}
```

Then, it can be used for hashing channels like this:

```kotlin
val deferredChecksum = CompletableDeferred()
val channel = HashingByteReadChannel(...)
val completingChannel = CompletingByteReadChannel<String>(deferredChecksum, channel) { channel.digest().encodeBase64String() }
```

What do you think about this?

runtime/io/common/src/aws/smithy/kotlin/runtime/io/HashingByteReadChannel.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+@InternalApi
+public class HashingByteReadChannel(private val hash: HashFunction, private val chan: SdkByteReadChannel) : SdkByteReadChannel by chan {
+    public override suspend fun read(sink: SdkBuffer, limit: Long): Long {
+        val buffer = SdkBuffer()
+
+        val rc = chan.read(buffer, limit)
+        if (rc == -1L) { return rc } // got 0 bytes, no need to hash anything
+
+        val dataToHash = buffer.commonPeek().readByteArray()
```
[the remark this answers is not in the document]
lauzadis: `peek` should work. The code-completion / suggestions in my IDE don't work great with the new I/O refactor, and I only picked `commonPeek` because it showed up and `peek` didn't.

runtime/io/common/src/aws/smithy/kotlin/runtime/io/HashingByteReadChannel.kt
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * A channel which hashes data as it is being read
+ * @param hash The [HashFunction] to hash data with
+ * @param chan the [SdkByteReadChannel] to hash
+ */
+@InternalApi
+public class HashingByteReadChannel(private val hash: HashFunction, private val chan: SdkByteReadChannel) : SdkByteReadChannel by chan {
+    public override suspend fun read(sink: SdkBuffer, limit: Long): Long {
+        val buffer = SdkBuffer()
```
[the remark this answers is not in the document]
lauzadis: I'm not sure how expensive it is, but it would be easy to refactor and reuse it between invocations, so I'll do that.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/DeferredHeaders.kt
```diff
@@ -0,0 +1,70 @@
...
+    override fun entries(): Set<Map.Entry<String, List<Deferred<String>>>> = emptySet()
+    override fun contains(name: String): Boolean = false
+    override fun isEmpty(): Boolean = true
+}
+
+/**
+ * Perform a deep copy of this map, specifically duplicating the value lists so that they're insulated from changes.
+ * @return A new map instance with copied value lists.
+ */
+internal fun Map<String, List<Deferred<String>>>.deepCopy() = mapValues { (_, v) -> v.toMutableList() }
```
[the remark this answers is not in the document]
lauzadis: Only because that's what [StringValuesMap does](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/util/StringValuesMap.kt#L100) but I can see if it's actually necessary.

In general, for the DeferredHeaders implementation, I copied the `Headers` implementation and replaced `String` with `Deferred`.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsRequestMiddleware.kt
```diff
@@ -0,0 +1,117 @@
...
+     * To use this, the HttpBody's underlying data source *must* be either a [HashingSource] or [HashingByteReadChannel],
+     * which means the HttpBody must also be either an [HttpBody.SourceContent] or [HttpBody.ChannelContent]. An exception
+     * will be thrown otherwise.
+     * @return the Base64 encoded checksum of an HttpBody
+     */
+    public val HttpBody.checksum: String get() = when (this) {
+        is HttpBody.SourceContent -> { (readFrom() as HashingSource).digest().encodeBase64String() }
+        is HttpBody.ChannelContent -> { (readFrom() as HashingByteReadChannel).digest().encodeBase64String() }
+        else -> throw ClientException("HttpBody type is not supported")
+    }
```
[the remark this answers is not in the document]
lauzadis: It's not used, I refactored and did not remove this

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsRequestMiddleware.kt
```diff
@@ -0,0 +1,117 @@
...
+                    .toHashingBody(checksumAlgorithm, req.subject.body.contentLength)
+                    .toCompletingBody(deferredChecksum, req.subject.body.contentLength)
+            }
+
+            req.subject.trailingHeaders.append(headerName, deferredChecksum)
+        } else if (req.subject.headers[headerName] == null) {
+            logger.debug { "Calculating checksum" }
+
+            val checksum = req.subject.body.readAll()?.hash(checksumAlgorithm)?.encodeBase64String()
+                ?: throw RuntimeException("Failed to calculate checksum")
```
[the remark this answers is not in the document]
lauzadis: This `readAll()` will only be called if the body is ineligible for upload with `aws-chunked` content encoding, which means it is smaller than 1MB.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsResponseMiddleware.kt
```diff
@@ -0,0 +1,104 @@
...
+        val checksumHeader = CHECKSUM_HEADER_VALIDATION_PRIORITY_LIST
+            .firstOrNull { call.response.headers.contains(it) } ?: run {
+            logger.warn { "User requested checksum validation, but the response headers did not contain any valid checksums" }
+            return call
+        }
+
+        if (MULTIPART_CHECKSUM_HEADER_REGEX.matches(checksumHeader)) {
+            logger.info { "Skipping validation of multipart response checksum $checksumHeader" }
+            return call
+        }
```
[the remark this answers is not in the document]
lauzadis: Yes that's right, I can remove this since it looks for exact matches for the header names.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/util/DeferredValuesMap.kt
```diff
@@ -0,0 +1,178 @@
...
+ */
+package aws.smithy.kotlin.runtime.http.util
+
+import aws.smithy.kotlin.runtime.util.InternalApi
+import kotlinx.coroutines.Deferred
+
+/**
+ * Mapping of String to a List of lazy values
+ */
+public interface DeferredValuesMap<T : Any> {
```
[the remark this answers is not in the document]
lauzadis: Sure, I'll try that refactor

runtime/hashing/common/src/aws/smithy/kotlin/runtime/hashing/HashFunction.kt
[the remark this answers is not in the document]
ianbotsf: Oh nice there's a defined format in the spec. 👍

runtime/io/common/src/aws/smithy/kotlin/runtime/io/CompletingByteReadChannel.kt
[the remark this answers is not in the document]
ianbotsf: That definitely seems like a cleaner separation of concerns (although I'm now noticing neither the original nor this have error handling and we'd probably want some). We could probably even add a `HashingByteReadChannel.asCompleting()` extension function to handle the extraction/encoding of the final digest.

Will we ever need to hash a channel without completing a `Deferred`? Are there other known use cases for a completing channel besides hashing?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsRequestMiddleware.kt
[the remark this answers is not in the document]
ianbotsf: It means the content size _is known_ and is smaller than 1MB. If `contentLength` is `null` this branch will trigger right?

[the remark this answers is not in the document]
lauzadis: Yes you're right. What are some cases when `contentLength` could be `null`? I don't think I've ever come across it before during testing

runtime/io/common/src/aws/smithy/kotlin/runtime/io/CompletingByteReadChannel.kt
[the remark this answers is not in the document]
aajtodd: **comment**: I'm not sure I would even expose this from `io`. I'd probably just err on the side of making it a small utility as part of the implementation (`internal` to signing/chunk implementation). I don't think we have any other use cases for this in the runtime and this is very simple wrapper code.

runtime/io/common/src/aws/smithy/kotlin/runtime/io/CompletingSource.kt
```diff
@@ -0,0 +1,24 @@
...
+import aws.smithy.kotlin.runtime.util.encodeBase64String
+import kotlinx.coroutines.CompletableDeferred
+
+/**
+ * An [SdkSource] which uses the digest of an underlying [HashingSource] to complete a [CompletableDeferred]
+ * @param deferredChecksum The [CompletableDeferred] to be completed when the underlying source is exhausted
+ * @param source the [HashingSource] which will be digested and used to complete [deferredChecksum]
+ */
+@InternalApi
+public class CompletingSource(private val deferredChecksum: CompletableDeferred<String>, private val source: HashingSource) : SdkSource by source {
```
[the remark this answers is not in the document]
aajtodd: **comment**: Same comment, this is so simple of wrapping code I'm not sure it needs to be made generic. It could just live with the as an internal implementation detail.

runtime/io/common/src/aws/smithy/kotlin/runtime/io/HashingByteReadChannel.kt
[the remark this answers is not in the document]
aajtodd: **question**: Couldn't this just be written using `HashingSink`?

e.g.

```kotlin
class HashingByteReadChannel(
    private val hash: HashFunction,
    private val delegate: SdkByteReadChannel
): SdkByteReadChannel by delegate {

    override suspend fun read(sink: SdkBuffer, limit: Long): Long {
        val hsink = HashingSink(hash, sink)
        val bufferedSink = hsink.buffer()
        val rc = delegate.read(bufferedSink.buffer, limit)
        bufferedSink.emit()
        return rc
    }

}

```

runtime/io/common/src/aws/smithy/kotlin/runtime/io/HashingByteReadChannel.kt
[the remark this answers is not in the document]
aajtodd: Not sure it's the I/O refactor, my IDE is having trouble with completions as well FWIW.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/DeferredHeaders.kt
```diff
@@ -0,0 +1,70 @@
...
+    public companion object {
+        public operator fun invoke(block: DeferredHeadersBuilder.() -> Unit): DeferredHeaders = DeferredHeadersBuilder()
+            .apply(block).build()
+
+        /**
+         * Empty [Headers] instance
+         */
+        public val Empty: DeferredHeaders = EmptyDeferredHeaders
+
+        public suspend fun DeferredHeaders.toHeaders(): Headers = when (this) {
```
aajtodd, line 23: **Quesiton**: What's the use case for this conversion? It seems...odd.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HttpBody.kt
```diff
@@ -144,6 +147,47 @@ public fun SdkSource.toHttpBody(contentLength: Long? = null): HttpBody =
...
+}
+
+/**
+ * Convert an [HttpBody.SourceContent] with a [HashingSource] or [HttpBody.ChannelContent] with a [HashingByteReadChannel]
+ * to a body which will use the digest to complete a [CompletableDeferred]
+ * @param completableDeferred the [CompletableDeferred] to complete once the underlying hashing body is exhausted
+ * @param contentLength the total content length of the source, if known
+ */
+@InternalApi
+public fun HttpBody.toCompletingBody(completableDeferred: CompletableDeferred<String>, contentLength: Long?): HttpBody =
```
aajtodd, line 176: **comment**: Same as before this doesn't feel general purpose enough and I'd maybe consider moving some of these completing things to where they are used.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsRequestMiddleware.kt
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * @param checksumAlgorithmName the name of the algorithm used to calculate the checksum
+ */
+@InternalApi
+public class FlexibleChecksumsRequestMiddleware(private val checksumAlgorithmName: String) : ModifyRequestMiddleware {
+    public override suspend fun modifyRequest(req: SdkHttpRequest): SdkHttpRequest {
+        val logger = coroutineContext.getLogger<FlexibleChecksumsRequestMiddleware>()
+
+        val headerName = "x-amz-checksum-${checksumAlgorithmName.lowercase()}"
+        logger.debug { "Resolved header name: $headerName" }
```
aajtodd, line 35: **nit**: "Resolved checksum header name"

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsRequestMiddleware.kt
```diff
@@ -0,0 +1,117 @@
...
+
+        // remove all checksum headers except for $headerName
+        // this handles the case where a user inputs a precalculated checksum, but it doesn't match the input checksum algorithm
+        req.subject.headers.removeAllChecksumHeadersExcept(headerName)
+
+        val checksumAlgorithm = checksumAlgorithmName.toHashFunction() ?: throw ClientException("Could not parse checksum algorithm $checksumAlgorithmName")
+
+        if (!checksumAlgorithm.isSupported) {
```
aajtodd, line 43: **Question**: if we've converted a checksum algorithm name to an actual runtime type when would this ever not be true?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsRequestMiddleware.kt
```diff
@@ -0,0 +1,117 @@
...
+            throw ClientException("Checksum algorithm $checksumAlgorithmName is not supported for flexible checksums")
+        }
+
+        if (req.subject.body.isEligibleForAwsChunkedStreaming) {
+            req.subject.header("x-amz-trailer", headerName)
+
+            val deferredChecksum = CompletableDeferred<String>()
```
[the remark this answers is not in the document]
aajtodd: Agreed we should tie this to _probably_ the execution context (coroutine context) [job](https://github.com/awslabs/smithy-kotlin/blob/209e835ad85004b72b977a439ef591f570378109/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/client/ExecutionContext.kt#L21). e.g. `req.context.coroutineContext.job`. It will get cancelled when the request is complete.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsRequestMiddleware.kt
```diff
@@ -0,0 +1,117 @@
...
+                req.subject.body = req.subject.body
+                    .toHashingBody(checksumAlgorithm, req.subject.body.contentLength)
+                    .toCompletingBody(deferredChecksum, req.subject.body.contentLength)
+            }
+
+            req.subject.trailingHeaders.append(headerName, deferredChecksum)
+        } else if (req.subject.headers[headerName] == null) {
+            logger.debug { "Calculating checksum" }
+
+            val checksum = req.subject.body.readAll()?.hash(checksumAlgorithm)?.encodeBase64String()
```
aajtodd, line 71: **correctness**: In addition to Ian's comment below about triggering when `contentLength` could be `null`. This is consuming the entire body without replacing it. This works for replayable streams only. Is that always true in this branch?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsResponseMiddleware.kt
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * Wraps the response in a hashing body, calculating the checksum as the response is streamed to the user.
+ * The checksum is validated after the user has consumed the entire body.
+ * Users can see which checksum was validated by referencing the `ResponseChecksumValidated` execution context variable.
+ */
+@InternalApi
+public class FlexibleChecksumsResponseMiddleware : ReceiveMiddleware {
+
+    public companion object {
+        public val ResponseChecksum: AttributeKey<Deferred<String>> = AttributeKey("ResponseChecksum")
```
aajtodd, line 43: **fix**: documentation for these keys if they are public and expected to be used

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsResponseMiddleware.kt
```diff
@@ -0,0 +1,104 @@
...
+            return call
+        }
+
+        if (MULTIPART_CHECKSUM_HEADER_REGEX.matches(checksumHeader)) {
+            logger.info { "Skipping validation of multipart response checksum $checksumHeader" }
+            return call
+        }
+
+        // let the user know which checksum will be validated
+        logger.debug { "Validating checksum in $checksumHeader" }
```
aajtodd, line 70: **nit**: `from`

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/operation/SdkHttpOperation.kt
```diff
@@ -99,6 +100,7 @@ public suspend fun <I, O, R> SdkHttpOperation<I, O>.execute(
         val output = handler.call(request)
         return block(output)
     } finally {
+        validateResponseChecksum(context)
         context.cleanup()
     }
```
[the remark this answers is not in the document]
aajtodd: Agreed this doesn't belong here. It should be done in the middleware if possible or change the implementation from middleware to `Interceptor` if necessary to get at more fine grained hooks (e.g. `modifyBeforeCompletion/readAfterExecution`).

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolClientGenerator.kt
aajtodd: Good point. Maybe we should discuss this

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolClientGenerator.kt
```diff
@@ -282,8 +280,33 @@ abstract class HttpProtocolClientGenerator(
...
-}
 
-// TODO https://github.com/awslabs/aws-sdk-kotlin/issues/557
-private fun OperationShape.checksumRequired(): Boolean =
-    hasTrait<HttpChecksumRequiredTrait>() || getTrait<HttpChecksumTrait>()?.isRequestChecksumRequired == true
+    /**
+     * Render optionally installing Md5ChecksumMiddleware.
+     * The Md5 middleware will only be installed if the operation requires a checksum and the user has not opted-in to flexible checksums.
+     */
+    private fun OperationShape.renderIsMd5ChecksumRequired(writer: KotlinWriter) {
```
aajtodd, line 288: **nit**: It feels odd to have a `render` method off the operation shape. Also this could probably just be turned into a [default middleware](https://github.com/awslabs/smithy-kotlin/blob/f29e7502abc4dd817fcf2338e742e67ad1ecb498/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt#L64) such that `HttpProtocolClientGenerator` need not know anything about it.

aajtodd (changes requested):

runtime/auth/aws-signing-tests/common/src/aws/smithy/kotlin/runtime/auth/awssigning/tests/AwsChunkedTestBase.kt
```diff
@@ -365,9 +393,9 @@ abstract class AwsChunkedTestBase(
         val data = ByteArray(dataLengthBytes) { 0x7A.toByte() }
         var previousSignature: ByteArray = byteArrayOf()
 
-        val trailingHeaders = Headers {
-            append("x-amz-checksum-crc32", "AAAAAA==")
-            append("x-amz-arbitrary-header-with-value", "UMM")
+        val trailingHeaders = DeferredHeaders {
+            append("x-amz-checksum-crc32", CompletableDeferred("AAAAAA=="))
+            append("x-amz-arbitrary-header-with-value", CompletableDeferred("UMM"))
         }
```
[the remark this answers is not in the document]
lauzadis: @aajtodd I don't understand your comment, passing a `String` value to the `CompletableDeferred` constructor marks it as completed, so I don't see how it would be useful to return it.

runtime/io/common/src/aws/smithy/kotlin/runtime/io/HashingByteReadChannel.kt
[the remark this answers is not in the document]
lauzadis: Hmm, I only ever noticed it with the newer I/O stuff. This is an example of what I mean:

Screen Shot 2023-01-20 at 2 44 40 PM

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/DeferredHeaders.kt
lauzadis: It's mainly a convenience function for calling `.await()` on all of the `Deferred` values and converting them to `String`

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsRequestMiddleware.kt
lauzadis: No it's not, I'll need to handle that 👍

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsRequestMiddleware.kt
lauzadis: `.toHashFunction()` also supports parsing MD5 (intended for other usecases like HashingSource/ByteReadChannel), so if the user modifies the `ChecksumAlgorithm` enum to pass "md5", it will convert to the proper runtime type, but MD5 is not a supported algorithm for flexible checksums.

In general this issue arises because `.toHashFunction()` was intended to be used for more than just flexible checksums. I could solve it by making an `internal` function which does the exact same thing, without MD5. What do you think?

runtime/auth/aws-signing-tests/common/src/aws/smithy/kotlin/runtime/auth/awssigning/tests/AwsChunkedTestBase.kt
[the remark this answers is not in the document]
aajtodd: Right, I see. I think my concern here was if we are exposing `CompletableDeferred` instances from the `DeferredHeaders` type (e.g. via `get`). The "complete" side of a deferred should be retained by the producer and not be accessible by a consumer of `DeferredHeaders`. As long as that is the case we're good.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/DeferredHeaders.kt
aajtodd: It's ok I suppose but it was harder to follow the implementation in the chunked implementation. This should have some documentation about what it's doing and probably be given an `InternalApi` annotation.

runtime/io/common/src/aws/smithy/kotlin/runtime/io/CompletingByteReadChannel.kt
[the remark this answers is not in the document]
lauzadis: I refactored this by instead adding an optional `CompletableDeferred` parameter to the `HashingSource`/`HashingByteReadChannel`.

If it is non-null, it will be completed using the `digest()` of the hash function once the underlying data source is exhausted.

runtime/io/common/src/aws/smithy/kotlin/runtime/io/HashingByteReadChannel.kt
```diff
@@ -0,0 +1,40 @@
...
+ * A channel which hashes data as it is being read
+ * @param hash The [HashFunction] to hash data with
+ * @param chan the [SdkByteReadChannel] to hash
+ * @param deferred the optional [CompletableDeferred] to be completed using the Base64-encoded digest of [chan] after it is exhausted
+ */
+@InternalApi
+public class HashingByteReadChannel(
+    private val hash: HashFunction,
+    private val chan: SdkByteReadChannel,
+    private val deferred: CompletableDeferred<String>? = null,
```
aajtodd, line 23: Speaking my last peace on this and then I'll shut up.... I still don't think this is a concern of `HashingByteReadChannel` or `HashingSource`. I think this functionality should be isolated to where it's needed and be done as an additional (simple) wrapper.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
```diff
@@ -0,0 +1,125 @@
...
+ * In this case, a [LazyAsyncValue] will be added to the execution context which allows the trailing checksum to be sent
+ * after the entire body has been streamed.
+ *
+ * @param checksumAlgorithmNameInitializer a function which parses the input [I] to return the checksum algorithm name
+ */
+@InternalApi
+public class FlexibleChecksumsRequestInterceptor<I>(
+    private val checksumAlgorithmNameInitializer: (I) -> String?,
+) : HttpInterceptor {
+    private var checksumAlgorithmName = CompletableDeferred<String>()
```
aajtodd, line 35: correctness/question: Why `CompletableDeferred` here? This could just be `String?` and defaulted to `null` (or possibly late init var)

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
```diff
@@ -0,0 +1,125 @@
...
+    override suspend fun modifyBeforeRetryLoop(context: ProtocolRequestInterceptorContext<Any, HttpRequest>): HttpRequest {
+        val logger = coroutineContext.getLogger<FlexibleChecksumsRequestInterceptor<I>>()
+        if (!checksumAlgorithmName.isCompleted) {
+            logger.debug { "no checksum algorithm specified, skipping flexible checksums processing" }
+            return context.protocolRequest
+        }
+
+        val req = context.protocolRequest.toBuilder()
+
+        check(context.protocolRequest.body.contentLength != null && context.protocolRequest.body.contentLength!! > 0) {
```
aajtodd, line 53: correctness: `-1` indicates an unknown body length (i.e. transfer encoding chunked). Is this correct given that?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptor.kt
```diff
@@ -0,0 +1,131 @@
...
+ * Validate a response's checksum.
+ *
+ * Wraps the response in a hashing body, calculating the checksum as the response is streamed to the user.
+ * The checksum is validated after the user has consumed the entire body using a checksum validating body.
+ * Users can check which checksum was validated by referencing the `ResponseChecksumValidated` execution context variable.
+ *
+ * @param block A function which uses the input [I] to return whether response checksum validation should occur
+ */
+
+public class FlexibleChecksumsResponseInterceptor<I>(
```
aajtodd, line 41: fix: `@InternalApi`

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/interceptors/Md5ChecksumInterceptor.kt
```diff
@@ -0,0 +1,54 @@
...
+        } ?: true
+    }
+
+    override suspend fun modifyBeforeRetryLoop(context: ProtocolRequestInterceptorContext<Any, HttpRequest>): HttpRequest {
+        if (!shouldInjectMD5Header) {
+            return context.protocolRequest
+        }
+
+        val checksum = when (val body = context.protocolRequest.body) {
+            is HttpBody.Bytes -> body.bytes().md5().encodeBase64String()
```
aajtodd, line 44: correctness: this doesn't handle anything other than in-memory http bodies, is that correct?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
lauzadis: Yeah, this is correct because the `chunked` `Transfer-Encoding` gets set during signing, which is downstream of this check

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
lauzadis: Yeah, I played around with both of those options and decided to do `CompletableDeferred` for a more structured behavior between producing and consuming the value, but it could totally work as a `String?`. I'll do that refactor for simplicity

runtime/io/common/src/aws/smithy/kotlin/runtime/io/HashingByteReadChannel.kt
lauzadis: I had a bit of trouble isolating it only to where it's needed because both the request and response interceptors need to use it. I could place it in one and use it in the other, but that adds a weird linkage between them

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/interceptors/Md5ChecksumInterceptor.kt
lauzadis: The way it's implemented today only handles in-memory byte bodies, there is even a [test case specifically for that behavior](https://github.com/awslabs/smithy-kotlin/blob/3326e72797d868258cbb84a6aedb686fe306cb3a/runtime/protocol/http/common/test/aws/smithy/kotlin/runtime/http/middleware/Md5ChecksumTest.kt#L56-L70). 

Is it possible we would need to handle other types of bodies here? I'm not sure about the context around the original implementation of this and why it was only set to work on `HttpBody.Bytes`

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/middleware/FlexibleChecksumsRequestMiddleware.kt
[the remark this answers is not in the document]
ianbotsf: I don't know of many concrete use cases but [our interface defines `contentLength` as nullable](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HttpBody.kt#L16-L20). In theory, I could imagine accepting as an HTTP body a stream coming from some source where the total length is unknown because it is being calculated on the fly. If it's very large, this seems like it could cause an OOM.

Are we sure the logic for `isEligibleForAwsChunkedStreaming` is correct? Should we not stream a body where the `contentLength` is unknown?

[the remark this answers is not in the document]
lauzadis: Discussed offline, I added a rolling hash to be used when the `contentLength` is `null` but the body is replayable. 

The unsupported case is when the `contentLength` is `null` but the body is **non**-replayable -- in this case, we can't compute the checksum using a rolling hash because the body needs to be replaced after consumption. So, the SDK will just read the entire body of unknown-length and hope it does not cause an OOM error.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
```diff
@@ -92,10 +92,26 @@ public class FlexibleChecksumsRequestInterceptor<I>(
...
+                    // calculate the checksum using a rolling hash
+                    val buffer = SdkBuffer()
+                    val bufferSize: Long = 8192
+
+                    val channel = req.body.toSdkByteReadChannel()!!
+                    while (!channel.isClosedForRead) {
+                        channel.read(buffer, bufferSize)
+                        checksumAlgorithm.update(buffer.readToByteArray())
+                    }
+                    checksumAlgorithm.digest().encodeBase64String()
```
ianbotsf, line 106: **Style**: I suggest moving the checksum calculation to its own function (do we already have an extension function somewhere for this?) and only encoding the digest to Base64 in this method.

ianbotsf (approved):

runtime/protocol/http/common/test/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptorTest.kt
```diff
@@ -0,0 +1,188 @@
...
+import aws.smithy.kotlin.runtime.http.response.HttpCall
+import aws.smithy.kotlin.runtime.http.response.HttpResponse
+import aws.smithy.kotlin.runtime.io.*
+import aws.smithy.kotlin.runtime.time.Instant
+import aws.smithy.kotlin.runtime.util.encodeBase64String
+import aws.smithy.kotlin.runtime.util.get
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.params.ParameterizedTest
```
aajtodd, line 29: fix

This is JVM only. Either move the test to `jvm` test source set or refactor to use kotest for these which has KMP support for parameterized testing (IIRC) (or just use a simple list of your own test types and iterate over them, not as nice output but it gets the job done).

runtime/protocol/http/common/test/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptorTest.kt
```diff
@@ -0,0 +1,214 @@
...
+import aws.smithy.kotlin.runtime.http.response.HttpCall
+import aws.smithy.kotlin.runtime.http.response.HttpResponse
+import aws.smithy.kotlin.runtime.io.SdkSource
+import aws.smithy.kotlin.runtime.io.source
+import aws.smithy.kotlin.runtime.time.Instant
+import aws.smithy.kotlin.runtime.util.get
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.CsvSource
```
aajtodd, line 24: fix

Same comment. JVM only imports

runtime/protocol/http/common/test/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptorTest.kt
```diff
@@ -0,0 +1,214 @@
...
+import kotlin.test.*
+
+data class TestInput(val value: String)
+data class TestOutput(val body: HttpBody)
+
+inline fun <reified I> newTestOperation(serialized: HttpRequestBuilder): SdkHttpOperation<I, TestOutput> =
```
aajtodd, line 30: fix

This looks [duplicated](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/protocol/http/common/test/aws/smithy/kotlin/runtime/http/operation/TestOperation.kt#L19). Can we just import this or move it to a common location within the HTTP test package?

aajtodd (approved):
fix and ship!

runtime/protocol/http/common/test/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptorTest.kt
lauzadis: It's not entirely duplicated, this one is customized to return `response.body` in a `TestOutput` type but the original one just returned a pre-determined `O` type. I needed access to the `response.body` in the output so I could consume the whole body and get the checksum validated, but I can't pass the body in using a pre-determined `O` because it needs to get wrapped internally in the interceptors / middleware.

PR #773, opened 2022-12-30
feat(rt): unsigned `aws-chunked`

lauzadis:
Adds support for making unsigned requests using `aws-chunked`.

Users can specify an unsigned request by setting the `isUnsignedPayload` field when configuring `AwsSigningMiddleware`.

## Issue \#
N/A

## Description of changes
This change is required to enable sending unsigned requests using aws-chunked. This is particularly useful for flexible checksums because users may want to make unsigned requests but still provide a checksum as a trailing header. 

### DIFF
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/HashSpecification.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/HashSpecification.kt
@@ -40,6 +40,11 @@ public sealed class HashSpecification {
     public object StreamingAws4HmacSha256PayloadWithTrailers : HashLiteral("STREAMING-AWS4-HMAC-SHA256-PAYLOAD-TRAILER")
 
     /**
+     * The hash value used for streaming unsigned requests with trailers
+     */
+    public object StreamingUnsignedPayloadWithTrailers : HashLiteral("STREAMING-UNSIGNED-PAYLOAD-TRAILER")
+
+    /**
      * The hash value should indicate ???
      */
     public object StreamingAws4HmacSha256Events : HashLiteral("STREAMING-AWS4-HMAC-SHA256-EVENTS")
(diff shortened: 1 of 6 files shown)


runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/HashSpecification.kt
```diff
@@ -39,6 +39,11 @@ public sealed class HashSpecification {
      */
     public object StreamingAws4HmacSha256PayloadWithTrailers : HashLiteral("STREAMING-AWS4-HMAC-SHA256-PAYLOAD-TRAILER")
 
+    /**
+     * The hash value used for streaming unsigned requests with trailers
+     */
+    public object StreamingUnsignedPayloadWithTrailers : HashLiteral("STREAMING-UNSIGNED-PAYLOAD-TRAILER")
+
     /**
      * The hash value should indicate ???
```
aajtodd, line 48: not related to your PR but this comment could be updated, the sha256-events hash literal indicates an event stream (e.g. `vnd.amazon.event-stream` content type)

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/internal/AwsChunkedReader.kt
```diff
@@ -157,6 +162,34 @@ internal class AwsChunkedReader(
...
+        val unsignedChunk = SdkBuffer()
+
+        // headers
+        unsignedChunk.apply {
+            writeUtf8(bodyBuffer.size.toString(16))
+            writeUtf8("\r\n")
+        }
+
+        // append the body
+        unsignedChunk.write(bodyBuffer.readByteArray())
```
aajtodd, line 188: fix: Don't use `readByteArray`, when working with buffers you want to try and stick to methods that take advantage of the internals of the buffer as much as possible (e.g. anything taking an `SdkSource` or `SdkBuffer` for instance). Internally these types transfer data ownership when possible rather than copying but if you force `readByteArray()` it's may end up doing two copies (1 (maybe) to get the data out of `bodyBuffer` and another to copy it into `unsignedChunk`).

This can just be `unsignedChunk.writeAll(bodyBuffer)`

aajtodd (approved):
Fix and ship

ianbotsf (approved):
> Users can specify an unsigned request by setting the `isUnsignedPayload` field when configuring `AwsSigningMiddleware`.

How can users configure middleware (signing or otherwise)?

lauzadis:
> > Users can specify an unsigned request by setting the `isUnsignedPayload` field when configuring `AwsSigningMiddleware`.
> 
> How can users configure middleware (signing or otherwise)?

The only way I've found to do this is by modifying the code-generated service client, which is not the best user experience. Maybe we can expose some configuration field in the signer or somewhere else?

PR #774, opened 2023-01-03
feat(rt): add configurable DNS implementation

### LINKED ISSUE #745: Ref Arch - Define a DNS Interface and default implementation
We need to define a DNS interface and allow HTTP engines to be configured with it. This is required to decouple HTTP engines from DNS resolution. Currently both of the supported engines (OkHttp and CRT) in `smithy-kotlin` allow custom DNS resolvers to be plugged in. 

Reference architecture 3.1.1

aajtodd:
## Issue \#

closes #745
 
## Description of changes

* **feat**: add a new `HostResolver` interface that allows customizing DNS resolution. 
    * The default implementation delegates to the system/JVM implementation for now.
    * Wire up `HostResolver` to HTTP client engines. CRT is left off for now because there isn't really a way to customize DNS in CRT currently (not in the JVM bindings anyway). Ticket opened for them to address at some point.
* **refactor**: move `Host` into runtime core and refactor the definition to take advantage of the new `IpAddr` type.


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/IpAddr.kt
```diff
@@ -0,0 +1,416 @@
...
+    /**
+     * True if this is an [Ipv6] instance
+     */
+    public val isIpv6: Boolean
+        get() = this is Ipv6
+
+    /**
+     * An IPv4 address as defined by [RFC 791](https://www.rfc-editor.org/rfc/rfc791)
+     */
+    public data class Ipv4(
```
aajtodd, line 55: **bikeshed**

`IpAddr.Ipv4` or `IpAddr.V4`?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/IpAddr.kt
```diff
@@ -0,0 +1,416 @@
...
+
+package aws.smithy.kotlin.runtime.net
+
+import aws.smithy.kotlin.runtime.util.InternalApi
+
+/**
+ * An IP Address (either IPv4 or IPv6)
+ */
+@InternalApi
+public sealed class IpAddr {
```
aajtodd, line 14: **bikeshed**

The variants of this sealed class are currently nested classes (e.g. `IpAddr.Ipv4`). Alternatively we could move each subclass to it's own file (I tend to like scoped names as it usually helps readability/understanding but these are a little bit more involved).

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/HostResolver.kt
```diff
@@ -0,0 +1,54 @@
...
+         * The default DNS host resolver (usually the default for the platform, e.g. InetAddress for JVM)
+         */
+        public val Default: HostResolver = DefaultHostResolver
+    }
+
+    /**
+     * Resolves the address(es) for a particular host and returns the list
+     * of addresses
+     */
+    public suspend fun resolve(hostname: String): List<HostAddress>
```
aajtodd, line 41: **question**

Ref arch calls out this function taking a struct, e.g.:

```kotlin
data class HostResolverArgs(val hostname: String, val service?: String = null)

public interface HostResolver {
     public suspend fun resolve(args: HostResolverArgs): List<HostAddress>
}
```

I shied away from this because (1) there is no documented use for it yet and (2) we can always add an overload later. I have no real strong opinion though, thoughts?

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/net/DefaultHostResolverJVM.kt
```diff
@@ -0,0 +1,41 @@
...
+import kotlinx.coroutines.withContext
+import java.net.Inet4Address
+import java.net.Inet6Address
+import java.net.InetAddress
+
+internal actual object DefaultHostResolver : HostResolver {
+    override suspend fun resolve(hostname: String): List<HostAddress> = withContext(Dispatchers.IO) {
+        InetAddress.getAllByName(hostname).map { it.toHostAddress() }
+    }
+    override fun reportFailure(addr: HostAddress) { }
```
aajtodd, line 19: **question**: Should there be a default caching behavior? If so what?

lucix-aws (approved):
Looks great, although I think we should discuss caching.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/Text.kt
```diff
@@ -0,0 +1,128 @@
...
+            in 'a'..'z', in 'A'..'Z', in '0'..'9', '-', '_', '.', '~' -> return@forEachIndexed
+            '%' -> {
+                if (index > length - 2) return false
+                if (!this[index+1].isHexDigit() || !this[index+2].isHexDigit()) return false
+            }
+            else -> return false
+        }
+    }
+    return true
+}
```
ianbotsf, line 126: **Question**: Unused?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/IpAddr.kt
aajtodd: > I think I'm leaving towards changing this to  IpAddr.V4...

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/net/DefaultHostResolverJVM.kt
ianbotsf: I think there needs to be higher-level caching to meet the spirit of the reference architecture. `InetAddress` provides some caching but without the ability to invalidate or purge. If we're going to use that, it implies to me that we resolve what it reports (post-low-level caching) with what we know from invalidations and purges...or bypass `InetAddress` altogether and _still_ write our own caching layer.

ianbotsf (approved):
Generally looks pretty good but we should resolve the question of caching _or_ defer to a later PR.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/IpAddr.kt
ianbotsf: Oh good idea. 👍

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/HostResolver.kt
```diff
@@ -0,0 +1,54 @@
...
+     * The resolved internet address
+     */
+    val address: IpAddr,
+)
+
+/**
+ * Component capable of resolving host names to one or more internet addresses
+ */
+@InternalApi
+public interface HostResolver {
```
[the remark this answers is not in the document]
aajtodd: Hedging risk. We need to decouple DNS and HTTP clients but it's not clear how much a customer should actually be able to configure this. Also based on conversations with Henso there is a chance this interface evolves some. The important bit right now, architecturally speaking, is breaking out DNS into a component we control and that HTTP clients consume.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/IpAddr.kt
aajtodd: > If we stay with this structure

I'm not sure I follow what you're saying or implying with this comment.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/IpAddr.kt
aajtodd: Well if we break out the variants into separate files then I'm less inclined to name it just `V4` or `V6` because it won't be namespaced by `IpAddr` anymore. 

Let's simplify with some real alternatives to consider:

ALT1:
```kotlin
sealed class IpAddr {
    class V4(...): IpAddr()

    class V6(...): IpAddr()
}
```

usage
```kotlin
val v4Addr = IpAddr.V4(...)
```
-----

ALT2
```kotlin
sealed class IpAddr {
     // common stuff
}

// file: IpV4Addr.kt
class IpV4Addr(...): IpAddr()


// file: IpV6Addr.kt
class IpV6Addr(...): IpAddr()

```

usage
```kotlin
val v4Addr = IpV4Addr(...)
```

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/IpAddr.kt
```diff
@@ -0,0 +1,416 @@
...
+         * The four eight-bit integers that make up this address
+         */
+        override val octets: ByteArray,
+    ) : IpAddr() {
+
+        /**
+         * Creates a new IPv4 address from four eight-bit octets. The result
+         * represents the IP address `a.b.c.d`
+         */
+        public constructor(a: UByte, b: UByte, c: UByte, d: UByte) : this(byteArrayOf(a.toByte(), b.toByte(), c.toByte(), d.toByte()))
```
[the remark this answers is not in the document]
aajtodd: I think it's reasonable to use parse but it's also reasonable to be able to specify segments directly which correspond to `a.b.c.d` and `a.b.c.d.e.f.h.g` for IPV4 and IPV6 respectively.

> I don't see them used outside of tests

I don't find this to be a good reason to dictate API design choices. I chose an API that I think is reasonable to represent an IP address, whether we use all of it or not today.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/IpAddr.kt
```diff
@@ -0,0 +1,416 @@
...
+                    else -> null
+                }
+            }
+
+        /**
+         * Returns true if this is a multicast address as defined by
+         * [RFC 4291 2.7](https://www.rfc-editor.org/rfc/rfc4291#section-2.7).
+         */
+        public val isMulticast: Boolean
+            get() = segments[0] and 0xff00u == 0xff00u.toUShort()
```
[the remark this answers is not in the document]
aajtodd: Yeah I think there is definitely a bit more to be exposed. My aim was to get a fairly reasonable representation of an API for IP addresses regardless of what we consume. I can understand going the other way though and exposing a more minimal surface area until need arises. What is your preference and why? Also if you're preference is a minimal subset what subset do you propose?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/Text.kt
aajtodd: https://github.com/awslabs/smithy-kotlin/pull/774#discussion_r1063463854

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/HostResolver.kt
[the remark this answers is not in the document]
ianbotsf: Got it. Then I recommend making the client engine config property `@InternalApi` as well.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/IpAddr.kt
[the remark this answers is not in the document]
ianbotsf: My concern is support burden. Exposing these properties means we must support them and even keep them up to date if more multicast scopes are introduced. We've chosen in other places to limit our implementations and APIs to only the features our SDK truly needs (e.g., deserialization) instead of providing general-use library functionality.

My preference for a minimal subset would be only the things which we need now or know we're going to need shortly for upcoming implementations. Other than that I'd assume we're not going to need it and could add it later should that change.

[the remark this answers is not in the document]
aajtodd: This likely mostly comes down to removing the multicast flags then I think, the rest of this seems like the bare bones API I'd want to define (IPv4, IPv6, parsing, toString, isLoopBack, and isUnspecified).

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/IpV4Addr.kt
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * An IPv4 address as defined by [RFC 791](https://www.rfc-editor.org/rfc/rfc791)
+ */
+public data class IpV4Addr(
+    /**
+     * The four eight-bit integers that make up this address
+     */
+    override val octets: ByteArray,
+) : IpAddr() {
```
ianbotsf, line 16: **Nit**: Missed this the first time but this is not the correct way to document parameters. They should be annotated alongside the class/method documentation:

```kotlin
/**
 * An IPv4 address as defined by [RFC 791](https://www.rfc-editor.org/rfc/rfc791)
 * @param octets The four eight-bit integers that make up this address
 */
```

Doing it this way compiles but has no effect on IDE/HTML docs.

Same problem noted in `IpV6Addr`.

ianbotsf (approved):
Small problem left, fix and ship!

PR #776, opened 2023-01-11
feat: implement client config overrides for one or more operations

lucix-aws:
(aka per-op config)

## Description of changes
* Implement `withConfig` extension on base interface. Takes a builder and returns a new client with a patched config.
* Implement `SdkManaged`, which represents some opaque resource that is shared (`share()`) and released (`unshare()`) by clients where necessary without user input.
* Implement `SdkManagedCloseable`, which as the name implies, wraps `Closeable`s to be used as `SdkManaged` instances. In the case of `Closeable`, the last call to `unshare()` triggers the `close()`.
* Make HTTP client engine a required config field. When not specified, a `ManagedHttpClientEngine` (implements `SdkManaged`) is created to track share count across `withConfig`-spawned clients.
* Add `SdkManagedGroup`, which accepts a variable number of managed resources and can `unshare()` all of them at once. Default clients now use this to simplify the logic of `init` and `close`.
* Add new `isExtension` semantics for symbols to prevent writers from incorrectly full-namespacing extensions when multiple with the same name are present in a file.


runtime/auth/aws-credentials/common/src/aws/smithy/kotlin/runtime/auth/awscredentials/CredentialsProvider.kt
```diff
@@ -4,12 +4,18 @@
...
-public interface CredentialsProvider {
+public interface CredentialsProvider : SharedCloseable {
     /**
      * Request credentials from the provider
      */
     public suspend fun getCredentials(): Credentials
+
+    public override fun share() {}
+
+    public override fun close() {}
```
ianbotsf, line 20: **Question**: Why is this necessary? Won't this remove the need for implementations to implement `share` and `close`, potentially leading to resources that are never closed?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/engine/internal/ManagedHttpClientEngine.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    override fun close() { wrapped.close() }
+}
+
+/**
+ * Wraps an [HttpClientEngine] to implement [SharedCloseable] for tracking internal use across multiple clients.
+ */
+@InternalApi
+public fun HttpClientEngine.manage(): HttpClientEngine =
+    if (this is ManagedHttpClientEngine) this else ManagedHttpClientEngine(this)
```
ianbotsf, line 27: **Nit**: `this is ManagedHttpClientEngine` → `isManaged()`

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ClientConfigGenerator.kt
```diff
@@ -197,17 +201,30 @@ class ClientConfigGenerator(
...
+                    writeInline("#1L = this@#configClass.name:L.#1L", it.propertyName)
+                    // must convert back to MutableList for builder
+                    // FIXME ideally, we either model the config and builder params separately, or refactor generation to have the builder version be derivable
+                    if (it.symbol.fullName.startsWith(KotlinTypes.Collections.MutableList.fullName)) {
+                        writeInline(".toMutableList()")
+                    }
+                    write("")
+                }
+            }
+    }
```
ianbotsf, line 217: **Correctness**: Missing unit tests for this codegen.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ServiceGenerator.kt
```diff
@@ -159,6 +164,17 @@ class ServiceGenerator(private val ctx: RenderingContext<ServiceShape>) {
         writer.write("public #L", signature)
     }
 
+    private fun renderWithConfig() {
+        writer.dokka {
+            write("Create a new subclient starting from the current one's config.")
```
ianbotsf, line 169: **Correctness**: Don't use the word "subclient" to refer to the new clients. Prefer terms like "new client", "independent client", etc.

ianbotsf (changes requested):
The interfaces/types for shareable resources feel incorrect here. Every type (e.g., engine, credentials provider) that _could_ be shared "implements" `SharedCloseable` but may have no-op implementations. But then the only place we check if something _is_ shared is via `isManaged`—which gates calling `.share()`. If that's the case, why do we need a "shared" interface in the first place?

aajtodd (changes requested):
> The interfaces/types for shareable resources feel incorrect here. Every type (e.g., engine, credentials provider) that could be shared "implements" SharedCloseable but may have no-op implementations. But then the only place we check if something is shared is via isManaged—which gates calling .share(). If that's the case, why do we need a "shared" interface in the first place?

I agree I'm having a hard time currently understanding the way resources are managed. Perhaps we can get a walkthrough to level set understanding of how all these new types are supposed to be used. If I squint hard enough I can see it but I don't want to assume anything and want to make sure I fully understand this design.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolClientGenerator.kt
```diff
@@ -89,27 +92,24 @@ abstract class HttpProtocolClientGenerator(
...
     }
 
-    //  defaults to Ktor since it's the only available engine in smithy-kotlin runtime
     /**
-     * The client engine to default to when one is not given in config. This type *MUST* be default constructable
-     * or else you need to override [renderInit] and construct it manually
+     * Render the class initialization block, the base behavior of which is to accept and configure the HTTP client engine.
+     * Use [renderAdditionalInit] to add logic within this block.
      */
-    protected open val defaultHttpClientEngineSymbol: Symbol = buildSymbol {
```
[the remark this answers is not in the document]
lucix-aws: It was just moved (and formalized as a RuntimeTypes symbol) to the client config property, since the defaulting was relocated there.

aajtodd (changes requested):
Couple questions/comments. An idea to kick around (maybe discuss in slack if it's easier).

runtime/auth/aws-credentials/common/src/aws/smithy/kotlin/runtime/auth/awscredentials/CredentialsProvider.kt
```diff
@@ -4,12 +4,19 @@
  */
 package aws.smithy.kotlin.runtime.auth.awscredentials
 
+import aws.smithy.kotlin.runtime.io.Closeable
+
 /**
  * Represents a producer/source of AWS credentials
  */
-public interface CredentialsProvider {
+public interface CredentialsProvider : Closeable {
```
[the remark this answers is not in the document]
lucix-aws: i would argue it simplifies the implementation (going back to the point that there does exist a sensible `close()` default behavior which is to no-op) but i don't have a strong opinion

runtime/io/common/test/aws/smithy/kotlin/runtime/io/ManagedCloseableTest.kt
```diff
@@ -0,0 +1,60 @@
...
+    fun testCloseNoShare() {
+        val closeable = MockCloseableImpl()
+        val wrapped = ManagedCloseable(closeable)
+
+        wrapped.close()
+        assertTrue(closeable.isClosed)
+    }
+
+    @Test
+    fun testCloseWithShare() {
```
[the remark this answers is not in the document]
lucix-aws: Every user of the default calls share including the first -- so in your example, after c2.close(), the share count is still 1 and so it wouldn't be closed.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolClientGenerator.kt
```diff
@@ -89,27 +92,24 @@ abstract class HttpProtocolClientGenerator(
...
-     */
-    protected open fun renderInit(writer: KotlinWriter) {
-        writer.addImport(defaultHttpClientEngineSymbol)
-        writer.openBlock("init {", "}") {
-            writer.write("val httpClientEngine = config.httpClientEngine ?: #T()", defaultHttpClientEngineSymbol)
-            writer.write("client = sdkHttpClient(httpClientEngine, manageEngine = config.httpClientEngine == null)")
+            write("")
+            renderAdditionalInit(writer)
         }
     }
```
ianbotsf, line 109: **Question**: I only see HTTP engine being shared here. Will credentials provider sharing come in a future PR?

ianbotsf (commented):
**Question**: Looks like the [long-standing issue requesting `Closeable`/`use` be added to stdlib](https://youtrack.jetbrains.com/issue/KT-31066) is resolved with the commitment that `AutoCloseable` will be added in Kotlin 1.8.20. Does that change any of our thinking? Would we ever rip out our own `Closeable`/`use` and replace them with stdlib `AutoCloseable` (incl. as the base class for `ManagedCloseable`)?

runtime/auth/aws-credentials/common/src/aws/smithy/kotlin/runtime/auth/awscredentials/CredentialsProvider.kt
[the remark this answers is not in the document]
lucix-aws: With the introduction of `SdkManaged` I'm a much stronger advocate of having this implement `Closeable` by default:
* same reason as before w.r.t. no-op `close`
* `DefaultCredentialsProvider` itself implements `Closeable`. It makes perfect sense characteristically (in my mind) that a credentials provider might have some lifecycle with resources that need to be explicitly reaped(just like `HttpClientEngine` - it might _not_ need an explicit `close()` on a platform like js where management of connections/resources is likely entirely internal to the node runtime, but we mark it as implementing close anyway rather than just say having `OkHttpEngine` specifically be closeable)
* simplifies the implementation when wrapping into `SdkManagedCloseable` - if the default `close` wasn't there, we'd have to check and wrap with a no-op implementation

runtime/auth/aws-credentials/common/src/aws/smithy/kotlin/runtime/auth/awscredentials/CredentialsProvider.kt
```diff
@@ -4,12 +4,19 @@
...
     /**
      * Request credentials from the provider
      */
     public suspend fun getCredentials(): Credentials
+
+    /**
+     * Shutdown and cleanup any resources
+     */
+    override fun close() { }
```
ianbotsf, line 21: **Question**: Why is this necessary? Won't this remove the need for implementations to implement `close`, potentially leading to resources that are never closed?

ianbotsf (approved):

runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkManaged.kt
```diff
@@ -0,0 +1,56 @@
...
+ */
+@InternalApi
+public interface SdkManaged {
+    /**
+     * Invoked by a caller to declare usership of the resource.
+     */
+    public fun share()
+
+    /**
+     * Invoked when a caller releases the resource. Returns a boolean indicating whether the resource has been released
```
aajtodd, line 22: **nit**: The semantics here are just `share` and `unshare`, the concept of what happens to the resource when it is unshared is up to the implementaiton. Maybe remove the language here referring to "release", etc.

runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkManaged.kt
```diff
@@ -0,0 +1,56 @@
...
+     * false.
+     */
+    public fun unshare(): Boolean
+}
+
+/**
+ * Abstract class which implements usage count tracking for [SdkManaged].
+ */
+@InternalApi
+public abstract class SdkManagedImpl : SdkManaged {
```
aajtodd, line 33: **style**: Without doing an exhaustive search nearly all of our abstract classes in the runtime are named as either `AbstractXyz` or `XyzBase`. Let's not introduce another naming convention.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/engine/internal/ManagedHttpClientEngine.kt
```diff
@@ -0,0 +1,21 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.http.engine.internal
+
+import aws.smithy.kotlin.runtime.http.engine.HttpClientEngine
+import aws.smithy.kotlin.runtime.io.SdkManagedCloseable
+import aws.smithy.kotlin.runtime.util.InternalApi
+
+@InternalApi
+public class ManagedHttpClientEngine(
```
aajtodd, line 12: **Question**: I think this can be `internal` or perhaps even `private` now. We only care about `SdkManaged` as a marker interface on the type, the concrete implementation doesn't need to be public anymore I don't think.

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ServiceGenerator.kt
```diff
@@ -159,6 +164,18 @@ class ServiceGenerator(private val ctx: RenderingContext<ServiceShape>) {
         writer.write("public #L", signature)
     }
 
+    private fun renderWithConfig() {
+        writer.dokka {
+            write("Create a new, independent client starting from the current one's config.")
+            write("Defaulted closeable resources are shared between clients, and will only be closed when ALL clients using them are closed.")
```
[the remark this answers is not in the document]
aajtodd: bump

runtime/auth/aws-credentials/common/src/aws/smithy/kotlin/runtime/auth/awscredentials/CredentialsProvider.kt
lucix-aws: see https://github.com/awslabs/smithy-kotlin/pull/776#discussion_r1072877830

runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkManaged.kt
```diff
@@ -0,0 +1,56 @@
...
+
+import aws.smithy.kotlin.runtime.util.InternalApi
+import kotlinx.atomicfu.locks.SynchronizedObject
+import kotlinx.atomicfu.locks.synchronized
+
+/**
+ * Interface that the lifecycle of some resource is managed by the SDK at runtime.
+ */
+@InternalApi
+public interface SdkManaged {
```
aajtodd, line 15: **Question**: Would you say this belongs in `runtime-core` long term? (I understand why it is here right now).

aajtodd (approved):

runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkManaged.kt
lucix-aws: 100%

ianbotsf (approved):

runtime/io/common/src/aws/smithy/kotlin/runtime/io/Closeable.kt
```diff
@@ -43,3 +43,7 @@ public inline fun <C : Closeable, R> C.use(block: (C) -> R): R {
         }
     }
 }
+
+public fun Any.closeIfCloseable() {
```
aajtodd, line 47: `@InternalApi`?

aajtodd (approved):

PR #789, opened 2023-02-02
refactor: break out service client runtime into own module

aajtodd:
## Issue \#

N/A
## Description of changes

Split off client only runtime components into separate module from `runtime-core`


lauzadis (approved):

smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/RuntimeTypes.kt
```diff
@@ -186,6 +173,19 @@ object RuntimeTypes {
         }
     }
 
+    object SmithyClient : RuntimeTypePackage(KotlinDependency.SMITHY_CLIENT) {
```
ianbotsf, line 176: **Question**: Same question about naming as with module name.

tests/codegen/paginator-tests/build.gradle.kts
```diff
@@ -65,6 +65,7 @@ dependencies {
 
     compileOnly(project(":smithy-kotlin-codegen"))
     implementation(project(":runtime:runtime-core"))
+    implementation(project(":runtime:smithy-client"))
```
ianbotsf, line 68: **Question**: Do we need to explicitly depend on **runtime-core** if **smithy-client** pulls it in as an API dependency?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/Exceptions.kt
```diff
@@ -82,12 +81,6 @@ public open class ServiceErrorMetadata : ErrorMetadata() {
         public val RequestId: AttributeKey<String> = AttributeKey("RequestId")
     }
 
-    /**
-     * The name of the service that sent this error response
-     */
-    public val serviceName: String
-        get() = attributes.getOrNull(SdkClientOption.ServiceName) ?: ""
-
```
ianbotsf, line 90: **Question**: Why was this removed?

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/SdkClientOption.kt
```diff
@@ -31,6 +31,11 @@ public object SdkClientOption {
      * The client logging mode (see [SdkLogMode]
      */
     public val LogMode: ClientOption<SdkLogMode> = ClientOption("LogMode")
+
+    /**
+     * Whether endpoint discovery is enabled or not. Default is true
+     */
+    public val EndpointDiscoveryEnabled: ClientOption<Boolean> = ClientOption("EndpointDiscoveryEnabled")
```
ianbotsf, line 38: **Question**: Why was this added?

settings.gradle.kts
```diff
@@ -51,6 +51,7 @@ include(":runtime:serde")
 include(":runtime:serde:serde-form-url")
 include(":runtime:serde:serde-json")
 include(":runtime:serde:serde-xml")
+include(":runtime:smithy-client")
```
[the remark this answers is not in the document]
aajtodd: client isn't very specific though either we have multiple types of clients (e.g. http client). Smithy client was to signify it's applicable to generated smithy clients (just like a hypothetical `smithy-server` would be applicable to generated servers)

tests/codegen/paginator-tests/build.gradle.kts
aajtodd: Technically no but it harms nothing to have it this way.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/Exceptions.kt
aajtodd: Because I moved `SdkClientOption` with the rest of the client stuff out of core. I looked through codegen, runtime, and generated clients. We don't ever populate this. I think we ought to remove it given that all generated exceptions inherit from a service specific exception already. I can be talked into a different direction though. I have questions/feels in general about `SdkClientOption` right now.

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/SdkClientOption.kt
aajtodd: Moved from `aws-sdk-kotlin`. We can remove it right now completely since it's unused but it was solving a FIXME from `aws-sdk-kotlin`.

PR #790, opened 2023-02-02
fix: stop logging "null" when exceptions are not present in trace events

ianbotsf:
## Issue \#

(none)

## Description of changes

#785 mistakenly started logging `null` on all trace events where no exception was present. This change fixes that by specifically detecting whether an exception is present and dispatching to the correct logger method (i.e., the overload with or without an exception).

### DIFF
--- a/runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/LoggingTraceProbe.kt
+++ b/runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/LoggingTraceProbe.kt
@@ -7,36 +7,41 @@ package aws.smithy.kotlin.runtime.tracing
 import mu.KLogger
 import mu.KotlinLogging
 
+private fun EventLevel.loggerMethod(): (KLogger, () -> Any?) -> Unit = when (this) {
+    EventLevel.Fatal,
+    EventLevel.Error,
+    -> KLogger::error
+    EventLevel.Warning -> KLogger::warn
+    EventLevel.Info -> KLogger::info
+    EventLevel.Debug -> KLogger::debug
+    EventLevel.Trace -> KLogger::trace
+}
+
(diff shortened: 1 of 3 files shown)


runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/LoggingTraceProbe.kt
```diff
@@ -7,36 +7,41 @@ package aws.smithy.kotlin.runtime.tracing
...
-            "$spanId: $content$exception"
+        when (message.exception) {
+            null -> event.level.loggerMethod()(logger) { "$spanId: ${message.content()}" }
+            else -> event.level.throwableLoggerMethod()(logger, message.exception) { "$spanId: ${message.content()}" }
         }
     }
 
     override fun postEvent(span: TraceSpan, event: TraceEvent) {
-        if (event.data is TraceEventData.Message) log(span.hierarchicalId, event)
+        if (event.data is TraceEventData.Message) log(span.hierarchicalId, event, event.data)
```
aajtodd, line 44: **nit**: It seems slightly odd to do the unwrap and pass `event` and `event.data` given that `event` contains the data already which in theory makes the extra parameter redundant/unnecessary.

aajtodd (approved):

lucix-aws (approved):

runtime/tracing/tracing-core/common/src/aws/smithy/kotlin/runtime/tracing/LoggingTraceProbe.kt
ianbotsf: It _is_ slightly odd, I agree, but I did it here to take advantage of smart-casting by the compiler. Since `postEvent` already knows that `event.data is TraceEventData.Message`, we can skip the re-casting in `log`.

PR #795, opened 2023-02-07
feat: rework max connections for OkHttp engine

ianbotsf:
## Issue \#

(none)

## Description of changes

A few connections config-related changes for OkHttp:
* Make `maxConnections` actually do something by setting up a custom `Dispatcher` instead of using it for the connection pool's idle count
* Add a new `maxConnectionsPerHost` parameter to allow greater throughput to a single endpoint

### DIFF
--- a/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
@@ -70,19 +70,28 @@ private fun OkHttpEngineConfig.buildClient(): OkHttpClient {
         followSslRedirects(false)
 
         // see: https://github.com/ktorio/ktor/issues/1708#issuecomment-609988128
+        // TODO disable this once better transient exception handling is in place
         retryOnConnectionFailure(true)
 
         connectTimeout(config.connectTimeout.toJavaDuration())
         readTimeout(config.socketReadTimeout.toJavaDuration())
         writeTimeout(config.socketWriteTimeout.toJavaDuration())
 
-        // use our own pool configured with the settings taken from config
+        // use our own pool configured with the timeout settings taken from config
         val pool = ConnectionPool(
-            maxIdleConnections = config.maxConnections.toInt(),
+            maxIdleConnections = 5, // The default from the no-arg ConnectionPool() constructor
             keepAliveDuration = config.connectionIdleTimeout.inWholeMilliseconds,
             TimeUnit.MILLISECONDS,
         )
         connectionPool(pool)
+
+        // Configure a dispatcher that uses maxConnections as a proxy for maxRequests. Note that this isn't precisely
+        // the same since some protocols (e.g., HTTP2) may use a single connection for multiple requests.
+        val dispatcher = Dispatcher().apply {
+            maxRequests = config.maxConnections.toInt()
+            maxRequestsPerHost = config.maxConnectionsPerHost.toInt()
+        }
+        dispatcher(dispatcher)
 
         // Log events coming from okhttp. Allocate a new listener per-call to facilitate dedicated trace spans.
         eventListenerFactory { call -> HttpEngineEventListener(pool, config.hostResolver, call) }
--- a/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngineConfig.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngineConfig.kt
@@ -8,6 +8,11 @@ package aws.smithy.kotlin.runtime.http.engine.okhttp
 import aws.smithy.kotlin.runtime.http.engine.HttpClientEngineConfig
 
 public class OkHttpEngineConfig private constructor(builder: Builder) : HttpClientEngineConfig(builder) {
+    /**
+     * The maximum number of connections to open to a single host.
+     */
+    public val maxConnectionsPerHost: UInt = builder.maxConnectionsPerHost
+
     public companion object {
         /**
          * The default engine config. Most clients should use this.
@@ -19,6 +24,10 @@ public class OkHttpEngineConfig private constructor(builder: Builder) : HttpClie
     }
 
     public class Builder : HttpClientEngineConfig.Builder() {
+        /**
+         * The maximum number of connections to open to a single host. Defaults to 5.
+         */
+        public var maxConnectionsPerHost: UInt = 5u
 
         internal fun build(): OkHttpEngineConfig = OkHttpEngineConfig(this)
     }
--- a/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngineConfig.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngineConfig.kt
@@ -11,7 +11,7 @@ public class OkHttpEngineConfig private constructor(builder: Builder) : HttpClie
     /**
      * The maximum number of connections to open to a single host.
      */
-    public val maxConnectionsPerHost: UInt = builder.maxConnectionsPerHost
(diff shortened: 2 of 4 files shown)


runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
```diff
@@ -70,20 +70,29 @@ private fun OkHttpEngineConfig.buildClient(): OkHttpClient {
...
 
         connectTimeout(config.connectTimeout.toJavaDuration())
         readTimeout(config.socketReadTimeout.toJavaDuration())
         writeTimeout(config.socketWriteTimeout.toJavaDuration())
 
-        // use our own pool configured with the settings taken from config
+        // use our own pool configured with the timeout settings taken from config
         val pool = ConnectionPool(
-            maxIdleConnections = config.maxConnections.toInt(),
+            maxIdleConnections = 5, // The default from the no-arg ConnectionPool() constructor
```
aajtodd, line 82: Question: Should this be a knob? I don't see any documentation on how this parameter is used internally to okhttp.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngineConfig.kt
```diff
@@ -8,6 +8,11 @@ package aws.smithy.kotlin.runtime.http.engine.okhttp
 import aws.smithy.kotlin.runtime.http.engine.HttpClientEngineConfig
 
 public class OkHttpEngineConfig private constructor(builder: Builder) : HttpClientEngineConfig(builder) {
+    /**
+     * The maximum number of connections to open to a single host.
+     */
+    public val maxConnectionsPerHost: UInt = builder.maxConnectionsPerHost
```
aajtodd, line 14: Question: Should this be a knob on base config? I believe it's called out in SRA as `maxConcurrency` (though this name is more appropriate I think)

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
```diff
@@ -70,20 +70,29 @@ private fun OkHttpEngineConfig.buildClient(): OkHttpClient {
...
             keepAliveDuration = config.connectionIdleTimeout.inWholeMilliseconds,
             TimeUnit.MILLISECONDS,
         )
         connectionPool(pool)
 
+        // Configure a dispatcher that uses maxConnections as a proxy for maxRequests. Note that this isn't precisely
```
aajtodd, line 88: Comment: I don't think we have any better option but do we have any concerns with this imprecise mapping? H2 isn't heavily used yet but this mapping could certainly limit throughput under certain setups.

I did find one of the maintainers [take](https://github.com/square/okhttp/issues/7084#issuecomment-1042872209) on not exposing a proper max connections knob.

aajtodd (approved):
Looks good, couple non blocking comments/questions.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngineConfig.kt
```diff
@@ -19,6 +24,10 @@ public class OkHttpEngineConfig private constructor(builder: Builder) : HttpClie
     }
 
     public class Builder : HttpClientEngineConfig.Builder() {
+        /**
+         * The maximum number of connections to open to a single host. Defaults to 5.
+         */
+        public var maxConnectionsPerHost: UInt = 5u
```
aajtodd, line 30: Missed this in my review...

Question/Correctness: Should this match `maxConnections` by default?

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
ianbotsf: From what I've observed, this is the maximum number of connections that will be left idle in the pool. When another connection would be returned to the pool and that would exceed the `maxIdleConnections`, the oldest connection is found and evicted.

Exposing a knob here might help with sporadic traffic. Keeping a large reserve of idle connections may accelerate rapid bursts of activity after periods of inactivity. That said, it's hyper-specific. I vote for not exposing this knob for now and revisiting if we get feedback about it.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngineConfig.kt
ianbotsf: I didn't see an equivalent setting for CRT engine so I added it to OkHttp specifically. If you know of a CRT setting that maps (or if we're okay with this being ignored by CRT) then I have no problem relocating it.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngineConfig.kt
ianbotsf: It could. I have no feeling for sensible defaults here and so just picked the value that was previously in effect from the `Dispatcher` anyway.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
aajtodd: Ya IDK

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngineConfig.kt
aajtodd: I think it makes sense as the default (`maxConnections` that is) because when users aren't sharing an HTTP engine then the setting actually makes sense (most services will use a single host with S3 being a notable exception). 

Otherwise if a user configures `maxConnections` but not `maxConnectionsPerHost` they will actually be limited by the latter perhaps without realizing it. I think `maxConnectionsPerHost` probably makes the most sense to tune when you are sharing an engine across service clients.

Thoughts?

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngineConfig.kt
aajtodd: I think we can reasonably support this in CRT.

Connection managers are [already per/host in CRT engine](https://github.com/awslabs/smithy-kotlin/blob/66e95fb45a9a508677eac67dcbd210820eb4dea7/runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt#L83). Would we just want to take either the max/min of `maxConnections` and `maxConnectionsPerHost` for CRT?

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngineConfig.kt
ianbotsf: That makes a lot of sense actually. I'll update this to a nullable field in the builder and default it to `maxConnections` in the config initializer.

lauzadis (approved):

PR #799, opened 2023-02-09
feat: Add `readFully` extension method to `SdkSource` which reads an exact amount of bytes

lauzadis:
This PR adds a `readFully` extension method to `SdkSource`, which is used to read an exact amount of bytes
## Issue \#

Helps solve https://github.com/awslabs/aws-sdk-kotlin/issues/836

## Description of changes

This change is required because `SdkSource` has no method to read an exact amount of bytes. Before this PR, users of `SdkSource` must use `read` in a loop to ensure they read the number of bytes they expected. When this loop is not used it could result in reading fewer bytes than expected and this unchecked behavior could cause issues.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -44,4 +44,4 @@ kotlinLoggingVersion=3.0.0
 slf4jVersion=2.0.6
 
 # crt
-crtKotlinVersion=0.6.8-SNAPSHOT
+crtKotlinVersion=0.6.9-SNAPSHOT
--- a/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/io/SdkSource.kt
+++ b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/io/SdkSource.kt
@@ -70,7 +70,7 @@ public fun SdkSource.readFully(sink: SdkBuffer, byteCount: Long) {
         val rc = read(sink, byteCount - totalBytesRead)
 
         if (rc == -1L) {
-            throw EOFException("Unexpected EOF: expected ${byteCount - rc} more bytes; consumed: $rc")
+            throw EOFException("Unexpected EOF: expected ${byteCount - totalBytesRead} more bytes; consumed: $totalBytesRead")
         }
 
         totalBytesRead += rc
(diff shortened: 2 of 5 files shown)


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/io/SdkSource.kt
```diff
@@ -53,3 +53,13 @@ public expect suspend fun SdkSource.readToByteArray(): ByteArray
...
+
+/**
+ * Remove exactly [byteCount] bytes from this source and appends them to [sink].
+ * @param sink The sink to append bytes to
+ * @param byteCount the number of bytes to read from the source
+ * @throws [IllegalArgumentException] when [byteCount] is less than zero
+ * @throws [EOFException] when the source is exhausted before [byteCount] bytes could be read
+ */
+@Throws(IOException::class)
+public expect fun SdkSource.readFully(sink: SdkBuffer, byteCount: Long)
```
aajtodd, line 65: This doesn't need to be expect actual I don't think. It looks like it can be in common

aajtodd (approved):

gradle.properties
```diff
@@ -44,4 +44,4 @@ kotlinLoggingVersion=3.0.0
 slf4jVersion=2.0.6
 
 # crt
-crtKotlinVersion=0.6.8-SNAPSHOT
+crtKotlinVersion=0.6.9-SNAPSHOT
```
ianbotsf, line 47: **Question**: Is this change related or a mistaken commit?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/io/SdkSource.kt
```diff
@@ -53,3 +53,26 @@ public expect suspend fun SdkSource.readToByteArray(): ByteArray
...
+public fun SdkSource.readFully(sink: SdkBuffer, byteCount: Long) {
+    require(byteCount >= 0L) { "Invalid length ($byteCount) must be >= 0L" }
+
+    var totalBytesRead = 0L
+    while (totalBytesRead != byteCount) {
+        val rc = read(sink, byteCount - totalBytesRead)
+
+        if (rc == -1L) {
+            throw EOFException("Unexpected EOF: expected ${byteCount - rc} more bytes; consumed: $rc")
+        }
```
ianbotsf, line 74: **Correctness**: `rc` shouldn't be part of the error messages in either expected or consumed counts since, at this point in the code, it's `-1`. You probably want `totalBytesRead` in both places.

ianbotsf (changes requested):

ianbotsf (approved):

PR #810, opened 2023-02-23
fix: correct token bucket delays when experiencing system time jumps

### LINKED ISSUE #805: When system time is incorrect and in the past coroutine can be suspended for a long period of time
## Describe the bug
If system time rolls back (ie it was incorrect and then set back to correct value) `capacity` will have negative value and `delayMs` can be potentially as big as time difference between `lastTimestamp ` and current system time.

https://github.com/awslabs/smithy-kotlin/blob/7babb8883133b35d2f81c6b51cebc3c03773775f/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/StandardRetryTokenBucket.kt#L67

## Expected Behavior
Coroutine should not suspend longer than for `options.maxCapacity`.

## Current Behavior
Coroutine get suspended longer than for `options.maxCapacity`.

## Steps to Reproduce
1. Set system time 1 day in the future
2. Create a queue using `SqsClient.createQueue()`. It should return `SignatureDoesNotMatch` error
3. Reset time back to correct value
4. Make another call to `SqsClient.createQueue()`. It will suspend for `86335800` milliseconds

## Possible Solution
Wrap calculation of `refillMs` with `abs` in `StandardRetryTokenBucket`:

```kotlin
    private fun refillCapacity() {
        val refillMs = abs(now() - lastTimestamp)
        val refillSize = floor(options.refillUnitsPerSecond.toDouble() / MS_PER_S * refillMs).toInt()
        capacity = min(options.maxCapacity, capacity + refillSize)
    }
```

## Context
If time on device is wrong we can receive `SignatureDoesNotMatch` error from SQS and ask user to set time correctly. But when user changes time back there's no way to see that there's an issue as coroutine have got suspended and no exceptions being thrown.

## Your Environment
Smithy Kotlin version - 0.17.0-beta
Operating system - Android 12

ianbotsf:
## Issue \#

Closes #805 

## Description of changes

As noted in #805 the retry token bucket calculation is susceptible to changes in system time. Replace the `Clock` implementation with a Kotlin `TimeSource.Monotonic`, which offers intra-moment comparison capable of withstanding system time changes.

### DIFF
--- a/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/StandardRetryTokenBucket.kt
+++ b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/StandardRetryTokenBucket.kt
@@ -6,30 +6,31 @@
 package aws.smithy.kotlin.runtime.retries.delay
 
 import aws.smithy.kotlin.runtime.retries.policy.RetryErrorType
-import aws.smithy.kotlin.runtime.time.Clock
-import aws.smithy.kotlin.runtime.time.epochMilliseconds
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlin.math.ceil
 import kotlin.math.floor
 import kotlin.math.min
+import kotlin.time.ExperimentalTime
+import kotlin.time.TimeSource
 
 private const val MS_PER_S = 1_000
 
 /**
  * The standard implementation of a [RetryTokenBucket].
  * @param options The configuration to use for this bucket.
- * @param clock A clock to use for time calculations.
+ * @param timeSource A monotonic time source to use for calculating the temporal token fill of the bucket.
  */
-public class StandardRetryTokenBucket(
+@OptIn(ExperimentalTime::class)
+public class StandardRetryTokenBucket constructor(
     public val options: StandardRetryTokenBucketOptions = StandardRetryTokenBucketOptions.Default,
-    private val clock: Clock = Clock.System,
+    private val timeSource: TimeSource = TimeSource.Monotonic,
 ) : RetryTokenBucket {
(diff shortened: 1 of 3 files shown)


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/StandardRetryTokenBucket.kt
```diff
@@ -6,30 +6,31 @@
...
 private const val MS_PER_S = 1_000
 
 /**
  * The standard implementation of a [RetryTokenBucket].
  * @param options The configuration to use for this bucket.
- * @param clock A clock to use for time calculations.
+ * @param timeSource A monotonic time source to use for calculating the temporal token fill of the bucket.
  */
-public class StandardRetryTokenBucket(
+@OptIn(ExperimentalTime::class)
```
aajtodd, line 25: We've gotten ourselves in trouble in the past by using experimental stdlib features (https://github.com/awslabs/aws-sdk-kotlin/issues/622, https://github.com/awslabs/aws-sdk-kotlin/issues/736). I don't have a better suggestion here but something to consider w.r.t any versioning policy/strategy we define.

aajtodd (approved):

lauzadis (approved):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/StandardRetryTokenBucket.kt
ianbotsf: Discussed offline and decided to keep usage of this type for now, despite it being marked as experimental. I've added a task [aws-sdk-kotlin#860](https://github.com/awslabs/aws-sdk-kotlin/issues/860) to track revisiting all usages of experimental/unstable APIs before GA.

PR #816, opened 2023-03-07
fix: use service's `signingName` instead of `serviceId` for presigning

lauzadis:
Presigned URLs are being incorrectly signed with `serviceId` instead of `signingName` 

## Issue \#
https://github.com/awslabs/aws-sdk-kotlin/issues/868

## Description of changes
We are currently using the service's `serviceId` for the signing context, but this should actually be the `signingName`. For example, S3's `serviceId` is "S3" but the `signingName` is "s3". 

This leads to a signature mismatch / other errors depending on how signature mismatch is handled on the service-side.

Currently the issue only affects presigned URLs generated from a [presigned config object](https://github.com/awslabs/aws-sdk-kotlin/blob/main/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/PresignerGenerator.kt#L360), not [a client config](https://github.com/awslabs/aws-sdk-kotlin/blob/main/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/PresignerGenerator.kt#L312). 

---

### DIFF
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/Presigner.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/Presigner.kt
@@ -115,6 +115,7 @@ public suspend fun createPresignedRequest(
     serviceConfig: ServicePresignConfig,
     requestConfig: PresignedRequestConfig,
 ): HttpRequest {
+    // TODO re-evaluate whether SigningContext is a necessary concept: https://github.com/awslabs/smithy-kotlin/issues/755
     val givenSigningContext = SigningContext(serviceConfig.signingName, serviceConfig.region)
     val endpoint = serviceConfig.endpointProvider(givenSigningContext)
     val signatureType = when (requestConfig.presigningLocation) {
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/Presigner.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/Presigner.kt
@@ -115,7 +115,7 @@ public suspend fun createPresignedRequest(
     serviceConfig: ServicePresignConfig,
     requestConfig: PresignedRequestConfig,
 ): HttpRequest {
-    val givenSigningContext = SigningContext(serviceConfig.serviceId, serviceConfig.region)
+    val givenSigningContext = SigningContext(serviceConfig.signingName, serviceConfig.region)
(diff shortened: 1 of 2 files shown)


runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/Presigner.kt
```diff
@@ -115,7 +115,7 @@ public suspend fun createPresignedRequest(
     serviceConfig: ServicePresignConfig,
     requestConfig: PresignedRequestConfig,
 ): HttpRequest {
-    val givenSigningContext = SigningContext(serviceConfig.serviceId, serviceConfig.region)
+    val givenSigningContext = SigningContext(serviceConfig.signingName, serviceConfig.region)
     val endpoint = serviceConfig.endpointProvider(givenSigningContext)
```
ianbotsf, line 119: **Question**: Why should we pass the _signing name_ to the endpoint resolver? If we don't need service ID here then where do we need it?

ianbotsf (commented):
**Question**: Why is our [Polly presigner test not failing](https://github.com/awslabs/aws-sdk-kotlin/blob/b39d4702f83b65c69a944a89404bb28734840235/services/polly/e2eTest/src/PollyPresignerTest.kt#L27-L45)?

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/Presigner.kt
lauzadis: [This is the line](https://github.com/awslabs/smithy-kotlin/pull/816/files#diff-6de192aa05351178bc825d783da83b16a586059df26763343f277c1bebf326feR129) which made me think we should pass `signingName` to the endpoint resolver. On line 129 we are using the resolved endpoint's `.context?.service` field as a signing name.

lauzadis:
> **Question**: Why is our [Polly presigner test not failing](https://github.com/awslabs/aws-sdk-kotlin/blob/b39d4702f83b65c69a944a89404bb28734840235/services/polly/e2eTest/src/PollyPresignerTest.kt#L27-L45)?

Because it creates a presigned request using `client.config` rather than a `PollyPresignConfig`. I'll add a test for this other `.presign()` function variant

lauzadis:
I added a [test here](https://github.com/awslabs/aws-sdk-kotlin/pull/870)

aajtodd (approved):

ianbotsf (approved):

Issue #818, opened 2023-03-08
Missing Content-Length header on empty payloads with CrtHttpEngine


abhimanyuchugh:
## Describe the bug

S3 API seems to always expect `Content-Length` header in the http request, however when the http request is generated by the `CrtHttpEngine`, it explicitly excludes this header for empty payloads. We use empty payloads to upload a placeholder file to S3, but with the `CrtHttpEngine`, we are not able to do this. This works fine with the default `OkHttpEngine` which does include the zero `Content-Length` header for empty payloads.

This behaviour was explicitly added in this commit (https://github.com/awslabs/smithy-kotlin/commit/ed4324590d37f19ffebab06eade776ea2ab61c72) however the linked PR doesn't match the commit change, so not sure of the context behind the change.

## Expected Behavior

Include `Content-Length=0` HTTP header for empty payloads in CrtHttpRequests

## Current Behavior

`Content-Length` header is explicitly excluded from the CrtHttpRequest per https://github.com/awslabs/smithy-kotlin/commit/ed4324590d37f19ffebab06eade776ea2ab61c72

Stack trace:
```
2023-03-06T15:39:47,840+00:00 [pool-4-thread-1] DEBUG httpTraceMiddleware - sdkRequestId: 71776c3b-e991-4e56-a9cf-ec7417d59de7; service: S3; operation: PutObject; - HttpRequest:
PUT /s3-path?x-id=PutObject
Host: s3.eu-west-1.amazonaws.com
amz-sdk-invocation-id: ...
amz-sdk-request: attempt=1; max=3
User-Agent: aws-sdk-kotlin/0.17.12-beta os/linux/5.4.228-131.415.amzn2.x86_64 lang/kotlin/1.8.10 
x-amz-user-agent: aws-sdk-kotlin/0.17.12-beta
X-Amz-Content-Sha256: ...
X-Amz-Date: 20230306T153947Z
X-Amz-Security-Token: ...
Authorization: ...

2023-03-06T15:39:47,909+00:00 [pool-4-thread-1] DEBUG httpTraceMiddleware - sdkRequestId: 71776c3b-e991-4e56-a9cf-ec7417d59de7; service: S3; operation: PutObject; - HttpResponse:
HTTP 411: Length Required
x-amz-request-id: ...
x-amz-id-2: ...
Content-Type: application/xml
Transfer-Encoding: chunked
Date: Mon, 06 Mar 2023 15:39:47 GMT
Server: AmazonS3
[...]

PR #819, opened 2023-03-09
fix: allow `Content-Length=0` headers in the CRT HTTP engine

lauzadis:
Reverts [a previous commit](https://github.com/awslabs/smithy-kotlin/commit/ed4324590d37f19ffebab06eade776ea2ab61c72) which disallowed sending requests with zero `Content-Length` in the CRT HTTP engine. 

## Issue \#

#818 

## Description of changes

This change is required to allow sending `Content-Length=0` headers in the CRT engine, as it's done in the OkHttp engine.

### DIFF
--- a/runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/RequestUtil.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/RequestUtil.kt
@@ -60,11 +60,7 @@ internal fun HttpRequest.toCrtRequest(callContext: CoroutineContext): aws.sdk.ko
         headers.forEach { key, values -> appendAll(key, values) }
     }
 
-    val bodyLen = body.contentLength
-    val contentLength = when {
-        bodyLen != null -> if (bodyLen > 0) bodyLen.toString() else null
-        else -> headers[CONTENT_LENGTH_HEADER]
-    }
+    val contentLength = body.contentLength?.toString() ?: headers[CONTENT_LENGTH_HEADER]
     contentLength?.let { crtHeaders.append(CONTENT_LENGTH_HEADER, it) }
 
     return aws.sdk.kotlin.crt.http.HttpRequest(method.name, url.encodedPath, crtHeaders.build(), bodyStream)
--- a/runtime/protocol/http-client-engines/http-client-engine-crt/common/test/aws/smithy/kotlin/runtime/http/engine/crt/RequestConversionTest.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-crt/common/test/aws/smithy/kotlin/runtime/http/engine/crt/RequestConversionTest.kt
@@ -92,6 +92,20 @@ class RequestConversionTest {
     }
 
     @Test
+    fun testEngineAddsContentLengthHeaderForEmptyBody() {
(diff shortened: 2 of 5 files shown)


runtime/protocol/http-client-engines/http-client-engine-crt/common/test/aws/smithy/kotlin/runtime/http/engine/crt/RequestConversionTest.kt
```diff
@@ -91,20 +91,6 @@ class RequestConversionTest {
...
-            Url.parse("https://test.aws.com?foo=bar"),
-            Headers.Empty,
-            HttpBody.Empty,
-        )
-
-        val testContext = EmptyCoroutineContext + Job()
-        val crtRequest = request.toCrtRequest(testContext)
-        assertFalse(crtRequest.headers.contains("Content-Length"))
-    }
-
```
ianbotsf, line 107: **Suggestion**: I recommend keeping this test but inverting the expectation and the name.

runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/RequestUtil.kt
```diff
@@ -60,11 +60,7 @@ internal fun HttpRequest.toCrtRequest(callContext: CoroutineContext): aws.sdk.ko
         headers.forEach { key, values -> appendAll(key, values) }
     }
 
-    val bodyLen = body.contentLength
-    val contentLength = when {
-        bodyLen != null -> if (bodyLen > 0) bodyLen.toString() else null
-        else -> headers[CONTENT_LENGTH_HEADER]
-    }
+    val contentLength = body.contentLength?.toString() ?: headers[CONTENT_LENGTH_HEADER]
```
ianbotsf, line 63: **Question**: This is interesting...for CRT we fallback to an explicitly-provided `Content-length` header if one was provided.

It looks like OkHttp only derives content length from `RequestBody.contentLength`. If `RequestBody.contentLength == -1` (unknown), it [explicitly removes any previously-set `Content-length` header](https://github.com/square/okhttp/blob/ee1ba4656d80ed62485387217271c7fa31dc888e/okhttp/src/jvmMain/kotlin/okhttp3/internal/http/BridgeInterceptor.kt#L52-L55). Should we be falling back to the user-set `Content-length` header in OkHttp for cases where body length is unspecified?

lauzadis: I think we should. I checked if sending a `Content-Length` would conflict with chunked requests, but it's purely additive and *shouldn't* clash with chunked requests according to our internal spec. 

I can't imagine a case where the fallback behavior would be harmful

aajtodd: What if they set content length in an interceptor before becoming a chunked request?

lauzadis: I think that's still fine. A chunked request can have a content length header

ianbotsf (approved):

PR #821, opened 2023-03-22
feat: generated auth scheme resolver and refactor authentication execution

aajtodd:
## Issue \#

N/A

## Description of changes

The main changes in this PR are generated versions of the auth scheme resolver related interfaces + refactor/wireup of those components. 

* Generate (internal) implementations of:
    * `AuthSchemeProvider` - service client specific auth scheme provider interfaces
    * `DefaultAuthSchemeProvider` - default implementation of `AuthSchemeProvider`
    * `AuthSchemeParameters` - input parameters for generated auth scheme resolver
    * `IdentityProviderConfigAdapter` - adapts the generated client config to `IdentityProviderConfig`
* Removes existing `SdkOperationExecution.signer` in favor of `OperationAuthConfig` which is a tuple of (`authSchemeResolver`, `configuredAuthSchemes`, `identityProviderConfig`)

Auth schemes are modeled as being provided via integration with codegen delegating to each `AuthSchemeHandler` to deal with parts specific to each auth scheme (e.g. instantiation of the scheme, mapping identity provider, etc).

There are still quite a few TODO's/FIXME's or just plain unknowns. For now these generated components are internal, future PR will investigate what/how a customer will interact with this (which we may delay exposing depending).

For now I've generated some of these components as `object`'s since we don't 100% know what they will look like. Their relationship and use will remain largely unchanged though. Expect some refactoring.


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/AuthSchemeProviderAdapterGenerator.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    fun render(ctx: ProtocolGenerator.GenerationContext) {
+        val symbol = getSymbol(ctx.settings)
+        ctx.delegator.useSymbolWriter(symbol) { writer ->
+            // TODO - auth parameters will need bound per/request as applicable (e.g. like EP2.0 or generate one per/request).
+            //        This is a simplified version (using object) while design is in flux.
+            writer.withBlock("internal object #T: #T {", "}", symbol, RuntimeTypes.HttpClient.Operation.AuthSchemeResolver) {
+
+                withBlock(
+                    "override suspend fun resolve(request: #T): List<#T> {",
```
lucix-aws, line 36: Won't we potentially need to pass request input / client config as well? As we do in endpoint resolution.

lucix-aws (approved):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/Sigv4AuthSchemeIntegration.kt
```diff
@@ -0,0 +1,97 @@
...
+    ) {
+        val expr = if (op?.hasTrait<UnsignedPayloadTrait>() == true) {
+            "sigv4(unsignedPayload = true)"
+        }else {
+            "sigv4()"
+        }
+        writer.write(expr)
+    }
+
+    // FIXME: Move to runtime?
```
lauzadis, line 75: +1 it seems like this could be shared by all services by placing in the runtime

lauzadis (approved):

PR #827, opened 2023-04-04
feat: add intEnum support

### LINKED ISSUE #752: Add support for IntEnum
Implement support for [intEnum](https://smithy.io/2.0/spec/simple-types.html#intenum).

This should be relatively easy as we used sealed class rather than (kotlin) enum to implement enum support in preparation for eventually supporting other value types.

lucix-aws:
## Issue \#
Closes #752

## Description of changes
* Add smithy `intEnum` support (codegen + all serde/binding points)
* Modernize `EnumGenerator` codegen
* Upgrade smithy to pull in new protocol tests, which triggered a few incidental changes:
  * update test endpoint ruleset to pass new, more-strict validation
  * `ShapeValueGenerator` (protocol tests)
    * handle fractional values in timestamps
    * ensure list member symbols are imported

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
@@ -156,9 +156,14 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Unit>() {
     }
 
     override fun stringShape(shape: StringShape) {
+        // smithy will present both strings with legacy enum trait AND explicit (non-int) enum shapes in this manner
         if (shape.hasTrait<@Suppress("DEPRECATION") software.amazon.smithy.model.traits.EnumTrait>()) {
             writers.useShapeWriter(shape) { EnumGenerator(shape, symbolProvider.toSymbol(shape), it).render() }
         }
+    }
+
+    override fun intEnumShape(shape: IntEnumShape) {
+        writers.useShapeWriter(shape) { EnumGenerator(shape, symbolProvider.toSymbol(shape), it).render() }
     }
 
     override fun unionShape(shape: UnionShape) {
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/Naming.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/Naming.kt
@@ -59,17 +59,12 @@ private fun String.sanitizeClientName(): String =
 fun clientName(raw: String): String = raw.sanitizeClientName().toPascalCase()
 
 /**
- * Get the (un-validated) name of an enum variant from the trait definition
+ * Get the (un-validated) name of an enum variant.
+ *
+ * This value can come from an enum definition trait, or it could be a member name from an explicit enum shape.
  */
-fun EnumDefinition.variantName(): String {
-    val identifier = name.orElseGet {
-        // we don't want to be doing this...name your enums people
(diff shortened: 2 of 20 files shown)


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/Naming.kt
lauzadis, line 1: GitHub wouldn't let me comment on the specific line, but L20 is an unused import now

lauzadis (approved):
Looks good! I had a few clarifying questions

codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/EnumGeneratorTest.kt
```diff
@@ -153,18 +152,94 @@ public sealed class Baz {
...
+     * Documentation for this value
+     */
+    public object T2Micro : test.model.Baz() {
+        override val value: kotlin.Int = 1
+        override fun toString(): kotlin.String = value.toString()
+    }
+
+    public object T2Nano : test.model.Baz() {
+        override val value: kotlin.Int = 2
+        override fun toString(): kotlin.String = value.toString()
```
aajtodd, line 213: **Question**: Is this the `toString()` representation we want for this? I know we use the value for string enums as the `toString()` representation. I suppose that is also questionable (although I think we did that for serialization purposes but I don't recall for sure at this point).

aajtodd (approved):

codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/EnumGeneratorTest.kt
lucix-aws: Probably not. It mattered much less for strings, since the name and value were (generally) the same.

Seems perhaps we should commonize to `EnumSymbol.VariantSymbol`?  e.g.
```kt
override fun toString(): kotlin.String = "Baz.T2Nano"
```

Note that for serialization (both string and int enums) we use the variant value directly.

ianbotsf: Strictly the variant symbol (and _not_ the enum symbol plus `.`) is most consistent with how Kotlin enums are handled.

ianbotsf (approved):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
```diff
@@ -156,11 +156,16 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Unit>() {
...
+        // smithy will present both strings with legacy enum trait AND explicit (non-int) enum shapes in this manner
         if (shape.hasTrait<@Suppress("DEPRECATION") software.amazon.smithy.model.traits.EnumTrait>()) {
             writers.useShapeWriter(shape) { EnumGenerator(shape, symbolProvider.toSymbol(shape), it).render() }
         }
     }
 
+    override fun intEnumShape(shape: IntEnumShape) {
+        writers.useShapeWriter(shape) { EnumGenerator(shape, symbolProvider.toSymbol(shape), it).render() }
+    }
+
```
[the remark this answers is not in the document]
lucix-aws: You can't actually apply the old enum trait to an `integer` shape, the only way to get int enums is to be on IDL2 + `intEnum`.

ianbotsf (approved):

PR #831, opened 2023-04-13
feat!: rename `SdkLogMode` to `LogMode` and enable its resolution from environment

lauzadis:
This PR has some changes to enable resolving `LogMode` from environment. It adds an `equals()` method to `LogMode` which is used to compare the `mask` values, as that's what determines log mode. It also adds a `fromString()` method which is used to parse both single and composite `LogMode` values from a string.

## Issue \#

https://github.com/awslabs/aws-sdk-kotlin/issues/432

## Description of changes

This change is required to support resolving `LogMode` from a user's environment.


runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/SdkLogMode.kt
```diff
@@ -55,6 +55,8 @@ public sealed class SdkLogMode(private val mask: Int) {
     public operator fun plus(mode: SdkLogMode): SdkLogMode = Composite(mask or mode.mask)
     public operator fun minus(mode: SdkLogMode): SdkLogMode = Composite(mask and mode.mask.inv())
 
+    public override fun equals(other: Any?): Boolean = (other is SdkLogMode) and (mask == (other as SdkLogMode).mask)
```
aajtodd, line 58: **fix**: `(other is SdkLogMode) && (mask == (other as SdkLogMode.mask)`

aajtodd (changes requested):

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/SdkLogMode.kt
```diff
@@ -70,6 +72,15 @@ public sealed class SdkLogMode(private val mask: Int) {
...
         )
+
+        /**
+         * Parse an [SdkLogMode] from a String
+         * @return the parsed SdkLogMode, or `null` if no log mode could be parsed
+         */
+        public fun fromString(string: String): SdkLogMode? =
+            string.trim().split("|").mapNotNull {
+                allModes().firstOrNull { sdkLogMode -> it.equals(sdkLogMode.toString(), ignoreCase = true) }
+            }.takeIf { it.isNotEmpty() }?.let { it.reduce { acc, sdkLogMode -> acc + sdkLogMode } }
```
ianbotsf, line 83: **Question**: It's strange to me that `fromString` parses a different format than `toString` produces. Should we unify the formats?

lauzadis: Yes I think so. I'd rather update the `toString` if nobody opposes that? The current `SdkLogMode(...)` is nice and readable, but it's not intuitive as a configuration value.

Alternatively I could update the `fromString` to support both formats: `SdkLogMode(abc)` and just the inner contents `abc`

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/SdkClientConfig.kt
```diff
@@ -61,7 +61,7 @@ public interface SdkClientConfig {
          * performance considerations when dumping the request/response body. This is primarily a tool for
          * debug purposes.
          */
-        public var sdkLogMode: SdkLogMode
+        public var sdkLogMode: SdkLogMode?
```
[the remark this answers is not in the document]
lauzadis: Sure, clients still won't output raw requests or responses by default, right? We just need to add documentation about the value getting resolved from environment variable and JVM system property

codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ServiceClientConfigGeneratorTest.kt
```diff
@@ -255,85 +253,20 @@ public class Config private constructor(builder: Builder) {
...
+    override val httpClientEngine: HttpClientEngine = builder.httpClientEngine ?: DefaultHttpEngine().manage()
+    override val authSchemes: kotlin.collections.List<aws.smithy.kotlin.runtime.http.auth.HttpAuthScheme> = builder.authSchemes
+    public val customProp: Int? = builder.customProp
+    public val endpointProvider: EndpointProvider = requireNotNull(builder.endpointProvider) { "endpointProvider is a required configuration property" }
+    override val idempotencyTokenProvider: IdempotencyTokenProvider = builder.idempotencyTokenProvider ?: IdempotencyTokenProvider.Default
+    override val interceptors: kotlin.collections.List<aws.smithy.kotlin.runtime.http.interceptors.HttpInterceptor> = builder.interceptors
+    override val retryPolicy: RetryPolicy<Any?> = builder.retryPolicy ?: StandardRetryPolicy.Default
+    override val retryStrategy: RetryStrategy = builder.retryStrategy ?: StandardRetryStrategy()
+    override val sdkLogMode: SdkLogMode = builder.sdkLogMode ?: SdkLogMode.LogRequest
+    override val tracer: Tracer = builder.tracer ?: DefaultTracer(LoggingTraceProbe, clientName)"""
```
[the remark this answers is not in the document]
lauzadis: Since I changed the `SdkLogMode` config property to use `ConfigPropertyType.RequiredWithDefault` instead of defining a `defaultValue` on the `symbol`, the builder field now always defaults to `null`

I tried playing around with this test-case for a while and I ended up with this modification

runtime/smithy-client/api/smithy-client.api
```diff
@@ -115,24 +155,20 @@ public abstract interface class aws/smithy/kotlin/runtime/client/SdkClient$Build
...
 	public abstract fun setClientName (Ljava/lang/String;)V
+	public abstract fun setLogMode (Laws/smithy/kotlin/runtime/client/LogMode;)V
 	public abstract fun setRetryPolicy (Laws/smithy/kotlin/runtime/retries/policy/RetryPolicy;)V
 	public abstract fun setRetryStrategy (Laws/smithy/kotlin/runtime/retries/RetryStrategy;)V
-	public abstract fun setSdkLogMode (Laws/smithy/kotlin/runtime/client/SdkLogMode;)V
-}
-
-public final class aws/smithy/kotlin/runtime/client/SdkClientConfig$DefaultImpls {
-	public static fun getSdkLogMode (Laws/smithy/kotlin/runtime/client/SdkClientConfig;)Laws/smithy/kotlin/runtime/client/SdkLogMode;
 }
```
ianbotsf, line 172: **Comment**: TIL `DefaultImpls` is yet another synthetic class the compiler adds.

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/SdkClientOption.kt
```diff
@@ -29,9 +29,9 @@ public object SdkClientOption {
     public val IdempotencyTokenProvider: ClientOption<IdempotencyTokenProvider> = ClientOption("IdempotencyTokenProvider")
 
     /**
-     * The client logging mode (see [SdkLogMode]
+     * The client logging mode (see [LogMode]
      */
```
ianbotsf, line 33: **Nit**: Missing closing parenthesis (`)`)

ianbotsf (changes requested):
Looking pretty close, just a few small issues left to fix.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/TestPlatformProvider.kt
```diff
@@ -0,0 +1,42 @@
...
+package aws.smithy.kotlin.runtime.util
+
+/**
+ * An implementation of [PlatformProvider] meant for testing
+ * @param env Environment variable mappings
+ * @param props System property mappings
+ * @param fs Filesystem path to content mappings
+ * @param os Operating system info to emulate
+ */
+public class TestPlatformProvider(
```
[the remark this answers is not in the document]
lauzadis: I tried this but realized it would create a circular dependency between **:runtime-core** and **:testing** because `TestPlatformProvider` needs `FileSystem`, `OperatingSystem`, and a few other things in **:runtime-core**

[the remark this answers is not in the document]
ianbotsf: Oh you're right. Well then let's make it `@InternalApi`.

ianbotsf (approved):

aajtodd (approved):

PR #835, opened 2023-04-18
refactor!: update tracing API

aajtodd:
As part of dog fooding the tracing API's I found it difficult to map our implementation to open telemetry and other tracing collection systems. This PR is the result of revamping some of the APIs to be more consistent with how trace data is consumed.

## Issue \#

n/A

## Description of changes

* Update `TraceSpan` to include `Attributes` of arbitrary key/name pairs that describe various data about the span.
    * Attributes are used by various trace aggregation systems to search for specific traces (e.g. "show me all traces with `Put` method and HTTP status of 400 is done by with a search like `http.method == "PUT" && http.status == 400`).
* Don't make `TraceSpan` implementation(s) responsible for span ID uniqueness properties.
* Track span name separate from span ID (many spans may have the same name as they describe the same logical function/method/invocation/etc). 
* Add a `traceId` property that identifies the root level "trace" a span belongs to. This is useful for correlating all span events in trace probes that belong to the same "trace".
* Add enum `TraceSpanStatus` which tracks whether a span is considered "failed" or not. This signal is consumed by tracing collection systems that sample traces to intelligently include "interesting traces" (i.e. you may be setup to sample 10% of traces, we want to ensure that those traces always include failures though). 
* Add a multi sink trace probe and update `DefaultTracer` to take a variadic number of probes by default
* Set common [Semantic Attributes](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/) like `rpc.method` and `rpc.service`. These come from open telemetry but seem fairly uncontroversial (e.g. `http.status_code` and `http.method` are semantic HTTP conventions). There are even ones for AWS SDK's defined already.
* Add `spanCreated` event to `TraceProbe` API (`spanClosed` already existed this just adds the missing lifecycle that was needed to map to external systems)
* Refactor the coroutine extensions to not be off `coroutineContext`.
* Fix span hierarchy. The "wrapped" span construct was fundamentally flawed, all span construction needs to go through either a tracer or existing span `child(...)` API to appropriately trigger trace probe lifecycle events.
* Relocated generation of `ServiceId` and `SdkVersion` which were in `aws-sdk-kotlin` but are not SDK specific

## Questions

I'll try and summarize here but the code has a bunch of FIXME/TODO's in it.

1. Representation of `Attributes` being mutable vs exposed as immutable with setters to mutate.
2. Trace span metadata? Should it exist, should more be on it (e.g. span ID), or should we just move them to top level and/or attributes by convention?
3. Representation of log messages with attributes?
    * It might be nice to make this configurable on the probe?
4. `Tracer` is very limited in what it can do when creating a span (right now it can only set the name). I wonder if this shouldn't be more flexible to allow setting attributes, or parent, etc. We sort of abstract over this in coroutine utils.
5. There is still the possibility (as there was before) that `withSpan` coroutine utils create a child span instead of a root span that doesn't share the same set of probes. (FWIW I don't think this is all that concerning of a use case).
6. Should `TraceEvent` also be allowed to have attributes? Rust/OTeL tracing ecosystems seem to support this but I haven't yet figured out how it is used in practice.


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolClientGenerator.kt
```diff
@@ -245,18 +245,18 @@ abstract class HttpProtocolClientGenerator(
         val hasOutputStream = outputShape.map { it.hasStreamingMember(ctx.model) }.orElse(false)
         val inputVariableName = if (inputShape.isPresent) "input" else KotlinTypes.Unit.fullName
 
-        writer
-            .write(
-                """val rootSpan = config.tracer.createRootSpan("#L-${'$'}{op.context.#T}")""",
-                op.id.name,
-                RuntimeTypes.HttpClient.Operation.sdkRequestId,
-            )
```
ianbotsf, line 253: **Question**: What is the role of `sdkRequestId` now? The only other place I see it used is in the `amz-sdk-request` header but it's now disconnected from tracing on the client side.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolClientGenerator.kt
```diff
@@ -245,18 +245,18 @@ abstract class HttpProtocolClientGenerator(
...
+                """return config.tracer.#T(#S) { span ->""",
                 "}",
-                RuntimeTypes.KotlinCoroutines.coroutineContext,
-                RuntimeTypes.Tracing.Core.withRootTraceSpan,
+                RuntimeTypes.Tracing.Core.withSpan,
+                op.id.name
             ) {
+                write("span.#T(#S, op.context.#T!!)", RuntimeTypes.Tracing.Core.setAttribute, "rpc.method", RuntimeTypes.SmithyClient.operationName)
+                write("span.#T(#S, ServiceId)", RuntimeTypes.Tracing.Core.setAttribute, "rpc.service")
+                // FIXME - set client name
```
ianbotsf, line 256: **Question**: Is it still your intent to allow setting a client name to disambiguate amongst multiple clients for the same service? How do you think that will be sent downstream to probes?

aajtodd:
Closing in favor of reworked API in `observability` branch

Issue #661, opened 2022-06-15
TLS Configuration


aajtodd:
(split into #820)

Add APIs to configure HTTP engine TLS context.

This may be at the common `HttpClientEngineConfig` level or at individual engine config level depending on what makes sense.

Possible knobs:
* min TLS version
* ~~CA certs (trust manager)~~ (handled in #820)
* ~~Cipher preferences~~ (handled in #820)
* ~~Certificate pinning~~ (handled in #820)
* ~~Host name verifier~~ (handled in #820)

Not every engine will support all of these or even if they do may make it difficult to map a common API to.

After these changes are made, coordinate updates to the Developer Guide to clarify how to configure TLS (especially min version).

PR #841, opened 2023-04-21
chore: refactor environment settings and retry settings out of aws-sdk-kotlin into smithy-kotlin

ianbotsf:
## Issue \#

Pre-requisite for #661 

## Description of changes

This change refactors environmental settings and moves the core logic from **aws-sdk-kotlin**'s `AwsSdkSetting` into a new **smithy-kotlin** `EnvironmentSetting`. It also moves retry configuration settings down to **smithy-kotlin** in a new `ClientSettings` namespace.

**Companion PR**: [aws-sdk-kotlin#904](https://github.com/awslabs/aws-sdk-kotlin/pull/904)

### DIFF
--- a/runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/ClientSettings.kt
+++ b/runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/ClientSettings.kt
@@ -1,3 +1,7 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
 package aws.smithy.kotlin.runtime.client.config
 
 import aws.smithy.kotlin.runtime.InternalApi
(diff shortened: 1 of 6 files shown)


lucix-aws (approved):

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/ClientSettings.kt
```diff
@@ -0,0 +1,20 @@
...
+    /**
+     * The maximum number of request attempts to perform. This is one more than the number of retries, so
+     * maxAttempts = 1 will have 0 retries.
+     */
+    public val MaxAttempts: EnvironmentSetting<Int> = intEnvSetting("SDK_MAX_ATTEMPTS", "sdk.maxAttempts")
+
+    /**
+     * Which RetryMode to use for the default RetryPolicy, when one is not specified at the client level.
+     */
+    public val RetryMode: EnvironmentSetting<RetryMode> = enumEnvSetting<RetryMode>("SDK_RETRY_MODE", "sdk.retryMode")
```
lauzadis, line 19: question: should we use the `KOTLIN_SDK` namespace (i.e `KOTLIN_SDK_RETRY_MODE`) to avoid conflicts?

lauzadis (approved):

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/ClientSettings.kt
ianbotsf: No, I think that would make common behavior among multiple generic SDKs difficult. For instance, if Rust or JavaScript eventually supported this setting in a non-AWS SDK then each language would have different setting names and clients wouldn't be able to set a single value to set the setting across SDK languages.

lauzadis (approved):

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/ClientSettings.kt
aajtodd: Maybe `aws.smithy` namespace and the `AWS_SMITHY_` environment variable prefix should be claimed/used.

Issue #820, opened 2023-03-10
Enhanced TLS configuration


ianbotsf:
(split from #661)

Add additional APIs to configure HTTP engine TLS context.

This may be at the common `HttpClientEngineConfig` level or at individual engine config level depending on what makes sense.

Possible knobs:
* ~~min TLS version~~ (handled in #661)
* CA certs (trust manager)
* Cipher preferences
* Certificate pinning
* Host name verifier

Not every engine will support all of these or even if they do may make it difficult to map a common API to.

After these changes are made, coordinate updates to the Developer Guide to clarify how to configure TLS.

Once certificate management is available, complete the HTTP test suite to verify that TLS min versions are working correctly on local test servers with self-signed certificates (see `TODO` in **ConnectionTest.kt**).

terrelewis2:
Hi @ianbotsf, just check if there is any updates on this. We have a similar requirement of being able to attach client certificates with the SQSClient. Given that is not possible currently, what is a good workaround to this?

terrelewis2:
Thanks @ianbotsf for the reply
For additional context we are actually implementing mTLS in our app which requires us to attach a client side key-certificate pair.
I see that there's an open [issue](https://github.com/awslabs/aws-sdk-kotlin/issues/1306) in the aws-sdk-kotlin library around the same as well

PR #844, opened 2023-04-26
feat: add HTTP engine config for min TLS version

### LINKED ISSUE #661: TLS Configuration

ianbotsf:
## Issue \#

Closes #661 

## Description of changes

This change adds a new `TlsContext` subgroup in generic HTTP engine config, to which I've added a new `minVersion` field for setting the minimum allowable TLS version. I've also moved `alpn` into the TLS context since it seems related. Finally, to prevent confusion or namespacing issues, I've removed `TlsContext` from the CRT-specific engine config.

### Testing

Unfortunately, testing this in an automated way is not yet possible. I've started some test harness implementation but won't be able to complete it until HTTP engines also support certification configuration, which is necessary to allow self-signed certificates for connecting to private HTTPS servers.

### Breaking changes

* Moves [`alpn` from `HttpClientEngineConfig`](https://github.com/awslabs/smithy-kotlin/blob/6d780cadfb67cdb42de60b2b073f21ad70f8fd50/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/HttpClientEngineConfig.kt#L65-L68) to new `TlsContext` class
* Removes CRT-specific [`TlsContext` on `CrtHttpEngineConfig`](https://github.com/awslabs/smithy-kotlin/blob/6d780cadfb67cdb42de60b2b073f21ad70f8fd50/runtime/protocol/http-client-engines/http-client-engine-crt/common/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngineConfig.kt#L38-L41)

See the [BREAKING: Streamlined TLS configuration](https://github.com/awslabs/aws-sdk-kotlin/discussions/909) post for more details.


runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
```diff
@@ -69,6 +71,8 @@ private fun OkHttpEngineConfig.buildClient(): OkHttpClient {
         followRedirects(false)
         followSslRedirects(false)
 
+        connectionSpecs(listOf(minTlsConnectionSpec(config.minTlsVersion), ConnectionSpec.CLEARTEXT))
```
aajtodd, line 74: **Question**: Why do we need to add `CLEARTEXT` here?

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/HttpClientEngineConfig.kt
```diff
@@ -72,6 +73,8 @@ public open class HttpClientEngineConfig constructor(builder: Builder) {
      */
     public val proxySelector: ProxySelector = builder.proxySelector
 
+    public val minTlsVersion: TlsVersion = builder.minTlsVersion
```
aajtodd, line 76: **Question**: Should we be forward looking and define something like `TlsContext` to group TLS settings?

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/ClientSettings.kt
```diff
@@ -17,6 +18,12 @@ public object ClientSettings {
      */
     public val MaxAttempts: EnvironmentSetting<Int> = intEnvSetting("sdk.maxAttempts", "SDK_MAX_ATTEMPTS")
 
+    /**
+     * Specifies the minimum acceptable version of TLS to use when connecting to service endpoints.
+     */
+    public val MinTlsVersion: EnvironmentSetting<TlsVersion> =
+        enumEnvSetting<TlsVersion>("SDK_MIN_TLS", "sdk.minTls").orElse(TlsVersion.Tls1_2)
```
aajtodd, line 25: **concern**: I think we really ought to use a namespace we own like `aws.smithy` or `aws.sdk` (I was late to other PR but left same comment for retries).

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
ianbotsf: We're replacing the default list of all acceptable connections, not just TLS ones. If we don't add `CLEARTEXT`, then HTTP connections will be impossible.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/HttpClientEngineConfig.kt
lucix-aws: Some thoughts here--

I would prefer it from an organization perspective, but if we go that route I think we'd want to reconsider grouping all of the connection based things (timeouts, max conns) as well.

I think that would be liable to complicate operation config override for these settings (which AFAIK we wanted to revisit being able to do independently of setting the HTTP client). e.g. Instead of patching one field in the config, I now have to construct a new container, merge in what i want to stay the same, and apply overrides.

Not taking a hard stance here, more just trying to play out the different paths forward.

lucix-aws (approved):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/config/TlsVersion.kt
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Describes a version of TLS
+ */
+@Suppress("ktlint:enum-entry-name-case")
+public enum class TlsVersion {
+    /**
+     * TLS version 1
+     */
+    Tls1_0,
```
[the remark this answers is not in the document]
ianbotsf: It appears we're inconsistent between SCREAMING_CASE (e.g., [`AlpnId`](https://github.com/awslabs/smithy-kotlin/blob/cd5759e4a34439c873863160179a0018e87cb6d9/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/HttpClientEngineConfig.kt#L155-L176), [`PresigningLocation`](https://github.com/awslabs/smithy-kotlin/blob/cd5759e4a34439c873863160179a0018e87cb6d9/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/Presigner.kt#L75-L85), [`TimestampFormat`](https://github.com/awslabs/smithy-kotlin/blob/cd5759e4a34439c873863160179a0018e87cb6d9/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/time/TimestampFormat.kt#L10-L40), etc.) and PascalCase (e.g., [`ErrorType`](https://github.com/awslabs/smithy-kotlin/blob/cd5759e4a34439c873863160179a0018e87cb6d9/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/Exceptions.kt#L138-L142), [`Phase`](https://github.com/awslabs/smithy-kotlin/blob/cd5759e4a34439c873863160179a0018e87cb6d9/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/io/middleware/Phase.kt#L19-L21), [`OsFamily`](https://github.com/awslabs/smithy-kotlin/blob/cd5759e4a34439c873863160179a0018e87cb6d9/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Platform.kt#L36-L43), etc.). Interestingly, even [Kotlin's own style guide allows either "depending on usage"](https://kotlinlang.org/docs/coding-conventions.html#names-for-backing-properties:~:text=For%20enum%20constants%2C%20it%27s%20OK%20to%20use%20either%20uppercase%20underscore%2Dseparated%20names%20(screaming%20snake%20case)%20(enum%20class%20Color%20%7B%20RED%2C%20GREEN%20%7D)%20or%20upper%20camel%20case%20names%2C%20depending%20on%20the%20usage.). I can switch this to SCREAMING_CASE since that matches [Kotlin's documented examples](https://kotlinlang.org/docs/enum-classes.html) but I wonder if we need to be more consistent throughout our project.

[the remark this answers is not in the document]
aajtodd: I'd say we probably should be. Probably even worth a lint 🤔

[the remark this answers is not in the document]
ianbotsf: Yeah, agreed on lint rules for whatever we decide. The default **ktlint** ruleset already verifies enum names although [it allows both styles permitted by JB's Kotlin guidelines](https://github.com/pinterest/ktlint/blob/d195ffde7d5160616900d4787f92f4d08e6e1aa4/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/EnumEntryNameCaseRule.kt#L15). You may notice that I suppressed this rule on the enum since I wrote it in PascalCase but needed `_` as a digit separator. Maybe I should've gone for `Tls1Dot1` (/s).

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/ClientSettings.kt
ianbotsf: I'm going to defer (re)naming this setting until standards exist for cross-SDK envvar/sysprop config naming conventions.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/HttpClientEngineConfig.kt
ianbotsf: `TlsContext` added and I've also moved the existing `alpn` setting into it alongside the new TLS minimum version.

Concerns about complications for overriding client config are most definitely valid but won't be solved in this PR. We'll tackle that in the near future with a larger overhaul.

runtime/protocol/http-client-engines/test-suite/common/test/aws/smithy/kotlin/runtime/http/test/ConnectionTest.kt
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.http.test
+
+// TODO Finish once we have HTTP engine support for client certificates
```
aajtodd, line 7: Add a note to the TLS configuration issue to finish this.

aajtodd (approved):
Looks great

runtime/protocol/http-client-engines/test-suite/common/test/aws/smithy/kotlin/runtime/http/test/ConnectionTest.kt
ianbotsf: Added note to #820.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/config/EnvironmentSetting.kt
```diff
@@ -41,9 +42,10 @@ public data class EnvironmentSetting<T>(
  * property first, falling back to the environment variable if necessary. If neither is set, it returns the setting's
  * default value.
  * @param platform The [PlatformEnvironProvider] to use for reading system properties and environment variables.
+ * Defaults to [PlatformProvider.System].
  */
 @InternalApi
-public fun <T> EnvironmentSetting<T>.resolve(platform: PlatformEnvironProvider): T? {
+public fun <T> EnvironmentSetting<T>.resolve(platform: PlatformEnvironProvider = PlatformProvider.System): T? {
```
lauzadis, line 48: Nice change here, in my LogMode PR I passed `PlatformProvider.System` but it's better to add a default

lauzadis (approved):

PR #848, opened 2023-04-28
fix: correct bad waiter codegen caused by dropped projection scope

lucix-aws:
## Description of changes
* Correct typo when checking comparison nullability
* Fix misused shape cursor when traversing subexpressions
* Only rewind shape cursor for disparate subexpressions (i.e. ones that have no bearing on the resulting scope, such as both sides of a comparison)
* Correct null guarding behavior on flat map expressions
* Retain context of inner shape when performing projections, such that the shape cursor is not lost when indexing back into projected shape members


aajtodd (approved):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -57,10 +57,14 @@ class KotlinJmespathExpressionVisitor(
...
 
+    // traverses an independent expression (one whose resolved scope does not persist in the outer evaluation)
     private fun acceptSubexpression(expr: JmespathExpression): VisitedExpression {
-        shapeCursor.addLast(currentShape)
+        val pos = shapeCursor.size
         val out = expr.accept(this)
-        shapeCursor.removeLast()
+
+        val diff = shapeCursor.size - pos
+        repeat(diff) { shapeCursor.removeLast() } // reset the shape cursor
```
ianbotsf, line 66: **Question**: Are some of our `visit*` methods putting something on the stack and failing to remove it at the end? Should we be wrapping stack pushes such that they're always definitively popped at the end of a scope?

```kotlin
override fun visitFoo(expr: FooExpression): VisitedExpression {
    val shape = shapeForFoo(expr)
    return shapeCursor.withPush(shape) { // push
        // do some things for foo codegen
        VisitedExpression(foo, bar)
    } // pop
}
```

runtime/protocol/http-client/api/http-client.api
```diff
@@ -142,6 +142,9 @@ public final class aws/smithy/kotlin/runtime/http/middleware/HttpResponseExcepti
 	public final fun setStatusCode (Laws/smithy/kotlin/runtime/http/HttpStatusCode;)V
 }
 
+public final class aws/smithy/kotlin/runtime/http/middleware/ResolveEndpointKt {
+}
+
```
ianbotsf, line 147: **Question**: Why is this here?

lucix-aws: It appeared to be a missed api snapshot update, i'm not sure why things downstream were building without it.

ianbotsf: That doesn't seem right. There's no **ResolveEndpoint.kt** file in [**aws/smithy/kotlin/runtime/http/middleware**](https://github.com/awslabs/smithy-kotlin/tree/main/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/middleware) and this PR doesn't add one. CI is failing too. Is your local copy combining two changes?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
lucix-aws: The visitors aren't really written to be "self-popping" per se, inner expressions are generally opaque to us, so when you're visiting some expression, you don't always have context on whether you're safe to pop back after visiting some given sub-expression. Some binary expressions will have sub-tree on the left where we very much need the modified scope to stay in context when we go to the right (i.e. projections).

What we _do_ know without tree introspection is when some sub-tree has isolated scope, so we can jump into it, let it do its business, and then rewind.

I think it's feasible to have implemented it in the way you're describing, though I think there'd have been more code to have expressions pop their own state off when done based on surrounding context.

runtime/protocol/http-client/api/http-client.api
lucix-aws: Yeah you're right. Looks like it was a build artifact in my working copy. Cleaned, re-built, it cleared out of the snapshot -- pushed.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
ianbotsf: OK, we can leave it as is for now.

ianbotsf (approved):

PR #854, opened 2023-05-19
feat: redesign presigning API

### LINKED ISSUE #755: Revisit Presign Request Design
We've had a few issues with presign requests. We should take another hard look at our presign API to ensure we are ready to commit to it.

ianbotsf:
## Issue \#

Closes #755 

## Description of changes

This change reworks presigning APIs for SigV4 services. It reworks the backing runtime method used by all codegenned presigners to be simpler and easier to invoke from codegen. The following small changes are part of the overall rework:
* **Breaking**: Remove `PresignedRequestConfig`, `PresigningLocation`, `ServicePresignConfig`, `SigningContext`, `SigningContextualizedEndpoint`, and `SigningEndpointProvider`. They're unnecessary now that all presigning uses `AwsSigningConfig` generated by the presigners.
* Rework/simplify the backing runtime method `presignRequest` (previously `createPresignedRequest`) used by codegenned presigners
* Add new `NoHttpEngine` HTTP engine. This will be useful for callers who need to presign and thus require a service client but who do not want to spin up an actual HTTP instance because **a)** they don't need to, they have local credentials and **b)** it's expensive.
* Move `AuthScheme`, `AuthSchemesAttributeKey`, and `Endpoint.authScheme` from **aws-sdk-kotlin** (previously in [**AuthScheme.kt**](https://github.com/awslabs/aws-sdk-kotlin/blob/65fa003ea69803b1dc741fc9e298e3925a6fad01/aws-runtime/aws-endpoint/common/src/aws/sdk/kotlin/runtime/endpoint/AuthScheme.kt))

**Companion PR**: [aws-sdk-kotlin#921](https://github.com/awslabs/aws-sdk-kotlin/pull/921)


runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/NoHttpEngine.kt
```diff
@@ -0,0 +1,28 @@
...
+import aws.smithy.kotlin.runtime.http.request.HttpRequest
+import aws.smithy.kotlin.runtime.http.response.HttpCall
+import aws.smithy.kotlin.runtime.operation.ExecutionContext
+
+/**
+ * Provides an implementation of an [HttpClientEngine] which is guaranteed never to make actual network calls. (Attempts
+ * to make network calls result in an exception.) Use this when you need to configure a service client, but you are
+ * certain no HTTP engine is actually needed.
+ */
+public class NoHttpEngine : HttpClientEngineBase("none") {
```
aajtodd, line 17: **Question**: I know we discussed this as an option but I didn't necessarily think we would own it/provide it. I imagined it more as an escape hatch available if absolutely needed. Is this something we actually need to provide out of the box?

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/endpoints/AuthScheme.kt
```diff
@@ -0,0 +1,44 @@
...
+ * Static attribute key for AWS endpoint auth schemes.
+ */
+@InternalApi
+public val AuthSchemesAttributeKey: AttributeKey<List<AuthScheme>> = AttributeKey("authSchemes")
+
+/**
+ * A set of signing constraints for an AWS endpoint.
+ */
+@InternalApi
+public sealed class AuthScheme {
```
aajtodd, line 20: **Suggestion**: I realize this is namespaced to `endpoints` package but given the identity/auth SRA changes this is kind of overloaded as a type. Is there a better name for this? I think long term endpoints will not be allowed to influence auth scheme, only signing context (i.e. IF `sigv4` is chosen as auth scheme, this additional context should be fed to the signer).

* `SigningContext` 
* `EndpointSigningContext`
* `AuthSchemeContext`

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/NoHttpEngine.kt
ianbotsf: Ah then I misunderstood. I found it useful to have a `NoHttpEngine` implementation when updating the [S3](https://github.com/awslabs/aws-sdk-kotlin/pull/921/files#diff-4d9bd50b3adffc65d327527a09402127b64f92c1a9d13ad84e894ba939f0e308) and [Polly](https://github.com/awslabs/aws-sdk-kotlin/pull/921/files#diff-dd429df13fe7c3ec049d98dbbf70c8f86c3266e904b01c6362f93ab252e050d0) presigning unit tests but I could always make this a test-only utility.

aajtodd: I see, I'm not 100% opposed I just hadn't pictured us shipping it. I don't necessarily see harm in it. We could discuss

ianbotsf: Discussed offline and decided to _not_ publish `NoHttpEngine` for now. We can revisit when we get more specific user requests.

PR #858, opened 2023-05-24
feat: add support for non-standard pagination termination via a truncation flag

ianbotsf:
## Issue \#

(none)

## Description of changes

This change adds support for non-standard pagination termination conditions via a new custom trait `PaginationTruncationMember`.

**Companion PR**: [aws-sdk-kotlin#928](https://github.com/awslabs/aws-sdk-kotlin/pull/928)


aajtodd (approved):
This looks good and gets us part of the way there to cover a few other paginated APIs we can't currently support because they use multi valued input/output markers.

I know the following are hand written in Go for this reason:


    Route53
        ListResourceRecordSets
    S3
        ListObjectVersions
        ListMultipartUploads

I think we could automate these by building on what you have here probably (IIRC all three of these use the `IsTruncated` flag. The missing piece is that they are not annotated with the `@paginated` trait because they have more than one marker that needs to be passed).

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
```diff
@@ -141,18 +145,27 @@ class PaginatorGenerator : KotlinIntegration {
...
-                        write("isFirstPage = false")
                         write("cursor = result.$nextMarkerLiteral")
+
+                        val truncationMember = outputShape.members().singleOrNull {
+                            it.hasTrait(PaginationTruncationMember.ID)
+                        }
+                        val hasNextPageCondition = when (truncationMember) {
+                            null -> "cursor?.isNotEmpty() == true"
+                            else -> "result.${truncationMember.defaultName()} == true"
+                        }
```
lauzadis, line 166: nit: consider `val hasNextPageCondition = truncationMember?.let { "result.${it.defaultName()}" } ?: "cursor?.isNotEmpty() == true"`

ianbotsf: Oh yep and I can extract the `== true` part as well. Good eye.

PR #860, opened 2023-05-31
fix: validate modeled HTTP status codes

lauzadis:
HTTP status codes are modeled on operations using the `@http` trait's `code` field, which [defaults to 200](https://smithy.io/2.0/spec/http-bindings.html?highlight=http#http-trait).

Currently we validate HTTP responses by checking if the code is in the 2xx range. Instead, we should validate using the modeled response code. 

- Change validation logic from `!response.status.isSuccess()` to `response.status.value != {expected HTTP status code}` **only if** the modeled code differs from the default value of 200.
- Remove `expectedHttpStatus` from the HTTP operation's execution context, since it was not being used anywhere. (Optional change, up for debate if we should keep it)

## Issue \#

https://github.com/awslabs/smithy-kotlin/issues/836

## Description of changes

This change is required to only validate the modeled HTTP response code. Currently the SDK may consider some responses "valid" even if the service did not intend for them to be valid.

### DIFF
--- a/.changes/2fda4a38-9ab6-4f22-8142-b8b269c52451.json
+++ b/.changes/2fda4a38-9ab6-4f22-8142-b8b269c52451.json
@@ -1,0 +1,5 @@
+{
+    "id": "2fda4a38-9ab6-4f22-8142-b8b269c52451",
+    "type": "misc",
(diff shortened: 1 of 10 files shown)


lauzadis:
We've paused work on htis

PR #864, opened 2023-06-14
feat: telemetry coroutine extensions

aajtodd:
## Issue \#

N/A

## Description of changes

This PR expands the telemetry API with coroutine extensions that the actual generated SDK clients and runtime will mostly use.

* Adds a new `ContextManager` interface. This is used to bridge the gap between the external calling code/observability backend and the SDK notion of context which relies on coroutine context for API boundary propagation. 
    * Consider this experimental for now as we figure out the best way to deal with some of the rubber meets the road issues around actually plugging in an observability backend like OTeL
* Adds coroutine extensions for logging, trace span creation, and propagation of `Context`, `TelemetryProvider`, etc.

## Example Instrumentation

The following is a strawman example of what instrumenting an operation might look like. This will likely change and/or get abstracted to be less verbose/more reusable. 

```kotlin
private suspend fun testOperation(provider: TelemetryProvider) {
    val tracer = provider.tracerProvider.getOrCreateTracer("S3")
    val currentContext = provider.contextManager.current()

    val telemetryCtx = TelemetryProviderContext(provider) + TelemetryContextElement(currentContext)
    val span = tracer.createSpan(
        "S3.GetObject",
        attributesOf {
            "rpc.method" to "GetObject"
            "rpc.service" to "S3"
        },
        SpanKind.CLIENT,
        currentContext,
    )

    withSpan(span, telemetryCtx) {
        span.setAttribute("rpc.system", "aws-api")
        testChild()
    }
}

private suspend fun testChild() {
    // FIXME - would require getting a tracer for child spans...
    // thats _ok_ but probably want to use an existing service client tracer instance if possible?
}

```

This leaves some open questions on how we want 

### DIFF
--- a/runtime/observability/telemetry-api/common/src/aws/smithy/kotlin/runtime/telemetry/trace/CoroutineContextTraceExt.kt
+++ b/runtime/observability/telemetry-api/common/src/aws/smithy/kotlin/runtime/telemetry/trace/CoroutineContextTraceExt.kt
@@ -1,0 +1,76 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package aws.smithy.kotlin.runtime.telemetry.trace
+
+import aws.smithy.kotlin.runtime.InternalApi
+import aws.smithy.kotlin.runtime.telemetry.context.telemetryContext
+import aws.smithy.kotlin.runtime.util.Attributes
+import aws.smithy.kotlin.runtime.util.emptyAttributes
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.withContext
+import kotlin.coroutines.AbstractCoroutineContextElement
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.coroutines.cancellation.CancellationException
+import kotlin.coroutines.coroutineContext
+
+/**
+ * A [CoroutineContext] element that carries a [TraceSpan].
+ * @param traceSpan The active trace span for this context.
+ */
+@InternalApi
+public data class TraceSpanContext(public val traceSpan: TraceSpan) : AbstractCoroutineContextElement(TraceSpanContext) {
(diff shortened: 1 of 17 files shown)


runtime/observability/telemetry-api/common/src/aws/smithy/kotlin/runtime/telemetry/trace/CoroutineContextTraceExt.kt
```diff
@@ -0,0 +1,76 @@
...
+    withContext(context + TraceSpanContext(span)) {
+        block(span)
+    }
+} catch (ex: Exception) {
+    if (ex !is CancellationException) {
+        span.setStatus(SpanStatus.ERROR)
+    }
+    span.recordException(ex, true)
+    throw ex
+} finally {
```
ianbotsf, line 74: **Question**: When getting a `CancellationException`, we'll skip setting the span status to `ERROR` but we'll still call `recordException` which will set the `error` attribute to `true`. Is this correct?

ianbotsf (approved):

runtime/observability/telemetry-api/common/src/aws/smithy/kotlin/runtime/telemetry/trace/CoroutineContextTraceExt.kt
aajtodd: Nope good catch.

ianbotsf: I suppose it depends on if we intend to complete canceled spans or not. Assuming we do intend to complete them, then I think marking _something_ is important (your `cancelled = true` attribute is a good idea) but definitely not the exception...that's an implementation detail of the language/coroutines, not an indication of unexpected behavior in the code.

aajtodd: I'll fix this in next PR

lauzadis (approved):

PR #869, opened 2023-06-19
refactor: remove old telemetry APIS

aajtodd:
## Issue \#

N/A

## Description of changes

* Remove previous logging and trace APIs
* Temporary TODO for one instance where we don't have a coroutine context to pull a logger from. 


runtime/protocol/http-test/jvm/src/aws/smithy/kotlin/runtime/httptest/TestWithLocalServer.kt
```diff
@@ -36,7 +33,7 @@ public abstract class TestWithLocalServer {
                 attempt++
                 try {
                     server.start()
-                    logger.info { "test server listening on: $testHost:$serverPort" }
+                    println("test server listening on: $testHost:$serverPort")
```
lauzadis, line 36: should `println` be used here? I see you removed the log statement entirely in `OkHttpProxySelector`, should that also have a `println`?

lauzadis (approved):

runtime/protocol/http-test/jvm/src/aws/smithy/kotlin/runtime/httptest/TestWithLocalServer.kt
aajtodd: This is in a test, I'm less worried about it. The `OkHttpProxySelector` would need to use logs but it lacks a `CoroutineContext` to pull a logger from. I _think_ this is actually ok though because the [HttpEngineEventListener](https://github.com/awslabs/smithy-kotlin/blob/a250c3e3e3e54ef35990a1609fb380a91b70cf4b/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/HttpEngineEventListener.kt#L53)  for the request is going to capture the same connect failure that was being logged.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpUtils.kt
```diff
@@ -159,11 +158,7 @@ internal class OkHttpProxySelector(
             else -> emptyList()
         }
     }
-
-    override fun connectFailed(uri: URI?, sa: SocketAddress?, ioe: IOException?) {
-        val logger = Logger.getLogger<OkHttpProxySelector>()
-        logger.error { "failed to connect to proxy: uri=$uri; socketAddress: $sa; exception: $ioe" }
-    }
+    override fun connectFailed(uri: URI?, sa: SocketAddress?, ioe: IOException?) {}
```
ianbotsf, line 161: **Question**: Do we intend to re-enable logging here once a default Slf4j binding is available? If so, I suggest leaving a `// FIXME` and possibly the commented code too for a starting place. As it is, this method looks pointless and like precisely the kind of thing I would delete if I came across it.

ianbotsf (approved):

PR #873, opened 2023-06-21
feat: add adaptive retries

ianbotsf:
## Issue \#

Closes [aws-sdk-kotlin#701](https://github.com/awslabs/aws-sdk-kotlin/issues/701)

## Description of changes

Adds adaptive retry support, including integration tests and DSL support.
[aws-sdk-kotlin#958](https://github.com/awslabs/aws-sdk-kotlin/pull/958)

### DIFF
--- a/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/AdaptiveClientRateLimiter.kt
+++ b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/AdaptiveClientRateLimiter.kt
@@ -1,0 +1,269 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.retries.delay
+
+import aws.smithy.kotlin.runtime.retries.policy.RetryErrorType
+import aws.smithy.kotlin.runtime.util.DslFactory
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlin.math.floor
+import kotlin.math.max
+import kotlin.math.min
+import kotlin.math.pow
+import kotlin.time.*
+import kotlin.time.Duration.Companion.seconds
+
+/**
+ * A client-side rate limiter backed by a token bucket. This limiter adaptively updates the refill rate of the bucket
+ * based on the number of successful transactions vs throttling errors. This limiter applies a smoothing function in an
+ * attempt to converge on the ideal transaction rate feasible for downstream systems.
+ */
+@OptIn(ExperimentalTime::class)
+public class AdaptiveClientRateLimiter internal constructor(
+    public override val config: Config = Config.Default,
+    private val timeSource: TimeSource,
+    private val rateMeasurer: AdaptiveRateMeasurer,
+    private val rateCalculator: CubicRateCalculator,
+) : ClientRateLimiter {
+    /**
+     * Initializes a new [AdaptiveClientRateLimiter]
+     * @param config The configuration parameters for this limiter
+     */
+    public constructor(config: Config = Config.Default) : this(
+        config,
+        TimeSource.Monotonic,
+        AdaptiveRateMeasurer(config, TimeSource.Monotonic),
+        CubicRateCalculator(config, TimeSource.Monotonic),
+    )
+
+    public companion object : DslFactory<Config.Builder, AdaptiveClientRateLimiter> {
+        /**
+         * Initializes a new [AdaptiveClientRateLimiter]
+         * @param block A DSL block which sets the configuration parameters for this limiter
+         */
+        public override operator fun invoke(block: Config.Builder.() -> Unit): AdaptiveClientRateLimiter =
+            AdaptiveClientRateLimiter(Config(Config.Builder().apply(block)))
+    }
+
+    private var capacity = 0.0
+    private var lastTimeMark: TimeMark? = null
+    internal var refillUnitsPerSecond = 0.0
+    private var maxCapacity = 0.0
+    private val mutex = Mutex()
+
+    override suspend fun acquire(cost: Int): Unit = mutex.withLock {
+        if (rateCalculator.throttlingEnabled) {
+            refillCapacity()
+            if (cost <= capacity) {
+                capacity -= cost
+            } else {
+                val extraRequiredCapacity = cost - capacity
+                val delayDuration = (extraRequiredCapacity / refillUnitsPerSecond).seconds
+                delay(delayDuration)
+                capacity = 0.0
+            }
+        }
+    }
+
+    private fun refillCapacity() {
+        lastTimeMark?.let {
+            val refillSeconds = it.elapsedNow().toDouble(DurationUnit.SECONDS)
+            val refillCapacity = refillUnitsPerSecond * refillSeconds
+            capacity = min(maxCapacity, capacity + refillCapacity)
+        }
+        lastTimeMark = timeSource.markNow()
+    }
+
+    override suspend fun update(errorType: RetryErrorType?): Unit = mutex.withLock {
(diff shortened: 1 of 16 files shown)


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/ClientRateLimiter.kt
```diff
@@ -0,0 +1,47 @@
...
+ * attempts).
+ *
+ * Callers should invoke [acquire] with the cost of their transaction prior to executing. If
+ * there are not enough permits available currently, the call may delay.
+ *
+ * Once the transaction is completed, callers should invoke [update] and indicate the type of error (if any) the
+ * transaction yielded. The rate limiter may use this information to adjust the number of available permits or the rate
+ * of permit renewal.
+ */
+public interface ClientRateLimiter {
```
aajtodd, line 23: **nit**: Why not just `RateLimiter`?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/ClientRateLimiter.kt
```diff
@@ -0,0 +1,47 @@
...
+    public val config: Config
+
+    /**
+     * Acquire a "permit" to conduct a transaction. If not enough permits are available, this method may delay.
+     * @param cost The relative cost of the transaction. Some transactions may take more work or represent a bundle of
+     * smaller transactions.
+     */
+    public suspend fun acquire(cost: Int)
```
aajtodd, line 31: **comment**: It seems odd to have an `acquire` but no `release` sibling. I'm assuming `update` is that but it feels sort of clunky (to me anyway) to not have the API symmetry.

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/retries/impl/Utils.kt
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.retries.impl
+
+import com.charleskorn.kaml.Yaml
+import kotlinx.serialization.DeserializationStrategy
+
+fun <T> Map<String, String>.deserialize(serializer: DeserializationStrategy<T>): Map<String, T> =
```
aajtodd, line 10: **nit**: `deserializeYaml` would be more appropriate since this isn't namespaced in a way to tell us the expected format.

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/RetryMode.kt
```diff
@@ -18,12 +18,14 @@ public enum class RetryMode {
...
     /**
-     * The standard retry mode. With this, the client will use the [StandardRetryStrategy][aws.smithy.kotlin.runtime.retries.StandardRetryStrategy]
+     * The standard retry mode. With this, the client will use the
+     * [StandardRetryStrategy][aws.smithy.kotlin.runtime.retries.StandardRetryStrategy].
      */
     STANDARD,
 
     /**
-     * Not implemented yet. https://github.com/awslabs/aws-sdk-kotlin/issues/701
+     * Adaptive retry mode. With this, the client will use the
```
aajtodd, line 27: **Question**: When would we choose one over the other? This seems like guidance in the developer guide would be good. In fact I haven't seen much from other SDKs here either on guidance and tuning. We should do better in our own guide.

aajtodd (approved):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/ClientRateLimiter.kt
ianbotsf: Generally, rate limiters don't allow "releasing" capacity. You must acquire a permit but the only thing that reincrements permits is time, not successes or failures. (A _concurrency_ limiter would have symmetrical acquire and release methods.) See for example [Guava's `RateLimiter`](https://guava.dev/releases/19.0/api/docs/index.html?com/google/common/util/concurrent/RateLimiter.html).

The `update` method is not analogous to releasing capacity. It's a mechanism for dynamically adjusting the refill rate based on the type of response received from the last piece of work.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/AdaptiveClientRateLimiter.kt
```diff
@@ -0,0 +1,269 @@
...
+            throttlingEnabled = true
+            cubicThrottle(lastMaxRate)
+        } else {
+            cubicSuccess()
+        }
+
+        return calculatedRate
+    }
+
+    internal fun calculateTimeWindow() = ((lastMaxRate * (1 - config.beta)) / config.scaleConstant).pow(1.0 / 3)
```
lauzadis, line 232: You could also use `kotlin.math.cbrt` for the cube root

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/retries/impl/StandardRetryIntegrationTestResources.kt
```diff
@@ -184,5 +187,38 @@ val standardRetryIntegrationTestCases = mapOf(
                 expected:
                   outcome: max_attempts_exceeded
                   retry_quota: 480
-        """.trimMargin(),
+        """.trimIndent(),
+)
+
+/* ktlint-disable annotation spacing-between-declarations-with-annotations */
```
lauzadis, line 193: Should this rule be permanently disabled? I noticed we run into this a lot

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/retries/impl/AdaptiveRetryIntegrationTestResources.kt
```diff
@@ -0,0 +1,184 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.retries.impl
+
+import aws.smithy.kotlin.runtime.retries.policy.RetryErrorType
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+val adaptiveRetryCubicTestCases = mapOf(
```
lauzadis, line 11: Can you add the source of these test cases? Not a direct link but just a hint for future reference

lauzadis (approved):

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/retries/impl/StandardRetryIntegrationTestResources.kt
ianbotsf: I still favor this rule the vast majority of the time. Kotlin is a denser language than Java with the capability to encode much more meaning in shorter lines but that can quickly hamper readability and maintainability. Rules like this keep us from attempting to be obscuredly brief unless we have a really good reason.

In my mind, those good reasons include many similar declarations requiring the same annotation which are otherwise lightly differentiated and have no user-facing documentation. These serialization classes are a good example of that because they only exist to hold data in the format already specified by the JSON struct...they carry no business logic, aren't user-facing, etc. The only other place I see we've disabled this rule is for the [typed environment setting implementations](https://github.com/awslabs/smithy-kotlin/blob/b4580e3d4a6fbf81023ba6379e1704a86b6b1c1b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/config/EnvironmentSetting.kt#L53-L58) which again are not user-facing and are largely rote (i.e., they differ only by type).

In the end, I like this rule and would strongly prefer keeping it—even at the expense of removing these overrides and putting the attributes on separate lines for these simple cases.

PR #875, opened 2023-06-22
feat: big number support

### LINKED ISSUE #213: Add MPP compatible big number support
We are currently generating JVM only types `BigInteger` and `BigDecimal`. 

We should just add a simple wrapper around string to the `software.aws.clientrt.smithy` package and generate to that. We can then have conversion utils in the JVM impl.

e.g.

```
// common
data class BigInteger(val value: String)

data class BigDecimal(val value: String)

// jvm
BigInteger.toJvmBigInt(): java.math.BigInt
BigDecimal.toJvmBigDecimal(): java.math.BigDecimal

```

There is no MPP compatible type in the standard library and so we would otherwise have to add a dependency. Rather than do that it makes more sense to just delegate this to the customer if they actually need to do calculations with the results. Big numbers are often used in cryptographic calculations which would drive different requirements for different people.

*internal id*: 174761915

ianbotsf:
## Issue \#

Closes #213 

## Description of changes

This change adds new `BigInteger` and `BigDecimal` classes which delegate to stdlib implementations on JVM. It also finalizes codegen support for the new big number classes.


lauzadis (approved):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/content/BigDecimal.kt
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.content
+
+public expect class BigDecimal(value: String) : Number {
```
aajtodd, line 7: **correctness/question**: Should this implement `Number`?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/content/BigDecimal.kt
```diff
@@ -0,0 +1,9 @@
...
+    public fun toPlainString(): String
```
aajtodd, line 8: **Question**: Why?

tests/compile/src/test/resources/kitchen-sink-model.smithy
```diff
@@ -404,6 +405,27 @@ structure PrefixHeadersIO {
...
 }
 
+@http(method: "POST", uri: "/primitives")
+operation Primitives {
+    input: PrimitivesIO,
+    output: PrimitivesIO
+}
+
+// A structure containing all the primitive types
+structure PrimitivesIO {
```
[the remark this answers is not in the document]
aajtodd: Aren't those gone in IDL 2?

aajtodd (approved):

tests/compile/src/test/resources/kitchen-sink-model.smithy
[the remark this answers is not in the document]
ianbotsf: Seemingly they are still part of [the prelude in V2](https://smithy.io/2.0/spec/model.html#prelude). I'll play around with adding them.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/content/BigDecimal.kt
ianbotsf: Certainly up for debate. In my opinion, they should because they are numbers (taxonomically). Kotlin's stdlib [`Number` alias](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-number/) automatically adds methods for conversion to other types (e.g., `toInt`, `toLong`, etc.) which seems useful. Plus I believe there are areas in our code where we check if something is a `Number` and act accordingly.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/content/BigDecimal.kt
ianbotsf: Ugh, because Java's `BigDecimal.toString()` returns a string in exponential notation (e.g., `1.23e20`) after the magnitude gets large enough. That's unsuitable for serialization because it would result in a loss of precision. So we need a way to get the full, unabridged decimal representation and JVM's version of that is [`toPlainString()`](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#toPlainString--). I don't love this, by the way—it feels conspicuous and strongly tied to a quirk of the JVM implementation.

Any suggestions for a more graceful way to handle it?

aajtodd: I think thats fine if it's the accepted norm on JVM, its what most users will be using so makes sense to be consistent there than introduce a new concept.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/content/BigDecimal.kt
aajtodd: I suppose `Number` documents that these methods may truncate. I guess it's fine 🤷

PR #879, opened 2023-06-28
fix: correctly deserialize consecutive XML flat maps 

lauzadis:
This PR fixes a bug in the XML deserializer which fails to process consecutive flat maps properly. Before this change, the deserializer would seek forward until it found a matching map key. Consecutive flat maps sharing the same key names would be lumped together unintentionally. 

The fix is to seek using `XmlSerialName` as a comparator since this is the only thing distinguishing different flat maps. A new deserialize method `peekSeek` is added which seeks forward and only consumes tokens if they are a match.

## Issue \#

https://github.com/awslabs/aws-sdk-kotlin/issues/962

## Description of changes

This fixes a bug when deserializing consecutive flat maps. See the new "consecutive flat maps" test case for an example of an XML tree for which parsing previously failed. 

### DIFF
--- a/runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
+++ b/runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
@@ -105,11 +105,7 @@ internal class XmlMapDeserializer(
 
     override fun hasNextEntry(): Boolean {
         val compareTo = when (descriptor.hasTrait<Flattened>()) {
-            // Prefer seeking to XmlSerialName if the trait exists
-            true -> when (descriptor.hasTrait<XmlSerialName>()) {
-                true -> descriptor.expectTrait<XmlSerialName>().name
-                false -> mapTrait.key
-            }
+            true -> descriptor.findTrait<XmlSerialName>()?.name ?: mapTrait.key // Prefer seeking to XmlSerialName if the trait exists
             false -> mapTrait.entry
         }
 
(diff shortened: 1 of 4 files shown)


runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
```diff
@@ -98,10 +104,19 @@ internal class XmlMapDeserializer(
...
     override fun hasNextEntry(): Boolean {
-        // Seek to either the entry or key token depending on the flatness of the map
+        val compareTo = when (descriptor.hasTrait<Flattened>()) {
+            // Prefer seeking to XmlSerialName if the trait exists
+            true -> when (descriptor.hasTrait<XmlSerialName>()) {
+                true -> descriptor.expectTrait<XmlSerialName>().name
+                false -> mapTrait.key
+            }
+            false -> mapTrait.entry
+        }
```
ianbotsf, line 114: **Style**: I generally find `when`/`true`/`false` to be less readable than `if`/`else` or null-coalescing. In this case, `when (descriptor.hasTrait<XmlSerialName>()) { ... }` can be replaced with:

```kotlin
descriptor.findTrait<XmlSerialName>()?.name ?: mapTrait.key
```

ianbotsf (approved):
This seems right to me. Do we have similar problems in our other deserializers (e.g., JSON)?

lauzadis:
> Do we have similar problems in our other deserializers (e.g., JSON)?

Thankfully JSON doesn't have this problem because it only has one way to represent a map

aajtodd (approved):

PR #885, opened 2023-07-07
Mark TelemetryProvider API experimental

aajtodd:
## Issue \#

N/A

## Description of changes

Mark the `TelemetryProvider` API as experimental and optin where necessary. This explicitly doesn't mark the `TelemetryProvider` interface or configuration types as experimental as that would make configuring the SDK itself experimental. Instead we mark every field/method on the interface as `Experimental`. As we stabilize each telemetry signal implementation we can start to remove the annotation (e.g. once we are happy with metrics and the set of metrics exported we might remove it from `meterProvider`). 

### DIFF
--- a/runtime/observability/telemetry-api/common/src/aws/smithy/kotlin/runtime/telemetry/TelemetryProvider.kt
+++ b/runtime/observability/telemetry-api/common/src/aws/smithy/kotlin/runtime/telemetry/TelemetryProvider.kt
@@ -5,6 +5,7 @@
 
 package aws.smithy.kotlin.runtime.telemetry
 
+import aws.smithy.kotlin.runtime.ExperimentalApi
 import aws.smithy.kotlin.runtime.telemetry.context.Context
 import aws.smithy.kotlin.runtime.telemetry.context.ContextManager
 import aws.smithy.kotlin.runtime.telemetry.logging.LoggerProvider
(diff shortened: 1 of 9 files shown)


runtime/observability/telemetry-api/common/src/aws/smithy/kotlin/runtime/telemetry/TelemetryProvider.kt
```diff
@@ -25,24 +26,29 @@ public interface TelemetryProvider {
...
     public val loggerProvider: LoggerProvider
 
     /**
      * Get the [ContextManager] used to get the current [Context]
      */
+    @ExperimentalApi
     public val contextManager: ContextManager
 }
 
+@OptIn(ExperimentalApi::class)
```
ianbotsf, line 51: **Question**: Shouldn't we opt into this annotation globally [the way we do with `@InternalApi`](https://github.com/awslabs/smithy-kotlin/blob/3e1d8618b0090b218461cffdcce20619d6b4119b/runtime/auth/aws-signing-default/build.gradle.kts#L27C1-L29)?

aajtodd: I decided not to because I wanted to see how far reaching the API becomes. Turns out there weren't that many places that need it.

ianbotsf (approved):

0marperez (approved):

PR #887, opened 2023-07-11
feat: enable endpoint discovery

### LINKED ISSUE #146: Endpoint Discovery
Design and implement dynamic endpoint discovery.

See also: 
    * [Endpoint discovery](https://awslabs.github.io/smithy/1.0/spec/aws/aws-core.html#client-endpoint-discovery)
    * /seps/accepted/shared/endpoint-discovery/
    * /seps/accepted/shared/sts_regionlization/

ianbotsf:
## Issue \#

Closes #146 

## Description of changes

Enables endpoint discovery for services that specify the correct traits in their models. Specifically:
* Codegens a new endpoint resolver that delegates to the service-default implementation and then replaces the host by querying a specific operation
* Caches the discovered endpoint by region and access key ID
* Adds a new config parameter for using endpoint discovery for services where it is optional (defaults to false)
* Adds an interceptor to detect when an endpoint is no longer valid and remove it from the cache

**Note** that this does not _fully_ implement [the Smithy spec](https://smithy.io/2.0/aws/aws-core.html#client-endpoint-discovery) because it assumes that no operation name or identifiers are passed to the discover endpoints operation, which is true for all AWS services that currently support endpoint discovery (i.e., DynamoDB, Timestream Query, and Timestream Write).

**Companion PR**: [aws-sdk-kotlin#976](https://github.com/awslabs/aws-sdk-kotlin/pull/976)


codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/discovery/EndpointDiscovererGeneratorTest.kt
```diff
@@ -0,0 +1,149 @@
...
+        )
+    }
+
+    @Test
+    fun testAsEndpointResolver() {
+        val actual = render()
+
+        actual.shouldContainOnlyOnceWithDiff(
+            """
+                internal suspend fun asEndpointResolver(client: TestClient, delegate: EndpointResolverAdapter) = EndpointResolver { request ->
```
aajtodd, line 39: **correctness**: Why does the conversion `asEndpointResolver` need to be `suspend`?

ianbotsf: Good catch, it does not need to be.

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/util/ReadThroughCacheTest.kt
```diff
@@ -0,0 +1,69 @@
...
+        fun uncachedValue() = ExpiringValue(counter++, clock.now() + 2.seconds)
+        val cache = ReadThroughCache<String, Int>(1.minutes, clock)
+
+        // Basic read through
+        assertEquals(0, cache.get("a") { uncachedValue() })
+        assertEquals(1, cache.get("b") { uncachedValue() })
+
+        // Basic cache verification
+        assertEquals(0, cache.get("a") { uncachedValue() })
+        assertEquals(1, cache.get("b") { uncachedValue() })
```
lauzadis, line 30: here instead of `uncachedValue()` it could be better to pass a function that throws an exception, to check that the cached value is read instead of the pass-through function being invoked

lauzadis (approved):

PR #890, opened 2023-07-12
fix: correct URL encoding in endpoint resolution

### LINKED ISSUE #888: Resolved endpoint URLs are incorrectly decoded
## Describe the bug
The endpoint provider generator [calls `Url.parse`](https://github.com/awslabs/smithy-kotlin/blob/3e1d8618b0090b218461cffdcce20619d6b4119b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/DefaultEndpointProviderGenerator.kt#L149) on constructed URLs. `Url.parse` [delegates to `urlParseImpl`](https://github.com/awslabs/smithy-kotlin/blob/3e1d8618b0090b218461cffdcce20619d6b4119b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/Url.kt#L42), which [URL-decodes path components](https://github.com/awslabs/smithy-kotlin/blob/3e1d8618b0090b218461cffdcce20619d6b4119b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/UrlParser.kt#L28). This leads to incorrectly decoding values which have already been URL-_encoded_ by the `uriEncode` built-in endpoint function.

## Expected Behavior
URL-encoded path components generated by endpoint providers should not be decoded when constructing a `Url`.

ianbotsf:
## Issue \#

Closes #888 

## Description of changes

This change corrects a bug in URL encoding in endpoint resolution which caused parsing to URL-_decode_ strings which were already properly encoded.

### DIFF
--- a/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/Url.kt
+++ b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/Url.kt
@@ -110,7 +110,7 @@ public data class UserInfo(public val username: String, public val password: Str
  */
 public enum class UrlDecodingBehavior {
     /**
-     * Identifies that a URL string does not its path component need to be URL-decoded.
+     * Identifies that a URL string does not need its path component need to be URL-decoded.
      */
     DO_NOT_DECODE_PATH,
 
--- a/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/Url.kt
+++ b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/Url.kt
@@ -108,16 +108,16 @@ public data class UserInfo(public val username: String, public val password: Str
 /**
  * Identifies the type of decoding behavior desired when parsing a URL.
  */
-public enum class UrlDecodingBehavior(internal val maybeDecode: (String) -> String) {
+public enum class UrlDecodingBehavior {
     /**
-     * Identifies that a URL string does not need to be URL-decoded.
+     * Identifies that a URL string does not its path component need to be URL-decoded.
      */
-    DO_NOT_DECODE({ it }),
+    DO_NOT_DECODE_PATH,
 
     /**
      * Identifies that a URL string needs to be URL-decoded.
      */
-    DECODE_COMPONENTS(String::urlDecodeComponent),
+    DECODE_COMPONENTS,
 }
 
 /**
--- a/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/Url.kt
+++ b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/Url.kt
@@ -39,8 +39,8 @@ public data class Url(
     }
 
     public companion object {
-        public fun parse(url: String): Url = parse(url, UrlDecodingBehavior.DECODE_COMPONENTS)
-        public fun parse(url: String, decodingBehavior: UrlDecodingBehavior): Url = urlParseImpl(url, decodingBehavior)
+        public fun parse(url: String): Url = parse(url, UrlDecoding.DecodeAll)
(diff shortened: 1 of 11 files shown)


lucix-aws (approved):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/UrlParser.kt
```diff
@@ -25,20 +25,20 @@ internal fun urlParseImpl(url: String): Url =
...
-                path = "/${it.urlDecodeComponent()}"
+                path = "/${decodingBehavior.maybeDecode(it)}"
             }
         }
 
         if (next.startsWith("?")) {
             next = next.capture(1 until next.firstIndexOrEnd("#")) {
                 it.splitAsQueryString().entries.forEach { (k, v) ->
-                    parameters.appendAll(k.urlDecodeComponent(), v.map(String::urlDecodeComponent))
+                    parameters.appendAll(decodingBehavior.maybeDecode(k), v.map { decodingBehavior.maybeDecode(it) })
```
lucix-aws, line 35: Question: Do we want this decode/no-decode to be a blanket setting?

The reason for shutting it off on the path is clear - but it seems like you'd always want the query decoded (and it would always be safe to do so).

ianbotsf: Yeah you're probably right. Will revise.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/UrlParser.kt
```diff
@@ -23,16 +24,20 @@ internal fun urlParseImpl(url: String): Url =
...
-                path = "/${it.urlDecodeComponent()}"
+                path = "/${pathDecode(it)}"
             }
         }
 
         if (next.startsWith("?")) {
             next = next.capture(1 until next.firstIndexOrEnd("#")) {
                 it.splitAsQueryString().entries.forEach { (k, v) ->
-                    parameters.appendAll(k.urlDecodeComponent(), v.map(String::urlDecodeComponent))
+                    parameters.appendAll(k.urlEncodeComponent(), v.map(String::urlDecodeComponent))
```
aajtodd, line 40: **correctness**: Why are we now encoding the key?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/Url.kt
```diff
@@ -104,6 +105,21 @@ private fun encodePath(
  */
 public data class UserInfo(public val username: String, public val password: String)
 
+/**
+ * Identifies the type of decoding behavior desired when parsing a URL.
+ */
+public enum class UrlDecodingBehavior {
```
aajtodd, line 111: **comment**: This has odd semantics since component may be from path or query parameters. Only path decoding consults this enum in the parse implementation which sort of reaffirms the oddness of it.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/Url.kt
```diff
@@ -104,6 +105,21 @@ private fun encodePath(
...
+    /**
+     * Identifies that a URL string does not its path component need to be URL-decoded.
```
aajtodd, line 113: **nit**: 

> Identifies that a URL string does not **need** its path component need to be URL-decoded.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/UrlParser.kt
ianbotsf: 😳 That's an accidental change. Will fix.

aajtodd (approved):

PR #897, opened 2023-07-19
fix: s3 custom treatment getbucketlocation response

0marperez:
## Issue \#
[194](https://github.com/awslabs/aws-sdk-kotlin/issues/194) - Custom treatment of `GetBucketLocation` response

## Description of changes

This was previously fixed using a customization but it left out error deserialization and deserialization of both wrapped and unwrapped non-error responses. It also overrode code unnecessarily so declared section is moved elsewhere now

Also look at: https://github.com/awslabs/aws-sdk-kotlin/pull/989


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/traits/UnwrappedXMLOutput.kt
```diff
@@ -0,0 +1,22 @@
...
+import software.amazon.smithy.model.node.ObjectNode
+import software.amazon.smithy.model.shapes.ShapeId
+import software.amazon.smithy.model.traits.AnnotationTrait
+
+/**
+ * Indicates the annotated structure is unwrapped XML output.
+ *
+ * Refer to this [s3 specific example](https://smithy.io/2.0/aws/customizations/s3-customizations.html#aws-customizations-s3unwrappedxmloutput-trait)
+ */
+class UnwrappedXMLOutput(node: ObjectNode) : AnnotationTrait(ID, node) {
```
ianbotsf, line 17: **Style**: Generally our names should follow PascalCase, including treating abbreviations like "XML" as regular words for the purposes of capitalization (e.g., `UnwrappedXmlOutput`).

runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
```diff
@@ -328,6 +331,63 @@ internal class XmlStructDeserializer(
...
 }
 
+/**
+ * Deserializes XML structures that are unwrapped
+ * See: https://smithy.io/2.0/aws/customizations/s3-customizations.html#aws-customizations-s3unwrappedxmloutput-trait
+ *
+ * @param objDescriptor associated [SdkObjectDescriptor] which represents the expected structure
+ * @param reader underlying [XmlStreamReader] from which tokens are read
+ */
+internal class UnwrappedXmlStructDeserializer(
```
ianbotsf, line 341: **Question**: Could be `private`?

runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
```diff
@@ -328,6 +331,63 @@ internal class XmlStructDeserializer(
...
+
+    override fun deserializeLong(): Long = deserializeValue { it.toLongOrNull() ?: throw DeserializationException("Unable to deserialize $it") }
+
+    override fun deserializeFloat(): Float = deserializeValue { it.toFloatOrNull() ?: throw DeserializationException("Unable to deserialize $it") }
+
+    override fun deserializeDouble(): Double = deserializeValue { it.toDoubleOrNull() ?: throw DeserializationException("Unable to deserialize $it") }
+
+    override fun deserializeString(): String = deserializeValue { it }
+
+    override fun deserializeBoolean(): Boolean = deserializeValue { it.toBoolean() }
```
ianbotsf, line 369: **Correctness**: I see a lot of duplicate code between `UnwrappedXmlStructDeserializer` and `XmlStructDeserializer`, which we should avoid. Let's either:
* make `XmlStructDeserializer` an `open class` so that the `UnwrappedXmlStructDeserializer` can inherit it and override the specific things that need to change –**or**–
* make an abstract base class for XML deserialization and make both `XmlStructDeserializer` and  `UnwrappedXmlStructDeserializer` inherit it, implementing only the functionality they individually need

runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
```diff
@@ -383,16 +443,29 @@ private fun SdkObjectDescriptor.fieldTokenMatcher(fieldDescriptor: SdkFieldDescr
...
 internal inline fun <reified TExpected : XmlToken> XmlStreamReader.takeNextAs(): TExpected {
     val token = this.nextToken() ?: throw DeserializationException("Expected ${TExpected::class} but instead found null")
     requireToken<TExpected>(token)
     return token as TExpected
 }
 
-// require that the given token be of type [TExpected] or else throw an exception
+/**
+ * Require that the given token be of type [TExpected] or else throw an exception
+ */
```
ianbotsf, line 466: **Comment**: Nice, thanks for correcting comments which should be KDocs!

ianbotsf (changes requested):

runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
aajtodd: Do we even need to override this? Can't this just be handled in the initial `deserializeStruct` method which is doing the token advance to the next element? 🤔

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt
```diff
@@ -626,6 +631,10 @@ abstract class HttpBindingProtocolGenerator : ProtocolGenerator {
             .closeBlock("}")
     }
 
+    object payloadDeserializer : SectionId {
```
aajtodd, line 634: **fix**: Shouldn't be necessary anymore

runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
```diff
@@ -213,65 +214,79 @@ internal class XmlListDeserializer(
...
-        if (inNestedMode()) {
-            // Returning from a nested struct call.  Nested deserializer consumed
-            // tokens so clear them here to avoid processing stale state
-            parsedFieldLocations.clear()
-        }
+        if (unwrapped) {
+            val FIELD_INDEX: Int = 0
+            return if (reader.peek() is XmlToken.Text) FIELD_INDEX else null
+        } else {
+            if (firstIteration) reader = reader.subTreeReader(XmlStreamReader.SubtreeStartDepth.CHILD); firstIteration = false
```
ianbotsf, line 233: **Question**: Using a mutable `reader` and changing its value on the first iteration seems clunky to me. Could detecting the wrapping mode be done during class initialization?

```kotlin
private class XmlStructDeserializer(
    ...
    reader: XmlStreamReader, // **not** a private val, just a constructor param
    ...
    unwrapped: Boolean, // **not** a private val, just a constructor param
) : Deserializer.FieldIterator {
    private val reader = if (unwrapped) reader else reader.subTreeReader()
```

runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
```diff
@@ -213,65 +214,78 @@ internal class XmlListDeserializer(
...
+    private val reader: XmlStreamReader = if (unwrapped) reader else reader.subTreeReader(XmlStreamReader.SubtreeStartDepth.CHILD)
+
     override fun findNextFieldIndex(): Int? {
-        if (inNestedMode()) {
-            // Returning from a nested struct call.  Nested deserializer consumed
-            // tokens so clear them here to avoid processing stale state
-            parsedFieldLocations.clear()
-        }
+        if (unwrapped) {
+            val FIELD_INDEX: Int = 0
```
aajtodd, line 231: **style**: Move this to a constant in the file e.g. `private const val FIRST_FIELD_INDEX: Int = 0`

runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
```diff
@@ -226,10 +226,13 @@ private class XmlStructDeserializer(
 
     private val reader: XmlStreamReader = if (unwrapped) reader else reader.subTreeReader(XmlStreamReader.SubtreeStartDepth.CHILD)
 
+    companion object {
```
aajtodd, line 229: You don't need a companion object for this. Constants are compile time replacements. This declares an empty companion object essentially.

aajtodd (approved):
Fix and ship

runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
```diff
@@ -213,65 +216,77 @@ internal class XmlListDeserializer(
...
+
     override fun findNextFieldIndex(): Int? {
-        if (inNestedMode()) {
-            // Returning from a nested struct call.  Nested deserializer consumed
-            // tokens so clear them here to avoid processing stale state
-            parsedFieldLocations.clear()
-        }
+        if (unwrapped) {
+            return if (reader.peek() is XmlToken.Text) FIRST_FIELD_INDEX else null
+        } else {
```
lauzadis, line 234: nit/style: since L233 is a `return if`, I think you could get rid of this huge `else` block 

```kt
if (unwrapped) {
    return if (reader.peek() is XmlToken.Text) FIRST_FIELD_INDEX else null
}

// rest of the else block, without the `else`...
```

This should also make the diff cleaner and easier to review

runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlDeserializer.kt
```diff
@@ -213,65 +216,77 @@ internal class XmlListDeserializer(
...
-        return when (val nextField = parsedFieldLocations.removeFirst()) {
-            is FieldLocation.Text -> {
-                val value = when (val peekToken = reader.peek()) {
-                    is XmlToken.Text -> reader.takeNextAs<XmlToken.Text>().value ?: ""
-                    is XmlToken.EndElement -> ""
-                    else -> throw DeserializationException("Unexpected token $peekToken")
+        if (unwrapped) {
+            val value = reader.takeNextAs<XmlToken.Text>().value ?: ""
+            return transform(value)
+        } else {
```
lauzadis, line 267: Same comment about refactoring the `else` block here

lauzadis (approved):

ianbotsf (approved):

PR #901, opened 2023-07-19
feat: add content length validation interceptor

lauzadis:
This PR adds an interceptor which validates the `Content-Length` response header against the response body. If the lengths don't match, an `IllegalStateException` is thrown.

This interceptor is currently only used for S3 `GetObject` responses.

## Issue \#

https://github.com/awslabs/aws-sdk-kotlin/issues/745

## Description of changes

This PR is required to support validating `Content-Length` response headers for S3 `GetObject` responses.

### DIFF
--- a/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ResponseLengthValidationInterceptor.kt
+++ b/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ResponseLengthValidationInterceptor.kt
@@ -14,6 +14,10 @@ import aws.smithy.kotlin.runtime.http.response.HttpResponse
 import aws.smithy.kotlin.runtime.http.toHttpBody
 import aws.smithy.kotlin.runtime.io.*
 
+/**
+ * An interceptor which compares the `Content-Length` header value against the length of the returned body. Throws an
+ * [IllegalStateException] if there is a mismatch.
+ */
 @InternalApi
 public class ResponseLengthValidationInterceptor : HttpInterceptor {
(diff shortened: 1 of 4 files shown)


aajtodd (approved):

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ResponseLengthValidationInterceptor.kt
```diff
@@ -0,0 +1,84 @@
...
+            validateContentLength(expectedContentLength, bytesReceived)
+        } else {
+            bytesReceived += it
+        }
+    }
+}
+
+private fun validateContentLength(expected: Long, actual: Long) {
+    check(expected == actual) { "Total bytes consumed ($actual) does not match expected ($expected)." }
+}
```
ianbotsf, line 84: **Correctness**: This error seems a little opaque for users. While I think the actual/expected byte counts are interesting, we should probably mention that this is likely because a connection closed prematurely or something like that. We should probably also throw an `IOException` (or something similar) instead of `IllegalStateException`.

ianbotsf (changes requested):

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ResponseLengthValidationInterceptor.kt
aajtodd: We have an [EOFException](https://github.com/awslabs/smithy-kotlin/blob/71db9ef7a8cfd28d79ee2d542d62827d1d3e6a17/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/io/Exceptions.kt#L12) defined in the runtime that inherits from IOException

ianbotsf (approved):

PR #906, opened 2023-07-26
feat: make user-supplied region available to credentials providers

0marperez:
## Issue \#
[583](https://github.com/awslabs/aws-sdk-kotlin/issues/583) - Make user-supplied region available to config resolution providers

## Description of changes

There are several places where config resolution providers require bootstrapping region, which is possibly present in user-supplied client config. This fixes this issue by making region available via attributes/execution context 

Also look at: https://github.com/awslabs/aws-sdk-kotlin/pull/997


runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/operation/SdkOperationExecution.kt
```diff
@@ -275,10 +275,13 @@ internal class AuthHandler<Input, Output>(
             "auth.scheme_id" to authScheme.schemeId.id
         }
 
+        // signing properties need to propagate from AuthOption to signer
```
ianbotsf, line 278: **Nit**: This comment only tells half the story now. We should update it to something like "properties need to propagate from AuthOption to signer and identity provider".

ianbotsf (approved):

aajtodd (approved):
Approved but agree with Ian's comment suggestion.

PR #927, opened 2023-08-17
refactor!: prefix endpoint and auth providers and make auth scheme provider public

aajtodd:
## Issue \#

n/a

## Description of changes

* **BREAKING**: prefix generated endpoint provider and parameters with the service client name (e.g. `EndpointProvider` -> `S3EndpointProvider`)
* **refactor**: minor cleanup and alignment to latest SRA auth APIs (e.g. `AuthSchemOption` -> `AuthOption`)
* **refactor**: make generated auth scheme provider type(s) public and part of service client config

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/AuthSchemeProviderGenerator.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/AuthSchemeProviderGenerator.kt
@@ -71,7 +71,7 @@ open class AuthSchemeProviderGenerator {
                 "private val operationOverrides = mapOf<#T, List<#T>>(",
                 ")",
                 KotlinTypes.String,
-                RuntimeTypes.Auth.Identity.AuthSchemeOption,
+                RuntimeTypes.Auth.Identity.AuthOption,
             ) {
                 operationsWithOverrides.forEach { op ->
                     val authHandlersForOperation = authIndex.effectiveAuthHandlersForOperation(ctx, op)
@@ -82,7 +82,7 @@ open class AuthSchemeProviderGenerator {
             withBlock(
                 "private val serviceDefaults = listOf<#T>(",
                 ")",
-                RuntimeTypes.Auth.Identity.AuthSchemeOption,
+                RuntimeTypes.Auth.Identity.AuthOption,
             ) {
                 val defaultHandlers = authIndex.effectiveAuthHandlersForService(ctx)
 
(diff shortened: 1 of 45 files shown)


runtime/auth/identity-api/common/src/aws/smithy/kotlin/runtime/auth/AuthOption.kt
```diff
@@ -12,14 +12,22 @@ import aws.smithy.kotlin.runtime.util.emptyAttributes
...
+    public val attributes: Attributes
+}
+
+public fun AuthOption(id: AuthSchemeId, attributes: Attributes = emptyAttributes()): AuthOption =
+    AuthOptionImpl(id, attributes)
+
+private data class AuthOptionImpl(
+    override val schemeId: AuthSchemeId,
+    override val attributes: Attributes,
+) : AuthOption
```
ianbotsf, line 33: **Comment**: We should consider soon whether to do this for other public data classes like `Credentials`, `HttpResponse`, `Url`, etc.

ianbotsf (approved):

runtime/auth/identity-api/common/src/aws/smithy/kotlin/runtime/auth/AuthOption.kt
aajtodd: Agreed. I'm wondering if we should do this for generated `EndpointParameters` as well which is currently a `class`.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/AuthSchemeProviderGenerator.kt
```diff
@@ -47,18 +46,25 @@ open class AuthSchemeProviderGenerator {
...
             RuntimeTypes.Auth.Identity.AuthSchemeProvider,
             paramsSymbol,
         )
     }
 
     private fun renderDefaultImpl(ctx: ProtocolGenerator.GenerationContext, writer: KotlinWriter) {
+        // FIXME - probably can't remain an object
         writer.withBlock(
-            "internal object #T : #T {",
+            "public object #T : #T {",
```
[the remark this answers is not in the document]
aajtodd: What if it ever needs a default argument/state at construction time? I suppose we could solve that by other means...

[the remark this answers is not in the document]
ianbotsf: Oh right. If think it _may_ need parameters or state in the future then I recommend we just make it a class now.

runtime/auth/identity-api/common/src/aws/smithy/kotlin/runtime/auth/AuthOption.kt
ianbotsf: I'm less worried about non-data classes because they can be carefully written to ensure forwards compatibility. Not opposed but not high-priority to me.

0marperez (approved):

PR #928, opened 2023-08-18
chore: convert dependencies to a version catalog

aajtodd:
## Issue \#

n/a

## Description of changes

Take advantage of gradle [version catalogs](https://docs.gradle.org/current/userguide/platforms.html#sub:central-declaration-of-dependencies)


build.gradle.kts
```diff
@@ -13,12 +13,13 @@ buildscript {
     }
 
     val kotlinVersion: String by project
+    // NOTE: buildscript classpath for the root project is the parent classloader for the subprojects, we
+    // only need to add e.g. atomic-fu and build-plugins here for imports and plugins to be available in subprojects.
     dependencies {
         classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion")
```
ianbotsf, line 19: **Question**: Can this dependency be moved into `libs`?

ianbotsf (approved):

lauzadis (approved):

aajtodd:
> Something I just thought of -- could we make the version catalog in the shared aws-kotlin-repo-tools instead, or is that not possible / too much abstraction?

You could but IMO it's not worth it right now. They can have their own dependency versions. aws-sdk-kotlin will get it's own version catalog that has some subset of the same dependencies/versions but also include smithy-kotlin modules.

PR #931, opened 2023-08-18
feat: add support for multi select hash to JMESPath visitor

### LINKED ISSUE #930: Add support for "MultiSelectHash" to JMESPath visitor

0marperez:
## Issue \#
closes #930 

## Description of changes

- Added support for multi select hash 
- Added missing list from multi select list to match JMESPath syntax
- `ensureNullGuard` is now `addNullGuard` and is added to all temporary val's
- `subField` function now allows accessing members of shapes that "are not there" (multi select hash creates anonymous objects with arbitrary attributes that don't show up in the model)
- comparator for `anyStringEquals` and `allStringEquals` now casts to a list of strings before comparing (anonymous objects have attributes of generic type)

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/AcceptorGenerator.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/AcceptorGenerator.kt
@@ -89,12 +89,12 @@ private fun KotlinWriter.renderPathAcceptor(wi: WaiterInfo, directive: String, i
         val comparison = when (matcher.comparator!!) {
             PathComparator.STRING_EQUALS -> "${actual.identifier} == ${expected.dq()}"
             PathComparator.BOOLEAN_EQUALS -> "${actual.identifier} == ${expected.toBoolean()}"
-            PathComparator.ANY_STRING_EQUALS -> "${actual.identifier}?.any { it == ${expected.dq()} } ?: false"
+            PathComparator.ANY_STRING_EQUALS -> "(${actual.identifier} as List<String>?)?.any { it == ${expected.dq()} } ?: false"
 
             // NOTE: the isNotEmpty check is necessary because the waiter spec says that `allStringEquals` requires
             // at least one value unlike Kotlin's `all` which returns true if the collection is empty
             PathComparator.ALL_STRING_EQUALS ->
-                "!${actual.identifier}.isNullOrEmpty() && ${actual.identifier}.all { it == ${expected.dq()} }"
+                "!(${actual.identifier} as List<String>).isNullOrEmpty() && ${actual.identifier}.all { it == ${expected.dq()} }"
         }
         write(comparison)
     }
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
@@ -14,17 +14,13 @@ import software.amazon.smithy.kotlin.codegen.core.CodegenContext
 import software.amazon.smithy.kotlin.codegen.core.KotlinWriter
 import software.amazon.smithy.kotlin.codegen.core.RuntimeTypes
 import software.amazon.smithy.kotlin.codegen.core.withBlock
-import software.amazon.smithy.kotlin.codegen.model.hasTrait
 import software.amazon.smithy.kotlin.codegen.model.isEnum
 import software.amazon.smithy.kotlin.codegen.model.targetOrSelf
-import software.amazon.smithy.kotlin.codegen.model.traits.OperationInput
-import software.amazon.smithy.kotlin.codegen.model.traits.OperationOutput
 import software.amazon.smithy.kotlin.codegen.utils.dq
 import software.amazon.smithy.kotlin.codegen.utils.getOrNull
 import software.amazon.smithy.kotlin.codegen.utils.toCamelCase
 import software.amazon.smithy.model.shapes.ListShape
 import software.amazon.smithy.model.shapes.MapShape
-import software.amazon.smithy.model.shapes.MemberShape
 import software.amazon.smithy.model.shapes.Shape
 import kotlin.contracts.ExperimentalContracts
 import kotlin.contracts.contract
@@ -88,7 +84,7 @@ class KotlinJmespathExpressionVisitor(
         if (right is CurrentExpression) return VisitedExpression(leftName, leftShape) // nothing to map
 
         val outerName = bestTempVarName("projection")
-        val flatMapExpr = ensureNullGuard(leftShape, "flatMap")
+        val flatMapExpr = addNullGuard(leftShape, "flatMap")
         writer.openBlock("val #L = #L#L {", outerName, leftName, flatMapExpr)
 
         shapeCursor.addLast(innerShape?.targetMemberOrSelf ?: leftShape.targetMemberOrSelf)
@@ -96,7 +92,7 @@ class KotlinJmespathExpressionVisitor(
         shapeCursor.removeLast()
 
         val innerCollector = when (right) {
-            is MultiSelectListExpression -> innerResult.identifier // Already a list
+            is MultiSelectListExpression, is MultiSelectHashExpression -> innerResult.identifier // Already a list
             else -> "listOfNotNull(${innerResult.identifier})"
         }
         writer.write(innerCollector)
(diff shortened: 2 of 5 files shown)


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -368,20 +400,12 @@ class KotlinJmespathExpressionVisitor(
...
-                append("?.$expr")
-                elvisExpr?.let { append(" ?: $it") }
-            }
-        } else {
-            ".$expr"
+    private fun addNullGuard(shape: Shape?, expr: String, elvisExpr: String? = null): String =
+        buildString {
+            append("?.$expr")
+            elvisExpr?.let { append(" ?: $it") }
         }
```
ianbotsf, line 407: **Question**: Why are we adding null guards to potentially-non-null expressions? We [made the codegen visitor null-aware](https://github.com/awslabs/smithy-kotlin/pull/842) in response to a waiters bug and I worry that regressing that behavior could introduce new bugs.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/AcceptorGenerator.kt
```diff
@@ -89,12 +89,12 @@ private fun KotlinWriter.renderPathAcceptor(wi: WaiterInfo, directive: String, i
...
             PathComparator.BOOLEAN_EQUALS -> "${actual.identifier} == ${expected.toBoolean()}"
-            PathComparator.ANY_STRING_EQUALS -> "${actual.identifier}?.any { it == ${expected.dq()} } ?: false"
+            PathComparator.ANY_STRING_EQUALS -> "(${actual.identifier} as List<String>?)?.any { it == ${expected.dq()} } ?: false"
 
             // NOTE: the isNotEmpty check is necessary because the waiter spec says that `allStringEquals` requires
             // at least one value unlike Kotlin's `all` which returns true if the collection is empty
             PathComparator.ALL_STRING_EQUALS ->
-                "!${actual.identifier}.isNullOrEmpty() && ${actual.identifier}.all { it == ${expected.dq()} }"
+                "!(${actual.identifier} as List<String>).isNullOrEmpty() && ${actual.identifier}.all { it == ${expected.dq()} }"
         }
```
ianbotsf, line 98: **Question**: Why is this new cast to `List<String>` necessary?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -107,26 +103,29 @@ class KotlinJmespathExpressionVisitor(
...
+        if (member != null) {
+            val memberTarget = ctx.model.expectShape(member.target)
+            unwrapExpr = when {
+                memberTarget.isEnum -> "value"
+                memberTarget.isEnumList -> "map { it.value }"
+                memberTarget.isEnumMap -> "mapValues { (_, v) -> v.value }"
+                memberTarget.isBlobShape || memberTarget.isTimestampShape ->
+                    throw CodegenException("acceptor behavior for shape type ${memberTarget.type} is undefined")
+                else -> null
+            }
```
lauzadis, line 120: To avoid using a `var` here, you could use a null-check  on `member` like this:

```kt
val unwrapExpr = member?.let {
	val memberTarget = ctx.model.expectShape(member.target)
	when {
		memberTarget.isEnum -> "value"
		memberTarget.isEnumList -> "map { it.value }"
		memberTarget.isEnumMap -> "mapValues { (_, v) -> v.value }"
		memberTarget.isBlobShape || memberTarget.isTimestampShape ->
			throw CodegenException("acceptor behavior for shape type ${memberTarget.type} is undefined")
                else -> null
	}
}
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -107,26 +103,29 @@ class KotlinJmespathExpressionVisitor(
...
         }
+
         val codegen = buildString {
             append("$parentName$nameExpr")
-            unwrapExpr?.let { append(ensureNullGuard(member, it)) }
+            unwrapExpr?.let { append(addNullGuard(member, it)) }
         }
 
-        shapeCursor.addLast(member)
+        if (member != null) shapeCursor.addLast(member)
```
lauzadis, line 128: Same comment here about using the Kotlin null helper methods: `member?.let { shapeCursor.addLast(it) }`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -244,12 +243,44 @@ class KotlinJmespathExpressionVisitor(
...
+        val attributes = StringBuilder()
+        var i = 0
+        expression.expressions.forEach {
+            if (i == expression.expressions.size - 1) {
+                attributes.append("val ${it.key}: T")
+            } else {
+                attributes.append("val ${it.key}: T, ")
+            }
+            i++
+        }
```
lauzadis, line 255: Kotlin has a [`joinToString` function](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/join-to-string.html) which handles this nicely. The default is a comma-separated string so you don't need to customize the parameters, just the string transformation.

```kt
val attributes = expression.expressions.joinToString {
	"val ${it.key}: T" 
}
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -244,12 +243,44 @@ class KotlinJmespathExpressionVisitor(
...
+        i = 0
+        expression.expressions.forEach {
+            val identifier = addTempVar(it.key, acceptSubexpression(it.value).identifier)
+            if (i == expression.expressions.size - 1) {
+                identifiers.append(identifier)
+            } else {
+                identifiers.append("$identifier, ")
+            }
+            i++
+        }
```
lauzadis, line 272: Same comment about `joinToString`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/AcceptorGenerator.kt
0marperez: Long explanation but when multi select hash makes a selection that selection is stored in an object. Something like this: `class Selection<T>(val x: T, val y: T)`. 

The `.isNullOrEmpty()` function can only be applied to arrays, lists and such. 

So at runtime when it was detecting the selection properties as generic (x or y) I was getting errors (because x or y didn't show up as a list, array, etc), that cast fixed them.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
0marperez: Thanks, I wish I knew about this earlier

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -244,20 +246,41 @@ class KotlinJmespathExpressionVisitor(
     }
 
     override fun visitMultiSelectHash(expression: MultiSelectHashExpression): VisitedExpression {
-        throw CodegenException("MultiSelectHashExpression is unsupported")
+        val properties = mutableListOf<String>()
+        expression.expressions.forEach {
+            properties.add("val ${it.key}: T")
+        }
+        writer.write("class Selection<T>(${properties.joinToString()})")
```
ianbotsf, line 253: **Nit**: Mutable lists are rarely necessary within a block. This, for instance, can be written as a single `joinToString` translation:

```kotlin
val properties = expression.expressions.joinToString { "val ${it.key}: T" }
writer.write("class Selection<T>($properties)")
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -244,20 +246,41 @@ class KotlinJmespathExpressionVisitor(
...
+        val listName = bestTempVarName("multiSelect")
+        writer.openBlock("val #L = listOfNotNull(", listName)
+        writer.openBlock("run {")
+
+        val identifiers = mutableListOf<String>()
+        expression.expressions.forEach {
+            val identifier = addTempVar(it.key, it.value.accept(this).identifier)
+            identifiers.add(identifier)
+        }
+        writer.write("Selection(${identifiers.joinToString()})")
```
ianbotsf, line 264: **Nit**: Also could be simplified to avoid mutability:

```kotlin
val identifiers = expression.expressions.joinToString { addTempVar(it.key, it.value.accept(this).identifier) }
writer.write("Selection($identifiers)")
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -244,20 +246,41 @@ class KotlinJmespathExpressionVisitor(
...
+
+        writer.closeBlock("},")
+        writer.closeBlock(")")
```
ianbotsf, line 267: **Style**: Favor `withBlock` over explicit `openBlock`/`closeBlock`:

```kotlin
writer.withBlock("val #L = listOfNotNull(", ")", listName) {
    withBlock("run {", "},") {
        // ...generate `Selection` class...
    }
}
```

ianbotsf (approved):
Minor feedback, fix and ship!

PR #932, opened 2023-08-21
refactor(rt): remove client options

aajtodd:
## Issue \#

n/a

## Description of changes

Minor cleanup to remove `ClientOption` alias and `ClientOptionsBuilder`. These originally had wider use cases planned but we have since gone in different directions. This PR scrubs these types from our public API and updates the one place they were used for anything.


lauzadis (approved):

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/operation/HttpOperationContext.kt
```diff
@@ -66,22 +63,30 @@ public open class HttpOperationContext {
      * Convenience builder for constructing HTTP client operations
      */
     @InternalApi
-    public open class Builder : ClientOptionsBuilder() {
-
+    public class Builder(private val ctx: ExecutionContext = ExecutionContext()) : MutableAttributes by ctx {
```
ianbotsf, line 66: **Question**: It's not clear to me what value this `Builder` class adds to `ExecutionContext` nor why it's an inner class of `HttpOperationContext`. How is this better than using `ExecutionContext` directly?

aajtodd: Thats fair and I tend to agree but I was trying to limit the scope of the changes since we use the current builder type and the `var` fields therein in a lot of tests as well as codegen. We can always change this later since it's internal but I wanted to remove `ClientOption` from the public API since it was never fleshed out and felt incomplete.

aajtodd: That being said we can refactor further if you feel we should just do it now.

ianbotsf: I think we should just do it now.

aajtodd: How do you envision we set things like `operationName` now? Just direclty via `set(HttpOperationContext.OperationName, "<name>")`, furtherwhere do we validate the required-ness of them.

ianbotsf: Yes, I think `set()` works just fine. Offhand, I'd guess the best place to verify they're set is in [`SdkHttpOperationBuilder.build`](https://github.com/awslabs/smithy-kotlin/blob/0e9417298e8a24886879ac84238c58d6c9ee5eb4/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/operation/SdkHttpOperation.kt#L127-L132).

ianbotsf (approved):

PR #933, opened 2023-08-22
fix: lower engine concurrency to closer to what it was when test was written to fix windows issue

aajtodd:
## Issue \#

n/a

## Description of changes

Lower the engine concurrency to closer to what it was when this test was written (16 for okhttp). This is a band aid to fix the flakiness of this test on windows CI since I haven't been able to reproduce it locally.

### DIFF
--- a/runtime/protocol/http-client-engines/test-suite/common/test/aws/smithy/kotlin/runtime/http/test/AsyncStressTest.kt
+++ b/runtime/protocol/http-client-engines/test-suite/common/test/aws/smithy/kotlin/runtime/http/test/AsyncStressTest.kt
@@ -10,6 +10,7 @@ import aws.smithy.kotlin.runtime.http.readAll
 import aws.smithy.kotlin.runtime.http.request.HttpRequest
 import aws.smithy.kotlin.runtime.http.response.complete
 import aws.smithy.kotlin.runtime.http.test.util.AbstractEngineTest
+import aws.smithy.kotlin.runtime.http.test.util.engineConfig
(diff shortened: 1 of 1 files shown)


runtime/protocol/http-client-engines/test-suite/common/test/aws/smithy/kotlin/runtime/http/test/AsyncStressTest.kt
```diff
@@ -22,6 +23,9 @@ class AsyncStressTest : AbstractEngineTest() {
     fun testConcurrentRequests() = testEngines {
         // https://github.com/awslabs/aws-sdk-kotlin/issues/170
         concurrency = 1_000
+        engineConfig {
+            maxConcurrency = 32u
```
lauzadis, line 27: Your PR description says 16 but this is 32, which is correct?

lauzadis (approved):

runtime/protocol/http-client-engines/test-suite/common/test/aws/smithy/kotlin/runtime/http/test/AsyncStressTest.kt
aajtodd: It was 16 I bumped it to 32 (hence the _closer_). We can move it to 16 or continue increasing it until it breaks. I didn't realize it was 16 until I looked it up after testing it.

ianbotsf (approved):
If this is a bandaid fix, let's leave a `FIXME` in here and possibly a backlog item.

PR #937, opened 2023-08-23
feat: add functions that operate on numbers (abs, floor, ceil) to JMESPath visitor

### LINKED ISSUE #936: Add support for functions that take in numbers to JMESPath visitor (abs, floor, ceil)

0marperez:
## Issue \#

closes #936 

## Description of changes

-Added functions that operate on numbers (abs, floor and ceil)
-Small fix for something I missed in the index function added as well


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -226,6 +226,39 @@ class KotlinJmespathExpressionVisitor(
             VisitedExpression(ident)
         }
 
+        "abs" -> {
+            codegenReq(expression.arguments.size == 1) { "Unexpected number of arguments to $expression" }
+
+            val number = acceptSubexpression(expression.arguments[0])
```
ianbotsf, line 232: **Suggestion**: Now that we have more than two functions, a pattern begins to emerge in how we handle arguments. I think we could create extension functions which shorten/simplify getting the args based on arity:

```kotlin
fun ExpressionFunction.singleArg(): JmespathExpression {
    codegenReq(arguments.size == 1) { "Unexpected number of arguments to $this" }
    return acceptSubexpression(arguments[0])
}

fun ExpressionFunction.twoArgs(): Pair<JmespathExpression, JmespathExpression> {
    codegenReq(arguments.size == 2) { "Unexpected number of arguments to $this" }
    return acceptSubexpression(arguments[0]) to acceptSubexpression(arguments[1])
}
```

Then we could clean up the args parts of the `when` branches in `visitFunction`:

```kotlin
override fun visitFunction(expression: FunctionExpression): VisitedExpression = when (expression.name) {
    "contains" -> {
        val (subject, search) = expression.twoArgs()
        ...
    }

    "length" -> {
        val subject = expression.singleArg()
        ...
    }

    ...
}
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -226,6 +226,39 @@ class KotlinJmespathExpressionVisitor(
...
+        "ceil" -> {
+            codegenReq(expression.arguments.size == 1) { "Unexpected number of arguments to $expression" }
+
+            val number = acceptSubexpression(expression.arguments[0])
+
+            val absExpr = ensureNullGuard(number.shape, "let { kotlin.math.ceil(it.toDouble()) }")
+            val ident = addTempVar("ceil", "${number.identifier}$absExpr")
+
+            VisitedExpression(ident, number.shape)
+        }
```
ianbotsf, line 260: **Nit**: By happy coincidence, the JMESPath names for these functions are the same in Kotlin. We can take advantage of that to collapse these three branches into one:

```kotlin
override fun visitFunction(expression: FunctionExpression): VisitedExpression = when (expression.name) {
    "contains" -> ...
    "length" -> ...

    "abs", "floor", "ceil" -> {
        codegenReq(expression.arguments.size == 1) { "Unexpected number of arguments to $expression" }

        val number = acceptSubexpression(expression.arguments[0])

        val expr = ensureNullGuard(number.shape, "let { kotlin.math.${expression.name}(it.toDouble()) }")
        val ident = addTempVar(expression.name, "${number.identifier}$expr")

        VisitedExpression(ident, number.shape)
    }
}
```

tests/codegen/waiter-tests/src/test/kotlin/com/test/WaiterTest.kt
```diff
@@ -1014,6 +1014,112 @@ class WaiterTest {
         },
     )
 
+    // function abs
+    @Test fun testAbsFunctionShortEquals() = successTest(
+        WaitersTestClient::waitUntilAbsFunctionShortEquals,
+        GetEntityResponse { primitives = EntityPrimitives { short = -1 } },
+        GetEntityResponse { primitives = EntityPrimitives { short = 1 } },
+        GetEntityResponse { primitives = EntityPrimitives { short = 2 } },
+    )
```
ianbotsf, line 1023: **Suggestion**: Our test code and our test model are too big. I suggest looking into ways to split up the code/model by JMESPath feature. For instance, we could have separate test classes for `abs`, `ceil`, `floor`, primitive equality, comparators, list indexing, etc.

Similarly, we could break up the model into smaller files. This would be a lot trickier because right now all these waiters exist on a single operation in a single service. We could consider separate services/operations for each group. We could also consider defining multiple files that just contain groups of waiter definitions and then stitch them together into **waiter-operations.smithy** during the build process.

ianbotsf (approved):

PR #938, opened 2023-08-24
fix: cancel `HttpCall` if closing the response body fails

lauzadis:
See https://github.com/awslabs/smithy-kotlin/issues/935 for the full context.

This PR fixes an `IllegalStateException` that's thrown when trying to close the `engineResponse.body` while a different coroutine is blocked trying to `read()` the body. 

The fix works by checking if closing the response body fails (throwing `IllegalStateException`). If it fails, then we can first cancel the `Call` (as [recommended by OkHttp devs](https://github.com/square/okhttp/issues/4110#issuecomment-402436150)), wait a short period of time, and *then* close the response body successfully.

## Issue \#

Closes https://github.com/awslabs/smithy-kotlin/issues/935

## Description of changes

This change is required to support closing event streams before they have been fully consumed.

### DIFF
--- a/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
@@ -51,7 +51,8 @@ public class OkHttpEngine(
         val engineResponse = mapOkHttpExceptions { engineCall.executeAsync() }
 
         callContext.job.invokeOnCompletion {
-            engineResponse.body.close()
+            engineCall.cancel()
+            runCatching { engineResponse.body.close() }
         }
 
         val response = engineResponse.toSdkResponse()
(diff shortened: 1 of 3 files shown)


runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
```diff
@@ -51,7 +55,14 @@ public class OkHttpEngine(
         val engineResponse = mapOkHttpExceptions { engineCall.executeAsync() }
 
         callContext.job.invokeOnCompletion {
-            engineResponse.body.close()
+            // FIXME: https://github.com/awslabs/smithy-kotlin/issues/935
+            val closeResult = runCatching { engineResponse.body.close() }
+
+            if (closeResult.exceptionOrNull() is IllegalStateException) {
+                engineCall.cancel()
+                runBlocking { delay(1.nanoseconds) }
```
aajtodd, line 63: [invokeOnCompletion](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-job/invoke-on-completion.html) shouldn't block according to the documentation.

lauzadis: Hmm... this might be difficult to work around then. Without this `delay` the IllegalStateException is still thrown on `.close()`, even when running `engineCall.cancel()` beforehand.

aajtodd: Yeah I _think_ the way it might work if we can get the following in place:
 1. tie the background coroutine to the call scope
 2. detect when we need to call `cancel` on the call (will require making a call an interface or something so that we can abstract this per/engine)
 3. in `HttpCall.complete` call cancel when necessary _before_ calling `join()` on the call context job. 

This would make the existing completion handler that closes the response body only be invoked _after_ the corouting doing the `read()` has exited.

This is obviously a bigger change.

lauzadis:
Fixed by https://github.com/awslabs/smithy-kotlin/pull/939

PR #940, opened 2023-08-29
feat: add functions (sum, avg, join, starts_with, ends_with) to JMESPath visitor

0marperez:
## Description of changes

Added:
-Sum
-Avg
-Join
-Starts with
-Ends with

Refactored:
-Separated operations from structures in smithy file
-Wrote function to avoid code duplication in JMESPath visitor functions code

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
@@ -213,33 +213,53 @@ class KotlinJmespathExpressionVisitor(
         return acceptSubexpression(this.arguments[0]) to acceptSubexpression(this.arguments[1])
     }
 
+    private fun VisitedExpression.dotFunction(expression: FunctionExpression, expr: String, elvisExpr: String? = null): VisitedExpression {
+        val expr = ensureNullGuard(this.shape, expr, elvisExpr)
+        val ident = addTempVar(expression.name, "${this.identifier}$expr")
+
+        return VisitedExpression(ident, this.shape)
+    }
+
     override fun visitFunction(expression: FunctionExpression): VisitedExpression = when (expression.name) {
         "contains" -> {
             val (subject, search) = expression.twoArgs()
-
-            val containsExpr = ensureNullGuard(subject.shape, "contains(${search.identifier})", "false")
-            val ident = addTempVar("contains", "${subject.identifier}$containsExpr")
-
-            VisitedExpression(ident)
+            subject.dotFunction(expression, "contains(${search.identifier})", "false")
         }
 
         "length" -> {
             writer.addImport(RuntimeTypes.Core.Utils.length)
             val subject = expression.singleArg()
-
-            val lengthExpr = ensureNullGuard(subject.shape, "length", "0")
-            val ident = addTempVar("length", "${subject.identifier}$lengthExpr")
-
-            VisitedExpression(ident)
+            subject.dotFunction(expression, "length", "0")
         }
 
         "abs", "floor", "ceil" -> {
             val number = expression.singleArg()
+            number.dotFunction(expression, "let { kotlin.math.${expression.name}(it.toDouble()) }")
+        }
 
-            val functionExpr = ensureNullGuard(number.shape, "let { kotlin.math.${expression.name}(it.toDouble()) }")
-            val ident = addTempVar(expression.name, "${number.identifier}$functionExpr")
+        "sum" -> {
+            val numbers = expression.singleArg()
+            numbers.dotFunction(expression, "sum()")
+        }
 
-            VisitedExpression(ident, number.shape)
+        "avg" -> {
+            val numbers = expression.singleArg()
+            numbers.dotFunction(expression, "average()")
+        }
+
+        "join" -> {
+            val (glue, list) = expression.twoArgs()
+            list.dotFunction(expression, "joinToString(${glue.identifier})")
+        }
+
+        "starts_with" -> {
+            val (subject, prefix) = expression.twoArgs()
+            subject.dotFunction(expression, "startsWith(${prefix.identifier})")
+        }
+
+        "ends_with" -> {
+            val (subject, suffix) = expression.twoArgs()
+            subject.dotFunction(expression, "endsWith(${suffix.identifier})")
         }
 
         else -> throw CodegenException("Unknown function type in $expression")
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
@@ -159,7 +159,7 @@ class KotlinJmespathExpressionVisitor(
             }
 
             val unSafeComparatorExpr = "compareTo(${right.identifier}) ${expression.comparator} 0"
-            val comparatorExpr = if (left.nullable) buildString { append("?.$unSafeComparatorExpr") } else buildString { append(".$unSafeComparatorExpr") }
+            val comparatorExpr = if (left.nullable) "?.$unSafeComparatorExpr" else ".$unSafeComparatorExpr"
 
             append("${left.identifier}$comparatorExpr")
         }
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
@@ -158,7 +158,10 @@ class KotlinJmespathExpressionVisitor(
                 append("if ($isNullExpr) null else ")
             }
 
-            append("${left.identifier}.compareTo(${right.identifier}) ${expression.comparator} 0")
+            val unSafeComparatorExpr = "compareTo(${right.identifier}) ${expression.comparator} 0"
+            val comparatorExpr = if (left.nullable) buildString { append("?.$unSafeComparatorExpr") } else buildString { append(".$unSafeComparatorExpr") }
+
+            append("${left.identifier}$comparatorExpr")
         }
 
         val ident = addTempVar("comparison", codegen)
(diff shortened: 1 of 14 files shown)


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -213,33 +213,53 @@ class KotlinJmespathExpressionVisitor(
...
-            val ident = addTempVar(expression.name, "${number.identifier}$functionExpr")
+        "sum" -> {
+            val numbers = expression.singleArg()
+            numbers.dotFunction(expression, "sum()")
+        }
+
+        "avg" -> {
+            val numbers = expression.singleArg()
+            numbers.dotFunction(expression, "average()")
+        }
```
ianbotsf, line 248: **Correctness**: I don't think this matches the [JMESPath spec for `avg`](https://jmespath.org/specification.html#avg) given an empty array. It should return `null` but I believe this returns `NaN`.

tests/codegen/waiter-tests/model/function-avg.smithy
```diff
@@ -0,0 +1,84 @@
...
+                matcher: {
+                    output: {
+                        path: "avg(lists.shorts) == `10`",
+                        expected: "false",
+                        comparator: "booleanEquals"
+                    }
+                }
+            }
+        ]
+    },
```
ianbotsf, line 20: **Nit**: The names of these waiters don't match the matcher conditions. They're all called "AvgEquals" but the success criteria are when the averages _do not equal_ 10.

tests/codegen/waiter-tests/model/function-sum.smithy
```diff
@@ -0,0 +1,84 @@
...
+                matcher: {
+                    output: {
+                        path: "sum(lists.shorts) == `10`",
+                        expected: "false",
+                        comparator: "booleanEquals"
+                    }
+                }
+            }
+        ]
+    },
```
ianbotsf, line 20: **Nit**: Same note as above about waiter names vs success criteria.

tests/codegen/waiter-tests/src/main/kotlin/com/test/DefaultWaitersTestClient.kt
```diff
@@ -89,4 +89,34 @@ class DefaultWaitersTestClient<T>(resultList: List<Result<T>>) : WaitersTestClie
         @Suppress("UNCHECKED_CAST")
         return if (nextResult.isSuccess) (nextResult as Result<GetSubFieldProjectionResponse>).getOrNull()!! else throw nextResult.exceptionOrNull()!!
     }
+
+    override suspend fun getFunctionSumEquals(input: GetFunctionSumEqualsRequest): GetFunctionSumEqualsResponse {
+        val nextResult = results.next()
+        @Suppress("UNCHECKED_CAST")
+        return if (nextResult.isSuccess) (nextResult as Result<GetFunctionSumEqualsResponse>).getOrNull()!! else throw nextResult.exceptionOrNull()!!
```
ianbotsf, line 96: **Nit**: Seems like we could write a generic convenience method to collapse all this duplication in every operation.

ianbotsf (changes requested):

lauzadis (approved):
Looks good! I had no more comments outside of what Ian had already, especially the comment about a generic `<T>` convenience method

0marperez:
About the `avg` function test for an empty list (`EmptyIntegerListAvgNotEquals`). JMESPath doesn't have a way to check if something is null so I just compared it to a number and set the success criteria to false. You can verify this behavior by playing around with JMESPath on their website and I also tried it in a waiter btw.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -158,7 +158,10 @@ class KotlinJmespathExpressionVisitor(
                 append("if ($isNullExpr) null else ")
             }
 
-            append("${left.identifier}.compareTo(${right.identifier}) ${expression.comparator} 0")
+            val unSafeComparatorExpr = "compareTo(${right.identifier}) ${expression.comparator} 0"
+            val comparatorExpr = if (left.nullable) buildString { append("?.$unSafeComparatorExpr") } else buildString { append(".$unSafeComparatorExpr") }
```
ianbotsf, line 162: **Nit**: `buildString { append(foo) }` can be rewritten as `foo`:

```kotlin
val comparatorExpr = if (left.nullable) "?.$unSafeComparatorExpr" else ".$unSafeComparatorExpr"
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -158,7 +158,10 @@ class KotlinJmespathExpressionVisitor(
...
+
+            append("${left.identifier}$comparatorExpr")
```
ianbotsf, line 164: **Nit**: `buildString { append(foo) }` can be rewritten as `foo`:

```kotlin
val comparatorExpr = if (left.nullable) "?.$unSafeComparatorExpr" else ".$unSafeComparatorExpr"
```

`buildString` is still a nice function and may be useful here if extended around the wider logic:

```kotlin
val comparison = buildString {
    append(left.identifier)
    if (left.nullable) append('?')
    append(".compareTo(${right.identifier}) ${expression.comparator} 0")
}
append(comparison)
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -213,33 +216,55 @@ class KotlinJmespathExpressionVisitor(
...
+            val numbers = expression.singleArg()
+            numbers.dotFunction(expression, "sum()")
+        }
+
+        "avg" -> {
+            val numbers = expression.singleArg()
+            val average = numbers.dotFunction(expression, "average()").identifier
+            val isNaN = ensureNullGuard(numbers.shape, "isNaN() == true")
+            VisitedExpression(addTempVar("averageOrNull", "if($average$isNaN) null else $average"), nullable = true)
+        }
```
ianbotsf, line 253: **Suggestion**: [`takeUnless`](https://kotlinlang.org/docs/scope-functions.html#takeif-and-takeunless) can simplify this for you:

```kotlin
"avg" -> {
    val numbers = expression.singleArg()
    numbers.dotFunction(expression, "average()").dotFunction(expression, "takeUnless { it.isNan() }")
}
```

Chaining `dotFunction` twice with the same `expression` is a little cheesy because it would result in codegen like:

```kotlin
val average = theList?.average()
val average2 = average?.takeUnless { it.isNan() }
```

If you go this route, it might be worth adding an overload of `dotFunction` that can take a string:

```kotlin
private fun VisitedExpression.dotFunction(preferredName: String, expr: String, elvisExpr: String? = null): VisitedExpression {
    val dotFunctionExpr = ensureNullGuard(shape, expr, elvisExpr)
    val ident = addTempVar(preferredName, "$identifier$dotFunctionExpr")
    
    return VisitedExpression(ident, shape)
}

private fun VisitedExpression.dotFunction(expression: FunctionExpression, expr: String, elvisExpr: String? = null) =
    dotFunction(expression.name, expr, elvisExpr)

...

"avg" -> {
    val numbers = expression.singleArg()
    numbers.dotFunction(expression, "average()").dotFunction("averageOrNull, "takeUnless { it.isNan() }")
}
```

tests/codegen/waiter-tests/src/main/kotlin/com/test/DefaultWaitersTestClient.kt
```diff
@@ -18,75 +18,43 @@ class DefaultWaitersTestClient<T>(resultList: List<Result<T>>) : WaitersTestClie
...
+
+    override suspend fun getFunctionEndsWithEquals(input: GetFunctionEndsWithEqualsRequest): GetFunctionEndsWithEqualsResponse = findSuccess()
 
-    override suspend fun getSubFieldProjection(input: GetSubFieldProjectionRequest): GetSubFieldProjectionResponse {
+    private fun <Response> findSuccess(): Response {
         val nextResult = results.next()
         @Suppress("UNCHECKED_CAST")
-        return if (nextResult.isSuccess) (nextResult as Result<GetSubFieldProjectionResponse>).getOrNull()!! else throw nextResult.exceptionOrNull()!!
+        return if (nextResult.isSuccess) (nextResult as Result<Response>).getOrNull()!! else throw nextResult.exceptionOrNull()!!
     }
```
ianbotsf, line 59: **Comment**: These look great now!

Also, `findSuccess` is good advice for us all! 😆

ianbotsf (approved):

ianbotsf:
> JMESPath doesn't have a way to check if something is null...

I don't understand. The [spec's section on equality](https://jmespath.org/specification.html?highlight=preserve%5C%20structure#equality-operators) seems to indicate that `null` is _only_ equal to itself. You weren't able to get an equality path to pass `booleanEquals`?

```
matcher: {
    output: {
        path: "avg(lists.integers) == null",
        expected: "true",
        comparator: "booleanEquals"
    }
}
```

0marperez:
No, when trying it in the waiter I get an error message:
```
6 | @waitable(
  | ^
··|
94| operation GetFunctionAvgEquals {

Waiter `EmptyIntegerListAvgEquals`, acceptor 0: Problem found in JMESPath
expression (avg(lists.integers) == null): Object field 'null' does not exist in
object with properties [primitives, lists, twoDimensionalLists, maps] (1:24)
```

And in the JMESPath website you can see null is there:
Screenshot 2023-08-30 at 10 25 46 AM
But when comparing it to another null it doesn't work:
Screenshot 2023-08-30 at 10 26 02 AM
Screenshot 2023-08-30 at 10 26 17 AM

ianbotsf:
> No, when trying it in the waiter I get an error message:
> 
> ```
> 6 | @waitable(
>   | ^
> ··|
> 94| operation GetFunctionAvgEquals {
> 
> Waiter `EmptyIntegerListAvgEquals`, acceptor 0: Problem found in JMESPath
> expression (avg(lists.integers) == null): Object field 'null' does not exist in
> object with properties [primitives, lists, twoDimensionalLists, maps] (1:24)
> ```

Interesting. It looks like Smithy's JMESPath type checker is getting to [`visitField`](https://github.com/smithy-lang/smithy/blob/14d4d6887159f39b05c5795c8cc63e045c200cd4/smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/TypeChecker.java#L160) when I'd expect it to be hitting [`visitLiteral`](https://github.com/smithy-lang/smithy/blob/14d4d6887159f39b05c5795c8cc63e045c200cd4/smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/TypeChecker.java#L195). I suspect this might be a bug in Smithy and I recommend you file an issue to them.

Looks like what you've got is the best we can do for now!

PR #945, opened 2023-08-31
feat(rt): add InputStream adapter for ByteStream

aajtodd:
## Issue \#

closes https://github.com/awslabs/aws-sdk-kotlin/issues/617

## Description of changes

Adds a conversion to go from `ByteStream` to `java.io.InputStream`

### DIFF
--- a/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannelJVM.kt
+++ b/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannelJVM.kt
@@ -1,0 +1,53 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.io
+
+import kotlinx.coroutines.runBlocking
+import java.io.InputStream
+import java.util.*
+
+/**
+ * Create a blocking [InputStream] that blocks everytime the channel suspends at [SdkByteReadChannel.read]
+ */
+public fun SdkByteReadChannel.toInputStream(): InputStream = InputAdapter(this)
+
+private const val DEFAULT_READ_BYTES = 8192L
+private class InputAdapter(private val ch: SdkByteReadChannel) : InputStream() {
+
+    private val buffer = SdkBuffer()
+
+    override fun read(): Int {
+        if (ch.isClosedForRead && buffer.size == 0L) return -1
+
+        if (buffer.size == 0L) {
+            val rc = readBlocking()
+            if (rc == -1L) return -1
+        }
+
+        return buffer.readByte().toInt() and 0xff
+    }
+    override fun read(b: ByteArray, off: Int, len: Int): Int {
+        Objects.checkFromIndexSize(off, len, b.size)
+        if (ch.isClosedForRead && buffer.size == 0L) return -1
+        if (buffer.size == 0L) {
+            val rc = readBlocking()
+            if (rc == -1L) return -1
+        }
+
+        return buffer.read(b, off, len)
+    }
+
+    private fun readBlocking(): Long =
+        runBlocking {
+            ch.read(buffer, DEFAULT_READ_BYTES)
+        }
+
+    override fun available(): Int = ch.availableForRead
+
+    override fun close() {
+        super.close()
+        ch.cancel(null)
+    }
+}
--- a/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannelJVM.kt
+++ b/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannelJVM.kt
@@ -28,6 +28,7 @@ private class InputAdapter(private val ch: SdkByteReadChannel) : InputStream() {
 
         return buffer.readByte().toInt() and 0xff
     }
+
     override fun read(b: ByteArray, off: Int, len: Int): Int {
         if (off < 0 || len < 0 || len > b.size - off) {
             throw IndexOutOfBoundsException()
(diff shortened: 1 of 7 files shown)


runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannelJVM.kt
```diff
@@ -0,0 +1,53 @@
...
+            if (rc == -1L) return -1
+        }
+
+        return buffer.read(b, off, len)
+    }
+
+    private fun readBlocking(): Long =
+        runBlocking {
+            ch.read(buffer, DEFAULT_READ_BYTES)
+        }
```
ianbotsf, line 45: **Question**: Looks like we _only_ ever read `DEFAULT_READ_BYTES` (8K) bytes at a time from the channel, even when the `read` call may have had a `len` higher than that. Given that `runBlocking` may be expensive, is it worth passing the requested length through to the channel read? Or making the chunk size configurable?

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannelJVM.kt
```diff
@@ -0,0 +1,53 @@
...
+import java.io.InputStream
+import java.util.*
+
+/**
+ * Create a blocking [InputStream] that blocks everytime the channel suspends at [SdkByteReadChannel.read]
+ */
+public fun SdkByteReadChannel.toInputStream(): InputStream = InputAdapter(this)
+
+private const val DEFAULT_READ_BYTES = 8192L
+private class InputAdapter(private val ch: SdkByteReadChannel) : InputStream() {
```
ianbotsf, line 17: **Question**: This adapter works via `runBlocking` on every `read`. That will work obvi but seems like it may be slow. Would it be better to spin up a background coroutine to handle reads that lives for the lifetime of the adapter?

lauzadis, line 17: Should we support `mark` / `reset`? I think it's not supported by default (`markSupported()` defaults to `false`). It can probably be additive work if requested

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/content/ByteStreamInputStreamTest.kt
```diff
@@ -0,0 +1,88 @@
...
+                override fun readFrom(): SdkByteReadChannel = SdkByteReadChannel(input)
+                override val contentLength: Long = input.size.toLong()
+            }.toInputStream()
+        }
+    }
+}
+
+class ByteStreamBufferInputStreamTest : ByteStreamInputStreamTest(ByteStreamFactory.BYTE_ARRAY)
+class ByteStreamSourceStreamInputStreamTest : ByteStreamInputStreamTest(ByteStreamFactory.SDK_SOURCE)
+class ByteStreamChannelSourceInputStreamTest : ByteStreamInputStreamTest(ByteStreamFactory.SDK_CHANNEL)
```
ianbotsf, line 38: **Comment**: Nice abstraction.

lauzadis (approved):

ianbotsf (approved):

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannelJVM.kt
aajtodd: `InputStream.read` is a blocking call. `runBlocking` is designed to bridge blocking and non-blocking worlds (it does so by blocking the current thread). Spinning up a coroutine will result in tying up two threads and/or more context switches to accomplish the same thing.

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannelJVM.kt
aajtodd: Why do you feel that `runBlocking` is expensive?


------

The [len](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#read-byte:A-int-int-) argument is documented as

> len - the maximum number of bytes to read.

We could change it but I chose 8K because thats the underlying segment size used by okio. I'm not sure this will matter all that much in practice but if you disagree I'm happy to change it.

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/SdkByteReadChannelJVM.kt
aajtodd: I think I'd vote to make it additive based on customer feedback/requests. Anyone feel different?

ianbotsf: Sounds good to me.

ianbotsf: As far as I understand (could be wrong, ofc) [`runBlocking` _starts a new coroutine_ and blocks the current thread](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/run-blocking.html#:~:text=Runs%20a%20new%20coroutine%20and%20blocks%20the%20current%20thread%20until%20its%20completion). Doing that on every 8KB chunk _sounds_ slower than starting a new coroutine _once_. Sure, it'll still block on every `read` call but the cost of setting up a new coroutine was already paid.

aajtodd: TL;DR I'm not at all worried about the overhead of `runBlocking` or starting a new coroutine.

----
`runBlocking` does start a new coroutine and blocks the current thread. This is fine though, coroutines are extremely lightweight. The only benchmarking I can find is [here](https://stackoverflow.com/questions/50230466/kotlin-withcontext-vs-async-await/50231191#50231191) which would suggest **140** byte allocation and **100 nanoseconds**. 

Introducing a background coroutine will likely introduce additional context switches and a more complicated implementation to synchronize the coroutine and blocking `read` call. It also needs a `CoroutineScope` to launch into which would also mean taking an additional parameter and dealing with cancellation, error propagation, etc.

PR #952, opened 2023-09-14
feat: add functions (min, max, to_array, to_string, to_number, not_null, reverse) to JMESPath visitor

0marperez:
## Description of changes

-Added min function
-Added max function
-Added to_array function
-Added to_string function
-Added to_number function
-Added not_null function
-Added reverse funtion


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -286,6 +286,44 @@ class KotlinJmespathExpressionVisitor(
...
+        "min" -> {
+            writer.addImport(RuntimeTypes.Core.Utils.min)
+            val list = expression.singleArg()
+            list.dotFunction(expression, "min()")
+        }
+
+        "reverse" -> {
+            val listOrString = expression.singleArg()
+            listOrString.dotFunction(expression, "reversed()")
+        }
```
ianbotsf, line 304: **Question**: Do these not need null guards?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -482,6 +520,29 @@ class KotlinJmespathExpressionVisitor(
...
+
+        forEach {
+            writer.withBlock("if (${it.identifier} != null && !$found) {", "}") {
+                write("$answer = ${it.identifier}")
+                write("$found = true")
+            }
+        }
+
+        return answer
+    }
```
ianbotsf, line 544: **Suggestion**: This should be possible without mutability and intermediate variables like `found`:

```kotlin
private fun List<VisitedExpression>.getNotNull(): String {
    val notNull = bestTempVarName("notNull")
    writer.withBlock("val $notNull = listOfNotNull(", ").firstOrNull()") {
        write("${it.identifier},")
    }

    return notNull
}
```

Which produces code like:

```kotlin
val notNull1 = listOfNotNull(
    foo,
    bar,
    baz,
).firstOrNull()
```

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Convenience.kt
```diff
@@ -35,3 +35,25 @@ public fun truthiness(value: Any?): Boolean = when (value) {
     null -> false
     else -> true
 }
+
+@InternalApi
+public fun List<Number>.max(): Double? =
+    @Suppress("UNCHECKED_CAST")
+    (this as List<Double>).maxOrNull()
```
ianbotsf, line 42: **Question**: Why is casting to `Double` correct here? Shouldn't the max of `[1L, 2L, 3L]` be `3L` instead of `3.0`?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Convenience.kt
```diff
@@ -35,3 +35,25 @@ public fun truthiness(value: Any?): Boolean = when (value) {
...
+    (this as List<Double>).minOrNull()
+
+@InternalApi
+public fun List<String>.min(): String? = this.minOrNull()
+
+@InternalApi
+public fun Number.toNumber(): Double = this.toDouble()
+
+@InternalApi
+public fun String.toNumber(): Double? = this.toDoubleOrNull()
```
ianbotsf, line 59: **Question**: Same as above, why is `Double` always right here?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
0marperez: The `dotFunction` function takes care of that:

    private fun VisitedExpression.dotFunction(expression: FunctionExpression, expr: String, elvisExpr: String? = null, isObject: Boolean = false): VisitedExpression {
        val dotFunctionExpr = ensureNullGuard(shape, expr, elvisExpr)
         ...
    }

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Convenience.kt
0marperez: So other parts of the code require a specific type of number and I chose double because I think all the number types we deal with can be represented as a Double (as long as they are small enough). I think I can re-write this though

0marperez: There's also the question of getting a number as a string, what type of number to choose for it

ianbotsf (approved):

aajtodd (approved):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Convenience.kt
```diff
@@ -35,3 +35,24 @@ public fun truthiness(value: Any?): Boolean = when (value) {
...
+public fun Float.toNumber(): Float = this
+
+@InternalApi
+public fun Double.toNumber(): Double = this
+
+@InternalApi
+public fun String.toNumber(): Double? = this.toDoubleOrNull()
+
+@InternalApi
+public fun Any.toNumber(): Nothing? = null
```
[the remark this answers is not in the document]
0marperez: According to the JMESPath [docs](https://jmespath.org/proposals/functions.html#to-number) for `to_number` if it's used on anything other than a string or number it should return `null`. And from looking at the Kotlin docs for `Nothing` :

> The nullable variant of this type, Nothing?, has exactly one possible value, which is null

PR #956, opened 2023-09-19
fix: ignore __type field when deserializing unions in json protocols

0marperez:
## Issue \#
closes aws-sdk-kotlin#1044

## Description of changes
-Added a flag for JSON protocols
-Added a `__type` to union object descriptors
-Ignoring `__type` unless it's member explicitly in union shape

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/JsonSerdeDescriptorGenerator.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/JsonSerdeDescriptorGenerator.kt
@@ -8,6 +8,7 @@ package software.amazon.smithy.kotlin.codegen.rendering.serde
 import software.amazon.smithy.kotlin.codegen.core.RenderingContext
 import software.amazon.smithy.kotlin.codegen.core.RuntimeTypes
 import software.amazon.smithy.kotlin.codegen.core.addImport
+import software.amazon.smithy.kotlin.codegen.core.withBlock
 import software.amazon.smithy.kotlin.codegen.model.getTrait
 import software.amazon.smithy.kotlin.codegen.utils.dq
 import software.amazon.smithy.model.shapes.MemberShape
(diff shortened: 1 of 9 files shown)


ianbotsf (changes requested):
I think we should move more of this custom logic into the the JSON-specific classes. Burying this in abstract or base implementations defeats the purpose of polymorphism and makes the code harder to follow.

There's a bunch of ways this could be done. One way, I think, is the descriptor logic could be moved into `JsonSerdeDescriptorGenerator`, which could manually add a `__type` item to the `memberShapes` list. The logic for rendering the field deserializer could be moved into a JSON-specific union deserializer generator which inherits `DeserializeUnionGenerator` and overrides `renderShapeDeserializer`.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/JsonSerdeDescriptorGenerator.kt
```diff
@@ -24,6 +25,73 @@ open class JsonSerdeDescriptorGenerator(
     private val supportsJsonNameTrait: Boolean = true,
 ) : AbstractSerdeDescriptorGenerator(ctx, memberShapes) {
 
+    override fun render() {
```
aajtodd, line 28: **fix**: This logic is specific to AWS protocols and doesn't belong in `smithy-kotlin`

runtime/serde/serde-json/common/test/aws/smithy/kotlin/runtime/serde/json/JsonDeserializerTest.kt
```diff
@@ -362,6 +362,110 @@ class JsonDeserializerTest {
         assertTrue(found, "unknown field not enumerated")
     }
 
+    private class BasicUnionTest {
```
aajtodd, line 365: These tests aren't really providing any additional value (especially since this behavior is covered by protocol tests anyway).

aajtodd (changes requested):
These changes are to account for AWS specific protocol behaviors and as such mostly should be implemented in `aws-sdk-kotlin` codegen.

I also think we can simplify this and not need to special case `__type` as a pseudo field. Instead we can just define a new deserialization trait to ignore it all together and support that in the runtime deserializer.

e.g.

```kotlin
  val X_DESCRIPTOR = SdkFieldDescriptor(SerialKind.Integer, JsonSerialName("x"))
  val Y_DESCRIPTOR = SdkFieldDescriptor(SerialKind.Integer, JsonSerialName("y"))
  val OBJ_DESCRIPTOR = SdkObjectDescriptor.build {
      field(X_DESCRIPTOR)
      field(Y_DESCRIPTOR)
      trait(IgnoreKeys("__type"))   // new object trait to signal that `__type` should just be ignored and not even enumerated in the `findNextFieldIndex()` logic.
  }


deserializer.deserializeStruct(OBJ_DESCRIPTOR) {
    loop@ while (true) {
        when (findNextFieldIndex()) {
            X_DESCRIPTOR.index -> x = value = MyUnion.X(deserializeInt())
            Y_DESCRIPTOR.index -> y = MyUnion.Y(deserializeInt())
            null -> break@loop
            else -> value MyUnion.SdkUnknown.also { skipValue() }
        }
    }
}

```

This should require way less changes in codegen since the AWS protocols just need to override `JsonSerdeDescriptorGenerator.getObjectDescriptorTraits()` and be relatively easy to add to to the `JsonDeserializer` in the runtime.

PR #958, opened 2023-09-25
feat(codegen): generate client side error correction

aajtodd:
## Issue \#

n/a

## Description of changes

Implement support for [client side error correction](https://smithy.io/2.0/spec/aggregate-types.html?highlight=error%20correction#client-error-correction) which sets defaults when `@required` is present for output shape members. This is to enable making these shape members non-nullable such that if a server doesn't actually send a required member in a response we don't explode at deserialization. 

NOTE: Nullablity hasn't changed in this PR, that will come when we enable `CLIENT_CAREFUL` check mode.


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/ClientErrorCorrection.kt
```diff
@@ -0,0 +1,61 @@
...
+     */
+    fun defaultValue(
+        ctx: CodegenContext,
+        member: MemberShape,
+        writer: KotlinWriter,
+    ): String {
+        val target = ctx.model.expectShape(member.target)
+        val targetSymbol = ctx.symbolProvider.toSymbol(target)
+
+        if (target.isEnum) {
```
0marperez, line 34: Just out of curiosity, why is enum outside of the when loop?

aajtodd: IDL v1 didn't have a dedicated enum shape, it just had `string` shapes with the `@enum` trait. the `ShapeType` doesn't account for this as it would gladly tell you that an enum is just a `ShapeType.STRING`.

0marperez: Is this something that should be commented on there? For people that have to maintain this in the future?

aajtodd: Sure I added a note and some additional test coverage from trying to make sure my understanding was correct. We handle this in a few spots already. In IDL v1 there is now enum shape so we'll never encounter `ShapeType.ENUM`. It's perhaps not as important anymore to cover IDLv1 and maybe could be removed but I know it bit us at one point in time so it's probably best to continue covering it.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
```diff
@@ -285,6 +279,20 @@ class StructureGenerator(
                         .write("this.#L = #Q.invoke(block)", memberName, memberSymbol)
                         .closeBlock("}")
                 }
+
+                write("")
+
+                // render client side error correction function to set @required members
+                withBlock(
+                    "internal fun correctErrors(): Builder {",
```
lauzadis, line 287: Should this be called something less negative like `fillMissingMembers()`?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/ClientErrorCorrection.kt
```diff
@@ -0,0 +1,63 @@
...
+            ShapeType.LONG -> "0L"
+            ShapeType.FLOAT -> "0f"
+            ShapeType.DOUBLE -> "0.0"
+            ShapeType.BIG_INTEGER -> writer.format("#T(\"0\")", RuntimeTypes.Core.Content.BigInteger)
+            ShapeType.BIG_DECIMAL -> writer.format("#T(\"0\")", RuntimeTypes.Core.Content.BigDecimal)
+            ShapeType.DOCUMENT -> "null"
+            ShapeType.UNION -> writer.format("#T.SdkUnknown", targetSymbol)
+            ShapeType.LIST,
+            ShapeType.SET,
+            -> "emptyList()"
```
lauzadis, line 56: Why not use `emptySet()`? 
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/empty-set.html

lauzadis (approved):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/ClientErrorCorrection.kt
aajtodd: Sets are [deprecated ](https://smithy.io/2.0/guides/building-codegen/mapping-shapes-to-languages.html?highlight=setshape#ignore-set-shapes-from-smithy-1-0)in IDLv2. I could probably remove this but for completeness I left it in.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
aajtodd: I named it to be close to the [specification](https://smithy.io/2.0/spec/aggregate-types.html#client-error-correction) that it comes from. Open to suggestions

0marperez (approved):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
ianbotsf, line 287: **Question**: We always call this immediately before `build()`. Are there any times we'd build a builder and _not_ call `correctErrors()`? If not, can we make this method private and invoke it directly from `build`?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt
```diff
@@ -618,6 +614,12 @@ abstract class HttpBindingProtocolGenerator : ProtocolGenerator {
                         renderDeserializeResponseCode(ctx, it, writer)
                     }
             }
+            .call {
+                // Render client side error correction for `@required` members.
+                // NOTE: nested members bound via the document/payload will be handled by the deserializer for the relevant
+                // content type. All other members (e.g. bound via REST semantics) will be corrected here.
+                writer.write("builder.correctErrors()")
+            }
```
ianbotsf, line 622: **Nit**: `call` is unnecessary.

codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGeneratorTest.kt
```diff
@@ -197,6 +197,10 @@ class StructureGeneratorTest {
                 public fun quux(block: com.test.model.Qux.Builder.() -> kotlin.Unit) {
                     this.quux = com.test.model.Qux.invoke(block)
                 }
+                
+                internal fun correctErrors(): Builder {
+                    return this
+                }
```
ianbotsf, line 203: **Nit**: Missing tests for a structure with _actual_ errors to correct.

ianbotsf (approved):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
aajtodd: Yes, right now we only invoke this during deserialization (which is the error being corrected). It's unclear if we would default `@required` members or throw an error for input types.

PR #959, opened 2023-09-25
feat(codegen): generate paginators and waiters with default parameters for all optional inputs

aajtodd:
## Issue \#

n/a

## Description of changes

* **feat**: We generate service client operation methods with a default parameter whenever the input shape to the operation has all optional members. This PR adds the same default parameter convenience for paginators and waiters.
* **fix**: Fix test utility to actually return `null` when no index is found (`indexOf` returns `-1` when not found, our test logic was predicating off `null` and silently passing tests that should fail).
* **fix**: https://github.com/awslabs/smithy-kotlin/pull/951 added API settings for internal vs external visibility but we missed paginators and waiters as part of this, updated codegen to respect the visibility for generated paginators and waiters.


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
```diff
@@ -124,6 +116,12 @@ class PaginatorGenerator : KotlinIntegration {
         """.trimIndent()
         val docReturn = "@return A [kotlinx.coroutines.flow.Flow] that can collect [${outputSymbol.name}]"
 
+        val inputParameter = if (paginationInfo.input.hasAllOptionalMembers) {
+            writer.format("initialRequest: #1T = #1T{}", inputSymbol)
```
lauzadis, line 120: nit (style): `#1T { }`

```kt
initialRequest: PutObjectRequest = PutObjectRequest{}

vs.

initialRequest: PutObjectRequest = PutObjectRequest { }
```

If you think it's a good suggestion, also applies to the same change in waiters

lauzadis (approved):

ianbotsf (approved):

0marperez (approved):

PR #963, opened 2023-09-26
fix(rt): slf4j 1.x compatability

aajtodd:
## Issue \#

fixes https://github.com/awslabs/aws-sdk-kotlin/issues/993

## Description of changes

This PR provides automatic fallback to SLF4J 1.x compatible logging implementation when 2.x API can't be loaded. 2.x is binary compatible with 1.x but it _does_ contain some new APIs that aren't available in 1.x (namely the fluent builder APIs). We could just map everything to a 1.x compatible API but it's not guaranteed to be the best mapping since underlying logging implementations may be more efficient than our compatibility layer can achieve. 

------

## Alternative

I was able to solve this differently by creating a separate 1.x implementation and then using Gradle component metadata rules to replace the underlying 2.x implementation we ship by default. I decided that fallback by default is a better customer experience though.

```kotlin
// file: build.gradle.kts of some consuming project

class LoggingCapability : ComponentMetadataRule {
    private val loggingModules = setOf("logging-slf4j2", "logging-slf4j")
    override fun execute(context: ComponentMetadataContext) = context.details.run {
        if (id.group == "aws.smithy.kotlin" && loggingModules.contains(id.name)){
            allVariants {
                withCapabilities {
                    addCapability("aws.smithy.kotlin", "slf4j-logger", id.version)
                }
            }
        }
    }
}

dependencies {
    // activate the "LoggingCapability" rule
    components.all(LoggingCapability::class.java)
    
    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutinesVersion")
    implementation("aws.sdk.kotlin:s3:$sdkVersion")
    implementation("aws.sdk.kotlin:dynamodb:$sdkVersion")

   // force 1.x on our classpath, this is just to reproduce the issue, in practice customers have been experiencing this
   // due to 1.x being what ends up on the classpath _somehow_ (e.g. shipped with gradle)
    implementation("org.slf4j:slf4j-api:1.7.36") {
        version {
            strictly("1.7.36")
        }
    }
    
    // hypothetical 1.x based implementation with same package name/types expected by default provider 
    implementation("aws.smithy.kotlin:logging-slf4j:$smithyKotlinVersion")

    // slf4j 1.x log4j
    implementation("org.apache.logging.log4j:log4j-slf4j-impl:2.20.0")
    implementation("org.apache.logging.log4j:log4j-core:2.20.0")
}

configurations.all {
    resolutionStrategy.capabilitiesResolution.withCapability("aws.smithy.kotlin:slf4j-logger") {
        val toBeSelected = candidates.firstOrNull { it.id.let { id -> id is ModuleComponentIdentifier && id.module == "logging-slf4j" } }
        toBeSelected?.let { select(it) }
        because("need slf4j 1.x bindings")
    }
}
```


NatanLifshitz:
Awesome, this issue was blocking us from updating past 0.28.1.

runtime/observability/logging-slf4j2/build.gradle.kts
```diff
@@ -2,7 +2,7 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
-description = "Logging provider based on SLF4J 2"
+description = "Logging provider based on SLF4J 2.x"
```
ianbotsf, line 5: **Nit**: This description is no longer fully accurate. It's based on 2.x unless it's not.

runtime/observability/logging-slf4j2/jvm/src/aws/smithy/kotlin/runtime/telemetry/logging/slf4j/Slf4j1xLoggerAdapter.kt
```diff
@@ -0,0 +1,78 @@
...
+        msg = message
+    }
+
+    override fun setKeyValuePair(key: String, value: Any) {
+        kvp[key] = value
+    }
+
+    override fun setContext(context: Context) {
+        // TODO - add a way to get the current trace context and set the key/value pair on it?
+    }
```
ianbotsf, line 52: **Nit**: Could be moved down to `AbstractSlf4jLoggerAdapter`.

runtime/observability/logging-slf4j2/jvm/src/aws/smithy/kotlin/runtime/telemetry/logging/slf4j/Slf4jLoggerProvider.kt
```diff
@@ -5,73 +5,24 @@
...
 
-    override fun setMessage(message: String) {
-        delegate.setMessage(message)
+    private val slf4jLoggerAdapter: (org.slf4j.Logger) -> Logger = try {
+        Class.forName("org.slf4j.spi.LoggingEventBuilder")
+        ::Slf4j2xLoggerAdapter
+    } catch (ex: ClassNotFoundException) {
+        LoggerFactory.getLogger(Slf4jLoggerProvider::class.java).warn("falling back to SLF4J 1.x compatible binding")
+        ::Slf4j1xLoggerAdapter
     }
```
ianbotsf, line 22: **Style**: May be clearer as a function.

ianbotsf (approved):

runtime/observability/logging-slf4j2/jvm/src/aws/smithy/kotlin/runtime/telemetry/logging/slf4j/Slf4j1xLoggerAdapter.kt
aajtodd: Unfortunately it can't because this is on the `LogRecordBuilder` implementation which is the thing that differs between the two.

runtime/observability/logging-slf4j2/jvm/src/aws/smithy/kotlin/runtime/telemetry/logging/slf4j/Slf4jLoggerProvider.kt
aajtodd: Not opposed but can you clarify what you are referring to? What would you prefer to see here?

runtime/observability/logging-slf4j2/jvm/src/aws/smithy/kotlin/runtime/telemetry/logging/slf4j/Slf4j1xLoggerAdapter.kt
ianbotsf: Oh right, sorry I was thinking this was still in `Slf4j1xLoggerAdapter`. Nvm.

runtime/observability/logging-slf4j2/jvm/src/aws/smithy/kotlin/runtime/telemetry/logging/slf4j/Slf4jLoggerProvider.kt
ianbotsf: A class-level `val` that returns a function is more complex to read than simply a `fun`:

```kotlin
private fun slf4jLoggerAdapter(delegate: org.slf4j.Logger) = try {
    Class.forName("org.slf4j.spi.LoggingEventBuilder")
    Slf4j2xLoggerAdapter(delegate)
} catch (_: ClassNotFoundException) {
    LoggerFactory.getLogger(Slf4jLoggerProvider::class.java).warn("falling back to SLF4J 1.x compatible binding")
    Slf4j1xLoggerAdapter(delegate)
}
```

aajtodd: I was trying to avoid invoking the class loader on every logger instantiation.

aajtodd: I could refactor it though to something like:


```kotlin
private val slf4j2Found = try {
    Class.forName("org.slf4j.spi.LoggingEventBuilder")
    true
} catch(_: ClassNotFoundException) {
     LoggerFactory.getLogger(Slf4jLoggerProvider::class.java).warn("falling back to SLF4J 1.x compatible binding")
    false  
}

   override fun getOrCreateLogger(name: String): Logger {
        val sl4fjLogger = LoggerFactory.getLogger(name)
        return if (slf4j2Found) { 
             Slf4j2xLoggerAdapter(slf4jLogger)
        } else {
             Slf4j1xLoggerAdapter(slf4jLogger)
        }
    }

```

ianbotsf: Oh right. I didn't think about caching the class loader invocation. OK, your original code works better then.

lauzadis (approved):

PR #964, opened 2023-09-27
fix: ignore __type when deserializing union for AWS Json protocols

0marperez:
## Issue \#

closes [aws-sdk-kotlin#1044](https://github.com/awslabs/aws-sdk-kotlin/issues/1044)

## Description of changes

-Added `IgnoreKey` trait and adapted `JsonDeserializer` to it 


runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/SdkFieldDescriptor.kt
```diff
@@ -6,23 +6,6 @@ package aws.smithy.kotlin.runtime.serde
...
 
-/**
- * This tag interface provides a mechanism to attach type-specific metadata to any field.
- * See [aws.smithy.kotlin.runtime.serde.xml.XmlList] for an example implementation.
- *
- * For example, to specify that a list should be serialized in XML such that values are wrapped
- * in a tag called "boo", pass an instance of XmlList to the FieldDescriptor of `XmlList(elementName="boo")`.
- */
-@InternalApi
-public interface FieldTrait
```
aajtodd, line 17: Why are these moved?

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/SerdeTraits.kt
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Indicates to deserializers to ignore field/key
+ *
+ * @param key The key to ignore in the payload
+ * @param regardlessOfInModel If true will ignore key even though the model indicates key should be there
+ *
+ */
+@InternalApi
+public data class IgnoreKey(public val key: String, public val regardlessOfInModel: Boolean = true) : FieldTrait
```
aajtodd, line 28: This should be in `serde-json` as it only applies there right? (see `JsonFieldTraits.kt`)

aajtodd: **Question**: Why do we need `regardlessOfInModel`, I would assume if we set `IgnoreKey` as a trait on the object descriptor it's going to be ignored regardless of whether it's modeled. This seems unnecessary to me but I'd like to understand what the motivation was.

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/SdkFieldDescriptor.kt
0marperez: Since the file was named `SdkFieldDescriptor` I thought it would make more sense to have field traits in a different place, I can change it to how it was though

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/SerdeTraits.kt
0marperez: I was thinking of a case where we have a union with a member called `__type`. If that happens we would not deserialize that member, because we're ignoring that key regardless of what the model says. The default is true anyway so it behaves as expected unless set explicitly to false.

0marperez: I didn't put it in a JSON specific file because I think this might be useful in the future for other deserializers like the XML one. I don't know if we have a use case already but that was my reasoning behind it.

aajtodd: > I was thinking of a case where we have a union with a member called __type. If that happens we would not deserialize that member, because we're ignoring that key regardless of what the model says

Shouldn't this just be a choice by codegen though? If we want it deserialized because it's in the model don't add the `IgnoreKey` trait.

0marperez: That makes sense but the issue this is trying to solve: https://github.com/smithy-lang/smithy/pull/1945
is that sometimes a unnecessary `__type` is serialized. What you're suggesting is to always ignore `__type`. But there could be a cases where `__type` could be in the model and we would be ignoring it because sometimes there's an extra one there unnecessarily

aajtodd: > What you're suggesting is to always ignore __type. But there could be a cases where __type could be in the model and we would be ignoring it because sometimes there's an extra one there unnecessarily

I don't follow, if `__type` is in the model and we _want_ to deserialize it then just don't emit the `IgnoreKey` trait during codegen. What am I missing here?

0marperez: I guess that logic could live in codegen, now it's in the deserializer

runtime/serde/serde-json/common/test/aws/smithy/kotlin/runtime/serde/json/JsonDeserializerIgnoresKeysTest.kt
```diff
@@ -0,0 +1,172 @@
...
+
+import aws.smithy.kotlin.runtime.serde.SdkFieldDescriptor
+import aws.smithy.kotlin.runtime.serde.SdkObjectDescriptor
+import aws.smithy.kotlin.runtime.serde.SerialKind
+import aws.smithy.kotlin.runtime.serde.deserializeStruct
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class JsonDeserializerIgnoresKeysTest {
+    class IgnoresKeysTest {
```
lauzadis, line 16: this double-nested class seems unnecessary, I think you can just declare these at the top-level without using a `companion object` for simplicity

runtime/serde/serde-json/common/test/aws/smithy/kotlin/runtime/serde/json/JsonDeserializerIgnoresKeysTest.kt
```diff
@@ -0,0 +1,172 @@
...
+        assertEquals(2, y)
+        assertEquals(null, z)
+    }
+
+    class IgnoresMultipleKeysTest {
+        companion object {
+            val W_DESCRIPTOR = SdkFieldDescriptor(SerialKind.Integer, JsonSerialName("w"))
+            val X_DESCRIPTOR = SdkFieldDescriptor(SerialKind.Integer, JsonSerialName("x"))
+            val Y_DESCRIPTOR = SdkFieldDescriptor(SerialKind.Integer, JsonSerialName("y"))
+            val Z_DESCRIPTOR = SdkFieldDescriptor(SerialKind.Integer, JsonSerialName("z"))
```
lauzadis, line 127: duplicated, can re-use the others descriptors above

lauzadis (approved):

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonDeserializer.kt
```diff
@@ -198,7 +198,13 @@ private class JsonFieldIterator(
                 val token = reader.nextTokenOf<JsonToken.Name>()
                 val propertyName = token.value
                 val field = descriptor.fields.find { it.serialName == propertyName }
-                field?.index ?: Deserializer.FieldIterator.UNKNOWN_FIELD
+
+                if (descriptor.traits.contains(IgnoreKey(propertyName))) {
```
ianbotsf, line 202: **Style**: Consider `if (IgnoreKey(propertyName) in descriptor.traits)`

runtime/serde/serde-json/common/test/aws/smithy/kotlin/runtime/serde/json/JsonDeserializerIgnoresKeysTest.kt
```diff
@@ -0,0 +1,161 @@
...
+import aws.smithy.kotlin.runtime.serde.deserializeStruct
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class JsonDeserializerIgnoresKeysTest {
+    private val X_DESCRIPTOR = SdkFieldDescriptor(SerialKind.Integer, JsonSerialName("x"))
+    private val Y_DESCRIPTOR = SdkFieldDescriptor(SerialKind.Integer, JsonSerialName("y"))
+    private val Z_DESCRIPTOR = SdkFieldDescriptor(SerialKind.Integer, JsonSerialName("z"))
+    private val OBJ_DESCRIPTOR = SdkObjectDescriptor.build {
+        trait(IgnoreKey("z")) // <----
```
ianbotsf, line 20: **Question**: What does the comment `// <----` mean? Same question on L120-121.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/JsonParserGenerator.kt
```diff
@@ -24,6 +24,13 @@ open class JsonParserGenerator(
 
     open val defaultTimestampFormat: TimestampFormatTrait.Format = TimestampFormatTrait.Format.EPOCH_SECONDS
 
+    open fun descriptorGenerator(
+        ctx: ProtocolGenerator.GenerationContext,
+        shape: Shape,
+        members: List<MemberShape>,
+        writer: KotlinWriter,
+    ): JsonSerdeDescriptorGenerator = JsonSerdeDescriptorGenerator(ctx.toRenderingContext(protocolGenerator, shape, writer), members, supportsJsonNameTrait)
```
ianbotsf, line 32: **Style**: Line length > 120 chars

ianbotsf (approved):
Looks good overall!

runtime/serde/serde-json/common/test/aws/smithy/kotlin/runtime/serde/json/JsonDeserializerIgnoresKeysTest.kt
```diff
@@ -0,0 +1,161 @@
...
+package aws.smithy.kotlin.runtime.serde.json
+
+import aws.smithy.kotlin.runtime.serde.SdkFieldDescriptor
+import aws.smithy.kotlin.runtime.serde.SdkObjectDescriptor
+import aws.smithy.kotlin.runtime.serde.SerialKind
+import aws.smithy.kotlin.runtime.serde.deserializeStruct
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class JsonDeserializerIgnoresKeysTest {
```
aajtodd, line 15: **Suggestion**: Every test has explicitly modeled field descriptors. You're missing coverage for unmodeled fields that we want to skip rather than enumerate as unknown.

aajtodd (approved):
Looks good, one suggestion but otherwise approved.

PR #966, opened 2023-09-29
feat: add type function to JMESPath visitor

0marperez:
## Description of changes

-Added `type` function to JMESPath visitor

NOTES: 
-Added flag to ignore null guards to `dotFunction` 
-Smithy's JMESPath parser won't allow string comparison so I had to get around that using the `Types` structure

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
@@ -219,8 +219,14 @@ class KotlinJmespathExpressionVisitor(
     private fun FunctionExpression.args(): List<VisitedExpression> =
         this.arguments.map { acceptSubexpression(it) }
 
-    private fun VisitedExpression.dotFunction(expression: FunctionExpression, expr: String, elvisExpr: String? = null, isObject: Boolean = false): VisitedExpression {
-        val dotFunctionExpr = ensureNullGuard(shape, expr, elvisExpr)
+    private fun VisitedExpression.dotFunction(
+        expression: FunctionExpression,
+        expr: String,
+        elvisExpr: String? = null,
+        isObject: Boolean = false,
+        enforceNullGuard: Boolean = true,
+    ): VisitedExpression {
+        val dotFunctionExpr = if (enforceNullGuard) ensureNullGuard(shape, expr, elvisExpr) else ".$expr"
         val ident = addTempVar(expression.name.toCamelCase(), "$identifier$dotFunctionExpr")
 
         shape?.let { shapeCursor.addLast(shape) }
@@ -320,6 +326,12 @@ class KotlinJmespathExpressionVisitor(
             writer.addImport(RuntimeTypes.Core.Utils.toNumber)
             val arg = expression.singleArg()
             arg.dotFunction(expression, "toNumber()")
+        }
+
+        "type" -> {
+            writer.addImport(RuntimeTypes.Core.Utils.type)
+            val arg = expression.singleArg()
+            arg.dotFunction(expression, "type()", enforceNullGuard = false)
         }
 
         else -> throw CodegenException("Unknown function type in $expression")
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
@@ -224,9 +224,9 @@ class KotlinJmespathExpressionVisitor(
         expr: String,
         elvisExpr: String? = null,
         isObject: Boolean = false,
-        enforceNullGuard: Boolean = true,
+        ensureNullGuard: Boolean = true,
     ): VisitedExpression {
-        val dotFunctionExpr = if (enforceNullGuard) ensureNullGuard(shape, expr, elvisExpr) else ".$expr"
+        val dotFunctionExpr = if (ensureNullGuard) ensureNullGuard(shape, expr, elvisExpr) else ".$expr"
         val ident = addTempVar(expression.name.toCamelCase(), "$identifier$dotFunctionExpr")
 
         shape?.let { shapeCursor.addLast(shape) }
@@ -331,7 +331,7 @@ class KotlinJmespathExpressionVisitor(
         "type" -> {
             writer.addImport(RuntimeTypes.Core.Utils.type)
             val arg = expression.singleArg()
-            arg.dotFunction(expression, "type()", enforceNullGuard = false)
+            arg.dotFunction(expression, "type()", ensureNullGuard = false)
         }
 
         else -> throw CodegenException("Unknown function type in $expression")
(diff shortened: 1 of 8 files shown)


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -219,8 +219,14 @@ class KotlinJmespathExpressionVisitor(
...
         this.arguments.map { acceptSubexpression(it) }
 
-    private fun VisitedExpression.dotFunction(expression: FunctionExpression, expr: String, elvisExpr: String? = null, isObject: Boolean = false): VisitedExpression {
-        val dotFunctionExpr = ensureNullGuard(shape, expr, elvisExpr)
+    private fun VisitedExpression.dotFunction(
+        expression: FunctionExpression,
+        expr: String,
+        elvisExpr: String? = null,
+        isObject: Boolean = false,
+        enforceNullGuard: Boolean = true,
```
lauzadis, line 227: nit/naming: `ensureNullGuard`

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/JMESPath.kt
```diff
@@ -56,3 +56,19 @@ public inline fun <reified T> Collection<Collection<T>>.flattenIfPossible(): Col
...
+ *
+ * See: [JMESPath spec](https://jmespath.org/specification.html#type)
+ */
+@InternalApi
+public fun Any?.type(): String = when (this) {
+    is String -> "string"
+    is Boolean -> "boolean"
+    is List<*>, is Array<*> -> "array"
+    is Short, is Int, is Long, is Float, is Double -> "number"
+    is HashMap<*, *>, is Any -> "object"
```
lauzadis, line 71: question: why `HashMap` instead of the more-general `Map`?

tests/codegen/waiter-tests/src/test/kotlin/com/test/FunctionTypeTest.kt
```diff
@@ -0,0 +1,156 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.test
+
+import com.test.model.*
+import com.test.utils.successTest
+import com.test.waiters.*
+import org.junit.jupiter.api.Test
```
lauzadis, line 11: Use `kotlin.test.Test` instead, it's better for KMP compatibility (even if we only plan to run codegen on JVM)

tests/codegen/waiter-tests/src/test/kotlin/com/test/FunctionTypeTest.kt
```diff
@@ -0,0 +1,156 @@
...
+
+class FunctionTypeTest {
+    @Test fun testTypeFunctionStringEquals() = successTest(
```
lauzadis, line 14: style: consider `@Test` on a separate line

0marperez: I think it makes more sense to have `@Test` like that for now. Only because the rest of the [tests](https://github.com/awslabs/smithy-kotlin/tree/3934d48a9e084e23b1250501b0deb8dafe2a1c03/tests/codegen/waiter-tests/src/test/kotlin/com/test) use the same style. I can refactor all to use that style just in another PR to keep things easy to review

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/JMESPath.kt
0marperez: This is related to this [other function](https://github.com/awslabs/smithy-kotlin/blob/3934d48a9e084e23b1250501b0deb8dafe2a1c03/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt#L506) that merges objects and returns a hash map. I didn't use a map for that, but I really don't know if it makes a difference. Is there a reason to use `Map` instead?

tests/codegen/waiter-tests/src/test/kotlin/com/test/FunctionTypeTest.kt
lauzadis: Ok, I agree with keeping it consistent

tests/codegen/waiter-tests/src/test/kotlin/com/test/FunctionTypeTest.kt
0marperez: Now that I notice that I see I have to do the same for the rest of the [tests](https://github.com/awslabs/smithy-kotlin/tree/3934d48a9e084e23b1250501b0deb8dafe2a1c03/tests/codegen/waiter-tests/src/test/kotlin/com/test)

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/JMESPath.kt
lauzadis: I lied earlier, [HashMap](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-hash-map/) is an implementation of **Mutable**Map. I think it's better to use the base class as a type rather than an implementation of it. Let's see what others think about it / if it matters.

ianbotsf: If we're truly combining this logic with `is Any` then it doesn't matter—we should just remove `is HashMap<*, *>` since that type is a subtype of `Any`.

In principle though, @lauzadis is right: we should favor type checking at the highest reasonable level. Both in general and specifically for collections we should typically code to APIs (i.e., interfaces) rather than implementations (i.e., specific classes). I can see that `mergeProperties` creates a `HashMap` but there will be other kinds of maps in the object graph, including some created directly by deserializers which will delegate down to stdlib-internal implementations.

tests/codegen/waiter-tests/model/function-type.smithy
```diff
@@ -0,0 +1,169 @@
...
+        acceptors: [
+            {
+                state: "success",
+                matcher: {
+                    output: {
+                        path: "type(primitives.string) == types.string",
+                        expected: "true",
+                        comparator: "booleanEquals"
+                    }
+                }
```
ianbotsf, line 18: **Question**: The matcher conditions in this file seem odd to me. Why compare against another structure's members when we have to explicitly set their values? Would `path: "type(primitives.string) == \"string\""` be simpler? Or even something that uses `stringEquals` directly?

```smithy
matcher: {
    output: {
        path: "type(primitives.string)",
        expected: "string",
        comparator: "stringEquals"
    }
}
```

0marperez: My bad, I forgot about `stringEquals` for some reason

ianbotsf (approved):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/JMESPath.kt
```diff
@@ -56,3 +56,19 @@ public inline fun <reified T> Collection<Collection<T>>.flattenIfPossible(): Col
...
+ * Returns a JS type as a string
+ *
+ * See: [JMESPath spec](https://jmespath.org/specification.html#type)
+ */
+@InternalApi
+public fun Any?.type(): String = when (this) {
+    is String -> "string"
+    is Boolean -> "boolean"
+    is List<*>, is Array<*> -> "array"
+    is Short, is Int, is Long, is Float, is Double -> "number"
```
aajtodd, line 70: **nit**: `is Number` would be better here

aajtodd (approved):

PR #969, opened 2023-10-03
feat: add functions (sort, sort_by, max_by, min_by, map) to JMESPath visitor

### LINKED ISSUE #596: Implement full support for JMESPath expression types
### Community Note

## Describe the Feature

Implementing waiters support in #572 added support for the subset of JMESPath expression types needed to support existing AWS services but several expression types were left for future implementation. Specifically, the following need to be supported in codegen:

* ~and expressions~
* current node expressions
* expression type expressions
* functions expressions (besides `contains` and `length`, which are currently supported)
* index expressions
* literal expressions (besides string, number, boolean, and null, which are currently supported
* multiselect hash expressions
* ~not expressions~
* ~or expressions~
* slice expressions

See [KotlinJmespathExpressionVisitor.kt](https://github.com/awslabs/smithy-kotlin/blob/b79e1ef788fea8e8debfc816023b5cbc720a8ca3/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt) for a list of supported/unsupported expression types.

The lack of these expression types means that future waiters which use them will not codegen correctly.

## Proposed Solution

Implement support for the missing expression types.

None.

## Additional Context

None.

0marperez:
## Issue \#

closes #596 
## Description of changes

-Added sort to JMESPath visitor
-Added sort_by to JMESPath visitor
-Added max_by to JMESPath visitor
-Added min_by to JMESPath visitor
-Added map to JMESPath visitor

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
@@ -105,7 +105,9 @@ class KotlinJmespathExpressionVisitor(
         return VisitedExpression(outerName, leftShape, innerResult.shape)
     }
 
-    private fun subfield(expression: FieldExpression, parentName: String, isObject: Boolean = false): VisitedExpression {
+    private data class SubFieldData(val name: String, val codegen: String, val member: Shape?)
+
+    private fun subfieldLogic(expression: FieldExpression, parentName: String, isObject: Boolean = false): SubFieldData {
         val member = currentShape.targetOrSelf(ctx.model).getMember(expression.name).getOrNull()
 
         val name = expression.name.toCamelCase()
@@ -130,8 +132,16 @@ class KotlinJmespathExpressionVisitor(
         }
 
         member?.let { shapeCursor.addLast(it) }
-        return VisitedExpression(addTempVar(name, codegen), member)
+        return SubFieldData(name, codegen, member)
     }
+
+    private fun subfield(expression: FieldExpression, parentName: String, isObject: Boolean = false): VisitedExpression {
+        val (name, codegen, member) = subfieldLogic(expression, parentName, isObject)
+        return VisitedExpression(addTempVar(name, codegen), member, nullable = currentShape.isNullable)
+    }
+
+    private fun subfieldCodegen(expression: FieldExpression, parentName: String, isObject: Boolean = false): String =
+        subfieldLogic(expression, parentName, isObject).codegen
 
     override fun visitAnd(expression: AndExpression): VisitedExpression {
         writer.addImport(RuntimeTypes.Core.Utils.truthiness)
@@ -159,14 +169,16 @@ class KotlinJmespathExpressionVisitor(
                 append("if ($isNullExpr) null else ")
             }
 
-            val unSafeComparatorExpr = "compareTo(${right.identifier}) ${expression.comparator} 0"
+            val unSafeComparatorExpr = "compareTo(${right.identifier})"
             val comparatorExpr = if (left.nullable) "?.$unSafeComparatorExpr" else ".$unSafeComparatorExpr"
 
             append("${left.identifier}$comparatorExpr")
         }
 
-        val ident = addTempVar("comparison", codegen)
-        return VisitedExpression(ident)
+        val unSafeComparatorValue = addTempVar("unSafeComparator", codegen)
+        val safeComparatorValue = addTempVar("safeComparator", "if ($unSafeComparatorValue == null) null else $unSafeComparatorValue ${expression.comparator} 0")
+
+        return VisitedExpression(safeComparatorValue)
     }
 
     override fun visitCurrentNode(expression: CurrentExpression): VisitedExpression {
@@ -257,9 +269,7 @@ class KotlinJmespathExpressionVisitor(
 
         "avg" -> {
             val numbers = expression.singleArg()
-            val average = numbers.dotFunction(expression, "average()").identifier
-            val isNaN = ensureNullGuard(numbers.shape, "isNaN() == true")
-            VisitedExpression(addTempVar("averageOrNull", "if($average$isNaN) null else $average"), nullable = true)
+            numbers.dotFunction(expression, "average()")
         }
 
         "join" -> {
@@ -334,7 +344,34 @@ class KotlinJmespathExpressionVisitor(
             arg.dotFunction(expression, "type()", ensureNullGuard = false)
         }
 
+        "sort" -> {
+            writer.addImport(RuntimeTypes.Core.Utils.jmespathSort)
+            val arg = expression.singleArg()
+            arg.dotFunction(expression, "jmespathSort()")
+        }
+
+        "sort_by" -> mappingFunction(expression, "sortBy", this::sortBy)
+
+        "max_by" -> mappingFunction(expression, "maxBy", this::maxBy)
+
+        "min_by" -> mappingFunction(expression, "minBy", this::minBy)
+
+        "map" -> mappingFunction(expression, "map", this::map, true)
+
         else -> throw CodegenException("Unknown function type in $expression")
+    }
+
+    private fun mappingFunction(
+        expression: FunctionExpression,
+        variableName: String,
+        function: (JmespathExpression, VisitedExpression) -> String,
+        invertedArgs: Boolean = false,
+    ): VisitedExpression {
+        val (argIndex, expressionIndex) = if (invertedArgs) 1 to 0 else 0 to 1
+
+        codegenReq(expression.arguments.size == 2) { "Unexpected number of arguments to $this" }
+        val arg = expression.arguments[argIndex].accept(this)
+        return VisitedExpression(addTempVar(variableName, function(expression.arguments[expressionIndex], arg)))
     }
 
     override fun visitIndex(expression: IndexExpression): VisitedExpression {
@@ -547,6 +584,37 @@ class KotlinJmespathExpressionVisitor(
 
         return notNull
     }
+
+    private fun mapFunctionLogic(
+        expression: JmespathExpression,
+        arg: VisitedExpression,
+        resultName: String,
+        operation: String,
+        stringOrNumberCheck: Boolean = true,
+    ): String {
+        val argName = arg.identifier
+        val result = bestTempVarName(resultName)
+        writer.withBlock("val $result = $argName?.$operation{", "}") {
+            val expressionValue = addTempVar("expression", subfieldCodegen((expression as ExpressionTypeExpression).expression as FieldExpression, "it"))
+            if (stringOrNumberCheck) {
+                write("if ($expressionValue as Any !is Number && $expressionValue as Any !is String) throw Exception(\"Result of applying expression should be string or number\")")
+            }
+            write(expressionValue)
+        }
+        return result
+    }
+
+    private fun sortBy(expression: JmespathExpression, arg: VisitedExpression): String =
+        mapFunctionLogic(expression, arg, "sorted", "sortedBy")
+
+    private fun maxBy(expression: JmespathExpression, arg: VisitedExpression): String =
+        mapFunctionLogic(expression, arg, "max", "maxBy")
+
+    private fun minBy(expression: JmespathExpression, arg: VisitedExpression): String =
+        mapFunctionLogic(expression, arg, "min", "minBy")
+
+    private fun map(expression: JmespathExpression, arg: VisitedExpression): String =
+        mapFunctionLogic(expression, arg, "map", "map", false)
 
     private val Shape.isNullable: Boolean
         get() = this is MemberShape &&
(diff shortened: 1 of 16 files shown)


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -334,9 +344,36 @@ class KotlinJmespathExpressionVisitor(
...
 
+    private fun mappingFunction(
+        expression: FunctionExpression,
+        variableName: String,
+        function: (JmespathExpression, VisitedExpression) -> String,
+        invertedArgs: Boolean = false,
+    ): VisitedExpression {
+        val (argIndex, expressionIndex) = if (invertedArgs) 1 to 0 else 0 to 1
+
+        codegenReq(expression.arguments.size == 2) { "Unexpected number of arguments to $this" }
```
lauzadis, line 372: Let's print the number of expected arguments and the number of actual arguments received

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -159,14 +169,16 @@ class KotlinJmespathExpressionVisitor(
                 append("if ($isNullExpr) null else ")
             }
 
-            val unSafeComparatorExpr = "compareTo(${right.identifier}) ${expression.comparator} 0"
+            val unSafeComparatorExpr = "compareTo(${right.identifier})"
```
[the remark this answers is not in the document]
aajtodd: Also this generates code like:

```kotlin
            val unSafeComparator = if (integer == null) null else integer?.compareTo(number)
            val safeComparator = if (unSafeComparator == null) null else unSafeComparator == 0
            safeComparator == true
```

Where you're just taking the null value if it's null otherwise comparing it to the expected value. You can compare `null` directly though

```kotlin
val comparator = integer == 0
```

Another example:
```kotlin
            val lists = it.lists
            val strings = lists?.strings
            val length = strings?.length ?: 0
            val number = 0.0
            val unSafeComparator = if (length == null) null else length.compareTo(number)
            val safeComparator = if (unSafeComparator == null) null else unSafeComparator > 0
            safeComparator == true
```

it can be condensed to:

```kotlin
            val safeComparator = length > 0
```

If the left expression of the comparision is nullable it can be changed to:

```
val safeComparator = length?.let { it > 0 }
```

Which will yield `Boolean?` as the result.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -334,9 +344,36 @@ class KotlinJmespathExpressionVisitor(
...
         else -> throw CodegenException("Unknown function type in $expression")
     }
 
+    private fun mappingFunction(
+        expression: FunctionExpression,
+        variableName: String,
+        function: (JmespathExpression, VisitedExpression) -> String,
+        invertedArgs: Boolean = false,
+    ): VisitedExpression {
+        val (argIndex, expressionIndex) = if (invertedArgs) 1 to 0 else 0 to 1
```
lauzadis, line 370: Can you explain why `map` has inverted arguments? I'm wondering if there is a "nicer" way to handle this?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -548,6 +585,37 @@ class KotlinJmespathExpressionVisitor(
...
+        operation: String,
+        stringOrNumberCheck: Boolean = true,
+    ): String {
+        val argName = arg.identifier
+        val result = bestTempVarName(resultName)
+        writer.withBlock("val $result = $argName?.$operation{", "}") {
+            val expressionValue = addTempVar("expression", subfieldCodegen((expression as ExpressionTypeExpression).expression as FieldExpression, "it"))
+            if (stringOrNumberCheck) {
+                write("if ($expressionValue as Any !is Number && $expressionValue as Any !is String) throw Exception(\"Result of applying expression should be string or number\")")
+            }
```
lauzadis, line 601: Same comment about `invertedArgs` here, why is this needed for `map` only?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -548,6 +585,37 @@ class KotlinJmespathExpressionVisitor(
...
+    private fun mapFunctionLogic(
+        expression: JmespathExpression,
+        arg: VisitedExpression,
+        resultName: String,
+        operation: String,
+        stringOrNumberCheck: Boolean = true,
+    ): String {
+        val argName = arg.identifier
+        val result = bestTempVarName(resultName)
+        writer.withBlock("val $result = $argName?.$operation{", "}") {
```
lauzadis, line 597: nit: space between `$operation` and `{`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -548,6 +585,37 @@ class KotlinJmespathExpressionVisitor(
...
+            val expressionValue = addTempVar("expression", subfieldCodegen((expression as ExpressionTypeExpression).expression as FieldExpression, "it"))
+            if (stringOrNumberCheck) {
+                write("if ($expressionValue as Any !is Number && $expressionValue as Any !is String) throw Exception(\"Result of applying expression should be string or number\")")
```
lauzadis, line 600: Can we throw a more descriptive exception type than the base `Exception`?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/JMESPath.kt
```diff
@@ -72,3 +72,27 @@ public fun Any?.type(): String = when (this) {
...
+@JvmName("FloatJmespathSort")
+public fun List<Float>.jmespathSort(): List<Float> = this.sorted()
+
+@InternalApi
+@JvmName("LongJmespathSort")
+public fun List<Long>.jmespathSort(): List<Long> = this.sorted()
+
+@InternalApi
+@JvmName("DoubleJmespathSort")
+public fun List<Double>.jmespathSort(): List<Double> = this.sorted()
```
lauzadis, line 98: Can this be made generic? You would lose the `@JvmName` annotation but also remove redundant functions.
If the annotation is not required, I'd try to make these generic for any `T`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -548,6 +585,37 @@ class KotlinJmespathExpressionVisitor(
         return notNull
     }
 
+    private fun mapFunctionLogic(
```
lauzadis, line 588: I'd call this `applyFunction`, since `mapFunctionLogic` could be confused for the actual `map` function, and I think the `Logic` suffix is unnecessary

lauzadis: Would it make sense to make this an extension function of `JmespathExpression`? Then you can use it like this

```kt
private fun JmespathExpression.applyFunction(
   ...
) { ... }

expression.applyFunction(arg, "sorted", "sortedBy")
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
aajtodd, line 600: **correctness**:

This check is wrong:

```kotlin
            val sorted = structs?.sortedBy{
                val expression = it?.integer
                if (expression as Any !is Number && expression as Any !is String) throw Exception("Result of applying expression should be string or number")
                expression
            }
```

`expression` is of type `Any?` so `as Any` is incorrect. Expression is likely never going to be null anyway though so I guess in practice it will be `Any` at runtime. If we can we should generate the code as `it.integer` not `it?` here since `.sortedBy` will only run if `structs` isn't null in the first place (due to `?`). 

Backing up for a second, why do we even need to generate the `stringOrNumber` check in the first place? I see in the [JMESPath spec](https://jmespath.org/specification.html#sort-by):

> If the result of evaluating the expr against the current array element results in type other than a number or a string, an invalid-type error will occur.

But why do we need to worry about this? Seems like this would be an invalid model most likely.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
aajtodd, line 597: **nit**: add space between `$operation` and opening brace `{`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
0marperez: The [map](https://jmespath.org/specification.html#map) expression can result in any type but for the rest ([max_by](https://jmespath.org/specification.html#max-by), `min_by`, etc..) the expression should result in a string or a number.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
0marperez: > But why do we need to worry about this? Seems like this would be an invalid model most likely.

The expression resulting in anything other than a string or number is not being checked for at the model level. I think it makes sense to have that check. Someone could add an expression that results in something else and it would fail without an exception explaining why

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/JMESPath.kt
```diff
@@ -72,3 +72,27 @@ public fun Any?.type(): String = when (this) {
     null -> "null"
     else -> throw Exception("Undetected type for: $this")
 }
+
+@InternalApi
+@JvmName("StringJmespathSort")
+public fun List<String>.jmespathSort(): List<String> = this.sorted()
```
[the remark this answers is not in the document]
0marperez: I added them because JMESPath's `sort` can only be applied to numbers and strings. I think it makes sense to check the type and then throw an exception explaining that if it's the wrong type of list

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
aajtodd: Also Smithy is doing some validation on the JMESPath expressions in the model: https://github.com/smithy-lang/smithy/blob/main/smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/TypeChecker.java

I don't think we need to add runtime validations here, this is the kind of stuff that belongs as an error in the model such that it never can even get into the hands of a customer.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/JMESPath.kt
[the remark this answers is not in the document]
aajtodd: Again I don't think we want to be responsible for runtime checking things that could be caught by model validation. These just inflate the size of the runtime core artifact with no real value. I think we should remove them and just call sorted directly which is already defined with generics

aajtodd (approved):

ianbotsf (approved):

PR #972, opened 2023-10-06
feat: implement clock skew interceptor

lauzadis:
Implements a clock skew interceptor which detects clock skews and applies corrections. The clock skew is stored as part of the interceptor's state so it can be applied to all future requests.

## Issue \#

Closes https://github.com/awslabs/aws-sdk-kotlin/issues/213

## Description of changes

### DIFF
--- a/runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/AwsHttpSigner.kt
+++ b/runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/AwsHttpSigner.kt
@@ -12,6 +12,7 @@ import aws.smithy.kotlin.runtime.auth.awssigning.internal.setAwsChunkedBody
 import aws.smithy.kotlin.runtime.auth.awssigning.internal.setAwsChunkedHeaders
 import aws.smithy.kotlin.runtime.auth.awssigning.internal.useAwsChunkedEncoding
 import aws.smithy.kotlin.runtime.http.HttpBody
+import aws.smithy.kotlin.runtime.http.operation.HttpOperationContext
 import aws.smithy.kotlin.runtime.http.request.HttpRequest
 import aws.smithy.kotlin.runtime.http.request.HttpRequestBuilder
 import aws.smithy.kotlin.runtime.util.get
@@ -118,7 +119,7 @@ public class AwsHttpSigner(private val config: Config) : HttpSigner {
         val contextSignedBodyHeader = attributes.getOrNull(AwsSigningAttributes.SignedBodyHeader)
 
         // operation signing config is baseConfig + operation specific config/overrides
-        val signingConfig = AwsSigningConfig {
+        var signingConfig = AwsSigningConfig {
             region = attributes[AwsSigningAttributes.SigningRegion]
             service = attributes.getOrNull(AwsSigningAttributes.SigningService) ?: checkNotNull(config.service)
             credentials = signingRequest.identity as Credentials
@@ -151,6 +152,15 @@ public class AwsHttpSigner(private val config: Config) : HttpSigner {
                 // use the payload to compute the hash
                 else -> HashSpecification.CalculateFromPayload
             }
+        }
+
+        // Apply clock skew if necessary
+        signingConfig = if (attributes.contains(HttpOperationContext.ClockSkew)) {
+            val builder = signingConfig.toBuilder()
+            builder.signingDate = builder.signingDate?.plus(attributes[HttpOperationContext.ClockSkew])
+            builder.build()
+        } else {
+            signingConfig
         }
 
         if (signingConfig.useAwsChunkedEncoding) {
--- a/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
+++ b/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
@@ -47,7 +47,7 @@ public class ClockSkewInterceptor : HttpInterceptor {
         val logger = coroutineContext.logger<ClockSkewInterceptor>()
 
         if (currentSkew != Duration.ZERO) {
-            logger.info { "applying clock skew $currentSkew to client"}
+            logger.info { "applying clock skew $currentSkew to client" }
             context.executionContext[HttpOperationContext.ClockSkew] = currentSkew
         }
 
--- a/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
+++ b/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
@@ -1,3 +1,7 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
 package aws.smithy.kotlin.runtime.http.interceptors
 
 import aws.smithy.kotlin.runtime.client.ProtocolResponseInterceptorContext
@@ -58,7 +62,7 @@ public class ClockSkewInterceptor : HttpInterceptor {
             logger.warn { "client clock is skewed $skew, applying correction" }
             context.executionContext[HttpOperationContext.ClockSkew] = skew
         } else {
-            logger.info { "client clock ($clientTime) is not skewed from the server ($serverTime)"}
+            logger.info { "client clock ($clientTime) is not skewed from the server ($serverTime)" }
         }
 
         return context.protocolResponse
--- a/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
+++ b/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
@@ -4,6 +4,7 @@
  */
 package aws.smithy.kotlin.runtime.http.interceptors
 
+import aws.smithy.kotlin.runtime.client.ProtocolRequestInterceptorContext
 import aws.smithy.kotlin.runtime.client.ProtocolResponseInterceptorContext
 import aws.smithy.kotlin.runtime.http.operation.HttpOperationContext
 import aws.smithy.kotlin.runtime.http.request.HttpRequest
@@ -36,6 +37,26 @@ public class ClockSkewInterceptor : HttpInterceptor {
         public fun Instant.isSkewed(serverTime: Instant): Boolean = getSkew(serverTime).absoluteValue >= CLOCK_SKEW_THRESHOLD
     }
 
+    // Clock skew to be applied to all requests
+    private var currentSkew: Duration = Duration.ZERO
+
+    /**
+     * Apply the previously-computed skew, if it's set, to the execution context before signing
+     */
+    public override suspend fun modifyBeforeSigning(context: ProtocolRequestInterceptorContext<Any, HttpRequest>): HttpRequest {
+        val logger = coroutineContext.logger<ClockSkewInterceptor>()
+
+        if (currentSkew != Duration.ZERO) {
+            logger.info { "applying clock skew $currentSkew to client"}
+            context.executionContext[HttpOperationContext.ClockSkew] = currentSkew
+        }
+
+        return context.protocolRequest
+    }
+
+    /**
+     * After receiving a response, check if the client clock is skewed and apply a correction if necessary.
+     */
     public override suspend fun modifyBeforeDeserialization(context: ProtocolResponseInterceptorContext<Any, HttpRequest, HttpResponse>): HttpResponse {
         val logger = coroutineContext.logger<ClockSkewInterceptor>()
 
--- a/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
+++ b/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
@@ -1,0 +1,66 @@
+package aws.smithy.kotlin.runtime.http.interceptors
+
+import aws.smithy.kotlin.runtime.client.ProtocolResponseInterceptorContext
+import aws.smithy.kotlin.runtime.http.operation.HttpOperationContext
+import aws.smithy.kotlin.runtime.http.request.HttpRequest
+import aws.smithy.kotlin.runtime.http.response.HttpResponse
+import aws.smithy.kotlin.runtime.http.response.header
+import aws.smithy.kotlin.runtime.telemetry.logging.logger
+import aws.smithy.kotlin.runtime.time.Instant
+import kotlin.coroutines.coroutineContext
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.minutes
+
+public class ClockSkewInterceptor : HttpInterceptor {
+    public companion object {
+        /**
+         * How much must the clock be skewed before attempting correction
+         */
+        public val CLOCK_SKEW_THRESHOLD: Duration = 4.minutes
+
+        /**
+         * Get the current skew between the client time and [serverTime] as a [Duration].
+         * It may be negative if the serverTime is in the future.
+         * @param serverTime the server's time
+         */
+        public fun Instant.getSkew(serverTime: Instant): Duration {
+            return this.until(serverTime)
+        }
+
+        /**
+         * Determine whether the client's clock is skewed relative to the server.
+         * @param serverTime the server's time
+         */
+        public fun Instant.isSkewed(serverTime: Instant): Boolean {
+            return getSkew(serverTime).absoluteValue >= CLOCK_SKEW_THRESHOLD
+        }
+    }
+
+    public override suspend fun modifyBeforeDeserialization(context: ProtocolResponseInterceptorContext<Any, HttpRequest, HttpResponse>): HttpResponse {
+        val logger = coroutineContext.logger<ClockSkewInterceptor>()
+
+        val clientTime = context.protocolRequest.headers["x-amz-date"]?.let {
+            Instant.fromIso8601(it)
+        } ?: run {
+            logger.info { "client did not send \"x-amz-date\" header, skipping skew calculation" }
+            return context.protocolResponse
+        }
+
+        val serverTime = context.protocolResponse.header("Date")?.let {
+            Instant.fromRfc5322(it)
+        } ?: run {
+            logger.info { "service did not return \"Date\" header, skipping skew calculation" }
+            return context.protocolResponse
+        }
+
+        if (clientTime.isSkewed(serverTime)) {
+            val skew = clientTime.getSkew(serverTime)
+            logger.warn { "client clock is skewed $skew, applying correction" }
+            context.executionContext[HttpOperationContext.ClockSkew] = skew
+        } else {
+            logger.info { "client clock ($clientTime) is not skewed from the server ($serverTime)"}
+        }
+
+        return context.protocolResponse
+    }
+}
--- a/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
+++ b/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
@@ -24,7 +24,7 @@ public class ClockSkewInterceptor : HttpInterceptor {
 
         /**
          * Get the current skew between the client time and [serverTime] as a [Duration].
-         * It may be negative if the serverTime is in the future.
+         * It may be negative if the serverTime is in the past.
          * @param serverTime the server's time
          */
         public fun Instant.getSkew(serverTime: Instant): Duration = this.until(serverTime)
--- a/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
+++ b/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
@@ -78,9 +78,9 @@ public class ClockSkewInterceptor : HttpInterceptor {
         }
 
         if (clientTime.isSkewed(serverTime)) {
-            val skew = clientTime.getSkew(serverTime)
-            logger.warn { "client clock is skewed $skew, applying correction" }
-            context.executionContext[HttpOperationContext.ClockSkew] = skew
+            currentSkew = clientTime.getSkew(serverTime)
+            logger.warn { "client clock is skewed $currentSkew, applying correction" }
+            context.executionContext[HttpOperationContext.ClockSkew] = currentSkew
         } else {
             logger.info { "client clock ($clientTime) is not skewed from the server ($serverTime)" }
         }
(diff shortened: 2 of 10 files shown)


runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
```diff
@@ -0,0 +1,90 @@
...
+
+        /**
+         * Determine whether the client's clock is skewed relative to the server.
+         * @param serverTime the server's time
+         */
+        public fun Instant.isSkewed(serverTime: Instant): Boolean = getSkew(serverTime).absoluteValue >= CLOCK_SKEW_THRESHOLD
+    }
+
+    // Clock skew to be applied to all requests
+    private var currentSkew: Duration = Duration.ZERO
```
lauzadis, line 44: question: is this subject to race conditions, and should this be an `Atomic`?

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
```diff
@@ -0,0 +1,90 @@
...
+        return context.protocolRequest
+    }
+
+    /**
+     * After receiving a response, check if the client clock is skewed and apply a correction if necessary.
+     */
+    public override suspend fun modifyBeforeDeserialization(context: ProtocolResponseInterceptorContext<Any, HttpRequest, HttpResponse>): HttpResponse {
+        val logger = coroutineContext.logger<ClockSkewInterceptor>()
+
+        val clientTime = context.protocolRequest.headers["x-amz-date"]?.let {
```
aajtodd, line 66: **question**: Should we also be looking at `Date` header on the request and take first one that we find out of (`x-amz-date`, `Date`)?

Also should this check come after looking for server time and fallback to `Instant.now()`? That could be risky if the delta between when the server set the response header and when we get to here is large but in practice it should be way less than the skew threshold so it seems reasonable safe as fallback. This appears to be what Java [does](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/retry/ClockSkewAdjuster.java#L50).

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
```diff
@@ -0,0 +1,90 @@
...
+         * How much must the clock be skewed before attempting correction
+         */
+        public val CLOCK_SKEW_THRESHOLD: Duration = 4.minutes
+
+        /**
+         * Get the current skew between the client time and [serverTime] as a [Duration].
+         * It may be negative if the serverTime is in the past.
+         * @param serverTime the server's time
+         */
+        public fun Instant.getSkew(serverTime: Instant): Duration = this.until(serverTime)
```
aajtodd, line 34: These seem like they could be `internal`

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
```diff
@@ -0,0 +1,90 @@
...
+            context.executionContext[HttpOperationContext.ClockSkew] = currentSkew
+        }
+
+        return context.protocolRequest
+    }
+
+    /**
+     * After receiving a response, check if the client clock is skewed and apply a correction if necessary.
+     */
+    public override suspend fun modifyBeforeDeserialization(context: ProtocolResponseInterceptorContext<Any, HttpRequest, HttpResponse>): HttpResponse {
```
aajtodd, line 63: **question**: Should this only apply clock skew if there was an error? I suppose if we have breached the threshold it shouldn't hurt anything in theory to start compensating before it becomes an issue 🤔

aajtodd (changes requested):
Overall looks good, I'm going to hold off on the sibling PR until we get some things cleaned up here.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/policy/RetryErrorType.kt
```diff
@@ -30,4 +30,9 @@ public enum class RetryErrorType {
      * effect on the server
      */
     Transient,
+
+    /**
+     * An error indicating that the client and server clock are skewed. A retry may be attempted after correction.
+     */
+    Skew,
```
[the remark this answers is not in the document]
lauzadis: Sure, this new error type was created during my early iterating, I think `ClientSide` makes sense

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
lauzadis: From my tests, we don't send a `Date` header, but sure, I can check both headers and then fallback to `Instant.now()`

runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/AwsHttpSigner.kt
```diff
@@ -153,6 +154,15 @@ public class AwsHttpSigner(private val config: Config) : HttpSigner {
             }
         }
 
+        // Apply clock skew if necessary
+        signingConfig = if (attributes.contains(HttpOperationContext.ClockSkew)) {
+            val builder = signingConfig.toBuilder()
+            builder.signingDate = builder.signingDate?.plus(attributes[HttpOperationContext.ClockSkew])
```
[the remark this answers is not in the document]
lauzadis: It's actually never null by this point, because when building the config [we either take the configured value or `Instant.now()`.](https://github.com/awslabs/smithy-kotlin/blob/bc522c642af169e7e34c128a818af79a38f8b799/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt#L93)

runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/AwsHttpSigner.kt
```diff
@@ -153,6 +154,15 @@ public class AwsHttpSigner(private val config: Config) : HttpSigner {
             }
         }
 
+        // Apply clock skew if necessary
+        signingConfig = if (attributes.contains(HttpOperationContext.ClockSkew)) {
```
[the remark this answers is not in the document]
lauzadis: This is related to my other comment about `signingDate` defaulting to `Instant.now()`. We currently depend on that codepath in the Builder, I could refactor to default `Instant.now()` in this builder and then move this clock skew inside.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
lauzadis: Yes I was leaning towards applying skew even if there hasn't been an error yet

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/time/Instant.kt
```diff
@@ -43,6 +43,8 @@ public expect class Instant : Comparable<Instant> {
      */
     public operator fun minus(duration: Duration): Instant
 
+    public fun until(other: Instant): Duration
```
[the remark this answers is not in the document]
lauzadis: What's gained by making this an extension function, just more idiomatic? It's already used as an extension function `myInstant.until(otherInstant)`

runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/AwsHttpSigner.kt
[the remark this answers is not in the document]
aajtodd: I see, yeah I think we can move this into the builder still and just use `Instant.now() + skew`. We need to figure out how this behaves w.r.t pre-existing signing dates in the context.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/time/Instant.kt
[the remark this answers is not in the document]
aajtodd: Yeah just more idiomatic

runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/AwsHttpSigner.kt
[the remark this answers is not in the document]
lauzadis: We have no code that sets the signing date via context. I think if someone creates a custom interceptor to set signing date, we should just use that and not apply skew. What do you think?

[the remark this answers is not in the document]
aajtodd: I think that's fine as long as we document it. IIRC it's reason for existing is to facilitate testing.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
```diff
@@ -0,0 +1,96 @@
...
+        } ?: run {
+            logger.info { "service did not return \"Date\" header, skipping skew calculation" }
+            return context.protocolResponse
+        }
+
+        val clientTime = context.protocolRequest.headers["Date"]?.let {
+            Instant.fromRfc5322(it)
+        } ?: context.protocolRequest.headers["x-amz-date"]?.let {
+            Instant.fromIso8601(it)
+        } ?: Instant.now()
```
ianbotsf, line 83: **Correctness**: Defaulting to `Instant.now()` seems wrong. Since there are a bunch of suspension points around receiving, it could be many minutes later when we _process_ a response vs when the response _actually arrived_ at the client. This might calculate a client-local time that's vastly different than when the request was first serialized.

If we truly do want a fallback in case of no outgoing `Date`/`x-amz-date` header, I suggest we put it in the execution context during `modifyBeforeSigning`.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
ianbotsf: This is different from what Java v2 does. They only begin applying clock skew correction when they receive an exception indicating the error (see [`RetryableStageHelper`](https://github.com/aws/aws-sdk-java-v2/blob/13985e0668a9a0b12ad331644e3c4fd1385c2cd7/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/utils/RetryableStageHelper.java#L184-L186) and [`ClockSkewAdjuster`](https://github.com/aws/aws-sdk-java-v2/blob/13985e0668a9a0b12ad331644e3c4fd1385c2cd7/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/retry/ClockSkewAdjuster.java#L29-L61)).

This hardcodes us into believing that anything beyond 4 minutes of skew is intolerable no matter what the service thinks. Maybe some services choose to handle more. Maybe some services don't care about clock skew. I believe we should only handle skew updates during clock skew exceptions.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * After receiving a response, check if the client clock is skewed and apply a correction if necessary.
+     */
+    public override suspend fun modifyBeforeDeserialization(context: ProtocolResponseInterceptorContext<Any, HttpRequest, HttpResponse>): HttpResponse {
+        val logger = coroutineContext.logger<ClockSkewInterceptor>()
+
+        val serverTime = context.protocolResponse.header("Date")?.let {
+            Instant.fromRfc5322(it)
+        } ?: run {
+            logger.info { "service did not return \"Date\" header, skipping skew calculation" }
```
ianbotsf, line 75: **Nit**: Probably should be a `debug`-level warning.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
```diff
@@ -0,0 +1,96 @@
...
+     * Apply the previously-computed skew, if it's set, to the execution context before signing
+     */
+    public override suspend fun modifyBeforeSigning(context: ProtocolRequestInterceptorContext<Any, HttpRequest>): HttpRequest {
+        val logger = coroutineContext.logger<ClockSkewInterceptor>()
+
+        val skew = currentSkew
+        if (skew != Duration.ZERO) {
+            logger.debug { "applying clock skew $skew to client" }
+            context.executionContext[HttpOperationContext.ClockSkew] = skew
+        }
```
ianbotsf, line 61: **Nit**: Extreme corner case but if the skew is ever non-zero and then is re-adjusted to zero, this will fail to update the execution context for existing requests.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
```diff
@@ -0,0 +1,96 @@
...
+        } ?: Instant.now()
+
+        if (clientTime.isSkewed(serverTime)) {
+            val skew = clientTime.getSkew(serverTime)
+            logger.warn { "client clock is skewed $skew, applying correction" }
+            _currentSkew.getAndSet(skew)
+            context.executionContext[HttpOperationContext.ClockSkew] = skew
+        } else {
+            logger.debug { "client clock ($clientTime) is not skewed from the server ($serverTime)" }
+        }
```
ianbotsf, line 92: **Nit**: This seems like a `trace`-level message at best.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
```diff
@@ -0,0 +1,96 @@
...
+
+        val clientTime = context.protocolRequest.headers["Date"]?.let {
+            Instant.fromRfc5322(it)
+        } ?: context.protocolRequest.headers["x-amz-date"]?.let {
+            Instant.fromIso8601(it)
+        } ?: Instant.now()
+
+        if (clientTime.isSkewed(serverTime)) {
+            val skew = clientTime.getSkew(serverTime)
+            logger.warn { "client clock is skewed $skew, applying correction" }
```
ianbotsf, line 87: **Nit**: Include the skewed client time in the message.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/operation/HttpOperationContext.kt
```diff
@@ -50,6 +51,11 @@ public object HttpOperationContext {
      * Cached attribute level attributes (e.g. rpc.method, rpc.service, etc)
      */
     public val OperationAttributes: AttributeKey<Attributes> = AttributeKey("aws.smithy.kotlin#OperationAttributes")
+
+    /**
+     * The clock skew duration to apply to any time-based calculations in the operation (e.g. signing)
+     */
+    public val ClockSkew: AttributeKey<Duration> = AttributeKey("aws.smithy.kotlin#ClockSkew")
```
ianbotsf, line 58: **Correctness**: The KDocs say "any time-based calculations in the operation" but signature calculation is _the only_ place this is used. It's not applied to retry delays, deserialization, etc.

runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptorTest.kt
```diff
@@ -0,0 +1,133 @@
...
+        assertTrue(clientTime.isSkewed(serverTime))
+        assertEquals(1.days, clientTime.getSkew(serverTime))
+    }
+
+    @Test
+    fun testNegativeSkewed() {
+        val clientTime = Instant.fromRfc5322("Wed, 7 Oct 2023 16:20:50 -0400")
+        val serverTime = Instant.fromRfc5322("Wed, 6 Oct 2023 16:20:50 -0400")
+        assertTrue(clientTime.isSkewed(serverTime))
+        assertEquals(Duration.ZERO - 1.days, clientTime.getSkew(serverTime))
```
ianbotsf, line 48: **Nit**: `Duration.ZERO` is unnecessary. Just `-1.days` works.

ianbotsf (changes requested):

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/operation/HttpOperationContext.kt
lauzadis: Thanks for catching that. I wrote this description before doing the implementation, when I still had planned to update **all** time-based calculations in the SDK, but I think just signature calculation is enough for now.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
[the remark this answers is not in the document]
lauzadis: Instantiating a logger requires a `coroutineContext` which doesn't exist at the field level. Is there some alternate way to make a logger that I'm missing?

lauzadis: Handled by using `null` as the initial value for `currentSkew` and applying skew every time it's non-null. The side effect of this is if it's non-zero and then re-adjusted to zero, the zero skew will be applied to every request (it's not a problem, just a different code path)

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
lauzadis: We spoke offline and agreed we could ship this as-is and follow up if we see issues with it. 

The main concern is if a service starts enforcing sub-minute skew, the current implementation would fail to comply because the threshold is set at 4 minutes.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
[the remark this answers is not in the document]
ianbotsf: Oh you're right of course. Nevermind.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
```diff
@@ -0,0 +1,117 @@
...
+            logger.warn { "client clock ($clientTime) is skewed $skew from the server ($serverTime), applying correction" }
+            _currentSkew.getAndSet(skew)
+            context.executionContext[HttpOperationContext.ClockSkew] = skew
+
+            // Mark the exception as retryable
+            val ex = (context.response.exceptionOrNull() as? SdkBaseException)
+            ex?.sdkErrorMetadata?.attributes?.set(ErrorMetadata.Retryable, true)
+            ex?.let { return Result.failure(it) }
+        } else {
+            logger.trace { "client clock ($clientTime) is not skewed from the server ($serverTime)" }
```
aajtodd, line 112: **nit**: We probably don't need this log statement to say we didn't do anything.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/ClockSkewInterceptor.kt
```diff
@@ -0,0 +1,117 @@
...
+            return context.response
+        }
+
+        val clientTime = context.protocolRequest.headers["Date"]?.let {
+            Instant.fromRfc5322(it)
+        } ?: context.protocolRequest.headers["x-amz-date"]?.let {
+            Instant.fromIso8601(it)
+        } ?: context.executionContext[HttpOperationContext.ClockSkewApproximateSigningTime]
+
+        if (clientTime.isSkewed(serverTime, context.protocolResponse?.status?.description)) {
```
aajtodd, line 101: **correctness**: This is using the `HttpStatusCode` description which is not the same as the the AWS error code that the `isSkewed` implementation is looking for. The change to predicate off exceptions probably means we need to move this interceptor into `aws-sdk-kotlin/aws-runtime/aws-http` package instead of `smithy-kotlin` since this is now AWS specific.

aajtodd (changes requested):
Looks _mostly_ good but some correctness issues and question about where this should live to address.

runtime/protocol/aws-protocol-core/common/src/aws/smithy/kotlin/runtime/awsprotocol/ClockSkewInterceptor.kt
```diff
@@ -0,0 +1,119 @@
...
+            "RequestInTheFuture",
+        )
+
+        // Errors possibly caused by clock skew
+        private val POSSIBLE_CLOCK_SKEW_ERROR_CODES = listOf(
+            "PriorRequestNotComplete",
+            "RequestTimeout",
+            "RequestTimeoutException",
+            "InternalError",
+        )
```
ianbotsf, line 61: **Correctness**: These don't appear to be related to clock skew. Looks like [Java considers `InvalidSignatureException`, `SignatureDoesNotMatch`, and `AuthFailure` as possible errors](https://github.com/aws/aws-sdk-java-v2/blob/586b471405bd504d33389c93a436d33702bd7d8e/core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/AwsErrorCode.java#L57-L61).

ianbotsf (changes requested):

runtime/protocol/aws-protocol-core/common/src/aws/smithy/kotlin/runtime/awsprotocol/ClockSkewInterceptor.kt
lauzadis: 🤦 Good catch. I took the wrong list of exceptions from Java and then convinced myself they could be caused by clock skew.

runtime/protocol/aws-protocol-core/common/src/aws/smithy/kotlin/runtime/awsprotocol/ClockSkewInterceptor.kt
```diff
@@ -0,0 +1,118 @@
...
+        public val CLOCK_SKEW_THRESHOLD: Duration = 4.minutes
+
+        /**
+         * Determine whether the client's clock is skewed relative to the server.
+         * @return true if the service's response represents a definite clock skew error
+         * OR a *possible* clock skew error AND the skew exists. false otherwise.
+         * @param responseCodeDescription the server's response code description
+         * @param serverTime the server's time
+         */
+        internal fun Instant.isSkewed(serverTime: Instant, responseCodeDescription: String?): Boolean =
```
aajtodd, line 43: **nit**: `responseCodeDescription` -> `errorCode`

Also `responseCodeDescription` should probably be non-nullable since this always returns false otherwise which means it's really a required parameter for determining if a clock is skewed.

aajtodd (approved):

ianbotsf (approved):

PR #981, opened 2023-10-16
fix: Prevent NoSuchMethodError when a slf4j1 logger is used with slf4j2 in the classpath

NatanLifshitz:
The `useSlf4j2x` flag is changed to be more precise and check if the `Logger` instance actually supports slf4j2 methods, as opposed to checking if slf4j2 exists in the classpath which fails with a slf4j1 logger in a slf4j2 classpath. 

A test project exists to demonstrate this issue. It does not pass with the old code, but does pass with this PR's code. I will note it was difficult as hell to reproduce in a vacuum. 
## Issue \#
https://github.com/awslabs/aws-sdk-kotlin/issues/1077
## Description of changes
Gradle overrides most of the slf4j classes, but it overrides it with a slf4j1 API, which doesn't support slf4j2 methods. This means if you tried to run the AWS SDK for Kotlin in a Gradle buildscript, it would use Gradle's slf4fj1 logger, and the AWS SDK was probably pulling slf4j2 itself into the classpath, causing this problem. 


runtime/observability/logging-slf4j2/jvm/src/aws/smithy/kotlin/runtime/telemetry/logging/slf4j/Slf4jLoggerProvider.kt
```diff
@@ -7,23 +7,31 @@ package aws.smithy.kotlin.runtime.telemetry.logging.slf4j
...
+                // Considering Logger#atLevel only exists in slf4j2, this will throw in slf4j1 and proceed without throwing in slf4j2.
+                logger.atLevel(Level.DEBUG)
+                true
+            } catch (e: NoSuchMethodError) {
+                LoggerFactory.getLogger(Slf4jLoggerProvider::class.java).warn("falling back to SLF4J 1.x compatible binding")
+                false
+            }
+        }
+        return useSlf4j2x!!
     }
```
ianbotsf, line 30: **Correctness**: This new logic for determining whether to use 1.x or 2.x looks good but it's being stored in a mutable `var` and may potentially be invoked by multiple threads simultaneously. This could lead to more work being done than necessary.

Please refactor this to use a `val` initialized at object load time (like before). This will ensure that the `useSlf4j2x` value only gets initialized once no matter how many threads might be calling `getOrCreateLogger` simultaneously.

tests/integration/slf4j-hybrid-consumer/src/test/kotlin/aws/smithy/kotlin/test/LoggingSlf4jHybridTest.kt
```diff
@@ -0,0 +1,34 @@
...
+ */
+package aws.smithy.kotlin.test
+
+import aws.smithy.kotlin.runtime.telemetry.logging.LogLevel
+import aws.smithy.kotlin.runtime.telemetry.logging.getLogger
+import aws.smithy.kotlin.runtime.telemetry.logging.slf4j.Slf4jLoggerProvider
+import aws.smithy.kotlin.runtime.telemetry.logging.warn
+import kotlin.test.Test
+
+class LoggingSlf4jHybridTest {
```
ianbotsf, line 13: **Comment**: Thanks for submitting a test project that confirms the correct behavior. This is great!

ianbotsf (changes requested):

runtime/observability/logging-slf4j2/jvm/src/aws/smithy/kotlin/runtime/telemetry/logging/slf4j/Slf4jLoggerProvider.kt
NatanLifshitz: It's not possible to initialize this at the object load time because it depends on a `org.slf4j.Logger` instance.
There's a couple of ways to approach this to get more correct multithreading results: using an atomic boolean, using a wrapper class around `org.slf4j.Logger`, etc. 
But then you need to realize: what are we solving? What happens if multiple threads access this at the same time? The answer is as you said:
> This could lead to more work being done than necessary.
This "extra" work is just catching an exception (only in slf4j1) and calling a method. Wrapping this code up with more ceremony would be a pointless micro-optimization, but I could do it if you'd like.

tests/integration/slf4j-hybrid-consumer/src/test/kotlin/aws/smithy/kotlin/test/LoggingSlf4jHybridTest.kt
NatanLifshitz: Yeah this was 99% of the work

runtime/observability/logging-slf4j2/jvm/src/aws/smithy/kotlin/runtime/telemetry/logging/slf4j/Slf4jLoggerProvider.kt
NatanLifshitz: Actually, I've thought of a way to do this only at object load. See latest revision.

ianbotsf (approved):

PR #984, opened 2023-10-19
fix!: gate logging intermediate signing results which may contain sensitive information

aajtodd:
## Issue \#

n/a

## Description of changes

We gate logging HTTP requests and responses because headers and query parameters (and the body) could have sensitive information bound to them from the model. We currently log canonical request during signing which may contain many of the same details (headers, query params, etc). This PR adds `LogMode.LogSigning` which gates logging these details unless explicitly opted into.

### DIFF
--- a/.changes/b1a1bee4-e3e4-4520-ae8e-48d3b0e94ae9.json
+++ b/.changes/b1a1bee4-e3e4-4520-ae8e-48d3b0e94ae9.json
@@ -1,0 +1,5 @@
+{
+    "id": "b1a1bee4-e3e4-4520-ae8e-48d3b0e94ae9",
+    "type": "bugfix",
+    "description": "Do not log intermediate signature calculations without explicit opt-in via `LogMode.LogSignging`."
+}
--- a/.changes/b1a1bee4-e3e4-4520-ae8e-48d3b0e94ae9.json
+++ b/.changes/b1a1bee4-e3e4-4520-ae8e-48d3b0e94ae9.json
@@ -1,5 +1,5 @@
 {
     "id": "b1a1bee4-e3e4-4520-ae8e-48d3b0e94ae9",
     "type": "bugfix",
-    "description": "Do not log intermediate signature calculations without explicit opt-in via `LogMode.LogSignging`."
+    "description": "Do not log intermediate signature calculations without explicit opt-in via `LogMode.LogSigning`."
 }
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,7 +5,7 @@ kotlin.mpp.stability.nowarn=true
 kotlin.native.ignoreDisabledTargets=true
 
 # SDK
-sdkVersion=0.27.8-SNAPSHOT
+sdkVersion=0.28.0-SNAPSHOT
 
 # kotlin
 kotlinVersion=1.8.22
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,7 +5,7 @@ kotlin.mpp.stability.nowarn=true
 kotlin.native.ignoreDisabledTargets=true
 
 # SDK
-sdkVersion=0.28.0-SNAPSHOT
+sdkVersion=0.27.8-SNAPSHOT
 
 # kotlin
 kotlinVersion=1.8.22
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
@@ -164,6 +164,14 @@ public class AwsSigningConfig(builder: Builder) {
      */
     public val expiresAfter: Duration? = builder.expiresAfter
 
+    /**
+     * Flag enabling whether detailed trace logging is enabled (if the signer implementation supports it). When true
+     * signers should emit intermediate logging details such as the canonical request and string to sign at the trace
+     * level. This is an opt-in configuration because these intermediate outputs may contain sensitive fields bound
+     * to headers, URI, or query parmaeters.
+     */
+    public val enableTraceLogging: Boolean = builder.enableTraceLogging
+
     public fun toBuilder(): Builder = Builder().also {
         it.region = region
         it.service = service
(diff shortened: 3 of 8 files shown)


runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
```diff
@@ -164,6 +164,14 @@ public class AwsSigningConfig(builder: Builder) {
      */
     public val expiresAfter: Duration? = builder.expiresAfter
 
+    /**
+     * Flag enabling whether detailed trace logging is enabled (if the signer implementation supports it). When true
+     * signers should emit intermediate logging details such as the canonical request and string to sign at the trace
+     * level. This is an opt-in configuration because these intermediate outputs may contain sensitive fields bound
+     * to headers, URI, or query parmaeters.
+     */
+    public val enableTraceLogging: Boolean = builder.enableTraceLogging
```
aajtodd, line 173: Not sure what we wanted to call this, let the 🚲 🏚️  commence!

.changes/b1a1bee4-e3e4-4520-ae8e-48d3b0e94ae9.json
```diff
@@ -0,0 +1,5 @@
+{
+    "id": "b1a1bee4-e3e4-4520-ae8e-48d3b0e94ae9",
+    "type": "bugfix",
+    "description": "Do not log intermediate signature calculations without explicit opt-in via `LogMode.LogSignging`."
```
lauzadis, line 4: typo: LogSigning

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
```diff
@@ -164,6 +164,14 @@ public class AwsSigningConfig(builder: Builder) {
      */
     public val expiresAfter: Duration? = builder.expiresAfter
 
+    /**
+     * Flag enabling whether detailed trace logging is enabled (if the signer implementation supports it). When true
+     * signers should emit intermediate logging details such as the canonical request and string to sign at the trace
```
lauzadis, line 169: > and string to sign

It looks the string to sign always gets logged (which I think is what we agreed to do). So should we only reference the canonical request in these docs?

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
lauzadis: I like this name *if* we were controlling all `logger.trace` calls with it, but we **always** log the "string to sign" at the trace level regardless of this flag. I think we should either rename this to better describe exactly what gets logged, or block the string-to-sign logging with this flag too.

runtime/smithy-client/common/test/aws/smithy/kotlin/runtime/client/LogModeTest.kt
```diff
@@ -33,7 +33,7 @@ class LogModeTest {
     fun testToString() {
         val mode = LogMode.allModes().reduce { acc, curr -> acc + curr }
         assertTrue { LogMode.allModes().all { mode.isEnabled(it) } }
-        val expected = "LogRequest|LogRequestWithBody|LogResponse|LogResponseWithBody"
+        val expected = "LogRequest|LogRequestWithBody|LogResponse|LogResponseWithBody|LogSigning"
```
lauzadis, line 36: Could you also add this value to the composite string test on L46? (tests the reverse of this, `String`->`LogMode`)

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
aajtodd: Yeah this is why I wanted to bikeshed over it because I had a similar feeling. I don't want to call it something too specific like `logCanonicalRequests` in the event there are other signing steps or calculations that we may want to put behind this flag.

Open to suggestions

lauzadis: `enableSensitiveTraceLogging`?

gradle.properties
```diff
@@ -5,7 +5,7 @@ kotlin.mpp.stability.nowarn=true
 kotlin.native.ignoreDisabledTargets=true
 
 # SDK
-sdkVersion=0.27.8-SNAPSHOT
+sdkVersion=0.28.0-SNAPSHOT
```
ianbotsf, line 8: **Question**: Why did this necessitate a major version bump?

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
ianbotsf: I'd go the other way, actually. The `SdkLogMode` option is named `LogSigning`. At a glance, that indicates to me that _all_ signing logging will be enabled/disabled by that flag. (The KDocs add some nuance but I think they actually _undermine_ the name.)

I think we should just make that the case: canonical request, string to sign, _and_ final signature should be gated by `LogSigning`.

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
ianbotsf: In tandem with my feedback above about `LogSigning` controlling _all_ signing logging, I think this should just be called `enableLogging`.

gradle.properties
aajtodd: It didn't, it fixes a previous commit where it should have been bumped: https://github.com/awslabs/smithy-kotlin/commit/681379d3cdc6ef578ae3b03b770dda1fc4a35aa0

aajtodd: I was expecting the nullability PR to land with everything else which bumped the version, that won't land until next week (hopefully with the name changes which would have also bumped it)

aajtodd: Blerg, it's causing downstream CI issues though so I'm reverting. @lauzadis we just need to remember to bump this before release.

ianbotsf (approved):

PR #985, opened 2023-10-24
fix: run `test` and `jvmTest` in downstream CI

lauzadis:
Run `test` and `jvmTest` tasks in the downstream CI. `publishToMavenLocal` does not run a full build, so currently the unit tests are not run as part of this CI

## Description of changes

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -109,4 +109,5 @@ jobs:
         # the generated protocol test projects will use whatever the SDK thinks the version of smithy-kotlin should be
         sed -i "s/smithy-kotlin-version = .*$/smithy-kotlin-version = \"$SMITHY_KOTLIN_VERSION\"/" gradle/libs.versions.toml
         ./gradlew --parallel publishToMavenLocal
+        ./gradlew test jvmTest
         ./gradlew testAllProtocols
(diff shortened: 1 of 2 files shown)


aajtodd (approved):

.github/workflows/continuous-integration.yml
```diff
@@ -109,4 +109,5 @@ jobs:
         # the generated protocol test projects will use whatever the SDK thinks the version of smithy-kotlin should be
         sed -i "s/smithy-kotlin-version = .*$/smithy-kotlin-version = \"$SMITHY_KOTLIN_VERSION\"/" gradle/libs.versions.toml
         ./gradlew --parallel publishToMavenLocal
+        ./gradlew test jvmTest
```
ianbotsf, line 112: **Question**: Doesn't this run the Linux tests too, which we don't want?

lauzadis: No, `allTests` would run the Linux suite, `test` alone does not

aajtodd: Linux tests are disabled by our plugin right now: https://github.com/awslabs/aws-kotlin-repo-tools/blob/main/build-plugins/src/main/kotlin/aws/sdk/kotlin/gradle/kmp/ConfigureTargets.kt#L167

But also no `test` covers codegen and integration tests that use the `kotlin(jvm)` plugin, `jvmTest` covers KMP.

ianbotsf (approved):

Issue #659, opened 2022-06-15
Refactor URL encoded/raw API


aajtodd:
Take a closer look at how we deal with encoded vs raw URL paths and query parameters. APIs should be designed to (1) easy to understand whether a raw or encoded key/value is being consumed and (2) minimize the amount of encoding/decoding we have to do especially w.r.t signing.

* We should have explicit APIs to get at raw vs encoded URL path and query parameters
* We should have APIs to add already encoded query parameters (and potentially path segments)
* [encodedPath](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/Url.kt#L68) should really be called `encodedPathAndQuery` and drop fragment.
* Fix encoding to [also encode keys](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/QueryParameters.kt#L72), it looks like we currently only encode the value. This hasn't been an issue and likely won't be because keys come from the model and are constrained by the naming rules for identifiers in Smithy. We should fix it nevertheless
* Drop `encodeParameters` from `URL` and fix presigning
    * The signing process itself needs fixed probably to enumerate the delta. Signing returns query parameters already encoded. This could be as simple as _always_ maintaining query parameters in their encoded state rather than raw such that we can just add the missing (already encoded) keys/values as we do now

Inspiration from [okhttp](https://github.com/square/okhttp/blob/master/okhttp/src/jvmMain/kotlin/okhttp3/HttpUrl.kt#L524) and [ktor](https://github.com/ktorio/ktor/blob/main/ktor-http/common/src/io/ktor/http/Url.kt#L57).

PR #989, opened 2023-10-27
chore: refactor URL types and supporting classes

ianbotsf:
## Issue \#

Addresses #659 

## Description of changes

* **Breaking**: Completely rewrite URL APIs to clarify content encoding, when data is in which state, and to reduce the number of times data is encoded/decoded.
  * Remove several now-unused functions and classes mainly relating to URL encoding/decoding
* **Breaking**: Move several classes out of `aws.smithy.kotlin.runtime.util` into better packages like `text` and `collections`
* Add new `MultiMap`/`MutableMultiMap` collection types for supporting URL query parameters
* Add new collection views for providing live mapping of underlying collections. These facilitate providing, for example, a view of data in both encoded and decoded forms while still having a single source of truth.

### ⚠️ Callouts

* Tests are in a bit of a bad state right now. I ripped out a bunch of tests against the old APIs but don't yet have the same level of coverage for the new ones...at least at the unit test level. I'll add more but feel free to highlight particular areas you feel could use better support.
* Still looking for feedback on the ergonomics and usability of the new APIs. The updated tests/codegen should provide some examples.

**Companion PR**: https://github.com/awslabs/aws-sdk-kotlin/pull/1119


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/text/Scanner.kt
lauzadis, line 1: correctness: missing tests

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/text/Scanner.kt
```diff
@@ -0,0 +1,153 @@
...
+     * @param literals One or more strings to search for at/after the current position
+     * @param handler The handler to invoke on the substring *up to but not including* the nearest found element of
+     * [literals].
+     */
+    public fun requireAndSkip(vararg literals: String, handler: (String) -> Unit) {
+        val (literal, startIndex) = requireNotNull(findAnyOf(literals)) { "Cannot find any of ${literals.toList()}" }
+        processAndSkip(literal, startIndex, handler)
+    }
+
+    override fun toString(): String = "Scanner(remainingText='${text.substring(currentIndex)}')"
```
lauzadis, line 116: would it be useful to also print the text as it was given?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/text/encoding/Encoding.kt
```diff
@@ -0,0 +1,129 @@
...
+
+    public val name: String
+
+    public fun decode(encoded: String): String
+    public fun encode(decoded: String): String
+
+    public fun encodableFromDecoded(decoded: String): Encodable = Encodable(decoded, encode(decoded), this)
+    public fun encodableFromEncoded(encoded: String): Encodable {
+        val decoded = decode(encoded)
+        val reencoded = encode(decoded) // TODO is this right?
```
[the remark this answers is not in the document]
ianbotsf: That's an option and it's why I left the `TODO` in the code. My opinion is that we'd want to normalize any encoded representation we receive in case the encoding was flawed or non-standard.

For instance, if someone gives us the encoded string `"%61%62%63"`, its decoded representation is `"abc"` but its encoded representation _should also_ be `"abc"` since those characters don't require encoding. Worse yet, if someone gives us the encoded string `"💥"`, we'd want to reencode that as `"%F0%9F%92%A5"`.

I think the alternatives here are:
1. Re-encode any given encoded string to automatically normalize it (currently what the PR shows)
1. Accept the given encoded string as-is and accept that this may cause problems later on (e.g., during signing, in service responses, etc.)
1. Inspect the encoded string to verify its correctness, throw an exception if malformed (e.g., `"💥"`). If the encoded string is non-standardly encoded (e.g., `"%61%62%63"`):
    1. Throw an exception **–or–**
    1. Silently ignore

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/text/Scanner.kt
ianbotsf: Yes this PR omits basically all tests except for some I wrote to help me verify some basics of implementation. After we settle on the API design, I'll write the rest of the tests for this class and others.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/text/Scanner.kt
ianbotsf: I personally don't think so for two reasons:

1. It's likely to be verbose and you could wind up with to-string representations like `"Scanner(text='The quick brown fox jumped over the lazy dogs',remainingText='quick brown fox jumped over the lazy dogs')"`. Seems noisy.
1. The scanner is stateful and forward-only. Any text that was previously processed is effectively out of scope and seems irrelevant in a `toString` output. The only reason this class _doesn't_ discard already-processed text (e.g., by making `text` a `var` and replacing it via `substring`) is for performance.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/text/encoding/Encoding.kt
```diff
@@ -0,0 +1,129 @@
...
+public interface Encoding {
+    @InternalApi
+    public companion object {
+        public val UserInfo: Encoding = PercentEncoding("user info", VALID_UCHAR)
+        public val Path: Encoding = PercentEncoding("path", VALID_PCHAR)
+        public val Query: Encoding = PercentEncoding("query string", VALID_QCHAR, mapOf(' ' to '+'))
+        public val Fragment: Encoding = PercentEncoding("fragment", VALID_FCHAR)
+
+        internal val None = object : Encoding {
+            override val name = "(no encoding)"
```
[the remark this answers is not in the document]
ianbotsf: This value is mainly used for `Encodable.toString`. Any hypothetical to-string of a non-encoded value would look weird with an empty string name: `"Encodable(decoded=foo, encoded=foo, encoding=)"`.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/text/encoding/Encoding.kt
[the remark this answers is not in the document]
lauzadis: Ok, it makes sense to me to re-encode values which don't require encoding ("abc") or re-encoding emojis to ensure downstream usages will work correctly.

It just stood out to me from an API perspective because I would expect that `"%61%62%63" == encodableFromEncoded("%61%62%63").encoded`. Can we log a message in cases where the re-encoded value is not the same as the given encoded value?

aajtodd (commented):
Overall looks ok, no huge objections just some cleanup, polish, and of course tests. 

It's hard to tell how big of a change this will be. I still have some concerns on usability but those may be addressed easily IDK.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/text/encoding/Encoding.kt
```diff
@@ -0,0 +1,129 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.util.text.encoding
```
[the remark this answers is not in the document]
ianbotsf: I like the idea of rethinking the **util** package but I'm not quite clear on the suggestion. More verbosely, are you suggesting this:

* **aws.smithy.kotlin.runtime**
  * **collections**
    * MutableListView.kt
    * ValuesMap.kt
    * ...
  * **text**
    * Scanner.kt
    * ...
    * **encoding**
      * Base64.kt
      * Encoding.kt
      * Hex.kt
      * ...
  * **util**
    * ...(basically whatever's left)

If so, I like it and I can incorporate it into the wider `Url` refactor.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/text/encoding/Encoding.kt
```diff
@@ -0,0 +1,129 @@
...
+
+import aws.smithy.kotlin.runtime.InternalApi
+
+private val ALPHA = (('A'..'Z') + ('a'..'z')).toSet()
```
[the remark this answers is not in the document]
ianbotsf: The only other place I found similar things was in the existing URL parsing logic. Once that's replaced, I don't know that these will be shared with any other code.

That said, I don't have a problem making them public but I should probably move them to be namespaced under the `PercentEncoding` companion object.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/text/encoding/Encoding.kt
```diff
@@ -0,0 +1,129 @@
...
+private val DIGIT = ('0'..'9').toSet()
+private val UNRESERVED = ALPHA + DIGIT + setOf('-', '.', '_', '~')
+private val SUB_DELIMS = setOf('!', '$', '&', '\'', '(', ')', '*', ',', ';', '=')
+private val VALID_UCHAR = UNRESERVED + SUB_DELIMS
+private val VALID_PCHAR = VALID_UCHAR + setOf(':', '@')
+private val VALID_FCHAR = VALID_PCHAR + setOf('/', '?')
+private val VALID_QCHAR = VALID_FCHAR - setOf('&', '=')
+
+@InternalApi
+public interface Encoding {
```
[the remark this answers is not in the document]
ianbotsf: That's a good point. `Encoding` itself can theoretically represent _any_ encoding, including Base64, hex bytes, etc. I should move out `PercentEncoding` to its own file. The companion `val`s like `UserInfo`, `Path`, etc., should also move.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/text/encoding/Encoding.kt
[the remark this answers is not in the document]
ianbotsf: Yes, I'll add a warning and also make clear in the KDocs that this will _reencode_ encoded text for normalization and validation purposes.

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/net/newnet/UrlTest.kt
```diff
@@ -0,0 +1,116 @@
...
+    @Test
+    fun testEquivalenceOfQueryStrings() {
+        testEquivalence(
+            // "https://amazon.com?", // Old parser strips '?' with empty query params...is bug?
+            "https://amazon.com?foo=bar",
+            // "https://amazon.com?foo=bar&baz=qux", // Old parser sorts query by key...is bug?
+            "https://amazon.com?foo=f%F0%9F%98%81o",
+            // "https://amazon.com?foo=f+o%20o", // Old parser encodes qparam ' ' to %20 instead of '+'...is bug?
+        )
+    }
```
ianbotsf, line 62: **Question**: No one remarked on these commented-out test cases. Our existing URL parser has some non-standard behavior that was surprising to me, such as stripping empty query strings and empty fragments. The new parser (correctly, I believe) retains them. This matches the RFC but is a different from our current behavior and may cause unexpected problems.

Do we have any thoughts about RFC adherence vis-à-vis current behavior?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/text/encoding/Encoding.kt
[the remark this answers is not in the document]
aajtodd: Yes basically something like that. We clearly have text and collection related _things_ that are just sort of random and could be in their own packages at this point I think.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/text/encoding/Encoding.kt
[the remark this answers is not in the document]
aajtodd: Yeah I could be mistaken, if not we can leave them private.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/newnet/Url.kt
```diff
@@ -0,0 +1,264 @@
...
+            set(value) { path.parseEncoded(value) }
+
+        // Query parameters
+
+        private var queryParameters = url?.queryParameters?.toBuilder()
+
+        /**
+         * Remove all query parameters from this URL
+         */
+        public fun clearQueryParameters() {
```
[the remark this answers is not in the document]
aajtodd: Previously we handled this with the [forceQuery](https://github.com/awslabs/smithy-kotlin/blob/c15956eeddaed8dedd6640ac60e7d291416a92be/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/Url.kt#L87) parameter on the builder. I'm not suggesting to ignore this but it's not something we've really needed or used much so can probably be moved to a boolean flag like that (either under `Url.Builder` directly or as part of `QueryParameters[.Builder])`). 

Mostly I think it's an ergonomics thing and IMHO the non-nullably empty version of this makes it easier to consume and composes better (whereas making it nullable hoists a bunch of concerns of it's parts into the `Url.Builder` directly rather than letting the individual part deal with it).

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/newnet/Url.kt
```diff
@@ -0,0 +1,264 @@
...
+         */
+        public fun queryParameters(block: QueryParameters.Builder.() -> Unit) {
+            val queryParameters = this.queryParameters ?: QueryParameters.Builder().also { this.queryParameters = it }
+            queryParameters.apply(block)
+        }
+
+        /**
+         * Get or set the query parameters as a **decoded** string
+         */
+        public var queryParametersDecoded: String?
```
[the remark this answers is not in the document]
aajtodd: We _could_ I'm not sure I'd bother it's an odd case that is handled more generically by splitting a string as query params or (`parameters.addAll(QueryParameters.parse(uri.query))`). I can go either way it's just not something most code needs.

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/net/newnet/UrlTest.kt
aajtodd: Stripping empty query strings and fragments 🤷 . I think a parser could set these as empty and if we go with my suggestion of a flag you could set the flag but in practice it's not really going to matter for us I don't think.

The `%20` vs `+` depends on whether the content type is `application/x-www-form-urlencoded`. 

https://en.wikipedia.org/wiki/Percent-encoding#The_application.2Fx-www-form-urlencoded_type

Our existing [implementation](https://github.com/awslabs/smithy-kotlin/blob/66e95fb45a9a508677eac67dcbd210820eb4dea7/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/QueryParameters.kt#L69) branches on this but from what I can tell we have generally not been using form-url encoding. Your decode function will have to handle this for query params but for encode I'd probably default to `%20` and maybe provide a flag or something to enable encoding as `+` like we do now.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/url/QueryParameters.kt
```diff
@@ -0,0 +1,217 @@
...
+import aws.smithy.kotlin.runtime.collections.MutableMultiMap
+import aws.smithy.kotlin.runtime.collections.mutableMultiMapOf
+import aws.smithy.kotlin.runtime.collections.views.asView
+import aws.smithy.kotlin.runtime.text.encoding.Encodable
+import aws.smithy.kotlin.runtime.text.encoding.PercentEncoding
+
+/**
+ * Represents the parameters in a URL query string.
+ */
+public class QueryParameters private constructor(
```
aajtodd, line 17: **correctness**: This representation no longer allows for multiple values to be bound to the same parameter name

e.g. `?foo=1&foo=2&foo=3`

That was one of the reasons for sharing `ValuesMap` between headers and query is both have similar semantics between single values, multiple values and whether to set/override vs append. 

Unless I'm missing something (possible) I *think* this representation may be flawed currently.

runtime/auth/aws-signing-default/common/src/aws/smithy/kotlin/runtime/auth/awssigning/RequestMutator.kt
```diff
@@ -49,14 +47,14 @@ internal class DefaultRequestMutator : RequestMutator {
...
             AwsSignatureType.HTTP_REQUEST_VIA_QUERY_PARAMS -> {
-                with(canonical.request.url.parameters) {
-                    set("X-Amz-Signature", signatureHex)
+                canonical.request.url.parameters.decodedParameters.put("X-Amz-Signature", signatureHex)
 
-                    entries().forEach {
-                        remove(it.key)
-                        appendAll(it.key, it.value.map(String::urlReencodeComponent))
-                    }
+                /* TODO don't need to reencode because their already canonicalized by `Encodable`?
```
aajtodd, line 52: did we answer this?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/url/Url.kt
```diff
@@ -0,0 +1,347 @@
...
+            val requestRelativePath = encoded
+                .substring(splitAt)
+                .ensurePrefix("/") // Request-line URI requires '/' for empty path
+
+            return encoded to requestRelativePath
+        }
+    }
+
+    private val encoded: String
+    public val requestRelativePath: String
```
aajtodd, line 123: docs?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/request/HttpRequestBuilder.kt
```diff
@@ -113,7 +102,7 @@ public suspend fun dumpRequest(request: HttpRequestBuilder, dumpBody: Boolean):
     val buffer = SdkBuffer()
 
     // TODO - we have no way to know the http version at this level to set HTTP/x.x
-    buffer.writeUtf8("${request.method} ${request.url.encodedPath}\r\n")
+    buffer.writeUtf8("${request.method} ${request.url.requestRelativePath}\r\n")
```
aajtodd, line 105: **question**: Why would this (and elsewhere where we've replaced `encodedPath` with `requestRelativePath`) not be `path.encoded`?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/url/Url.kt
```diff
@@ -0,0 +1,347 @@
...
+ * @param fragment Optional fragment component of this URL (without the `#` character)
+ * @param userInfo Optional user authentication information for this URL
+ */
+public class Url private constructor(
+    public val scheme: Scheme,
+    public val host: Host,
+    public val port: Int,
+    public val path: UrlPath,
+    public val parameters: QueryParameters,
+    public val userInfo: UserInfo,
```
aajtodd, line 30: **suggestion** this could remain nullable, we don't use it much and unlike `QueryParameters` and `Headers` it's "empty" state is less useful.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/collections/ValuesMap.kt
```diff
@@ -2,9 +2,10 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
-package aws.smithy.kotlin.runtime.util
+package aws.smithy.kotlin.runtime.collections
 
 import aws.smithy.kotlin.runtime.InternalApi
+import aws.smithy.kotlin.runtime.util.CaseInsensitiveMap
```
aajtodd, line 8: **fix**: Should this not be in collections?

Also (and this can be a follow on PR) we should look at what's left in util and see if it still belongs there. On surface `Stack` and `ReadThroughCache` utils probably belongs in collections, possibly `Attributes`?. Also some of the other APIs probably belong in `io` or elsewhere.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/text/encoding/PercentEncoding.kt
```diff
@@ -0,0 +1,114 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.text.encoding
+
+import aws.smithy.kotlin.runtime.InternalApi
+
+@InternalApi
+public class PercentEncoding(
```
aajtodd, line 10: docs

aajtodd (commented):
Overall looks good

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/url/QueryParameters.kt
lauzadis: I *think* it's okay, because the query string keys must be case-sensitively unique according to Smithy. https://smithy.io/2.0/spec/http-bindings.html#httpquery-trait

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/url/QueryParameters.kt
```diff
@@ -0,0 +1,217 @@
...
+    private val delegate: MultiMap<Encodable, Encodable>,
+
+    /**
+     * A flag indicating whether to force inclusion of the `?` query separator even when there are no parameters (e.g.,
+     * `http://foo.com?` vs `http://foo.com`).
+     */
+    public val forceQuery: Boolean,
```
lauzadis, line 24: naming: `forceQuerySeparator`?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/text/encoding/Encoding.kt
```diff
@@ -0,0 +1,31 @@
...
+
+    public val name: String
+
+    public fun decode(encoded: String): String
+    public fun encode(decoded: String): String
+
+    public fun encodableFromDecoded(decoded: String): Encodable = Encodable(decoded, encode(decoded), this)
+    public fun encodableFromEncoded(encoded: String): Encodable {
+        val decoded = decode(encoded)
+        // val reencoded = encode(decoded) // TODO is this right?
```
lauzadis, line 28: nit: leftover TODO

lauzadis (approved):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/url/QueryParameters.kt
aajtodd: Right but the key is unique, this is for multiple _values_ assigned to a key.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt
```diff
@@ -314,46 +314,47 @@ abstract class HttpBindingProtocolGenerator : ProtocolGenerator {
...
                             tsLabel
                         } else {
                             "input.${binding.member.defaultName()}"
                         }
 
-                        val encodeSymbol = RuntimeTypes.Http.Util.encodeLabel
-                        val encodeFn = if (segment.isGreedyLabel) {
-                            writer.format("#T(greedy = true)", encodeSymbol)
+                        if (segment.isGreedyLabel) {
+                            write("addAll(#S.split(#S))", "\${$identifier}", '"')
```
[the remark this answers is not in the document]
ianbotsf: You're right on both accounts. I've fixed the `"` splitting to be `/` as intended. I'm fixed part of the label encoding but I still have more changes to make to support query parameters.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/url/UrlPath.kt
```diff
@@ -0,0 +1,209 @@
...
+            set(value) { parseDecoded(value) }
+
+        /**
+         * Get or set the URL path as an **encoded** string.
+         */
+        public var encoded: String
+            get() = asEncoded(segments, trailingSlash)
+            set(value) { parseEncoded(value) }
+
+        public val segments: MutableList<Encodable> = path?.segments?.toMutableList() ?: mutableListOf()
```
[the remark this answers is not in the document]
ianbotsf: We could make it `@InternalApi`. It seemed _slightly_ weird that the built `UrlPath` would have a `segments: List<Encodable>` member but the builder wouldn't have a mutable equivalent. But practically speaking, most times users will interact with the encoded/decoded variants.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/url/UrlPath.kt
```diff
@@ -0,0 +1,209 @@
...
+        /**
+         * Normalizes the segments of a URL path according to the following rules:
+         * * The returned path always begins with `/` (e.g., `a/b/c` → `/a/b/c`)
+         * * The returned path ends with `/` if the input path also does
+         * * Empty segments are discarded (e.g., `/a//b` → `/a/b`)
+         * * Segments of `.` are discarded (e.g., `/a/./b` → `/a/b`)
+         * * Segments of `..` are used to discard ancestor paths (e.g., `/a/b/../c` → `/a/c`)
+         * * All other segments are unmodified
+         */
+        public fun normalize() {
```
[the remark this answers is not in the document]
ianbotsf: The SDK does not call this anywhere by default except during signing (on a copy), Still, I added it here because it seems useful for users to be able to normalize paths which may be constructed dynamically, coming from external sources, etc.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/url/UrlPath.kt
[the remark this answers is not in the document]
aajtodd: No it's fine. It was more of a sanity check "do we need this" but if we make use of it somewhere then that's ok. I was imagining that most code was setting/getting it through `encodedSegments`/`decodedSegments`.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/url/QueryParameters.kt
ianbotsf: This representation does support multiple values bound to the same parameter name and there are multiple unit/integ tests which verify this (e.g., [**DefaultCanonicalizerTest**](https://github.com/awslabs/smithy-kotlin/pull/989/files#diff-510c6ad126e119e323ae8e8d5e25622fec6f0e975a4e17288049cbd27e1f407f)'s `testCanonicalQueryParams`).

The `QueryParameters` class implements `MultiMap<K, V>`, which itself extends `Map<K, List<V>>`. There are various convenience members which support things like putting/adding a single value to a key but under the hood values are stored in lists and support the full range of list-based operations at the map level.

runtime/auth/aws-signing-default/common/src/aws/smithy/kotlin/runtime/auth/awssigning/RequestMutator.kt
ianbotsf: My signing tests verified that we _do_ need to reencode in the request mutator. I've already updated the code in this class and I'll remove the TODO.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/request/HttpRequestBuilder.kt
ianbotsf: The previous `encodedPath` field was a misnomer. Beyond just the path, it also included the query parameters and fragment. Further, it also prepended a `/` if none was present. Replacing it with `path.encoded` (as I did initially based on the name) causes breaks in signing, protocol tests, and several other places.

I chose to name it `requestRelativePath` because the places this is used generally deal with forming an HTTP request line (e.g., `GET /requestRelativePath?foo=bar#baz HTTP/1.1`). It's still kinda wordy though...open to alternative naming suggestions.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/url/Url.kt
ianbotsf: Now that I've switched this to non-nullable I find I prefer it since it involves less null-coalescing operations for users. I could switch it back if you feel strongly but I have a slight preference for leaving it non-null.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/collections/ValuesMap.kt
ianbotsf: Yes, it should. I can move it and the others you mentioned to **collections** as well.

runtime/auth/aws-signing-default/common/src/aws/smithy/kotlin/runtime/auth/awssigning/Canonicalizer.kt
```diff
@@ -178,32 +187,37 @@ internal class DefaultCanonicalizer(private val sha256Supplier: HashSupplier = :
...
-
-private fun Pair<String, List<String>>.asQueryParamComponents(): List<String> =
-    second
-        .map { this@asQueryParamComponents.first to it.urlReencodeComponent() }
-        .sortedBy { it.second }
-        .map { "${it.first}=${it.second}" }
+internal fun Url.Builder.canonicalQueryParams(): String {
+    val canonicalized = parameters
+        .entries
+        .associate { (key, values) -> key.reencode().encoded to values.map { it.reencode().encoded } } // FIXME 🤮
```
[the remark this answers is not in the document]
ianbotsf: Oh whoops, I should've fixed that _before_ committing/pushing it.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpUtils.kt
```diff
@@ -137,6 +143,8 @@ internal class OkHttpProxyAuthenticator(
     }
 }
 
+private fun UserInfo.toBasicCredentials() = Credentials.basic(userName.decoded, password.decoded)
```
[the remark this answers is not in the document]
ianbotsf: True, it was a bit more useful when `UserInfo` was nullable. Will remove.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/url/QueryParameters.kt
aajtodd: Ahh ok I missed that part.

codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGeneratorTest.kt
```diff
@@ -47,14 +47,19 @@ internal class SmokeTestOperationSerializer: HttpSerialize<SmokeTestRequest> {
...
-            path = pathSegments.joinToString(separator = "/", prefix = "/")
-            parameters {
-                if (input.query1 != null) append("Query1", input.query1)
+            path.encodedSegments {
+                add(PercentEncoding.Path.encode("smoketest"))
+                "$label1".split("/").mapTo(this) { PercentEncoding.SmithyLabel.encode(it) }
+                add(PercentEncoding.Path.encode("foo"))
+            }
+            parameters.encodedParameters {
+                val labels = mutableMultiMapOf<String, String>()
```
aajtodd, line 56: **Question**: Why do we need an auxiliary data structure here?

Wonder if we can have an extension function or something that allows you to set the encoding to use.

```kotlin
        public fun encodedParameters(
            encoding: Encoding = PercentEncoding.Query,
            block: MutableMultiMap<String, String>.() -> Unit
        ) {
            asView(
                Encodable::encoded,
                encoding::encodableFromEncoded,
                Encodable::encoded,
                encoding::encodableFromEncoded,
            ).apply(block)
        }

```

```kotlin
parameters.encodedParameters(PercentEncoding.SmithyLabel) {
     if (input.query1 != null) add("Query1", input.query1)
} 
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpStringValuesMapSerializer.kt
```diff
@@ -215,7 +215,7 @@ class HttpStringValuesMapSerializer(
 
 private val HttpBinding.Location.addFnName: String
     get() = when (this) {
-        HttpBinding.Location.QUERY, HttpBinding.Location.QUERY_PARAMS -> "add" // uses MutableMultiMap
+        HttpBinding.Location.QUERY, HttpBinding.Location.QUERY_PARAMS -> "labels.add" // uses MutableMultiMap
```
aajtodd, line 218: this feels too coupled to where it's used

lauzadis (approved):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt
```diff
@@ -314,46 +314,50 @@ abstract class HttpBindingProtocolGenerator : ProtocolGenerator {
...
-                            write("#S.split(#S).mapTo(this) { #L(it) }", "\${$identifier}", '/', encodeFn)
+                            if (segment.isGreedyLabel) {
+                                write("#S.split(#S).mapTo(this) { #L(it) }", "\${$identifier}", '/', encodeFn)
+                            } else {
+                                write("add(#L(#S))", encodeFn, "\${$identifier}")
+                            }
                         } else {
-                            write("add(#L(#S))", encodeFn, "\${$identifier}")
+                            // literal
+                            val encodeFn = format("#T.Path.encode", RuntimeTypes.Core.Text.Encoding.PercentEncoding)
```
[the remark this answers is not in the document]
ianbotsf: Confusingly, path literals are encoded differently from path labels. In particular, literals have a wider variety of characters they accept without encoding while labels are more conservative. I couldn't find explicit documentation supporting this so I fall back to [the RFC 3982 specification of how paths should be encoded](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3).

Note that [at least one protocol test](https://github.com/smithy-lang/smithy/blob/d457aabb80feb4088caa3ac27d337b84e3ebc43d/smithy-aws-protocol-tests/model/restJson1/http-labels.smithy#L260) enforces that path literal segments are not over-encoded. In this test, encoding literals with `SmithyLabel` would result in the path `/ReDosLiteral/abc/%28a%2B%29%2B` which fails to match the expected value of `/ReDosLiteral/abc/(a+)+`.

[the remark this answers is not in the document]
aajtodd: You might add some of this context as a comment in the code

aajtodd (approved):

PR #990, opened 2023-10-27
fix: retry a better set of HTTP exceptions in CRT engine

ianbotsf:
## Issue \#

(none)

## Description of changes

Improve the logic for which CRT HTTP exceptions are retryable.

**Companion PR**: https://github.com/awslabs/aws-crt-kotlin/pull/82


runtime/protocol/http-client-engines/http-client-engine-crt/jvm/test/aws/smithy/kotlin/runtime/http/engine/crt/RequestUtilTest.kt
```diff
@@ -11,36 +11,35 @@ import kotlin.test.assertTrue
 class RequestUtilTest {
     @Test
     fun testExceptionRetryability() {
-        setOf(
+        mapOf(
             // All IO errors are retryable
-            "AWS_IO_SOCKET_NETWORK_DOWN",
-            "AWS_ERROR_IO_OPERATION_CANCELLED",
-            "AWS_IO_DNS_NO_ADDRESS_FOR_HOST",
+            1050 to "AWS_IO_SOCKET_NETWORK_DOWN",
```
aajtodd, line 16: **question**: How stable are these error codes? The way they are defined and used in C means you generally are never going to rely on specific numbers. I don't know if they guarantee they'll never insert one in-between existing ones.

aajtodd (approved):

lauzadis (approved):

PR #1000, opened 2023-11-20
feat: request compression trait

### LINKED ISSUE #955: Support request compression
### Community Note

## Describe the Feature

Add support for request compression as [described in the Smithy spec](https://smithy.io/2.0/spec/behavior-traits.html?highlight=requestcompression#compression).

No.

## Proposed Solution

Implement the feature as described in the Smithy spec.

(none)

## Additional Context

(none)

0marperez:
## Issue \#
closes #955

## Description of changes
Adds interceptor for request compression and compression algorithm interface for user supplied compression algorithms

See also:
https://github.com/awslabs/aws-sdk-kotlin/pull/1120


runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionTraitInterceptor.kt
```diff
@@ -0,0 +1,65 @@
...
+private fun addHeader(request: HttpRequestBuilder, algorithmId: String) {
+    val previousEncodings = request.headers["Content-Encoding"]
+    val contentEncodingHeaderPrefix = previousEncodings?.let { "$previousEncodings, " } ?: ""
+
+    request.headers.remove("Content-Encoding")
+    request.header(
+        "Content-Encoding",
+        "$contentEncodingHeaderPrefix$algorithmId",
+    )
+}
```
lauzadis, line 65: simplification: this entire function can be replaced by `request.headers.append("Content-Encoding", algorithm.id)`. It handles multiple values for a single key and you don't have to worry about overwriting content.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionTraitInterceptor.kt
```diff
@@ -0,0 +1,65 @@
...
+private val VALID_COMPRESSION_RANGE = 0..10485760
+
+/**
+ * HTTP interceptor that compresses operation request payloads when eligible
+ */
+@InternalApi
+public class RequestCompressionTraitInterceptor(
+    private val compressionThreshold: Int,
+    private val requestedCompressionAlgorithms: List<String>,
+    private val supportedCompressionAlgorithms: List<CompressionAlgorithm>,
```
lauzadis, line 25: question: I think these are swapped [based on your usage in **aws-sdk-kotlin**](https://github.com/awslabs/aws-sdk-kotlin/pull/1120/files#diff-dea08dda1e260a9c2df5faed3bfe24787a318c0b121291ee5c2129791f223df8R70-R83)? Does the client config supply the "requested"  algorithms and the modeled trait provides the "supported" algorithms?

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionTraitInterceptor.kt
```diff
@@ -0,0 +1,65 @@
...
+                requestedCompressionAlgorithms.find { supported.id == it } != null
+            }?.let { algorithm ->
+
+                val request = context.protocolRequest.toBuilder()
+                request.body = algorithm.compress(request.body)
+                addHeader(request, algorithm.id)
+                return request.build()
+            }
+        }
+        return context.protocolRequest
```
lauzadis, line 48: style: simplification of control flow
```kt
       val algorithm = supportedCompressionAlgorithms.find { supported ->
            requestedCompressionAlgorithms.find { supported.id == it } != null
        }

        return if (algorithm != null && (context.protocolRequest.body.isStreaming || payloadSize!! >= compressionThreshold)) {
            val request = context.protocolRequest.toBuilder()
            request.body = algorithm.compress(request.body)
            request.headers.append("Content-Encoding", algorithm.id)
            request.build()
        } else context.protocolRequest
```

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/requestcompression/Gzip.kt
lauzadis, line 1: location: this should probably be in `runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/compression` (new folder)

same with `CompressionAlgorithm`

runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionTraitInterceptorTest.kt
```diff
@@ -0,0 +1,112 @@
...
+        op.roundTrip(client, Unit)
+
+        val call = op.context.attributes[HttpOperationContext.HttpCallList].first()
+        assertEquals(null, call.request.headers["Content-Encoding"])
+        assertEquals("<Foo>bar</Foo>", call.request.body.toByteStream()!!.toByteArray().decodeToString())
+    }
+
+    @Test
+    fun testInvalidCompressionThreshold() = runTest {
+        assertFailsWith<IllegalArgumentException> {
```
lauzadis, line 51: correctness: this `assertFailsWith` should be more tightly scoped (around `op.roundTrip(client, Unit)`)

runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionTraitInterceptorTest.kt
```diff
@@ -0,0 +1,112 @@
...
+import kotlin.test.Ignore
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
+class RequestCompressionTraitInterceptorTest {
+    private val client = SdkHttpClient(TestEngine())
+
+    @Test
+    fun testNoCompression() = runTest {
```
lauzadis, line 29: It would be nice to test "no compression" for distinct reasons:
- Compression threshold not met
- Supported algorithms not found
- Any more?

runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionTraitInterceptorTest.kt
lauzadis, line 1: Missing test for "Streaming payloads are always compressed"

lauzadis (changes requested):
Good start!

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/requestcompression/Gzip.kt
aajtodd: This speaks to the design/modeling of this type probably being too heavily tied to `HttpBody`. I'll review more thoroughly tomorrow with suggestions.

runtime/protocol/http-client/jvm/src/aws/smithy/kotlin/runtime/http/interceptors/requestcompression/GzipJvm.kt
```diff
@@ -0,0 +1,98 @@
...
+    private val gzipBuffer = SdkBuffer()
+    private val gzipOutputStream = GZIPOutputStream(gzipBuffer.outputStream())
+
+    override fun read(sink: SdkBuffer, limit: Long): Long { // TODO: Fix this
+        // Read "limit" bytes into byte array
+        val converter = SdkBuffer()
+        val bytesRead = source.read(converter, limit)
+        if (bytesRead == -1L) source.close()
+        val byteArray = converter.readByteArray()
+        converter.close()
```
aajtodd, line 55: **fix**: Not necessary

runtime/protocol/http-client/jvm/src/aws/smithy/kotlin/runtime/http/interceptors/requestcompression/GzipJvm.kt
```diff
@@ -0,0 +1,98 @@
...
+    return compressedBody
+}
+
+internal class GzipSdkSource(
+    private val source: SdkSource,
+) : SdkSource by source {
+    private val gzipBuffer = SdkBuffer()
+    private val gzipOutputStream = GZIPOutputStream(gzipBuffer.outputStream())
+
+    override fun read(sink: SdkBuffer, limit: Long): Long { // TODO: Fix this
```
aajtodd, line 49: **suggestion**: Rather than comment everywhere I propose the following (untested but should give you a general idea of the overall structure/flow):


```kotlin
private class GzipSdkSource (
    private val source: SdkSource
): SdkSource {
    private val gzipBuffer = SdkBuffer()
    private val gzipOutputStream = GZIPOutputStream(gzipBuffer.outputStream())

    override fun read(sink: SdkBuffer, limit: Long): Long {
        require(limit >= 0L)
        if (limit == 0L) return 0L

        val temp = SdkBuffer()
        val rc = source.read(temp, limit)

        if (rc == -1L) {
            // may trigger additional bytes written by gzip defalter
            gzipOutputStream.close()
        }

        // source is exhausted and nothing left buffered we are done
        if (rc == -1L && gzipBuffer.exhausted()) return -1L

        // compress what we read and add it to the buffer
        if (rc >= 0L) {
            gzipOutputStream.write(temp.readByteArray())
        }

        // read bytes read from compressed content
        return gzipBuffer.read(sink, limit)
    }

    override fun close() {
        source.close()
    }
}
```

The `SdkByteReadChannel` will look similar.

runtime/protocol/http-client/jvm/test/aws/smithy/kotlin/runtime/http/interceptors/requestcompression/GzipJvmTest.kt
```diff
@@ -0,0 +1,66 @@
...
+import aws.smithy.kotlin.runtime.io.source
+import aws.smithy.kotlin.runtime.io.toSdkByteReadChannel
+import aws.smithy.kotlin.runtime.io.use
+import kotlinx.coroutines.test.runTest
+import java.util.zip.GZIPInputStream
+import kotlin.test.Ignore
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class GzipJvmTest {
```
aajtodd, line 20: **fix**: marking for visibility since these are not implemented/todo still

aajtodd (changes requested):
Good start! Looks like some minor refactoring, some tests to finish implementing and codegen support is left.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionInterceptor.kt
```diff
@@ -0,0 +1,44 @@
...
+        val payloadSizeBytes = context.protocolRequest.body.contentLength
+        val algorithm = availableCompressionAlgorithms.find { available ->
+            supportedCompressionAlgorithms.find { available.id == it } != null
+        }
+
+        return if (algorithm != null && (context.protocolRequest.body.isStreaming || payloadSizeBytes?.let { it >= compressionThresholdBytes } == true)) {
+            algorithm.compress(context.protocolRequest)
+        } else {
+            context.protocolRequest
+        }
```
lauzadis, line 42: question: Should we log a message letting the user know why request compression could not be completed? (Could not find a supported algorithm / request size threshold not met)

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/requestcompression/CompressionAlgorithm.kt
```diff
@@ -0,0 +1,32 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package aws.smithy.kotlin.runtime.http.interceptors.requestcompression
+
+import aws.smithy.kotlin.runtime.http.request.HttpRequest
+
+/**
+ * Represents a compression algorithm. Is used by algorithms compressing request payloads on operations with the
```
lauzadis, line 11: nit: `CompressionAlgorithm` "Is used by algorithms..." is odd, I would say these algorithms are "used by an interceptor..."

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionInterceptor.kt
```diff
@@ -0,0 +1,44 @@
...
+        require(compressionThresholdBytes in VALID_COMPRESSION_THRESHOLD_BYTES_RANGE) { "compressionThresholdBytes ($compressionThresholdBytes) must be in the range $VALID_COMPRESSION_THRESHOLD_BYTES_RANGE" }
+    }
+
+    override suspend fun modifyBeforeRetryLoop(
+        context: ProtocolRequestInterceptorContext<Any, HttpRequest>,
+    ): HttpRequest {
+        val payloadSizeBytes = context.protocolRequest.body.contentLength
+        val algorithm = availableCompressionAlgorithms.find { available ->
+            supportedCompressionAlgorithms.find { available.id == it } != null
+        }
```
lauzadis, line 36: correctness: I think this logic does not comply with the specification: 
> The client MUST only compress the request using the first supported algorithm in the list in the trait and stop considering algorithms that appear later in the list.

e.g. 

```
supportedCompressionAlgorithms = ["gzip", "deflate"]
availableCompressionAlgorithms = ["deflate", "gzip"]
```

The SDK would select "deflate" but it should be "gzip" according to the spec.

runtime/protocol/http-client/jvm/src/aws/smithy/kotlin/runtime/http/interceptors/requestcompression/GzipJvm.kt
```diff
@@ -0,0 +1,48 @@
...
+
+        return compressedRequest.build()
+    }
+}
+
+internal fun compressBytes(bytes: ByteArray): ByteArray {
+    val byteArrayOutputStream = ByteArrayOutputStream()
+    val gzipOutputStream = GZIPOutputStream(byteArrayOutputStream)
+
+    gzipOutputStream.write(bytes); gzipOutputStream.close()
```
lauzadis, line 44: style: these can be on separate lines. same for L46 below

runtime/protocol/http-client/jvm/test/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionInterceptorTest.jvm.kt
lauzadis, line 1: naming: `RequestCompressionInterceptorTestJvm.kt` (remove `.jvm` from the name)

runtime/protocol/http-client/jvm/test/aws/smithy/kotlin/runtime/http/interceptors/requestcompression/GzipCompressionTest.kt
```diff
@@ -0,0 +1,197 @@
...
+import kotlin.test.Test
+import kotlin.test.assertContentEquals
+import kotlin.test.assertEquals
+
+class GzipCompressionTest {
+    @Test
+    fun testBytesCompression() {
+        runBytesCompressionTest("")
+        runBytesCompressionTest("<Foo>bar</Foo>")
+        runBytesCompressionTest("<Baz>foo</Baz>".repeat(100))
```
lauzadis, line 22: question: What is the purpose of testing two different non-zero content lengths?

runtime/protocol/http-client/jvm/test/aws/smithy/kotlin/runtime/http/interceptors/requestcompression/GzipCompressionTest.kt
```diff
@@ -0,0 +1,197 @@
...
+        )
+        runGzipSdkSourceCompressionTest(
+            "Lorem ipsum dolor sit amet, dolorem corpora iracundia has ea, duo cu stet alterum scriptorem, " +
+                "et qui putent tractatos. Ne epicurei gloriatur pro, et ornatus consulatu necessitatibus qui. " +
+                "Veri eripuit feugiat sed no, dicat ridens id quo. Mei ne putent impedit antiopam. Ad libris " +
+                "assueverit his. Quo te brute vitae iuvaret, ut nibh bonorum sea. Mel altera vocibus ei, no vel" +
+                " tantas postea doming.",
+            1005L,
+        )
+    }
```
lauzadis, line 62: correctness: these repeated `runXTest` should probably be parameterized, see other tests in the SDK which use `@ParameterizedTest` annotation

runtime/protocol/http-client/jvm/test/aws/smithy/kotlin/runtime/http/interceptors/requestcompression/GzipCompressionTest.kt
```diff
@@ -0,0 +1,197 @@
...
+
+    val compressedBytes = compressBytes(bytes)
+    val decompressedBytes = decompressGzipBytes(compressedBytes)
+
+    assertContentEquals(bytes, decompressedBytes)
+    assertEquals(bytesHash, decompressedBytes.crc32())
+}
+
+private fun runGzipSdkSourceCompressionTest(payload: String, limit: Long, skipReadFully: Boolean = false) {
+    val readingMethods = listOf<(GzipSdkSource, SdkBuffer, Long, Long) -> ByteArray?> (
```
lauzadis, line 116: I like the way you test different methods of reading the bodies

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HttpBody.kt
```diff
@@ -218,3 +218,12 @@ public fun HttpBody.toSdkByteReadChannel(scope: CoroutineScope? = null): SdkByte
...
 }
+
+/**
+ * Determines if a http body is streaming type or not.
+ */
+public val HttpBody.isStreaming: Boolean
+    get() = when {
+        this is HttpBody.ChannelContent || this is HttpBody.SourceContent || this.contentLength == null -> true
+        else -> false
+    }
```
lauzadis, line 229: nit: can be simplified to `(this is HttpBody.ChannelContent || this is HttpBody.SourceContent || this.contentLength == null)`

A `when` block seems overkill for a Boolean

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/GzipByteReadChannel.kt
```diff
@@ -0,0 +1,64 @@
...
+public class GzipByteReadChannel(
+    private val channel: SdkByteReadChannel,
+) : SdkByteReadChannel by channel {
+    private val gzipBuffer = SdkBuffer()
+    private val gzipOutputStream = GZIPOutputStream(gzipBuffer.outputStream())
+
+    /**
+     * Keeps track of whether a read operation has been made on this byte read channel
+     */
+    private var read: Boolean = false
```
lauzadis, line 20: question: This seems suspicious, why is it necessary?

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/GzipSdkSource.kt
```diff
@@ -0,0 +1,67 @@
...
+
+        val temp = SdkBuffer()
+        val rc = source.read(temp, limit)
+
+        gzipOutputStream.write(temp.readByteArray())
+        gzipBuffer.readAll(sink)
+
+        bytesRead += rc
+
+        if (bytesRead == bytesAvailable || rc == -1L) {
```
lauzadis, line 51: question/correctness: What happens when `bytesAvailable` is not provided (null)? Does this still work? Maybe it could be worth adding to your test suite

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/GzipSdkSource.kt
```diff
@@ -0,0 +1,67 @@
...
+    private val gzipBuffer = SdkBuffer()
+    private val gzipOutputStream = GZIPOutputStream(gzipBuffer.outputStream(), true)
+    private var bytesRead: Long = 0
+
+    /**
+     * Keeps track of whether a read operation has been made on this sdk source
+     */
+    private var read: Boolean = false
+
+    override fun read(sink: SdkBuffer, limit: Long): Long {
```
lauzadis, line 24: comment: In general these `read` functions seem complicated, I'll have to pull it down and take a longer look but I feel it could be simplified

lauzadis (changes requested):
Looks good! It's getting very close to completion. The main feedback I have is correctness in selecting the algorithm and simplification of the new GzipSdkSource / GzipByteReadChannel.

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/GzipByteReadChannel.kt
0marperez: When compressing an empty payload the `GzipByteReadChannel` sets `isClosedForRead` to true from the first read. The program will go into that first `if` block checking the value of `isClosedForRead` and would return -1 and do nothing with the `gzipOutputStream`. The reason it's needed is to write an empty byte array into the `gzipOutputStream` without that it can't be decompressed. A byte array that looks like "[]" can't be decompressed by gzip, it needs the header. It also closes the gzip output steam and buffer since they won't be needed anymore

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/GzipSdkSource.kt
0marperez: Yes, I think this still works, it would use the `rc` value to check for -1 and determine the source is read completely that way. I was testing it without the `bytesRead` and `bytesAvailable` logic before and it worked.  You're right about adding a test for that though, I think it should be in the test suite. 

The only reason that additional logic (`bytesAvailable` & `bytesRead`) is there is because when using `readFully(...)` when the source is read completely there is no -1 returned. The -1 is returned only after exactly all the bytes in the source are read and then another read operation is tried. `readFully()` considers -1 as an error, so the additional logic catches when we read the source completely and does all the closing and finalizing. 

The only case where this wouldn't work is when there is a chunked payload which has unknown length and someone would try to read the whole payload using `readFully()`. Then we would never know when to finalize and close the gzip output stream and buffer because a -1 throw an exception before we could do that.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/compression/RequestCompressionIntegration.kt
```diff
@@ -0,0 +1,102 @@
...
+        ctx: ProtocolGenerator.GenerationContext,
+        resolved: List<ProtocolMiddleware>,
+    ): List<ProtocolMiddleware> = resolved + requestCompressionTraitMiddleware
+
+    override fun additionalServiceConfigProps(ctx: CodegenContext): List<ConfigProperty> =
+        super.additionalServiceConfigProps(ctx) + listOf(
+            ConfigProperty {
+                name = "compressionAlgorithms"
+                documentation = """
+                The mutable list of compression algorithms supported by the SDK.
```
aajtodd, line 39: nit: "the list of compression algorithms..."

runtime/protocol/http-client/jvm/test/aws/smithy/kotlin/runtime/http/interceptors/requestcompression/GzipCompressionTest.kt
aajtodd: That only works on JVM FYI

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/GzipByteReadChannel.kt
```diff
@@ -0,0 +1,64 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.io
+
+import aws.smithy.kotlin.runtime.InternalApi
+import java.util.zip.GZIPOutputStream
+
+@InternalApi
+public class GzipByteReadChannel(
```
aajtodd, line 11: **fix**: docs

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/GzipByteReadChannel.kt
```diff
@@ -0,0 +1,64 @@
...
+        gzipOutputStream.write(temp.readByteArray())
+        gzipBuffer.readAll(sink)
+
+        if (isClosedForRead) {
+            gzipOutputStream.close()
+            gzipBuffer.readAll(sink)
+            gzipBuffer.close()
+        }
+
+        temp.close()
```
aajtodd, line 53: unnecessary, closing a buffer doesn't do anything

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/GzipByteReadChannel.kt
aajtodd: I'm not convinced we need this. If we're "compressing" an empty payload then it should be a no-op and we shouldn't compress anything or set a `Content-Encoding`.

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/GzipByteReadChannel.kt
```diff
@@ -0,0 +1,64 @@
...
+     * Keeps track of whether a read operation has been made on this byte read channel
+     */
+    private var read: Boolean = false
+
+    override suspend fun read(sink: SdkBuffer, limit: Long): Long {
+        require(limit >= 0L)
+        if (limit == 0L) return 0L
+
+        if (isClosedForRead) {
+            if (!read) { // Empty payload
```
aajtodd, line 27: **correctness*: Why wouldn't we always return `-1` here when we are closed?

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/GzipByteReadChannel.kt
```diff
@@ -0,0 +1,64 @@
...
+
+            return -1
+        }
+
+        if (!read) read = true
+
+        val temp = SdkBuffer()
+        val rc = channel.read(temp, limit)
+
+        gzipOutputStream.write(temp.readByteArray())
```
aajtodd, line 44: correctness:

```kotlin
if (rc == -1L && gzipBuffer.exhausted()) return -1L


```

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/GzipByteReadChannel.kt
```diff
@@ -0,0 +1,64 @@
...
+        gzipBuffer.readAll(sink)
+
+        if (isClosedForRead) {
```
aajtodd, line 47: **correctness**: Predicate off `rc` being `-1` instead

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/GzipByteReadChannel.kt
```diff
@@ -0,0 +1,64 @@
...
+            return -1
+        }
+
+        if (!read) read = true
+
+        val temp = SdkBuffer()
+        val rc = channel.read(temp, limit)
+
+        gzipOutputStream.write(temp.readByteArray())
+        gzipBuffer.readAll(sink)
```
aajtodd, line 45: **correctness**: User didn't ask to read all though they asked to read up to the limit.

```kotlin
        if (rc == -1L) {
            // may trigger additional bytes written by gzip defalter
            gzipOutputStream.close()
        }

        // source is exhausted and nothing left buffered we are done
        if (rc == -1L && gzipBuffer.exhausted()) return -1L

        // compress what we read and add it to the buffer
        if (rc >= 0L) {
            gzipOutputStream.write(temp.readByteArray())
        }

        // read bytes read from compressed content
        return gzipBuffer.read(sink, limit)

```

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/GzipSdkSource.kt
aajtodd: Agreed, these need simplified. See my previous comment https://github.com/awslabs/smithy-kotlin/pull/1000#discussion_r1407933393

runtime/protocol/http-client/jvm/test/aws/smithy/kotlin/runtime/http/interceptors/requestcompression/GzipCompressionTest.kt
```diff
@@ -0,0 +1,180 @@
...
+ */
+
+package aws.smithy.kotlin.runtime.http.interceptors.requestcompression
+
+import aws.smithy.kotlin.runtime.hashing.crc32
+import aws.smithy.kotlin.runtime.http.interceptors.decompressGzipBytes
+import aws.smithy.kotlin.runtime.io.*
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.params.ParameterizedTest
```
aajtodd, line 13: **fix**: We should make tests in `common` whenever possible.

runtime/protocol/http-client/jvm/test/aws/smithy/kotlin/runtime/http/interceptors/requestcompression/GzipCompressionTest.kt
```diff
@@ -0,0 +1,180 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package aws.smithy.kotlin.runtime.http.interceptors.requestcompression
```
aajtodd, line 6: **fix**: These tests appear to be testing the `GzipSdkSource` and `GzipByteReadChannel` abstractions directly, they belong in `runtime-core` tests closer to where those types are declared.

runtime/protocol/http-client/jvm/test/aws/smithy/kotlin/runtime/http/interceptors/requestcompression/GzipCompressionTest.kt
```diff
@@ -0,0 +1,180 @@
...
+import aws.smithy.kotlin.runtime.http.interceptors.decompressGzipBytes
+import aws.smithy.kotlin.runtime.io.*
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ValueSource
+import kotlin.test.assertContentEquals
+import kotlin.test.assertEquals
+
+class GzipCompressionTest {
```
aajtodd, line 18: **suggestion**: You have a good start here and as Matas said I like that you are trying to test different read methods. I think this test could be a lot cleaner if you refactor it to something like https://github.com/awslabs/smithy-kotlin/blob/ed7cad2d4f3f5d2d244efee9315b28a5ce63c148/runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/content/ByteStreamFlowTest.kt#L14-L18

Define an abstract test that has all the different content you want to compress and different methods of reading said content and instantiate it with different inputs (`SdkSource`, `SdkByteReadChannel`, `ByteArray`, etc). You'll have to create some utility types to do this but it would look something like:


```kotlin

fun interface GzipTestReader {
    suspend fun read(sink: SdkBuffer, limit: Long): Long
}

fun interface GzipTestReaderFactory {
    fun newReader(payload: ByteArray): GzipTestReader
    
    companion object {
        val SDK_SOURCE = GzipTestReaderFactory { payload ->
            val source = GzipSdkSource(payload.source(), payload.size.toLong())
            GzipTestReader { sink, limit ->  
                source.read(sink, limit)
            }
        }
        
        val BYTE_CHANNEL = GzipTestReaderFactory { payload ->
            val ch = SdkByteReadChannel(payload)
            val source = GzipByteReadChannel(ch)
            GzipTestReader { sink, limit ->
                source.read(sink, limit)
            }
        }
    }
}

class GzipSourceCompressionTest: GzipCompressionTest(GzipTestReaderFactory.SDK_SOURCE)
class GzipChannelCompressionTest: GzipCompressionTest(GzipTestReaderFactory.BYTE_CHANNEL)

abstract class GzipCompressionTest(
    val factory: GzipTestReaderFactory
) {
    @Test
    fun testReadLargeBody() {
        // use factory to create new instances and test read against it

    }

    @Test
    fun testReadChunked() {

    }

    @Test
    fun testReadOneByteAtTime() {

    }

    // ...


}

```

The advantage of doing this is being able to debug specific tests better as well as getting the benefits of parameterization and test scenario re-use across different readers.

aajtodd (changes requested):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/http/compression/CompressionAlgorithm.kt
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package aws.smithy.kotlin.runtime.http.compression
```
aajtodd, line 6: **fix**: This probably belongs in the `http` module (package name is fine). Same thing for the `Gzip` implementation.`runtime-core` should just have the primitives for implementing the compression (so things like `GzipSdkSource` or `GzipByteReadChannel`, etc).

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/http/compression/CompressionAlgorithm.kt
```diff
@@ -0,0 +1,46 @@
...
+     * The [IANA](https://www.iana.org/assignments/http-parameters/http-parameters.xhtml)
+     * has a list of registered encodings for reference.
+     */
+    public val contentEncoding: String
+
+    /**
+     * Compresses a bytes array
+     * @return Compressed [ByteArray]
+     */
+    public fun compressBytes(bytes: ByteArray): ByteArray
```
aajtodd, line 33: **fix**: If we're moving this to `http` then we can revert this back to `compress(request: HttpRequest): HttpRequest`.

If we want to leave this concept in `runtime-core` then I think we should refactor it to be `compress(stream: ByteStream): ByteStream` which will bring about some other changes to make that work for HTTP. We should talk about it.

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/io/GzipSdkSourceTest.kt
```diff
@@ -0,0 +1,110 @@
...
+        val compressedBytes = tempBuffer.readByteArray()
+        val decompressedBytes = decompressGzipBytes(compressedBytes)
+
+        assertContentEquals(bytes, decompressedBytes)
+        assertEquals(bytesHash, decompressedBytes.crc32())
+    }
+
+    @Test
+    fun testReadLargeBodyLargeLimit() = runTest {
+        val payload = "Hello World".repeat(100)
```
aajtodd, line 93: *fix*:: this is only 1100 bytes, I'd up the large body size in all these test to something greater than 16k, ideally something not evenly divisible by 1024. Internally SdkBuffer uses segment sizes of 8k IIRC so we're not even breaking into multiple segments with these sizes.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/config/CompressionClientConfig.kt
```diff
@@ -0,0 +1,89 @@
...
+
+public interface CompressionClientConfig {
+    public val requestCompression: RequestCompressionConfig
+
+    public interface Builder {
+        public fun requestCompression(block: RequestCompressionConfig.Builder?.() -> Unit) {
+            requestCompression.apply(block)
+        }
+
+        public var requestCompression: RequestCompressionConfig.Builder?
```
aajtodd, line 21: **correctness**: Should this not be `val` and also be non-nullable?

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/config/CompressionClientConfig.kt
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * The configuration properties for request compression.
+ */
+public class RequestCompressionConfig(builder: Builder) {
+    public companion object {
+        public inline operator fun invoke(block: Builder.() -> Unit): RequestCompressionConfig =
+            RequestCompressionConfig(Builder().apply(block))
+    }
+
+    public fun toBuilderApplicator(): Builder = Builder().apply {
```
aajtodd, line 34: **fix**: Should have the `@InternalApi` annotation

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/config/CompressionClientConfig.kt
```diff
@@ -0,0 +1,89 @@
...
+import aws.smithy.kotlin.runtime.http.compression.CompressionAlgorithm
+import aws.smithy.kotlin.runtime.http.compression.Gzip
+
+@DslMarker
+public annotation class CompressionClientConfigDsl
+
+public interface CompressionClientConfig {
+    public val requestCompression: RequestCompressionConfig
+
+    public interface Builder {
```
aajtodd, line 16: **fix**: Mark with the `@CompressionClientConfigDsl` annotation

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/config/CompressionClientConfig.kt
```diff
@@ -0,0 +1,89 @@
...
+        public fun requestCompression(block: RequestCompressionConfig.Builder?.() -> Unit) {
```
aajtodd, line 17: **correctness**: We shouldn't be passing a nullable type for this, it's not ergonomic for users

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/config/CompressionClientConfig.kt
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    public fun toBuilderApplicator(): Builder = Builder().apply {
+        compressionAlgorithms = this@RequestCompressionConfig.compressionAlgorithms.toMutableList()
+        disableRequestCompression = this@RequestCompressionConfig.disableRequestCompression
+        requestMinCompressionSizeBytes = this@RequestCompressionConfig.requestMinCompressionSizeBytes
+    }
+
+    /**
+     * The list of compression algorithms supported by the SDK.
```
aajtodd, line 41: **nit**: by the client

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/config/CompressionClientConfig.kt
```diff
@@ -0,0 +1,89 @@
...
+         * More compression algorithms can be added and may override an existing implementation.
+         * Use the `CompressionAlgorithm` interface to create one.
+         */
+        public var compressionAlgorithms: MutableList<CompressionAlgorithm> = mutableListOf(Gzip())
+
+        /**
+         * Flag used to determine when a request should be compressed or not.
+         * False by default.
+         */
+        public var disableRequestCompression: Boolean? = null
```
aajtodd, line 75: **fix**: These can be non-nullable and just be defaulted I think

aajtodd (changes requested):
Getting closer

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionInterceptor.kt
```diff
@@ -0,0 +1,71 @@
...
+        context: ProtocolRequestInterceptorContext<Any, HttpRequest>,
+    ): HttpRequest {
+        val payloadSizeBytes = context.protocolRequest.body.contentLength
+        val algorithm = findMatch(supportedCompressionAlgorithms, availableCompressionAlgorithms)
+
+        return if (algorithm != null && (context.protocolRequest.body.isStreaming || payloadSizeBytes?.let { it >= compressionThresholdBytes } == true)) {
+            algorithm.compressRequest(context.protocolRequest)
+        } else {
+            val logger = coroutineContext.logger<RequestCompressionInterceptor>()
+            val skipCause = if (algorithm == null) "requested compression algorithm(s) are not supported by the client" else "request size threshold ($compressionThresholdBytes) was not met"
```
lauzadis, line 46: correctness: "requested compression algorithm(s) are not supported by the **service**"

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionInterceptor.kt
```diff
@@ -0,0 +1,71 @@
...
+    private fun findMatch(
+        supportedCompressionAlgorithms: List<String>,
+        availableCompressionAlgorithms: List<CompressionAlgorithm>,
+    ): CompressionAlgorithm? {
+        supportedCompressionAlgorithms.forEach { supportedId ->
+            availableCompressionAlgorithms.forEach { algorithm ->
+                if (algorithm.id == supportedId) return algorithm
+            }
+        }
+        return null
```
lauzadis, line 63: style: this can be simplified to 
```
supportedCompressionAlgorithms.firstNotNullOfOrNull { id ->
    availableCompressionAlgorithms.find { it.id == id }
}
```

and then you probably don't need to break it out into it's own function, it can be done in the main interceptor logic

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/io/GzipSdkSourceTest.kt
```diff
@@ -0,0 +1,113 @@
...
+
+        assertContentEquals(bytes, decompressedBytes)
+        assertEquals(bytesHash, decompressedBytes.crc32())
+    }
+
+    @Test
+    fun testReadLargeBody() = runTest {
+        val payload = "Hello World".repeat(1600)
+
+        println("\n\n\n${payload.length}\n\n\n")
```
lauzadis, line 55: correctness: remove println

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/compression/GzipJVM.kt
```diff
@@ -0,0 +1,58 @@
...
+import java.io.ByteArrayOutputStream
+import java.util.zip.GZIPOutputStream
+
+/**
+ * The gzip compression algorithm.
+ * Used to compress http requests.
+ *
+ * See: https://en.wikipedia.org/wiki/Gzip
+ */
+@InternalApi
```
lauzadis, line 22: question: why is this `InternalApi`, isn't it customer-facing?

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/GzipByteReadChannelJVM.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    override suspend fun read(sink: SdkBuffer, limit: Long): Long {
+        require(limit >= 0L)
+        if (limit == 0L) return 0L
+
+        val temp = SdkBuffer()
+        val rc = channel.read(temp, limit)
+
+        if (rc == -1L) {
+            // may trigger additional bytes written by gzip defalter
```
lauzadis, line 35: nit typo: deflater

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/compression/RequestCompressionIntegration.kt
```diff
@@ -0,0 +1,111 @@
...
+            ConfigProperty {
+                name = "requestCompression"
+                symbol = RuntimeTypes.HttpClient.Config.RequestCompressionConfig
+                builderSymbol = RuntimeTypes.HttpClient.Config.RequestCompressionConfig.nestedBuilder
+                toBuilderExpression = ".toBuilderApplicator()"
+                baseClass = RuntimeTypes.HttpClient.Config.CompressionClientConfig
+                propertyType = ConfigPropertyType.Custom(
+                    render = { prop, writer ->
+                        writer.write(
+                            "override val #1L: #2T = builder.#1L?.let { #2T(builder.#1L!!) } ?: #2T{}",
```
lauzadis, line 50: question: is the `#1L!!` necessary? I would think since you're in the `#1L?.let { ... }` block, the compiler would know it's non-null

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HttpBody.kt
```diff
@@ -190,21 +199,25 @@ public suspend fun HttpBody.readAll(): ByteArray? = when (this) {
...
-        override val contentLength: Long? = body.contentLength
-        override val isOneShot: Boolean = body.isOneShot
-        override fun readFrom(): SdkByteReadChannel = body.readFrom()
-    }
-    is HttpBody.SourceContent -> object : ByteStream.SourceStream() {
-        override val contentLength: Long? = body.contentLength
-        override val isOneShot: Boolean = body.isOneShot
-        override fun readFrom(): SdkSource = body.readFrom()
+public fun HttpBody.toByteStream(): ByteStream? {
+    if (this is ByteStreamHttpBody) return this.stream // avoid wrapping multiple times
```
lauzadis, line 203: question: what's the issue with wrapping multiple times?

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HttpBody.kt
```diff
@@ -104,21 +104,30 @@ public fun String.toHttpBody(): HttpBody = encodeToByteArray().toHttpBody()
...
+    override val contentLength: Long? = stream.contentLength
+    override val isOneShot: Boolean = stream.isOneShot
+    override fun readFrom(): SdkByteReadChannel = stream.readFrom()
+}
+
+private class ByteStreamSourceHttpBody(override val stream: ByteStream.SourceStream) : ByteStreamHttpBody, HttpBody.SourceContent() {
+    override val contentLength: Long? = stream.contentLength
+    override val isOneShot: Boolean = stream.isOneShot
+    override fun readFrom(): SdkSource = stream.readFrom()
 }
```
lauzadis, line 131: question: what is the purpose of introducing these new `ByteStreamHttpBody` implementations

lauzadis (commented):
Looks good! I have some questions but it should be ready to ship very soon!

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/compression/RequestCompressionIntegration.kt
ianbotsf: It would be better to use `it` inside the `let` block anyway.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HttpBody.kt
aajtodd: If we already have a `ByteStream` that was converted to `HttpBody` we would end up with

`ByteStream` -> `HttpBody` -> `ByteStream` -> `HttpBody` 

We can avoid all of that by introducing explicit types instead of anonymous objects and check if it's already been wrapped and if so unwrap it instead of wrapping it again.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/config/CompressionClientConfig.kt
```diff
@@ -0,0 +1,92 @@
...
+package aws.smithy.kotlin.runtime.http.config
+
+import aws.smithy.kotlin.runtime.InternalApi
+import aws.smithy.kotlin.runtime.compression.CompressionAlgorithm
+import aws.smithy.kotlin.runtime.compression.Gzip
+
+@DslMarker
+public annotation class CompressionClientConfigDsl
+
+public interface CompressionClientConfig {
```
aajtodd, line 14: **fix**: Docs

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/config/CompressionClientConfig.kt
```diff
@@ -0,0 +1,92 @@
...
+        public var disableRequestCompression: Boolean? = null
+
+        /**
+         * The threshold in bytes used to determine if a request should be compressed or not.
+         * MUST be in the range 0-10,485,760 (10 MB). Defaults to 10,240 (10 KB).
+         */
+        public var requestMinCompressionSizeBytes: Long? = null
+    }
+
+    init {
```
aajtodd, line 87: **style**: Move this to top of the class before the companion object.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/compression/CompressRequest.kt
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.compression
```
aajtodd, line 5: **fix**: This should be in `runtime.http.compression`

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionInterceptor.kt
0marperez: wouldn't it be the client? because different clients can have different supported algorithms. Why the service?

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/config/CompressionClientConfig.kt
```diff
@@ -0,0 +1,92 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.http.config
```
aajtodd, line 5: **correctness/discussion**: Now that `CompressionAlgorithm` isn't tied to HTTP should this configuration type live in `smithy-client` instead?

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/compression/GzipJVM.kt
0marperez: I think I got a comment saying this should be `InternalApi`. I don't think customers should be using this, but I don't know if that goes with the definition of the `InternalApi` annotation. "This API is internal to the smithy-kotlin runtime and should not be used". Or if actually should be used by clients

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/GzipByteReadChannelJVM.kt
```diff
@@ -0,0 +1,60 @@
...
+            gzipOutputStream.close()
+            gzipBuffer.readAll(sink)
```
aajtodd, line 37: **correctness**: This may read past `limit`, can we not just remove this and fall through to L53 where we read from `gzipBuffer`?

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/io/GzipByteReadChannelTest.kt
```diff
@@ -0,0 +1,231 @@
...
+ */
+package aws.smithy.kotlin.runtime.io
+
+import aws.smithy.kotlin.runtime.hashing.crc32
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertContentEquals
+import kotlin.test.assertEquals
+
+expect fun decompressGzipBytes(bytes: ByteArray): ByteArray
```
aajtodd, line 13: **nit**: Move to it's own file in both common and JVM test since it is re-used in more than one test.

e.g.

```
// file: GzipTestUtils.kt
expect fun decompressGzipBytes(bytes: ByteArray): ByteArray

// file: GzipTestUtilsJVM.kt
actual fun decompressGzipBytes(bytes: ByteArray): ByteArray
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/compression/RequestCompressionIntegration.kt
0marperez: I think I added this because the compiler was giving me an error, which it shouldn't. I'll try it with `it` though

runtime/protocol/http-client/jvm/test/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionInterceptorTestJvm.kt
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.http.interceptors
+
+import aws.smithy.kotlin.runtime.io.use
+import java.util.zip.GZIPInputStream
+
+actual fun decompressGzipBytes(bytes: ByteArray): ByteArray =
+    GZIPInputStream(bytes.inputStream()).bufferedReader(Charsets.UTF_8).use { it.readText() }.toByteArray()
```
aajtodd, line 11: **suggestion**: There is a kotlin extension to read all bytes from an input stream that doesn't involve going through `readText`

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.io.-input-stream/read-bytes.html

aajtodd (changes requested):
Getting super close! This looks great

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionInterceptor.kt
lauzadis: If I understand the wording correctly, the client is the one requesting compression algorithms to be used. A "requested compression algorithm" is always supported by the client.

I think it can be reworded to "no **modeled** compression algorithms are supported by the client" or "(client's) requested compression algorithms are not supported by the **service**"

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/compression/GzipJVM.kt
lauzadis: By default we are adding the Gzip compression algorithm to the clients' supported compression algorithms list. Since we're adding it to clients, I believe it's not an internal API. I'm interested what other people think

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/config/CompressionClientConfig.kt
0marperez: I get an error saying `Variable cannot be initialized before declaration`. The highest it can be is under the variable declarations, is there a way around this? If I do just place it under the variable declarations do I refactor the `toBuilderApplicator` and companion object to be somewhere else?

aajtodd: You don't even need the init block you could just assign them directly

```kotlin

    public val compressionAlgorithms: List<CompressionAlgorithm> = builder.compressionAlgorithms

    ...

    public val disableRequestCompression: Boolean = builder.disableRequestCompression ?: false

    ...
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/compression/RequestCompressionIntegration.kt
ianbotsf: Yeah, the compiler would recognize that the builder property is a `var` and thus may change between when it's known to be not null from `?.let` and when it's used again.

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/compression/GzipJVM.kt
ianbotsf: I agree with Matas. This is a value users may want to add to their own lists of supported compressions. It shouldn't be `@InternalApi`.

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/io/GzipByteReadChannelTest.kt
0marperez: Should this be `InternalApi`?

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/config/CompressionClientConfig.kt
0marperez: I think that makes sense

0marperez: Under `aws.smithy.kotlin.runtime.client.config` or just `aws.smithy.kotlin.runtime.client` I would guess `config` but just double checking

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/io/GzipByteReadChannelTest.kt
aajtodd: No it's in a test sourceset and isn't published, it's only applicable to tests.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/config/CompressionClientConfig.kt
aajtodd: `config` is probably more correct but it would seem we weren't consistent at first glance

lauzadis (approved):

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/CompressionClientConfig.kt
```diff
@@ -0,0 +1,89 @@
...
+        public fun requestCompression(block: RequestCompressionConfig.Builder.() -> Unit) {
+            requestCompression.apply(block)
+        }
+    }
+}
+
+/**
+ * The configuration properties for request compression.
+ */
+public class RequestCompressionConfig(builder: Builder) {
```
aajtodd, line 33: **fix**: constructor should be `internal`

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/CompressionClientConfig.kt
```diff
@@ -0,0 +1,89 @@
...
+        compressionAlgorithms = this@RequestCompressionConfig.compressionAlgorithms.toMutableList()
+        disableRequestCompression = this@RequestCompressionConfig.disableRequestCompression
+        requestMinCompressionSizeBytes = this@RequestCompressionConfig.requestMinCompressionSizeBytes
+    }
+
+    /**
+     * A builder for [CompressionClientConfig]
+     */
+    @CompressionClientConfigDsl
+    public class Builder {
```
aajtodd, line 69: **fix**:  constructor should be `internal`

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/CompressionClientConfig.kt
```diff
@@ -0,0 +1,89 @@
...
+    public val disableRequestCompression: Boolean = builder.disableRequestCompression ?: false
+
+    /**
+     * The threshold in bytes used to determine if a request should be compressed or not.
+     * MUST be in the range 0-10,485,760 (10 MB). Defaults to 10,240 (10 KB).
+     */
+    public val requestMinCompressionSizeBytes: Long = builder.requestMinCompressionSizeBytes ?: 10_240
+
+    @InternalApi
+    public fun toBuilderApplicator(): Builder = Builder().apply {
```
aajtodd, line 59: **suggestion**: I think this can just be `toBuilder()`, the `toBuilderApplicator` name/pattern had more to do with supporting multiple different implementations.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/compression/RequestCompressionIntegration.kt
```diff
@@ -0,0 +1,117 @@
...
+        ctx: ProtocolGenerator.GenerationContext,
+        resolved: List<ProtocolMiddleware>,
+    ): List<ProtocolMiddleware> = resolved + requestCompressionTraitMiddleware
+
+    override fun additionalServiceConfigProps(ctx: CodegenContext): List<ConfigProperty> =
+        listOf(
+            ConfigProperty {
+                name = "requestCompression"
+                symbol = RuntimeTypes.SmithyClient.Config.RequestCompressionConfig
+                builderSymbol = RuntimeTypes.SmithyClient.Config.RequestCompressionConfig.nestedBuilder.toBuilder()
```
aajtodd, line 46: **correctness**: This config property doesn't look like the other properties we've defined DSL's for. I can't say it's wrong at a glance but take a look at [RetryStrategy](https://github.com/awslabs/smithy-kotlin/blob/f5feb5638394010118a5234b450bd4020920871c/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt#L90) and [HttpClient](https://github.com/awslabs/smithy-kotlin/blob/f5feb5638394010118a5234b450bd4020920871c/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt#L43) to see if this can't be simplified.

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/CompressionClientConfig.kt
0marperez: I think this needs to be pubic, it's used when converting the builder to a config instance `override val requestCompression: RequestCompressionConfig = RequestCompressionConfig(builder.requestCompression)`.

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/CompressionClientConfig.kt
0marperez: I think this needs to be public, it's used to set the default for the `requestCompression` builder. `override var requestCompression: RequestCompressionConfig.Builder = RequestCompressionConfig.Builder()`

aajtodd (approved):

PR #1001, opened 2023-11-24
fix: `AttributeKey` code generation

lauzadis:
Usage of `AttributeKey` in endpoints  code generation fails with `Not enough information to infer type variable T`.

Example usage:
```kt
return Endpoint(
    Url.parse("${url?.scheme}://${url?.authority}/${uriEncodedBucket}${url?.path}"),
    attributes = attributesOf {
        AttributeKey("backend") to "FOOBAR" // Not enough information to infer type variable T
        SigningContextAttributeKey to listOf(
            sigV4(
                serviceName = "s3",
                disableDoubleUriEncode = true,
                signingRegion = "${params.region}",
            ),
        )
    },
)
```

The fix is to provide the type for `AttributeKey<T>`.  New codegen will look like:
```kt
AttributeKey.create("backend", "FOOBAR") to "FOOBAR"
```
This is done to avoid type-parsing at codegen and just let the Kotlin compiler work it out.

## Description of changes


aajtodd (changes requested):
I don't think we need a new function for this. Also I'm not in favor of instantiating the value twice just to get the type information. 

We should be able to just leverage the other `Attributes` builder [function](https://github.com/awslabs/smithy-kotlin/blob/a9e46e55bb40dc408abf582adec0da6b10ca957c/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/collections/Attributes.kt#L194) for this which associates any string to a value `T`

```kotlin
  attributes = attributesOf {
      "backend" to "FOOBAR"
      "document" to buildDocument { ... }
      ...
  }
```

PR #1007, opened 2023-12-06
feat: generate operation samples

aajtodd:
## Issue \#

https://github.com/awslabs/aws-sdk-kotlin/issues/239

downstream: https://github.com/awslabs/aws-sdk-kotlin/pull/1143

## Description of changes

* Implements support for generating samples from the [examples trait](https://smithy.io/2.0/spec/documentation-traits.html#smithy-api-examples-trait).
* Cleans up `ShapeValueGenerator` a bit to make the output formatting nicer (we previously only used it for generating protocol tests so minor formatting issues weren't a big deal but they should be cleaner for docs).

### Notes
1. Code is output to `src/samples` directory. Samples are not automatically included by dokka, you have to update the configuration to get them included.
2. We decided to omit doing anything with the `output` of the examples trait and to omit error examples. It isn't clear what customer benefit there is to including them and instead decided to focus samples on making a request for the given operation.
3. This renders using the playground by default which won't actually work, see https://github.com/Kotlin/dokka/issues/3041. We will likely hold off merging this until that is disabled.

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
@@ -25,7 +25,7 @@ class ShapeValueGenerator(
 ) {
 
     /**
-     * Renders the generation of a shape value declaration for the given parameters inline
+     * Renders a shape value declaration for the given parameters inline
      * with the current writer.
      *
      * @param writer writer to write generated code with.
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
@@ -206,7 +206,7 @@ class ShapeValueGenerator(
                         val variantName = memberName.replaceFirstChar { c -> c.uppercaseChar() }
                         writer.writeInline("${currSymbol.name}.$variantName(")
                         generator.instantiateShapeInline(writer, memberShape, valueNode)
-                        writer.write(")")
+                        writer.writeInline(")")
                     }
                     else -> throw CodegenException("unexpected shape type " + currShape.type)
                 }
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
@@ -34,8 +34,12 @@ class ShapeValueGenerator(
      */
     fun instantiateShapeInline(writer: KotlinWriter, shape: Shape, params: Node) {
         if (shape.isStructureShape) {
-            classDeclaration(writer, shape.asStructureShape().get()) {
+            if (params.isNullNode) {
                 writeShapeValuesInline(writer, shape, params)
+            } else {
+                classDeclaration(writer, shape.asStructureShape().get()) {
+                    writeShapeValuesInline(writer, shape, params)
+                }
             }
         } else {
             writeShapeValuesInline(writer, shape, params)
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
@@ -74,7 +74,8 @@ class ShapeValueGenerator(
     private fun classDeclaration(writer: KotlinWriter, shape: StructureShape, block: () -> Unit) {
         val symbol = symbolProvider.toSymbol(shape)
         // invoke the generated DSL builder for the class
-        writer.writeInline("#L {\n", symbol.name)
+        writer.writeInline("#L {", symbol.name)
+            .ensureNewline()
             .indent()
             .call { block() }
             .dedent()
@@ -88,7 +89,8 @@ class ShapeValueGenerator(
 
         val collectionGeneratorFunction = symbolProvider.toSymbol(shape).expectProperty(SymbolProperty.IMMUTABLE_COLLECTION_FUNCTION)
 
-        writer.writeInline("$collectionGeneratorFunction(\n")
+        writer.writeInline("$collectionGeneratorFunction(")
+            .ensureNewline()
             .indent()
             .call { block() }
             .dedent()
@@ -108,7 +110,8 @@ class ShapeValueGenerator(
         collectionSymbol.references.forEach {
             writer.addImport(it.symbol)
         }
-        writer.writeInline("$generatorFn(\n")
+        writer.writeInline("$generatorFn(")
+            .ensureNewline()
             .indent()
             .call { block() }
             .dedent()
@@ -158,7 +161,8 @@ class ShapeValueGenerator(
         override fun objectNode(node: ObjectNode) {
             if (currShape.type == ShapeType.DOCUMENT) {
                 writer
-                    .writeInline("#T {\n", RuntimeTypes.Core.Content.buildDocument)
+                    .writeInline("#T {", RuntimeTypes.Core.Content.buildDocument)
+                    .ensureNewline()
                     .indent()
             }
 
@@ -187,14 +191,15 @@ class ShapeValueGenerator(
                         } else {
                             generator.instantiateShapeInline(writer, memberShape, valueNode)
                             if (i < node.members.size - 1) {
-                                writer.writeInline(",\n")
+                                writer.writeInline(",")
+                                    .ensureNewline()
                             }
                         }
                     }
                     is DocumentShape -> {
                         writer.writeInline("#S to ", keyNode.value)
                         generator.instantiateShapeInline(writer, currShape, valueNode)
-                        writer.writeInline("\n")
+                        writer.ensureNewline()
                     }
                     is UnionShape -> {
                         val member = currShape.getMember(keyNode.value).orElseThrow {
@@ -256,7 +261,8 @@ class ShapeValueGenerator(
                             node.elements.forEach {
                                 generator.instantiateShapeInline(writer, currShape, it)
                                 writer.unwrite(writer.newline)
-                                writer.writeInline(",\n")
+                                writer.writeInline(",")
+                                    .ensureNewline()
                             }
                         }
                     }
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
@@ -25,24 +25,40 @@ class ShapeValueGenerator(
 ) {
 
     /**
-     * Writes generation of a shape value type declaration for the given the parameters.
+     * Renders the generation of a shape value declaration for the given parameters inline
+     * with the current writer.
      *
      * @param writer writer to write generated code with.
      * @param shape the shape that will be declared.
      * @param params parameters to fill the generated shape declaration.
      */
-    fun writeShapeValueInline(writer: KotlinWriter, shape: Shape, params: Node) {
+    fun instantiateShapeInline(writer: KotlinWriter, shape: Shape, params: Node) {
+        if (shape.isStructureShape) {
+            classDeclaration(writer, shape.asStructureShape().get()) {
+                writeShapeValuesInline(writer, shape, params)
+            }
+        } else {
+            writeShapeValuesInline(writer, shape, params)
+        }
+    }
+
+    /**
+     * Renders the mapping of the shape fields to the given parameters
+     *
+     * @param writer writer to write generated code with.
+     * @param shape the shape that will be declared.
+     * @param params parameters to fill the generated shape declaration.
+     */
+    fun writeShapeValues(writer: KotlinWriter, shape: Shape, params: Node) {
+        writer.ensureNewline()
+        writeShapeValuesInline(writer, shape, params)
+        writer.ensureNewline()
+    }
+
+    private fun writeShapeValuesInline(writer: KotlinWriter, shape: Shape, params: Node) {
         val nodeVisitor = ShapeValueNodeVisitor(writer, this, shape)
         when (shape.type) {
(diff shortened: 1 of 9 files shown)


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/samples/KDocSamplesGenerator.kt
```diff
@@ -0,0 +1,143 @@
...
+        val operations = topDownIndex.getContainedOperations(ctx.settings.service)
+        operations.filter { it.hasTrait<ExamplesTrait>() }
+            .forEach { op ->
+                val examples = op.expectTrait<ExamplesTrait>()
+
+                val writer = KotlinWriter(samplePackage(ctx.settings))
+                writer.withBlock("class #L {", "}", sampleClassName(op)) {
+                    examples.examples
+                        // unclear what benefit error examples provide, omit for now
+                        .filterNot { it.error.isPresent }
```
ianbotsf, line 81: **Style**: Fluent call chains should have every `.` invocation on a subsequent line, including the first one:

```kotlin
examples
    .examples
    .filterNot { is.error.isPresent }
    …
```

Same comment below about `example.documentation`.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/samples/KDocSamplesGenerator.kt
```diff
@@ -0,0 +1,143 @@
...
+                                example.documentation
+                                    .getOrDefault(example.title)
+                                    .breakLongLines()
+                                    .forEach { line ->
+                                        write("// #L", line)
+                                    }
+
+                                renderNormalExample(ctx, writer, op, example)
+                            }
+                            write("")
```
ianbotsf, line 95: **Question**: Why blank lines before _and_ after the function?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/samples/KDocSamplesGenerator.kt
```diff
@@ -0,0 +1,143 @@
...
+            writer.withBlock("#L#LClient.#L {", "}", respPrefix, clientName, op.defaultName()) {
+                ShapeValueGenerator(ctx.model, ctx.symbolProvider).writeShapeValues(writer, input, example.input)
+            }
+        }
+
+        // TODO - not clear what benefit there is to the example output, omit it for now
+    }
+}
+
+internal fun String.breakLongLines(maxLineLengthChars: Int = 100): List<String> {
```
ianbotsf, line 120: **Suggestion**: Since these will most often be seen in very width-constrained surfaces (e.g., Dokka's width-constrained HTML, IDE tooltips, etc.) I might suggest an even smaller max line length than 100. We might have to experiment to see what's good in the average case but my guess is that 80 is better than 100.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/samples/KDocSamplesGenerator.kt
```diff
@@ -0,0 +1,143 @@
...
+    val words = Regex("\\w+[.,]?|\\\".*\\\"[.,]?|\\(.*\\)[.,]?").findAll(this).map(MatchResult::value)
```
ianbotsf, line 121: **Nit**: Move the `Regex` definition to module-level.

ianbotsf, line 121: **Style**: The regex string escaping can be simplified by using triple quotes. Additionally, I don't believe quotes have to be escaped in regex:

```kotlin
val wordsPattern = Regex("""\w+[.,]?|".*"[.,]?|\(.*\)[.,]?""")
```

ianbotsf, line 121: **Correctness**: The quoted string and parenthesized string cases use greedy matching, which will cause problems in strings with multiple quoted or parenthesized sections. For example, the pattern applied to `The "quick" brown fox jumped over the "lazy" dogs.` would match `"quick" brown fox jumped over the "lazy"`. I think you want to use `.*?` instead of `.*`.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/samples/KDocSamplesGenerator.kt
```diff
@@ -0,0 +1,143 @@
...
+    val lines = mutableListOf<String>()
+    val wordsOnLine = mutableListOf<String>()
+    var lineLength = 0
+
+    words.forEach { word ->
+        if (word.length + lineLength < maxLineLengthChars) {
+            if (wordsOnLine.isNotEmpty()) lineLength++
+            lineLength += word.length
+            wordsOnLine.add(word)
+        } else {
```
ianbotsf, line 131: **Correctness**: This miscalculates line length as merely the sum of word lengths. Spaces should be included too.

ianbotsf (changes requested):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
```diff
@@ -58,11 +74,12 @@ class ShapeValueGenerator(
     private fun classDeclaration(writer: KotlinWriter, shape: StructureShape, block: () -> Unit) {
         val symbol = symbolProvider.toSymbol(shape)
         // invoke the generated DSL builder for the class
-        writer.writeInline("#L {\n", symbol.name)
+        writer.writeInline("#L {", symbol.name)
+            .ensureNewline()
```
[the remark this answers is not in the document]
aajtodd: IIRC it has to do with indentation on `write` vs `writeInline`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
```diff
@@ -25,24 +25,40 @@ class ShapeValueGenerator(
...
         when (shape.type) {
-            ShapeType.STRUCTURE -> {
-                if (params.isNullNode) {
-                    params.accept(nodeVisitor)
-                } else {
-                    classDeclaration(writer, shape.asStructureShape().get()) {
-                        params.accept(nodeVisitor)
-                    }
-                }
-            }
```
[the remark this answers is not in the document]
aajtodd: It has to do with how it's used, both functions delegate to this new private function `writeShapeValuesInline`. You can see it invokes this function on L38

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
```diff
@@ -58,11 +74,12 @@ class ShapeValueGenerator(
...
         // invoke the generated DSL builder for the class
-        writer.writeInline("#L {\n", symbol.name)
+        writer.writeInline("#L {", symbol.name)
+            .ensureNewline()
             .indent()
             .call { block() }
             .dedent()
-            .write("")
+            .ensureNewline()
             .write("}")
```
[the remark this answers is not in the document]
aajtodd: It doesn't do that now (or before). This just fixes adding additional new lines when one already existed.

e.g.

Before:
```
Builder {
   member1 = "foo"

}
```

After:
```
Builder {
   member1 = "foo"
}
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/samples/KDocSamplesGenerator.kt
```diff
@@ -0,0 +1,143 @@
...
+ * [KotlinIntegration] that renders [KDoc samples](https://kotlinlang.org/docs/kotlin-doc.html#sample-identifier)
+ * and pre-processes the documentation to insert references to the generated sample identifiers for operations
+ * that have [examples](https://smithy.io/2.0/spec/documentation-traits.html#smithy-api-examples-trait)
+ */
+class KDocSamplesGenerator : KotlinIntegration {
+    override fun enabledForService(model: Model, settings: KotlinSettings): Boolean {
+        val topDownIndex = TopDownIndex.of(model)
+        val operations = topDownIndex.getContainedOperations(settings.service)
+        return operations.any { it.hasTrait<ExamplesTrait>() }
+    }
```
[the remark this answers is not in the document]
aajtodd: possibly, I'd be worried it doesn't check within the closure of the service shape though (e.g. a model can have multiple services defined for instance).

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/samples/KDocSamplesGenerator.kt
```diff
@@ -0,0 +1,143 @@
...
+
+    private fun sampleIdentifier(settings: KotlinSettings, op: OperationShape, index: Int): String =
+        listOf(
+            samplePackage(settings),
+            sampleClassName(op),
+            sampleFunctionName(index),
+        ).joinToString(separator = ".")
+
+    private fun sampleFunctionName(index: Int): String = "sample${index + 1}"
+    private fun sampleClassName(op: OperationShape): String = op.id.name
```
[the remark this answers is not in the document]
aajtodd: > Are these sample names visible in the rendered documentation 

Yes. 

> so 1 could be omitted for the first sample.

Sure.

>  Additionally, I think the function could be module-level and its name could include the operation, for example fooservice.samples.someOperationExample, fooservice.samples.someOperationExample2, etc.

It does include the operation name, it's `fooservice.samples.Operation.sample`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/samples/KDocSamplesGenerator.kt
aajtodd:
```
class Foo {
    @Sample
     fun foo() { ... }
     @Sample
     fun bar() { ... }
}
```

vs

```
class Foo {
    
     @Sample
     fun foo() { ... }
     
     @Sample
     fun bar() { ... }

}
```

aajtodd: Could just be before, seems really minor since nobody is going to actually look at these (only the function body is rendered into API ref documentation)

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/samples/KDocSamplesGenerator.kt
aajtodd: Spaces are accounted for on L128 `if (wordsOnLine.isNotEmpty()) lineLength++`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/samples/KDocSamplesGenerator.kt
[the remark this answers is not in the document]
ianbotsf: Yes I see that but my point was that `sample` as a method name is not very descriptive _and_ it's duplicated with a namespace component. Further, we don't use operation names as namespaces anywhere else and it seems weird to do so here. If these names/namespaces are going to be visible then I think they'd be better named something more like `fooservice.samples.someOperationExample`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/samples/KDocSamplesGenerator.kt
aajtodd: These don't show up in IDE tooltips and I'm not sure what we want to optimize for here. 80 is generally considered short these days. Many IDEs are set to 120 so 100 seemed reasonable. We can put it whatever we want though if you feel strongly.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ShapeValueGenerator.kt
```diff
@@ -25,24 +25,40 @@ class ShapeValueGenerator(
...
-     * Writes generation of a shape value type declaration for the given the parameters.
+     * Renders the generation of a shape value declaration for the given parameters inline
+     * with the current writer.
      *
      * @param writer writer to write generated code with.
      * @param shape the shape that will be declared.
      * @param params parameters to fill the generated shape declaration.
      */
-    fun writeShapeValueInline(writer: KotlinWriter, shape: Shape, params: Node) {
+    fun instantiateShapeInline(writer: KotlinWriter, shape: Shape, params: Node) {
```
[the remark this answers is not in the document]
aajtodd: True but it's more about intended use at the call site I think. If you are assigning the value to a variable in some writer then use instantiate, if you just want the values then use `writeShapeValues`.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/samples/KDocSamplesGenerator.kt
```diff
@@ -63,10 +76,11 @@ class KDocSamplesGenerator : KotlinIntegration {
             sampleFunctionName(index),
         ).joinToString(separator = ".")
 
-    private fun sampleFunctionName(index: Int): String = "sample${index + 1}"
+    private fun sampleFunctionName(index: Int): String = "sample" + if (index > 0) "$index" else ""
```
ianbotsf, line 79: **Nit**: The second sample is now called `sample1` but I think `sample2` was better (like before).

ianbotsf (approved):

lauzadis (approved):

PR #1012, opened 2023-12-12
refactor: migrate non AWS specific protocol generation

aajtodd:
## Issue \#

n/a

## Description of changes

This PR has a lot of codegen refactoring to support eventual migration of AWS protocol support from `aws-sdk-kotlin` to a new module in `smithy-kotlin`. The changes mostly involve declaring new sections that can be overridden and/or migrating logic that probably should have never been in `aws-sdk-kotlin` in the first place (e.g. sigv4 related stuff should have lived with the sigv4 integration). 

Summary below:

**refactor**: remove unused `KotlinIntegration::onShapeWriterUse` extension function
**refactor**: Replace `KotlinIntegration::renderBindEndpointBuiltins` with a new function `customizeEndpointResolution` which allows for more than just overriding rendering support for builtins. It allows registering custom property renderers and custom function support.
**refactor**: Remove exception base class from `ProtocolGenerator` in favor of declaring a section that can be overwritten. The logic for rendering the exception has been abstracted to `ServiceExceptionBaseClassGenerator` so that it can mostly be re-used when doing so. 
**refactor**: Migrate `mergeSequential` from `aws-sdk-kotlin` runtime to `smithy-kotlin` runtime. There wasn't anything AWS specific about this and it is only used for event streams (currently)
**refactor**: Sigv4 integration
    * Move various bits of codegen into section writers (support for protocol test service client setup)
    * Migrate logic from the AWS endpoint delegator into the new `EndpointCustomization` type
**refactor**: Move support for the `SDK::Endpoint` builtin into a `KotlinIntegration`
**refactor**: Move generation of `mergeServiceDefaults` into `HttpProtocolClientGenerator`. The function was never AWS specific, it just sets default operation execution context attributes.
**refactor**: Always generate an override of the operation `retryPolicy` in `HttpProtocolClientGenerator` since it is always registered as a service client configuration property.
**refactor**: Move the Smithy build transforms from `aws-sdk-kotlin` since they aren't specific to the SDK. Code/tests did not change, just the name of the transform.


lauzadis (approved):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/SdkEndpointBuiltinIntegration.kt
```diff
@@ -0,0 +1,68 @@
...
+            """.trimIndent()
+            propertyType = ConfigPropertyType.SymbolDefault
+        }
+    }
+
+    override fun enabledForService(model: Model, settings: KotlinSettings): Boolean {
+        val service = model.expectShape<ServiceShape>(settings.service)
+        val isAwsSdk = service.hasTrait<ServiceTrait>()
+        return isAwsSdk || service.getEndpointRules()?.parameters?.find { it.isBuiltIn && it.builtIn.get() == BUILTIN_NAME } != null
+    }
```
ianbotsf, line 49: **Nit**: `isAwsSdk` is probably no longer the best name for this value.

ianbotsf (approved):

PR #1017, opened 2023-12-21
feat: emit metrics from CRT HTTP engine

ianbotsf:
## Issue \#

https://github.com/awslabs/smithy-kotlin/issues/893

## Description of changes

This change emits metrics from the CRT HTTP engine similar to the ones emitted by the OkHttp engine.

**Companion PRs**: https://github.com/awslabs/aws-crt-java/pull/738, https://github.com/awslabs/aws-crt-kotlin/pull/86

### DIFF
--- a/runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt
@@ -13,9 +13,12 @@ import aws.smithy.kotlin.runtime.crt.SdkDefaultIO
 import aws.smithy.kotlin.runtime.http.HttpErrorCode
 import aws.smithy.kotlin.runtime.http.HttpException
 import aws.smithy.kotlin.runtime.http.engine.ProxyConfig
+import aws.smithy.kotlin.runtime.http.engine.internal.HttpClientMetrics
 import aws.smithy.kotlin.runtime.http.request.HttpRequest
 import aws.smithy.kotlin.runtime.io.Closeable
 import aws.smithy.kotlin.runtime.net.TlsVersion
+import aws.smithy.kotlin.runtime.telemetry.metrics.measureSeconds
+import kotlinx.atomicfu.atomic
 import kotlinx.coroutines.TimeoutCancellationException
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.Semaphore
@@ -27,8 +30,10 @@ import aws.smithy.kotlin.runtime.net.TlsVersion as SdkTlsVersion
 
 internal class ConnectionManager(
     private val config: CrtHttpEngineConfig,
+    private val metrics: HttpClientMetrics,
 ) : Closeable {
     private val leases = Semaphore(config.maxConnections.toInt())
+    private val pending = atomic(0L)
 
     private val crtTlsContext: TlsContext = TlsContextOptionsBuilder()
         .apply {
@@ -61,16 +66,22 @@ internal class ConnectionManager(
         val manager = getManagerForUri(request.uri, proxyConfig)
         var leaseAcquired = false
 
-        return try {
-            // wait for an actual connection
-            val conn = withTimeout(config.connectionAcquireTimeout) {
-                // get a permit to acquire a connection (limits overall connections since managers are per/host)
-                leases.acquire()
-                leaseAcquired = true
-                manager.acquireConnection()
-            }
+        metrics.queuedRequests = pending.incrementAndGet()
 
-            LeasedConnection(conn)
+        return try {
+            metrics.requestsQueuedDuration.measureSeconds {
+                // wait for an actual connection
+                val conn = withTimeout(config.connectionAcquireTimeout) {
+                    // get a permit to acquire a connection (limits overall connections since managers are per/host)
+                    leases.acquire()
+                    leaseAcquired = true
+                    metrics.connectionAcquireDuration.measureSeconds {
+                        manager.acquireConnection()
+                    }
+                }
+
+                LeasedConnection(conn)
+            }
         } catch (ex: Exception) {
             if (leaseAcquired) {
                 leases.release()
@@ -82,8 +93,18 @@ internal class ConnectionManager(
             }
 
             throw httpEx
+        } finally {
+            metrics.queuedRequests = pending.decrementAndGet()
+            emitConnections()
         }
     }
+
+    private fun emitConnections() {
+        val idleConnections = leases.availablePermits.toLong()
+        metrics.idleConnections = idleConnections
+        metrics.acquiredConnections = config.maxConnections.toLong() - idleConnections
+    }
+
     private suspend fun getManagerForUri(uri: Uri, proxyConfig: ProxyConfig): HttpClientConnectionManager = mutex.withLock {
         connManagers.getOrPut(uri.authority) {
             val connOpts = options.apply {
@@ -105,6 +126,7 @@ internal class ConnectionManager(
             HttpClientConnectionManager(connOpts)
         }
     }
+
     override fun close() {
         connManagers.forEach { entry -> entry.value.close() }
         crtTlsContext.close()
@@ -117,6 +139,7 @@ internal class ConnectionManager(
                 delegate.close()
             } finally {
                 leases.release()
+                emitConnections()
             }
         }
     }
--- a/runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/SdkStreamResponseHandler.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/SdkStreamResponseHandler.kt
@@ -10,16 +10,27 @@ import aws.sdk.kotlin.crt.io.Buffer
 import aws.smithy.kotlin.runtime.http.*
 import aws.smithy.kotlin.runtime.http.HeadersBuilder
 import aws.smithy.kotlin.runtime.http.HttpException
+import aws.smithy.kotlin.runtime.http.engine.EngineAttributes
+import aws.smithy.kotlin.runtime.http.engine.crt.io.reportingTo
+import aws.smithy.kotlin.runtime.http.engine.internal.HttpClientMetrics
 import aws.smithy.kotlin.runtime.http.response.HttpResponse
+import aws.smithy.kotlin.runtime.http.response.copy
 import aws.smithy.kotlin.runtime.io.SdkBuffer
 import aws.smithy.kotlin.runtime.io.SdkByteChannel
 import aws.smithy.kotlin.runtime.io.SdkByteReadChannel
+import aws.smithy.kotlin.runtime.io.source
+import aws.smithy.kotlin.runtime.operation.ExecutionContext
 import aws.smithy.kotlin.runtime.telemetry.logging.logger
+import aws.smithy.kotlin.runtime.telemetry.metrics.recordSeconds
+import aws.smithy.kotlin.runtime.time.Instant
+import aws.smithy.kotlin.runtime.time.fromEpochNanoseconds
 import aws.smithy.kotlin.runtime.util.derivedName
 import kotlinx.atomicfu.locks.reentrantLock
 import kotlinx.atomicfu.locks.withLock
-import kotlinx.coroutines.*
+import kotlinx.coroutines.DelicateCoroutinesApi
+import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.launch
 import kotlin.coroutines.CoroutineContext
 
 /**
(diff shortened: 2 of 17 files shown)


lauzadis (approved):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/time/Instant.kt
```diff
@@ -44,6 +44,13 @@ public expect class Instant : Comparable<Instant> {
      */
     public operator fun minus(duration: Duration): Instant
 
+    /**
+     * Returns a duration representing the amount of time between this and [other]. If [other] is before this instant,
+     * the resulting duration will be negative.
+     * @param other The [Instant] marking the end of the duration
+     */
+    public operator fun minus(other: Instant): Duration
```
aajtodd, line 52: tests?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/time/Instant.kt
```diff
@@ -102,4 +109,7 @@ public fun Instant.Companion.fromEpochMilliseconds(milliseconds: Long): Instant
     return fromEpochSeconds(secs, ns.toInt())
 }
 
+public fun Instant.Companion.fromEpochNanoseconds(ns: Long): Instant =
```
aajtodd, line 112: tests?

runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt
```diff
@@ -61,16 +66,22 @@ internal class ConnectionManager(
         val manager = getManagerForUri(request.uri, proxyConfig)
         var leaseAcquired = false
 
+        metrics.queuedRequests = pending.incrementAndGet()
```
aajtodd, line 69: **correctness**: If we are trying to acquire a connection then it's not queued right? I think queued would be before we hit the [requestLimiter](https://github.com/awslabs/smithy-kotlin/blob/2174a835e5bf088418db88368d4f888204ed9008/runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt#L63).

```kotlin
metrics.queuedRequests = pending.incrementAndGet()
requestLimiter.withPermit {
    metrics.queuedRequests = pending.decrementAndGet()
}

```

runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt
```diff
@@ -82,8 +93,18 @@ internal class ConnectionManager(
...
 
             throw httpEx
+        } finally {
+            metrics.queuedRequests = pending.decrementAndGet()
+            emitConnections()
         }
     }
+
+    private fun emitConnections() {
+        val idleConnections = leases.availablePermits.toLong()
```
aajtodd, line 103: **correctness**: this semaphore isn't tied in anyway to actual connections in CRT, you could have 100 max connections configured but that doesn't mean you have 100 idle connections already connected waiting to be used. I'm thinking this would have to come from the actual CRT connection manager.

aajtodd (changes requested):

runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt
```diff
@@ -82,8 +93,18 @@ internal class ConnectionManager(
             }
 
             throw httpEx
+        } finally {
+            metrics.queuedRequests = pending.decrementAndGet()
+            emitConnections()
         }
     }
+
+    private fun emitConnections() {
```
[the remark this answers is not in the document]
ianbotsf: Since the class already has the word "connection" in it, renaming to simply `emitMetrics`.

runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt
ianbotsf: Oh jeez, good point. Yes, this is clearly flawed.

That brings up an interesting wrinkle, though. We don't have just one CRT connection manager, we have [one for each host](https://github.com/awslabs/smithy-kotlin/blob/49a0089293068cdb9aec6d07fe019080c3b14117/runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt#L55). Does the Smithy metric for acquired connections equal the sum of acquired connections _across all_ underlying CRT connection managers?

Moreover, we have [a CRT config param called `maxConnections`](https://github.com/awslabs/smithy-kotlin/blob/49a0089293068cdb9aec6d07fe019080c3b14117/runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/CrtHttpEngineConfig.kt#L31-L34). Contrary to its name, it's [used as the maximum number of connections _for each_ underlying CRT connection manager](https://github.com/awslabs/smithy-kotlin/blob/49a0089293068cdb9aec6d07fe019080c3b14117/runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt#L50). Meaning if `maxConnections` is 100 and we connect to 3 hosts, there are actually 3 underlying CRT connection managers which _each_ have a max of 100 connections for a total of 300. How then do we measure the Smithy metric of idle connections? Is it the sum of idle connections _across all_ underlying CRT connection managers (which may be more than `maxConnections`)? Or is it `maxConnections` minus the sum of acquired connections _across all_ underlying CRT connection managers (which may be less than the actual number of open/idle connections)?

runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/SdkStreamResponseHandler.kt
```diff
@@ -210,4 +252,19 @@ internal class SdkStreamResponseHandler(
...
         }
     }
+
+    override fun onMetrics(stream: HttpStream, metrics: HttpStreamMetrics) {
+        val sendEnd = positiveInstantOrNull(metrics.sendEndTimestampNs)
+        val receiveStart = positiveInstantOrNull(metrics.receiveStartTimestampNs)
+
+        if (sendEnd != null && receiveStart != null) {
+            val ttfb = receiveStart - sendEnd
+            if (ttfb.isPositive()) {
```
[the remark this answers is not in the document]
ianbotsf: Not directly but event streams may begin receiving traffic before sending has concluded since the communication is bidirectional and duplexed. It seemed prudent in that situation to not report TTFB.

runtime/protocol/http-client-engines/http-client-engine-crt/jvm/test/aws/smithy/kotlin/runtime/http/engine/crt/SdkStreamResponseHandlerTest.kt
```diff
@@ -3,17 +3,23 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+@file:OptIn(ExperimentalApi::class)
```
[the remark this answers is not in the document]
ianbotsf: It can actually be removed. I forgot why I added this but it's unused now!

runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt
ianbotsf: OK, if "queued" means "before we attempt to acquire a connection" then I'm guessing that the `requestsQueuedDuration` measurement below is _also_ wrong. I'll move it too.

runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt
aajtodd: The fact CRT uses connection manager per host is an implementation detail, the config setting is at the engine level. This is what the `leases` semaphore is for is to control max connections across multiple managers (albeit it likely isn't perfect at ensuring we never cross the `maxConnections` threshold due to idle connections).

The reason I allowed each connection manager to be configured with `maxConnections` is because we don't know how many hosts an engine will be used to connect to, it may be 1 or it may be split across many hosts which will require many connection managers.

runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt
aajtodd: Had to look at the definitions I gave them again and I think this would be inline with what it says

> queued=waiting to be executed (e.g. waiting for thread to be available), in-flight=actively processing

If we are to the point of acquiring a connection we are "actively processing" the request. I can see where the definition could be interpreted differently though as in "I have all the resources needed at this point to execute the request" but I think establishing a connection (or waiting on one to be available) is part of overall request processing. Curious if others disagree.

runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt
ianbotsf: I see this as an area that may cause confusion later, especially if users begin correlating the idle connections metric with connections reported by the OS or JVM. I don't have a better suggestion though given how the CRT connection manager works. I'll switch the implementation to calculate the idle connections metric as `maxConnections` minus the sum of acquired connections across all underlying CRT connection managers.

ianbotsf: Oh and interrogating the number of active connections from the CRT connection managers requires another revision of **aws-crt-kotlin**: https://github.com/awslabs/aws-crt-kotlin/pull/88

lauzadis (approved):

runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt
```diff
@@ -82,8 +93,18 @@ internal class ConnectionManager(
...
+        } finally {
+            metrics.queuedRequests = pending.decrementAndGet()
+            emitMetrics()
         }
     }
+
+    private fun emitMetrics() {
+        val acquiredConnections = connManagers.values.sumOf { it.managerMetrics.leasedConcurrency }
+        metrics.acquiredConnections = acquiredConnections
+        metrics.idleConnections = config.maxConnections.toLong() - acquiredConnections
```
aajtodd, line 105: **fix**: This still isn't right, idle connections are connections established with the server not currently in use. If I'm not mistaken this is just [availableConcurrency](https://github.com/awslabs/aws-crt-java/blob/main/src/main/java/software/amazon/awssdk/crt/http/HttpManagerMetrics.java#L22C2-L22C2) which would require the same summing across managers.

runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt
aajtodd: This still doesn't look correct, I don't think `queuedRequests` is calculated in connection manager.

aajtodd (changes requested):

runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt
ianbotsf: OK, sum of `availableConcurrency` sounds doable, although it will exceed max connections under some circumstances.

runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/ConnectionManager.kt
ianbotsf: Ah I think I was confusing the semaphore inside `ConnectionManager` with the one inside `CrtHttpEngine`. Switching.

PR #1018, opened 2023-12-26
feat: expand use of `chunked` content encoding

lauzadis:
If a streaming body with an undefined content length is provided, upload it with `chunked` to ensure the hash can be properly calculated. This reduces occurrences of exceptions claiming `Stream must be replayable to calculate a body hash` 
## Issue \#

Addresses an issue in https://github.com/awslabs/aws-sdk-kotlin/issues/1157

## Description of changes

### DIFF
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/internal/AwsChunkedUtil.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/internal/AwsChunkedUtil.kt
@@ -43,8 +43,8 @@ internal fun SdkBuffer.writeTrailerSignature(signature: String) {
  */
 @InternalApi
 public val HttpBody.isEligibleForAwsChunkedStreaming: Boolean
-    get() = (this is HttpBody.SourceContent || this is HttpBody.ChannelContent) && contentLength != null &&
-        (isOneShot || contentLength!! > AWS_CHUNKED_THRESHOLD)
+    get() = (this is HttpBody.SourceContent || this is HttpBody.ChannelContent) &&
+        (isOneShot || (contentLength?.compareTo(AWS_CHUNKED_THRESHOLD) ?: 0) > 0)
 
 /**
  * @return a boolean representing if the signing configuration is configured (via [HashSpecification]) for aws-chunked content encoding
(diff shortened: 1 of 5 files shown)


runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/AwsHttpSigner.kt
```diff
@@ -163,6 +165,12 @@ public class AwsHttpSigner(private val config: Config) : HttpSigner {
                         HashSpecification.StreamingAws4HmacSha256Payload
                     }
                 }
+                request.headers["x-amz-checksum-sha256"] != null -> {
```
aajtodd, line 168: **suggestion**: Should probably be a constant value somewhere.

aajtodd: **fix**: Needs tests

aajtodd (changes requested):
This change seems _ok_ but I don't think it addresses the underlying issue. If a user submits a request without a flexible checksum this would still fail right? I think we need to solve that because it looks like the request in the issue without a checksum should succeed as well unless I'm missing something.

aajtodd (approved):

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/internal/AwsChunkedUtil.kt
```diff
@@ -43,8 +43,8 @@ internal fun SdkBuffer.writeTrailerSignature(signature: String) {
  */
 @InternalApi
 public val HttpBody.isEligibleForAwsChunkedStreaming: Boolean
-    get() = (this is HttpBody.SourceContent || this is HttpBody.ChannelContent) && contentLength != null &&
-        (isOneShot || contentLength!! > AWS_CHUNKED_THRESHOLD)
+    get() = (this is HttpBody.SourceContent || this is HttpBody.ChannelContent) &&
+        (isOneShot || (contentLength?.compareTo(AWS_CHUNKED_THRESHOLD) ?: 0) > 0)
```
ianbotsf, line 47: **Style**: `?.compareTo` works but is pretty verbose and still requires null-coalescing the value. Shorter would just be:

```kotlin
(contentLength ?: 0 > AWS_CHUNKED_THRESHOLD)
```

Same comment applies below.

ianbotsf (approved):

PR #1019, opened 2023-12-26
feat: support non-ASCII header values

lauzadis:
Support non-ASCII header values in the default OkHttp engine.

## Issue \#

https://github.com/awslabs/aws-sdk-kotlin/issues/1163

## Description of changes

OkHttp [refuses to support non-ASCII header values by default](https://github.com/square/okhttp/issues/2016). They have an [`addUnsafeNonAscii`](https://square.github.io/okhttp/5.x/okhttp/okhttp3/-headers/-builder/add-unsafe-non-ascii.html) method which will skip validation, allowing us to form OkHttp requests which have headers with non-ASCII values. 

This is done to align with [the HTTP spec](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.4) and the behavior of the CRT HTTP engine, which supports non-ASCII values by default.

### DIFF
--- a/runtime/protocol/http-client-engines/test-suite/common/test/aws/smithy/kotlin/runtime/http/test/NonAsciiHeaderValueTest.kt
+++ b/runtime/protocol/http-client-engines/test-suite/common/test/aws/smithy/kotlin/runtime/http/test/NonAsciiHeaderValueTest.kt
@@ -21,7 +21,7 @@ class NonAsciiHeaderValueTest : AbstractEngineTest() {
         test { _, client ->
             val req = HttpRequest {
                 testSetup()
-                url.path.decoded = "concurrent"
+                url.path.decoded = "non-ascii"
                 headers.append("non-ascii-header", "µ")
                 headers.append("another-one", "\uD83C\uDF0A")
             }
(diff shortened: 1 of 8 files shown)


aajtodd (changes requested):
Can we get an http engine [integration test](https://github.com/awslabs/smithy-kotlin/tree/main/runtime/protocol/http-client-engines/test-suite) that verifies behavior across engines?

runtime/protocol/http-client-engines/test-suite/common/test/aws/smithy/kotlin/runtime/http/test/NonAsciiHeaderValueTest.kt
```diff
@@ -0,0 +1,34 @@
...
+import aws.smithy.kotlin.runtime.http.HttpStatusCode
+import aws.smithy.kotlin.runtime.http.complete
+import aws.smithy.kotlin.runtime.http.request.HttpRequest
+import aws.smithy.kotlin.runtime.http.test.util.AbstractEngineTest
+import aws.smithy.kotlin.runtime.http.test.util.test
+import aws.smithy.kotlin.runtime.http.test.util.testSetup
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class NonAsciiHeaderValueTest : AbstractEngineTest() {
```
aajtodd, line 17: **nit/suggestion**: Most of the other tests are under a classification like "uploads" or "downloads". I suggest making this more generic like "HeaderTests"

runtime/protocol/http-client-engines/test-suite/common/test/aws/smithy/kotlin/runtime/http/test/NonAsciiHeaderValueTest.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    @Test
+    fun testNonAsciiHeaderValue() = testEngines {
```
aajtodd, line 20: **suggestion**: Link to the issue.

aajtodd (approved):
Couple suggestions but looks good.

Issue #894, opened 2023-07-18
Emit HTTP engine thread count as a metric


aajtodd:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue, please leave a comment

## Describe the Feature

Add a new metric `smithy.client.http.threads.count` representing the current thread pool count (if applicable for an engine)

## Is your Feature Request related to a problem?

Insight into the way an HTTP engine thread pool grows or shrinks w.r.t connections and load is a useful insight for tuning.

- [x] I may be able to implement this feature request

PR #1022, opened 2024-01-04
feat: support for OkHttp thread metrics

ianbotsf:
## Issue \#

#894 

## Description of changes

This change adds metrics for the OkHttp engine's thread pool. It should be considered WIP until the following open questions are addressed:

* How can we best test this in an automated fashion?
* This change only adds support for OkHttp. CRT doesn't seem to have support for interrogating metrics about HTTP client thread counts. Is that good enough for now or should we push for enhanced support in CRT?
* The additions to `HttpClientMetrics` eschew the existing pattern of pushing metrics to intermediate refs and the polling them from the async counter in favor of pulling the source data directly via a passed callback. This seems cleaner to me since it removes unnecessary state but I'm curious whether others think it's a good idea (and whether we'd potentially want to extend it to other async counters as appropriate).

### DIFF
--- a/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/internal/HttpClientMetrics.kt
+++ b/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/internal/HttpClientMetrics.kt
@@ -25,7 +25,12 @@ public object HttpClientMetricAttributes {
     public val AcquiredConnection: Attributes = attributesOf { "state" to "acquired" }
     public val QueuedRequest: Attributes = attributesOf { "state" to "queued" }
     public val InFlightRequest: Attributes = attributesOf { "state" to "in-flight" }
+    public val TotalThread: Attributes = attributesOf { "state" to "total" }
+    public val ActiveThread: Attributes = attributesOf { "state" to "active" }
 }
+
+@InternalApi
+public data class ThreadState(val total: Long, val active: Long)
 
 /**
  * Container for common HTTP engine related metrics. Engine implementations can re-use this and update
@@ -39,6 +44,7 @@ public object HttpClientMetricAttributes {
 public class HttpClientMetrics(
     scope: String,
     public val provider: TelemetryProvider,
+    threadStateCallback: (() -> ThreadState?)? = null,
 ) : Closeable {
     private val meter = provider.meterProvider.getOrCreateMeter(scope)
 
@@ -76,34 +82,6 @@ public class HttpClientMetrics(
         "The amount of time a connection has been open",
     )
 
-    private val connectionLimitHandle = meter.createAsyncUpDownCounter(
-        "smithy.client.http.connections.limit",
-        { it.record(_connectionsLimit.value) },
-        "{connection}",
-        "Max connections configured for the HTTP client",
-    )
-
-    private val connectionUsageHandle = meter.createAsyncUpDownCounter(
-        "smithy.client.http.connections.usage",
-        ::recordConnectionState,
-        "{connection}",
-        "Current state of connections (idle, acquired)",
-    )
-
-    private val requestsConcurrencyLimitHandle = meter.createAsyncUpDownCounter(
-        "smithy.client.http.requests.limit",
-        { it.record(_requestConcurrencyLimit.value) },
-        "{request}",
-        "Max concurrent requests configured for the HTTP client",
-    )
-
-    private val requestsHandle = meter.createAsyncUpDownCounter(
-        "smithy.client.http.requests.usage",
-        ::recordRequestsState,
-        "{request}",
-        "The current state of HTTP client request concurrency (queued, in-flight)",
-    )
-
     public val bytesSent: MonotonicCounter = meter.createMonotonicCounter(
         "smithy.client.http.bytes_sent",
         "By",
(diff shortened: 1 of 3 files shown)


runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/internal/HttpClientMetrics.kt
```diff
@@ -122,6 +100,41 @@ public class HttpClientMetrics(
...
+            ::recordRequestsState,
+            "{request}",
+            "The current state of HTTP client request concurrency (queued, in-flight)",
+        ),
+        threadStateCallback?.let { callback ->
+            meter.createAsyncUpDownCounter(
+                "smithy.client.http.threads.count",
+                { measurement -> recordThreadState(measurement, callback) },
+                "{thread}",
+                "Current state of HTTP engine threads (active, running)",
```
lauzadis, line 133: How are "running" threads different from "active" ones? I see only "total" and "active" are recorded.

lauzadis:
1. I think unit tests for the metrics being emitted would be a good start. I'm not sure, is it easy to change the number of active threads in the OkHttp client, or is it more an internal implementation detail in the engine? If the latter, I think it would be sufficient to just check that the metrics are being emitted.
2. It would be nice to have both engines' metrics aligned but it can probably be solved on the longer term
3. I think it's a good idea

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/internal/HttpClientMetrics.kt
ianbotsf: You're right, "active" and "running" are not the correct terms here. I can change those to "total" and "active".

ianbotsf:
> 1. I think unit tests for the metrics being emitted would be a good start. I'm not sure, is it easy to change the number of active threads in the OkHttp client, or is it more an internal implementation detail in the engine? If the latter, I think it would be sufficient to just check that the metrics are being emitted.

To be clear, you mean verify that a metric value _exists_ but not what it is? Because this is multi-threaded behavior and we don't control the implementation of the underlying thread pool, it'll be difficult to achieve a stable expectation for number of total/active threads in a test case.

lauzadis:
> > 1. I think unit tests for the metrics being emitted would be a good start. I'm not sure, is it easy to change the number of active threads in the OkHttp client, or is it more an internal implementation detail in the engine? If the latter, I think it would be sufficient to just check that the metrics are being emitted.
> 
> To be clear, you mean verify that a metric value _exists_ but not what it is? Because this is multi-threaded behavior and we don't control the implementation of the underlying thread pool, it'll be difficult to achieve a stable expectation for number of total/active threads in a test case.

Yes, just testing that a metric value exists if we don't have predictable behavior to test on.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/internal/HttpClientMetrics.kt
```diff
@@ -39,6 +44,7 @@ public object HttpClientMetricAttributes {
 public class HttpClientMetrics(
     scope: String,
     public val provider: TelemetryProvider,
+    threadStateCallback: (() -> ThreadState?)? = null,
```
aajtodd, line 47: This seems different for no apparent reason. It's not wrong I guess but feels odd that we would special case it since each engine can adapt however they want to the current way metrics are exposed.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/internal/HttpClientMetrics.kt
```diff
@@ -25,8 +25,13 @@ public object HttpClientMetricAttributes {
     public val AcquiredConnection: Attributes = attributesOf { "state" to "acquired" }
     public val QueuedRequest: Attributes = attributesOf { "state" to "queued" }
     public val InFlightRequest: Attributes = attributesOf { "state" to "in-flight" }
+    public val TotalThread: Attributes = attributesOf { "state" to "total" }
```
aajtodd, line 28: **correctness**:  The state attribute for this should probably follow the [usage](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-naming) convention where the sub states total to some limit. 
e.g. thread state usage metric `smithy.client.http.threads.usage` with states of `idle | consumed` and a separate metric for the limit `smithy.client.http.threads.limit`.

Also I suppose we should probably get this defined in the SRA before codifying it in an implementation.

We may not know the total thread limit in every case or thread pools may be dynamic and spin down threads which would break the "should" part of the convention but I think that's fine. In the ideal case the convention applies and even when it doesn't it's still how you'd want it graphed.

aajtodd (changes requested):
> How can we best test this in an automated fashion?
I'm not sure we can in every scenario. I had to do a lot of testing locally with an observability stack setup to actually view the metrics working. Let me know if you discover something better.

> This change only adds support for OkHttp. CRT doesn't seem to have support for interrogating metrics about HTTP client thread counts. Is that good enough for now or should we push for enhanced support in CRT?

I'd say do what we can for now (although see my comment about getting this in SRA first before finalizing it).


> The additions to HttpClientMetrics eschew the existing pattern of pushing metrics to intermediate refs and the polling them from the async counter in favor of pulling the source data directly via a passed callback. This seems cleaner to me since it removes unnecessary state but I'm curious whether others think it's a good idea (and whether we'd potentially want to extend it to other async counters as appropriate).

The reason it's structured this way is so that engines can update the values however they see fit (whether that be inc/dec or absolute value changes). See my other comment here but I don't think I see a reason to special case this new metric and would likely prefer to keep them consistent.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/internal/HttpClientMetrics.kt
ianbotsf: Sure, I'll change this to idle/active threads.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/internal/HttpClientMetrics.kt
ianbotsf: Yes I concede it's very different from what we have elsewhere and feels very specialized right now.

I think one problem is that we don't directly control the thread pool. We don't decide when new threads are created or when they swap from idle to active. All we can do is inspect the state at any given moment to know the current value. That sounds like what an `AsyncUpDownCounter` is intended to do via a callback.

We could start a background thread in the OkHttp engine to periodically inspect the thread pool and stuff the idle/active counts into `HttpClientMetrics`, which would then be read when the callback is run. But this means the value will be out of sync by whatever interval the background thread uses. If, for instance, the background thread runs every 10 seconds but the meter polls every 5 seconds, then up to two datapoints will be reported inaccurately.

Is your greater concern that this change exposes the concept of callbacks outside of `HttpClientMetrics` and inverts the collection model from a push to a pull? Is it the inconsistency of some metrics being exposed outside of `HttpClientMetrics` as callbacks while others are mutable properties inside? Or is it that this may not be flexible enough to adapt to other engines where thread updates are more deterministic?

PR #1023, opened 2024-01-05
feat: add SIGV4A auth scheme integration 

0marperez:
## Issue \#

N/A

## Description of changes

Added a custom exception for SIGV4A in default signer

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
@@ -50,7 +50,7 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Unit>() {
         LOGGER.info("Discovering KotlinIntegration providers...")
         integrations = ServiceLoader.load(KotlinIntegration::class.java, classLoader)
             .onEach { integration -> LOGGER.info("Loaded KotlinIntegration: ${integration.javaClass.name}") }
-            .filter { integration -> integration.enabledForService(context.model, settings) }
+            .filter { integration -> integration.enabledForService(context.model, settings) } // TODO: Change so we don't filter until previous integrations model modifications are complete
             .onEach { integration -> LOGGER.info("Enabled KotlinIntegration: ${integration.javaClass.name}") }
             .sortedBy(KotlinIntegration::order)
 
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
@@ -8,9 +8,15 @@ package software.amazon.smithy.kotlin.codegen
 import software.amazon.smithy.build.FileManifest
 import software.amazon.smithy.build.PluginContext
 import software.amazon.smithy.codegen.core.SymbolProvider
-import software.amazon.smithy.kotlin.codegen.core.*
+import software.amazon.smithy.kotlin.codegen.core.GenerationContext
+import software.amazon.smithy.kotlin.codegen.core.KotlinDelegator
+import software.amazon.smithy.kotlin.codegen.core.KotlinDependency
+import software.amazon.smithy.kotlin.codegen.core.toRenderingContext
 import software.amazon.smithy.kotlin.codegen.integration.KotlinIntegration
-import software.amazon.smithy.kotlin.codegen.model.*
+import software.amazon.smithy.kotlin.codegen.model.OperationNormalizer
+import software.amazon.smithy.kotlin.codegen.model.getEndpointRules
+import software.amazon.smithy.kotlin.codegen.model.getEndpointTests
+import software.amazon.smithy.kotlin.codegen.model.hasTrait
 import software.amazon.smithy.kotlin.codegen.rendering.*
 import software.amazon.smithy.kotlin.codegen.rendering.protocol.ApplicationProtocol
 import software.amazon.smithy.kotlin.codegen.rendering.protocol.ProtocolGenerator
@@ -47,7 +53,6 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Unit>() {
             .filter { integration -> integration.enabledForService(context.model, settings) }
             .onEach { integration -> LOGGER.info("Enabled KotlinIntegration: ${integration.javaClass.name}") }
             .sortedBy(KotlinIntegration::order)
-            .toList()
 
         LOGGER.info("Preprocessing model")
         // Model pre-processing:
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AsymmetricAuthSchemeIntegration.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AsymmetricAuthSchemeIntegration.kt
@@ -49,9 +49,6 @@ class SigV4AsymmetricAuthSchemeIntegration : KotlinIntegration {
     override fun additionalServiceConfigProps(ctx: CodegenContext): List<ConfigProperty> =
         if (modelHasSigV4aTrait(ctx)) listOf(credentialsProviderProp) else super.additionalServiceConfigProps(ctx)
 
-    override fun customizeEndpointResolution(ctx: ProtocolGenerator.GenerationContext): EndpointCustomization? =
-        if (modelHasSigV4aTrait(ctx)) SigV4AsymmetricEndpointCustomization else null
-
     override val sectionWriters: List<SectionWriterBinding> = listOf(
             SectionWriterBinding(HttpProtocolUnitTestRequestGenerator.ConfigureServiceClient, renderHttpProtocolRequestTestConfigureServiceClient),
             SectionWriterBinding(HttpProtocolUnitTestResponseGenerator.ConfigureServiceClient, renderHttpProtocolResponseTestConfigureServiceClient),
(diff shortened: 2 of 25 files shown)


runtime/auth/aws-signing-default/common/src/aws/smithy/kotlin/runtime/auth/awssigning/DefaultAwsSigner.kt
```diff
@@ -21,8 +21,16 @@ internal class DefaultAwsSignerImpl(
...
 
-        // TODO implement SigV4a
-        require(config.algorithm == AwsSigningAlgorithm.SIGV4) { "${config.algorithm} support is not yet implemented" }
+        // TODO: implement SigV4a
+        if (config.algorithm != AwsSigningAlgorithm.SIGV4) {
+            var exceptionMessage = "${config.algorithm} support is not yet implemented for the default signer."
+
+            if (config.algorithm == AwsSigningAlgorithm.SIGV4_ASYMMETRIC) {
+                exceptionMessage += " Please follow the AWS SDK for Kotlin developer guide to set it up with the CRT signer. **LINK TO GUIDE**"
+            }
```
ianbotsf, line 30: **Style**: You can avoid mutable string references and concatenation by using `buildString`:

```kotlin
val exceptionMessage = buildString {
    append("${config.algorithm} support is not yet implemented for the default signer.")

    if (config.algorithm == AwsSigningAlgorithm.SIGV4_ASYMMETRIC) {
        append(" Please follow the AWS SDK for Kotlin developer guide to set it up with the CRT signer. **LINK TO GUIDE**")
    }
}
```

ianbotsf, line 30: **Correctness**: I'd keep the message simple, both for brevity in logs and for accuracy. Our redirect link may not always go to the developer guide. I'd shoot for something closer to: `"<ALGORITHM> support is not yet implemented for the default signer. See <LINK> for more info."`

ianbotsf (changes requested):

runtime/auth/aws-signing-default/common/src/aws/smithy/kotlin/runtime/auth/awssigning/DefaultAwsSigner.kt
```diff
@@ -21,8 +21,16 @@ internal class DefaultAwsSignerImpl(
...
         val logger = coroutineContext.logger<DefaultAwsSignerImpl>()
 
-        // TODO implement SigV4a
-        require(config.algorithm == AwsSigningAlgorithm.SIGV4) { "${config.algorithm} support is not yet implemented" }
+        // TODO: implement SigV4a
+        if (config.algorithm != AwsSigningAlgorithm.SIGV4) {
+            var exceptionMessage = "${config.algorithm} support is not yet implemented for the default signer."
+
+            if (config.algorithm == AwsSigningAlgorithm.SIGV4_ASYMMETRIC) {
+                exceptionMessage += " Please follow the AWS SDK for Kotlin developer guide to set it up with the CRT signer. **LINK TO GUIDE**"
```
aajtodd, line 29: **correctness**: This doesn't really belong here. This is in the generic `smithy-kotlin` runtime and isn't specific to the AWS SDK for Kotlin. This belongs as _probably_ a custom interceptor on the SDK side that looks for `UnsupportedSigningAlogirhm` as the exception type and customizes the message there.

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningExceptions.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.auth.awssigning
+
+import aws.smithy.kotlin.runtime.InternalApi
+
+@InternalApi
+public class UnsupportedSigningAlgorithm(message: String) : Exception(message)
```
aajtodd, line 10: **fix**: This should probably inherit from `ClientException`.

aajtodd (changes requested):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
```diff
@@ -8,9 +8,15 @@ package software.amazon.smithy.kotlin.codegen
 import software.amazon.smithy.build.FileManifest
 import software.amazon.smithy.build.PluginContext
 import software.amazon.smithy.codegen.core.SymbolProvider
-import software.amazon.smithy.kotlin.codegen.core.*
```
0marperez, line 11: My Intellij does this automatically, I'll turn it off I think

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AsymmetricAuthSchemeIntegration.kt
```diff
@@ -0,0 +1,77 @@
...
+ */
+class SigV4AsymmetricAuthSchemeIntegration : KotlinIntegration {
+    override val order: Byte = -59
+
+    // Needs to be true due to the way integrations are filtered out before application and sigV4a customization.
+    // See 'CodegenVisitor' & 'SigV4AsymmetricTraitCustomization'
+    override fun enabledForService(model: Model, settings: KotlinSettings): Boolean = true
+
+    override fun authSchemes(ctx: ProtocolGenerator.GenerationContext): List<AuthSchemeHandler> =
+        if (modelHasSigV4aTrait(ctx)) listOf(SigV4AsymmetricAuthSchemeHandler()) else super.authSchemes(ctx)
```
lauzadis, line 35: correctness: no need to call `super` here, it can be `else listOf()`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
```diff
@@ -47,7 +53,6 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Unit>() {
             .filter { integration -> integration.enabledForService(context.model, settings) }
             .onEach { integration -> LOGGER.info("Enabled KotlinIntegration: ${integration.javaClass.name}") }
             .sortedBy(KotlinIntegration::order)
-            .toList()
```
lauzadis, line 50: question; why remove `toList()`?

0marperez: `sortedBy` returns a list so the `toList()` was unnecessary

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AsymmetricAuthSchemeIntegration.kt
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Register support for the `aws.auth#sigv4a` auth scheme.
+ */
+class SigV4AsymmetricAuthSchemeIntegration : KotlinIntegration {
+    override val order: Byte = -59
+
+    // Needs to be true due to the way integrations are filtered out before application and sigV4a customization.
+    // See 'CodegenVisitor' & 'SigV4AsymmetricTraitCustomization'
+    override fun enabledForService(model: Model, settings: KotlinSettings): Boolean = true
```
ianbotsf, line 32: **Question**: We don't have any other integrations where `enabledForService` _always_ returns `true`. The majority of them evaluate some criteria from the model and/or codegen context to make the determination about enablement. It's not clear why we need _two_ customizations to handle this. Why couldn't this be combined with `SigV4AsymmetricTraitCustomization` up in **aws-sdk-kotlin**?

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningExceptions.kt
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * See: [AwsSigningAlgorithm], [AwsSigner]
+ *
+ * @param isSigV4a Whether the unsupported signing algorithm is sigV4a
+ */
+@InternalApi
+public class UnsupportedSigningAlgorithm(
+    message: String,
+    public val isSigV4a: Boolean,
+) : ClientException(message)
```
ianbotsf, line 21: **Correctness**: The `isSigV4a` parameter feels too specific. It'd be better to simply include the algorithm which is unsupported (e.g., `AwsSigningAlgorithm.SIGV4_ASYMMETRIC`).

ianbotsf, line 21: **Style**: Typically exceptions have the word `Exception` in their name. As verbose as it is, I suggest `UnsupportedSigningAlgorithmException`.

ianbotsf (changes requested):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AsymmetricAuthSchemeIntegration.kt
0marperez: > We don't have any other integrations where enabledForService always returns true. The majority of them evaluate some criteria from the model and/or codegen context to make the determination about enablement

This one is a weird one, the code in `CodegenVisitor` first filters out the integrations that won't be enabled via the `enabledForService()`  function and AFTER it applies the changes from the integrations.

The issue with this is that we have another integration (`SigV4AsymmetricTraitCustomization`) that applies the SIGV4A trait (the condition we would use to apply this integration). So the integration would be filtered out if we used the `enabledForService()` function because the filtering happens before the customization adds the trait we'd be looking for .

> Why couldn't this be combined with SigV4AsymmetricTraitCustomization up in aws-sdk-kotlin?

The `SigV4AsymmetricTraitCustomization` is there to add the SIGV4A trait to the models that should have it. This one is how we would deal with the trait. 

Since the trait can be added to any model it makes sense to have them be independent, the rendering can be done by smithy and the customization can be done in the sdk only

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AsymmetricAuthSchemeIntegration.kt
```diff
@@ -25,7 +25,8 @@ import software.amazon.smithy.model.shapes.ShapeId
  * Register support for the `aws.auth#sigv4a` auth scheme.
  */
 class SigV4AsymmetricAuthSchemeIntegration : KotlinIntegration {
-    override val order: Byte = -59
+    // Needs to happen after the `SigV4AsymmetricTraitCustomization` (-60).
+    override val order: Byte = -50
```
ianbotsf, line 29: **Question**: This now has the same order (`-50`) as `SigV4AuthSchemeIntegration`...is that intentional?

lauzadis (approved):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AsymmetricAuthSchemeIntegration.kt
ianbotsf: Discussed offline and decided to add a `TODO` + backlog item for improving the filtering logic in `CodegenVisitor` so this `enabledForService` check can be more meaningful.

ianbotsf (approved):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AsymmetricAuthSchemeIntegration.kt
0marperez: Yeah, the order doesn't really matter between those two anymore, the `order` being the same number isn't really an issue either. They're sorted by `sortedBy(KotlinIntegration::order)`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AsymmetricAuthSchemeIntegration.kt
aajtodd: This could be modeled accurately by looking for the `sigv4a` trait and inspecting the endpoint rules.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AsymmetricAuthSchemeIntegration.kt
```diff
@@ -0,0 +1,78 @@
...
+    override val authSchemeIdSymbol: Symbol = buildSymbol {
+        name = "AuthSchemeId.AwsSigV4Asymmetric"
+        val ref = RuntimeTypes.Auth.Identity.AuthSchemeId
+        objectRef = ref
+        namespace = ref.namespace
+        reference(ref, SymbolReference.ContextOption.USE)
+    }
+
+    override fun identityProviderAdapterExpression(writer: KotlinWriter) {
+        writer.write("config.credentialsProvider")
```
aajtodd, line 51: **Question/Correctness**: We don't register support for this config property in this integration, what is guaranteeing that this config property exists?

PR #1030, opened 2024-02-05
chore: add @InternalApi to binary compatibility checks; introduce new PR check to verify no API-breaking changes

ianbotsf:
## Issue \#

https://github.com/awslabs/aws-sdk-kotlin/issues/1191

## Description of changes

This change tightens backwards compatibility guarantees by:
* Removing `@InternalApi` from the exemptions for the backwards compatibility verifier
  * Also updates the API dumps because a lot of previously-exempted APIs are now in scope
* Adds a PR check which verifies that either no API dump lines have been deleted _or_ the **acknowledge-api-break** label is applied to the PR
  * [Example run with no breaking changes](https://github.com/awslabs/smithy-kotlin/actions/runs/7791179880/job/21246672755) (succeeded)
  * [Example run with breaking changes, no label](https://github.com/awslabs/smithy-kotlin/actions/runs/7791240677/job/21246879734) (failed)
  * [Example run with breaking changes, **acknowledge-api-break** label](https://github.com/awslabs/smithy-kotlin/actions/runs/7791252823/job/21246921308?pr=1030) (succeeded)

### DIFF
--- a/.github/workflows/api-compat-verification.yml
+++ b/.github/workflows/api-compat-verification.yml
@@ -1,0 +1,33 @@
+name: API compatibility verification
+
+on:
+  pull_request:
+    types: [ opened, synchronize, reopened, labeled, unlabeled ]
+    branches: [ main ]
+
+jobs:
+  api-compat-verification:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Check for API compatibility
+        if: ${{ !contains(github.event.pull_request.labels.*.name, 'acknowledge-api-break') }}
+        run: |
+          git fetch origin ${{ github.base_ref }} --depth 1 && \
+          git diff remotes/origin/${{ github.base_ref }} --numstat "*.api" | awk '
+            BEGIN { s = 0 }
+            
+            # git diff numstat shows lines deleted in field 2, hence sum up field 2 across all items 
+            { s += $2 }
+            
+            # exit with the number of lines deleted as the result code so that `if ${{ failure() }}` works below
+            END { exit s }
+          '
+      - name: Error message
+        if: ${{ failure() }}
+        run: |
+          echo "::error ::This change modifies the public API in a way that may be backwards-incompatible. Carefully"
+          echo "::error ::review this pull request and either:
+          echo "::error ::* Revert the changes which caused the API incompatibility –or–"
+          echo "::error ::* Add the 'acknowledge-api-break' label to this PR (in rare cases warranting an API breakage)"
+          exit 1
--- a/.github/workflows/api-compat-verification.yml
+++ b/.github/workflows/api-compat-verification.yml
@@ -27,7 +27,7 @@ jobs:
         if: ${{ failure() }}
         run: |
           echo "::error ::This change modifies the public API in a way that may be backwards-incompatible. Carefully"
-          echo "::error ::review this pull request and either:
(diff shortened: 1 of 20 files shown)


.github/workflows/api-compat-verification.yml
```diff
@@ -0,0 +1,33 @@
+name: API compatibility verification
+
+on:
+  pull_request:
+    types: [ opened, synchronize, reopened, labeled, unlabeled ]
```
lauzadis, line 5: Is there any reason to specify these types? Our other workflows look like they apply to all types

.github/workflows/api-compat-verification.yml
```diff
@@ -0,0 +1,33 @@
...
+            { s += $2 }
+            
+            # exit with the number of lines deleted as the result code so that `if failure()` works below
+            END { exit s }
+          '
+      - name: Error message
+        if: ${{ failure() }}
+        run: |
+          echo "::error ::This change modifies the public API in a way that may be backwards-incompatible. Carefully"
+          echo "::error ::review this pull request and either:"
```
lauzadis, line 30: style: It might be better to combine L29 and L30 because GitHub's CI output autowraps the text
(I think it looks good on one line)

lauzadis (approved):

.github/workflows/api-compat-verification.yml
ianbotsf: I copied the job definition from [**changelog-verification.yml**](https://github.com/awslabs/smithy-kotlin/blob/b2350449f70a78f3c4c647f251b49d1cfc0b57e4/.github/workflows/changelog-verification.yml) which also explicitly specifies PR events. I think, in general, we could be more conservative in the subtypes of events which trigger action runs and reduce some of the churn/cost associated with them.

Do you have a reason to prefer running on all event subtypes?

lauzadis: No actually, I think the current `types` cover all the cases where this should run, so it's good to leave it as is. I just never saw these types specified (I did not look at the `changelog-verification.yml` job) and was curious

aajtodd: In practice for us these are likely about the same but it's fine as is, here is the [full list](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) for reference if you leave it unspecified and just have `pull_request:`

build.gradle.kts
```diff
@@ -116,8 +116,6 @@ configureLinting(lintPaths)
 
 // Binary compatibility
 apiValidation {
-    nonPublicMarkers.add("aws.smithy.kotlin.runtime.InternalApi")
```
aajtodd, line 119: In general I agree with this change but I am wondering about what we will do when we do need/want to break internal APIs. Presumably we'd deprecate the APIs and leave them in place for some period of time or perhaps indefinitely and switch codegen to use whatever the replacement is?

aajtodd (approved):

build.gradle.kts
ianbotsf: Discussed offline and decided largely on what @aajtodd suggested: if we need to deprecate an `@InternalApi` declaration, we'll mark the API as `@Deprecated` and leave it in place for some number of releases. It will be considered for (potentially-breaking) removal if it is determined to be actively harming something (e.g., security, critical bugs, etc.).

PR #1035, opened 2024-02-12
fix: http trace interceptor not keeping original HttpBody properties and LogMode not behaving as expected

aajtodd:
## Issue \#

fixes https://github.com/awslabs/aws-sdk-kotlin/issues/1198

## Description of changes

* **fix**: Preserve the original `HttpBody` properties like content length from the original `HttpBody` when duplicating it. The issue in awslabs/aws-sdk-kotlin#1198 seems to have been caused by the fact that the original body was an `AwsChunked` body with a `transfer-encoding: chunked` header. When we replace the body during the trace call it was replaced with a non-equivalent in-memory byte array body which always has a content length.
* **fix**: `LogRequestWithBody` and `LogResponseWithBody` should imply `LogRequest` and `LogResponse` respectively. I discovered this trying to debug this issue and seeing the incorrect behaviors for dumping the canonical request from S3

### DIFF
--- a/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/request/HttpRequestBuilder.kt
+++ b/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/request/HttpRequestBuilder.kt
@@ -113,7 +113,7 @@ public suspend fun dumpRequest(request: HttpRequestBuilder, dumpBody: Boolean): 
     request.headers.entries()
         .filterNot { it.key in skip }
         .forEach { (key, values) ->
-            buffer.writeUtf8(values.joinToString(separator = ";", prefix = "${key}: ", postfix = "\r\n"))
+            buffer.writeUtf8(values.joinToString(separator = ";", prefix = "$key: ", postfix = "\r\n"))
         }
 
     buffer.writeUtf8("\r\n")
--- a/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/request/HttpRequestBuilder.kt
+++ b/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/request/HttpRequestBuilder.kt
@@ -112,8 +112,8 @@ public suspend fun dumpRequest(request: HttpRequestBuilder, dumpBody: Boolean): 
     val skip = setOf("Host", "Content-Length")
     request.headers.entries()
         .filterNot { it.key in skip }
-        .forEach { entry ->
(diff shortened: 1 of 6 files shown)


lauzadis (approved):

0marperez (approved):

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/request/HttpRequestBuilder.kt
```diff
@@ -113,26 +112,35 @@ public suspend fun dumpRequest(request: HttpRequestBuilder, dumpBody: Boolean):
     val skip = setOf("Host", "Content-Length")
     request.headers.entries()
         .filterNot { it.key in skip }
-        .forEach {
-            buffer.writeUtf8(it.value.joinToString(separator = ";", prefix = "${it.key}: ", postfix = "\r\n"))
+        .forEach { entry ->
+            buffer.writeUtf8(entry.value.joinToString(separator = ";", prefix = "${entry.key}: ", postfix = "\r\n"))
         }
```
ianbotsf, line 117: **Style**: Consider using a destructured lambda declaration:

```kotlin
        .forEach { (key, value) ->
            buffer.writeUtf8(value.joinToString(separator = ";", prefix = "$key: ", postfix = "\r\n"))
```

runtime/protocol/http/common/test/aws/smithy/kotlin/runtime/http/HttpRequestBuilderTest.kt
```diff
@@ -67,13 +65,18 @@ class HttpRequestBuilderTest {
...
-        val actual = dumpRequest(builder, true)
-        assertTrue(builder.body is HttpBody.Bytes)
-        val expected = "GET /debug/test?foo=bar\r\nHost: test.amazon.com\r\nContent-Length: ${content.length}\r\nx-baz: quux;qux\r\n\r\n$content"
-        assertEquals(expected, actual)
+        val actualWithContent = dumpRequest(builder, true)
+        assertTrue(builder.body is HttpBody.SourceContent)
+        val expectedWithContent = "$expectedNoContent$content"
+        assertEquals(expectedWithContent, actualWithContent)
+
+        val actualReplacedContent = builder.body.readAll()?.decodeToString() ?: fail("expected content")
```
ianbotsf, line 78: **Nit**: Replace `fail` with `assertNotNull`:

```kotlin
val actualReplacedContent = assertNotNull(builder.body.readAll(), "expected content").decodeToString()
```

ianbotsf (approved):

PR #1043, opened 2024-02-27
feat: support sigv4aSigningRegionSet config setting

0marperez:
## Issue \#
See aws-sdk-kotlin#1235

## Description of changes
The `sigv4aSigningRegionSet` config setting is added in aws-sdk-kotlin#1235 and it's routed to the relevant signer in this PR


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AsymmetricAuthSchemeIntegration.kt
```diff
@@ -66,7 +66,7 @@ private class SigV4AsymmetricAuthSchemeHandler : AuthSchemeHandler {
 
     override fun instantiateAuthSchemeExpr(ctx: ProtocolGenerator.GenerationContext, writer: KotlinWriter) {
         val signingService = AwsSignatureVersion4Asymmetric.signingServiceName(ctx.service)
-        writer.write("#T(#T, #S)", RuntimeTypes.Auth.HttpAuthAws.SigV4AsymmetricAuthScheme, RuntimeTypes.Auth.Signing.AwsSigningStandard.DefaultAwsSigner, signingService)
+        writer.write("#T(#T, #S, config.sigv4aSigningRegionSet)", RuntimeTypes.Auth.HttpAuthAws.SigV4AsymmetricAuthScheme, RuntimeTypes.Auth.Signing.AwsSigningStandard.DefaultAwsSigner, signingService)
```
ianbotsf, line 69: **Correctness**: We're referencing a client config property here that doesn't exist anywhere in **smithy-kotlin**. I think the addition of that config property should be moved from **aws-sdk-kotlin** into this repo, possibly into this integration.

ianbotsf (changes requested):

runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/AwsHttpSigner.kt
```diff
@@ -108,6 +108,12 @@ public class AwsHttpSigner(private val config: Config) : HttpSigner {
          * The default predicate is to not reject signing any headers (i.e., `_ -> true`).
          */
         public var shouldSignHeader: ShouldSignHeaderPredicate = { _ -> true }
+
+        /**
+         * The sigV4a region set that was configured by the client.
+         * If value is not null then it should take precedence over the contextSigningRegionSet from the attributes
```
lauzadis, line 114: nit: should -> will

runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/AwsHttpSigner.kt
```diff
@@ -134,6 +140,7 @@ public class AwsHttpSigner(private val config: Config) : HttpSigner {
             algorithm = config.algorithm
 
             region = when {
+                algorithm == AwsSigningAlgorithm.SIGV4_ASYMMETRIC && config.clientSigv4aSigningRegionSet != null -> config.clientSigv4aSigningRegionSet!!.toSet().joinToString(",")
```
lauzadis, line 143: question: I noticed we call `.toSet()`, why not just make it a `Set` instead of a `List` in configuration?

runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/SigV4AsymmetricAuthScheme.kt
```diff
@@ -32,6 +32,19 @@ public class SigV4AsymmetricAuthScheme(
         },
     )
 
+    public constructor(
+        awsSigner: AwsSigner,
+        serviceName: String? = null,
+        clientSigningRegionSet: List<String>? = null,
```
lauzadis, line 38: naming suggestion: `signingRegionSet`

runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/SigV4AsymmetricAuthScheme.kt
```diff
@@ -32,6 +32,19 @@ public class SigV4AsymmetricAuthScheme(
...
+    ) : this (
+        AwsHttpSigner.Config().apply {
+            signer = awsSigner
+            service = serviceName
+            algorithm = AwsSigningAlgorithm.SIGV4_ASYMMETRIC
+            clientSigv4aSigningRegionSet = clientSigningRegionSet
```
lauzadis, line 44: design: I think instead of adding a new `clientSigv4aSigningRegionSet` parameter to the signer it would be nice if we could work this into the signing attributes. 

If the client configures a SigV4A signing region set here, we might be able to add that as a signing attribute (overriding any existing value) instead of directly routing it through

runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/AwsHttpSigner.kt
aajtodd: I'd err on the side of leaving it a list should order ever matter.

**correctness**: This is favoring the value set on the signer config but the context should take priority not config.

runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/SigV4AsymmetricAuthScheme.kt
aajtodd: We don't need a new constructor for this do we? Can we not just add a new default param?

runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/AwsHttpSigner.kt
```diff
@@ -108,6 +108,12 @@ public class AwsHttpSigner(private val config: Config) : HttpSigner {
          * The default predicate is to not reject signing any headers (i.e., `_ -> true`).
          */
         public var shouldSignHeader: ShouldSignHeaderPredicate = { _ -> true }
+
+        /**
+         * The sigV4a region set that was configured by the client.
+         * If value is not null then it should take precedence over the contextSigningRegionSet from the attributes
+         */
+        public var clientSigv4aSigningRegionSet: List<String>? = null
```
aajtodd, line 116: This should probably just be called `sigv4ASigningRegionSet`

runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/SigV4AsymmetricAuthScheme.kt
aajtodd: Agreed we definitely need the ability to set signing region set as an attribute. We already appear to have one `AwsSigningAttributes.SigningRegionSet`. We also need to remove `client` from all the names, these types may not always be used in client codegen.

aajtodd (changes requested):

runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/AwsHttpSigner.kt
ianbotsf: Reminder: `toSet()` and `setOf()` return a `LinkedHashMap` which guarantees the order is preserved (minus any removed duplicate elements).

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AsymmetricAuthSchemeIntegration.kt
```diff
@@ -34,6 +39,40 @@ class SigV4AsymmetricAuthSchemeIntegration : KotlinIntegration {
 
     override fun authSchemes(ctx: ProtocolGenerator.GenerationContext): List<AuthSchemeHandler> =
         if (modelHasSigV4aTrait(ctx)) listOf(SigV4AsymmetricAuthSchemeHandler()) else emptyList()
+
+    override fun additionalServiceConfigProps(ctx: CodegenContext): List<ConfigProperty> =
+        if (modelHasSigV4aTrait(ctx)) {
```
aajtodd, line 44: **question**: Why can't this be determined in `enabledForService`? I see a comment about how integrations are filtered but I'm still not sure I'm following.

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/SigV4aClientConfig.kt
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.client.config
```
aajtodd, line 5: **fix**: This doesn't belong in `smithy-client`. If anything it belongs in `aws-credentials` (which probably should have been named `aws-identity-api` or similar 🤷) or possibly `http-auth-aws`.

runtime/auth/http-auth-aws/common/src/aws/smithy/kotlin/runtime/http/auth/AwsHttpSigner.kt
```diff
@@ -135,6 +136,7 @@ public class AwsHttpSigner(private val config: Config) : HttpSigner {
             algorithm = config.algorithm
 
             region = when {
+                algorithm == AwsSigningAlgorithm.SIGV4_ASYMMETRIC && !configSigningRegionSet.isNullOrEmpty() -> configSigningRegionSet.joinToString(",")
```
aajtodd, line 139: **suggestion**: Add the context here in a comment about _why_ config takes precedence. 

```kotlin
when {
     // siging region set from client config overrides all other sources like endpoints because it's designed as an
     // escape hatch for customers to control/limit which regions a request will be valid for (e.g. since some services
     // use '*')
     algorithm == AwsSigningAlgorithm.SIGV4_ASYMMETRIC ...
     ...
}
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AsymmetricAuthSchemeIntegration.kt
lauzadis: relevant: https://github.com/awslabs/aws-sdk-kotlin/pull/1245#discussion_r1515002007

0marperez: SigV4a as a trait is added for some services that don't model their use in another integration, `SigV4AsymmetricTraitCustomization`. We could look for that trait in `enabledForService` but `CodegenVisitor` filters integrations out based on that `enabledForService` function before the model preprocessing can happen from integration `SigV4AsymmetricTraitCustomization`. The fix is simple but it breaks some seemingly unrelated tests.

0marperez: `CodegenVisitor.kt`
```
integrations = ServiceLoader.load(KotlinIntegration::class.java, classLoader)
    .onEach { integration -> LOGGER.info("Loaded KotlinIntegration: ${integration.javaClass.name}") }
    .filter { integration -> integration.enabledForService(context.model, settings) } // TODO: Change so we don't filter until previous integrations model modifications are complete
    .onEach { integration -> LOGGER.info("Enabled KotlinIntegration: ${integration.javaClass.name}") }
    .sortedBy(KotlinIntegration::order)


LOGGER.info("Preprocessing model")
// Model pre-processing:
// 1. Start with the model from the plugin context
// 2. Apply integrations
// 3. Flatten error shapes (see: https://github.com/awslabs/smithy/pull/919)
// 4. Normalize the operations
var resolvedModel = context.model
for (integration in integrations) {
    resolvedModel = integration.preprocessModel(resolvedModel, settings)
}

```

0marperez: I'm working on those tests right now btw

ianbotsf (approved):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AsymmetricAuthSchemeIntegration.kt
0marperez: This is fixed in [smithy-kotlin#1049](https://github.com/awslabs/smithy-kotlin/pull/1049)

aajtodd (approved):

PR #1047, opened 2024-02-29
fix: disable OkHttp response decompression

### LINKED ISSUE #1041: Document/disable OkHttp's automatic response decompression
## Describe the bug

Getting a compressed object from S3 with `S3Client.getObject` when using the default OkHttp engine returns the body _uncompressed_ instead of in its original, persisted form. This is because [OkHttp automatically accepts and decompresses GZipped responses](https://square.github.io/okhttp/features/calls/#rewriting-requests). This behavior is surprising and differs from CRT engine and the Java v2 SDK, both of which return the compressed object body.

## Expected Behavior

Users are not surprised by unexpectedly decompressed objects.

## Steps to Reproduce

This code shows the issue:

```kotlin
val uncompressed = ByteArray(1024) { it.mod(Byte.MAX_VALUE) }
val compressed = ByteArrayOutputStream()
    .also { GZIPOutputStream(it, true).apply {
        write(uncompressed)
        flush()
        close()
    } }
    .toByteArray()
println("Uncompressed length: ${uncompressed.size}") // Uncompressed length: 1024
println("Compressed length: ${compressed.size}") // Compressed length: 164

runBlocking {
    S3Client.fromEnvironment().use { s3 ->
        s3.putObject {
            bucket = "<some-bucket>"
            key = "compression-test"
            contentEncoding = "gzip"
            body = ByteStream.fromBytes(compressed)
        }

        s3.getObject(GetObjectRequest {
            bucket = "<some-bucket>"
            key = "compression-test"
        }) {
            println("Response length: ${it.body?.toByteArray()?.size}") // Response length: 1024
        }
    }
}
```

## Possible Solution

We should either:
* document this behavior more clearly
* offer HTTP engine settings to change it
* disable the OkHttp transparent decompression altogether

## Context

Currently available workarounds include setting `responseContentEncoding = "identity"` in the `GetObjectRequest`, adding an interceptor to accomplish the same thing, or using the CRT engine.

## Your Environment

* Smithy Kotlin version used: 1.0.15
* Platform (JVM/JS/Native): OpenJDK 8
* Operating System and version: Amazon Linux 2

ianbotsf:
## Issue \#

Closes #1041 

## Description of changes

This change disables [OkHttp's transparent response compression feature](https://square.github.io/okhttp/features/calls/#rewriting-requests) by explicitly setting `Accept-Encoding` header to `identity` when not otherwise set.


runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpUtils.kt
```diff
@@ -88,6 +80,19 @@ internal fun HttpRequest.toOkHttpRequest(
...
+private fun Headers.toOkHttpHeaders(): OkHttpHeaders = OkHttpHeaders.Builder().also { okHeaders ->
+    forEach { key, values ->
+        values.forEach { value ->
+            okHeaders.addUnsafeNonAscii(key, value)
+        }
+    }
+
+    if ("Accept-Encoding" !in this) {
+        // Disable OkHttp transparent response decompression. See https://github.com/awslabs/smithy-kotlin/issues/1041
+        okHeaders.addUnsafeNonAscii("Accept-Encoding", "identity")
```
aajtodd, line 92: **Question**: This doesn't cause issues with signatures? I didn't see anywhere in our canonicalizer or CRT where `Accept-Encoding` is ignored.

lauzadis: I think it's not an issue because `toOkHttpRequest` is called after canonicalization / signing, and `Accept-Encoding` is not sent as a signed header

lauzadis (approved):

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpUtils.kt
lauzadis: Oh ok, if you add `Accept-Encoding` before signing (like in an interceptor), yes it does end up as a signed header. All headers are signed by default unless someone configures `shouldSignHeader` predicate.

aajtodd: Right so that would result in a signature mismatch server side I would assume.

lauzadis: I tried adding an `Accept-Encoding: gzip` header _before_ signing for a request to S3 and it succeeded, can't say that will work for all services though.

ianbotsf: If some other feature/interceptor added an `Accept-Encoding` header, it would go through our regular signing logic and be included in the signature calculation. Because this PR adds the header post-signing _only if it doesn't already exist_, it will be unsigned.

According to [the public SigV4 spec](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html), `Accept-Encoding` is not _required_ to be signed:
> The `CanonicalHeaders` list must include the following:
> * HTTP `host` header.
> * If the `Content-Type` header is present in the request, you must add it to the `CanonicalHeaders` list.
> * Any `x-amz-*` headers that you plan to include in your request must also be added. For example, if you are using temporary security credentials, you need to include `x-amz-security-token` in your request. You must add this header in the list of `CanonicalHeaders`.

That wording implies then that any additional headers in the request _may_ be canonicalized and included in the signature but are not required to be. (The `SignedHeaders` component of the `Authorization` header would definitively identify which headers were included in the signature calculation.)

That said, it's possible that some service-side implementations of SigV4 verification may have different assumptions about which headers must be signed. I've run this PR against our (admittedly small) suite of E2E tests against Kinesis, Polly, S3, STS, Timestream, and Transcribe, all of which pass.

ianbotsf: FWIW, I just ran our service benchmark suite as well, which touches S3, SNS, STS, CloudWatch, CloudWatch Events, DynamoDB, Pinpoint, and IAM. Again, all calls succeed.

That the E2E tests and benchmarks all pass indicates to me that _these_ services, at least, do not require `Accept-Encoding` to be signed (which is in agreement with the SigV4 spec).

aajtodd: 👍 Cool I just wanted to confirm we'd verified this.

> The SignedHeaders component of the Authorization header would definitively identify which headers were included in the signature calculation.

Yes but I know in the past I want to say we still hit issues where services didn't seem to care. This all looks fine though and we've done due diligence.

aajtodd (approved):

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpUtils.kt
```diff
@@ -88,6 +80,19 @@ internal fun HttpRequest.toOkHttpRequest(
     return builder.build()
 }
 
+private fun Headers.toOkHttpHeaders(): OkHttpHeaders = OkHttpHeaders.Builder().also { okHeaders ->
```
aajtodd, line 83: I forgot to say this but my only suggestion would be to have added this as an integration test that every engine that receives a gzip response leaves it alone and intact.

ianbotsf: Good idea! Added a test to the engine test suite and verified it failed before the fix and succeeds now.

aajtodd (approved):

BREAKING-CHANGE.tmp.md
```diff
@@ -0,0 +1,91 @@
...
+**After the change** the `getObject` example prints:
+
+```
+Response length: 1024
+Content-Encoding: gzip
+```
+
+# How to migrate
+
+If you are using the default OkHttp engine and **do not want** to decompress response bodies, the no change is necessary.
```
lauzadis, line 68: typo: `the no change is necessary` -> `then no change is necessary

lauzadis (approved):

PR #1053, opened 2024-03-14
misc: remove "Content-Length" headers for `aws-chunked` requests

lauzadis:
## Description of changes

### DIFF
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/internal/AwsChunkedUtil.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/internal/AwsChunkedUtil.kt
@@ -66,7 +66,10 @@ public val AwsSigningConfig.useAwsChunkedEncoding: Boolean
 public fun HttpRequestBuilder.setAwsChunkedHeaders() {
     headers.append("Content-Encoding", "aws-chunked")
     headers["Transfer-Encoding"] = "chunked"
-    headers["X-Amz-Decoded-Content-Length"] = body.contentLength!!.toString()
+    headers["X-Amz-Decoded-Content-Length"] = checkNotNull(headers["Content-Length"] ?: body.contentLength?.toString()) {
+        "Expected \"Content-Length\" header or `body.contentLength` to be set for aws-chunked"
+    }
+    headers.remove("Content-Length") // Any precalculated Content-Length is no longer correct after we chunk the body up
 }
 
 /**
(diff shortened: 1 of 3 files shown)


runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/internal/AwsChunkedUtil.kt
```diff
@@ -66,7 +66,10 @@ public val AwsSigningConfig.useAwsChunkedEncoding: Boolean
 public fun HttpRequestBuilder.setAwsChunkedHeaders() {
     headers.append("Content-Encoding", "aws-chunked")
     headers["Transfer-Encoding"] = "chunked"
-    headers["X-Amz-Decoded-Content-Length"] = body.contentLength!!.toString()
+    headers["X-Amz-Decoded-Content-Length"] = checkNotNull(headers["Content-Length"] ?: body.contentLength?.toString()) {
+        "Expected \"Content-Length\" header or `body.contentLength` to be set for aws-chunked"
+    }
```
ianbotsf, line 71: **Nit**: `checkNotNull` throws `IllegalStateException` in the null case. We should probably throw `ClientException` or some subclass of that.

runtime/auth/http-auth-aws/common/test/aws/smithy/kotlin/runtime/http/auth/AwsHttpSignerTestBase.kt
ianbotsf, line 1: **Nit**: We should probably add another test here which exercises the logic when the `Content-Length` header is explicitly set.

ianbotsf (approved):
Minor issues. Fix and ship.

runtime/auth/http-auth-aws/common/test/aws/smithy/kotlin/runtime/http/auth/AwsHttpSignerTestBase.kt
lauzadis: The Content-Length header is already explicitly set before signing here in `buildOperation`: https://github.com/smithy-lang/smithy-kotlin/pull/1053/files#diff-d065def50baec1c8bae8f11c7c9930c93cdcc815cc16c97a00276477b4a95db8L70

ianbotsf: Excellent, never mind then!

PR #1063, opened 2024-04-08
fix: correctly handle async cancellation of call context in OkHttp engine

### LINKED ISSUE #1061: IllegalStateException "Unbalanced enter/exit" in okio AsyncTimeout
## Describe the bug

Running lots of SQS client calls in parallel throws `java.lang.IllegalStateException at AsyncTimeout.kt:58` when canceling the scope. Follow-up on my last few comments in #962 .

## Expected Behavior

Exception not thrown

## Current Behavior

Exception thrown :)

## Steps to Reproduce

Start with an empty `c7gd.xlarge` instance running [`ami-0c5789c17ae99a2fa`](https://eu-west-1.console.aws.amazon.com/ec2/home?region=eu-west-1#Images:visibility=public-images;imageId=ami-0c5789c17ae99a2fa)

Install java and docker:

```bash
sudo apt install default-jdk
sudo snap install docker
sudo groupadd docker
sudo usermod -aG docker $USER
```

(Reboot the ec2 instance at this point)

Clone the example project and run the tests:

```
git clone https://github.com/madisp/aws-kotlin-threads.git
cd aws-kotlin-threads
./gradlew test
```

## Possible Solution

I _think_ the issue is the `Job#invokeOnCompletion` call here - https://github.com/smithy-lang/smithy-kotlin/blob/5dddc09697d2dfc2c883f7a963392428e7338b8f/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt#L54. The completion handler will probably be called on the dispatcher thread of whoever is doing the cancel. `ResponseBody.close` will try to consume any unconsumed bytes and this can lead to multiple threads working with the same `ResponseBody`, correctly triggering a precondition check failure in `AsyncTimeout`.

## Context

I'm using `aws-sdk-kotlin` in a backend project and was experimenting with the limits of SQS clients in a single process.

## Your Environment

* Smithy Kotlin version used: 1.1.1
* Platform (JVM/JS/Native): JVM, Linux ARMv8
* Operating System and version: Ubuntu 22.04 LTS Jammy running on `c7gd.xlarge`.

ianbotsf:
## Issue \#

Resolves #1061 

## Description of changes

Update OkHttp engine call context cancellation to first cancel the call before attempting to close any response body.

### DIFF
--- a/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
@@ -51,6 +51,7 @@ public class OkHttpEngine(
         val engineResponse = mapOkHttpExceptions { engineCall.executeAsync() }
 
         callContext.job.invokeOnCompletion {
+            engineCall.cancel()
             engineResponse.body.close()
         }
 
--- a/runtime/protocol/http-client-engines/test-suite/common/test/aws/smithy/kotlin/runtime/http/test/AsyncStressTest.kt
+++ b/runtime/protocol/http-client-engines/test-suite/common/test/aws/smithy/kotlin/runtime/http/test/AsyncStressTest.kt
@@ -87,4 +87,45 @@ class AsyncStressTest : AbstractEngineTest() {
             assertEquals(engineJobsBefore.size, engineJobsAfter.size, message)
         }
     }
+
+    @Test
+    fun testJobCancellation() = testEngines {
+        // https://github.com/smithy-lang/smithy-kotlin/issues/1061
+
+        test { _, client ->
+            val req = HttpRequest {
+                testSetup()
+                url.path.decoded = "slow"
+            }
+
+            // Expect CancellationException because we're cancelling
+            assertFailsWith<CancellationException> {
+                coroutineScope {
+                    val parentScope = this
+
+                    println("Invoking call on ctx $coroutineContext")
+                    val call = client.call(req)
+
+                    val bytes = async {
+                        delay(100.milliseconds)
+                        println("Body of type ${call.response.body} on ctx $coroutineContext")
+                        try {
+                            call.response.body.readAll()
+                        } catch (e: Throwable) {
+                            // IllegalStateException: "Unbalanced enter/exit" will be thrown if body closed improperly
+                            assertIsNot<IllegalStateException>(e)
+                            null
+                        }
+                    }
+
+                    val cancellation = async {
+                        delay(400.milliseconds)
+                        parentScope.cancel("Cancelling!")
+                    }
+
+                    awaitAll(bytes, cancellation)
+                }
+            }
+        }
+    }
 }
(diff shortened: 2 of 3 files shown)


runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
```diff
@@ -51,6 +51,7 @@ public class OkHttpEngine(
         val engineResponse = mapOkHttpExceptions { engineCall.executeAsync() }
 
         callContext.job.invokeOnCompletion {
+            engineCall.cancel()
```
aajtodd, line 54: **correctness**: I'm not sure we want to trigger a cancellation on _every_ invocation including successful ones. This may trigger [other behaviors/paths](https://github.com/square/okhttp/blob/5f008f382d67c79a990b31d4baebbdbc24365e64/okhttp/src/main/kotlin/okhttp3/internal/connection/RealCall.kt#L126) in okhttp ([including lifecycle methods on interceptors](https://github.com/square/okhttp/blob/5f008f382d67c79a990b31d4baebbdbc24365e64/okhttp/src/main/kotlin/okhttp3/internal/connection/RealCall.kt#L138-L143)). 


What we probably want is to know if there was a response body and if it was consumed or not. If not then cancel and close otherwise we don't need to do anything. 

Alternatively we have the right hooks in place on [HttpCall](https://github.com/smithy-lang/smithy-kotlin/blob/6e221cc0f7a31622cbadd4559f22627f4d396106/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HttpCall.kt#L69). We could add an `internal open fun cleanup() { }` that we [override](https://github.com/smithy-lang/smithy-kotlin/blob/b866db4ed13bc1b37de99db4463a26f07a4d3737/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpUtils.kt#L185) to invoke `engineResponse.body.close()` on the same thread that the execution is happening on.

aajtodd (changes requested):

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
ianbotsf: Yeah that makes sense. Let me try modifying `HttpCall` first and see how that works. Thanks!

ianbotsf: Discussed offline and decided to try more intelligent `invokeOnCompletion` logic instead.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
```diff
@@ -50,15 +50,16 @@ public class OkHttpEngine(
...
-        }
-
         val response = engineResponse.toSdkResponse()
         val requestTime = Instant.fromEpochMilliseconds(engineResponse.sentRequestAtMillis)
         val responseTime = Instant.fromEpochMilliseconds(engineResponse.receivedResponseAtMillis)
 
-        return OkHttpCall(request, response, requestTime, responseTime, callContext, engineCall)
+        return OkHttpCall(request, response, requestTime, responseTime, callContext, engineCall).also { call ->
+            callContext.job.invokeOnCompletion { error ->
+                if (error != null) call.cancelInFlight()
```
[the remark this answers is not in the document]
ianbotsf: Yes we definitely want to cancel when we get a `CancellationException`. We're bridging two contexts here so we want to ensure resources belonging to the engine are cleaned up when the call context is cancelled. I'll leave a comment to that effect for future readers.

PR #1066, opened 2024-04-12
fix: lazily resolve proxy environment variables

lauzadis:
This PR updates the `EnvironmentProxySelector` to lazily resolve environment variables rather than immediately parsing them during initialization

## Issue \#

Resolves https://github.com/awslabs/aws-sdk-kotlin/issues/1281

## Description of changes


ianbotsf (approved):
**Nit**: Can we add a new test verifying that junk values in the vars/props are ignored when overriding the proxy selector in HTTP client config?

lauzadis:
@ianbotsf I added a test but it required making `EnvironmentProxySelector` `public` with an `@InternalApi` annotation because the tests are in a separate module, do you think it's worth keeping this way?

0marperez (approved):

PR #1067, opened 2024-04-16
fix: apply clock skew interceptor to clients created via `invoke`

ianbotsf:
## Issue \#

Addresses https://github.com/awslabs/aws-sdk-kotlin/issues/1211

## Description of changes

This change supports adding the clock skew interceptor to clients created via `invoke` (as opposed to via `fromEnvironment`) by creating a new `AbstractSdkClientFactory` superclass with a _non-_`suspend` `finalizeConfig` method and adjusting codegen accordingly.

**Companion PR**: https://github.com/awslabs/aws-sdk-kotlin/pull/1284

### DIFF
--- a/runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/AbstractSdkClientFactory.kt
+++ b/runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/AbstractSdkClientFactory.kt
@@ -1,0 +1,23 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.client
+
+public abstract class AbstractSdkClientFactory<
+    TConfig : SdkClientConfig,
+    TConfigBuilder : SdkClientConfig.Builder<TConfig>,
+    TClient : SdkClient,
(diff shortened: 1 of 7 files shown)


runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/AbstractSdkClientFactory.kt
```diff
@@ -0,0 +1,23 @@
...
+
+    /**
+     * Inject any client-specific config
+     */
+    protected open fun finalizeConfig(builder: TClientBuilder) { }
+
+    public override operator fun invoke(block: TConfigBuilder.() -> Unit): TClient = builder().apply {
+        config.apply(block)
+        finalizeConfig(this)
+    }.build()
```
lauzadis, line 22: Invoking an abstract class looks a little odd since they aren't meant to be instantiated. I don't have any suggestion here, it just stands out to me

ianbotsf: This isn't quite invoking an abstract class. It's an operator defined on an abstract class which will only be callable on a concrete instance. No one's ever going to call `AbstractSdkClientFactory.invoke { … }` or `AbstractSdkClientFactory().invoke { … }`—both are compile-time errors.

This simply enables classes which inherit `AbstractSdkClientFactory` (namely, service client companion objects) to be invoked (e.g., `S3Client.invoke { … }`).

Issue #1068, opened 2024-04-18
Smithy is not compatible with OkHttp 5.0.0-alpha.14


mattcreaser:
## Describe the bug
Smithy has a dependency on OkHttp 5.0.0-alpha.11. OkHttp has [made breaking changes in their newer Alpha versions](https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-500-alpha14), which is causing crashes for downstream consumers of Smithy.

## Expected Behavior
Application using Smithy should be able to run with latest OkHttp version.

## Current Behavior
```
java.lang.NoClassDefFoundError: Failed resolution of: Lokhttp3/JvmCallExtensionsKt;
                                                                                                    	at aws.smithy.kotlin.runtime.http.engine.okhttp.OkHttpEngine.roundTrip(OkHttpEngine.kt:51)
                                                                                                    	at aws.smithy.kotlin.runtime.http.engine.internal.ManagedHttpClientEngine.roundTrip(Unknown Source:2)
                                                                                                    	at aws.smithy.kotlin.runtime.http.SdkHttpClient$executeWithCallContext$2.invokeSuspend(SdkHttpClient.kt:44)
                                                                                                    	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                                                                                                    	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:585)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:802)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:706)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:693)
                                                                                                    	Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@ab2cfdd, Dispatchers.Default]
[...]

PR #1070, opened 2024-04-18
misc: upgrade to OkHttp 5.0.0-alpha.14, Okio 3.9.0, Kotlin 1.9.23

lauzadis:
Upgrade to the latest versions of OkHttp, Okio, and Kotlin.

Changes include needing to opt-in to `ExperimentalCoroutinesApi` for the `executeAsync` call and no longer being able to implement `BufferedSink` because it's now a `sealed interface`. 

We can only use `Buffer` in our tests now, [which is hardcoded to always open](https://github.com/square/okio/blob/e2ac4c05093e166bcb2e56fb8f834fc4f03d299b/okio/src/jvmMain/kotlin/okio/Buffer.kt#L504), so the `isClosed` checks are no longer possible to perform.
## Issue \#

Related to https://github.com/smithy-lang/smithy-kotlin/issues/1068

## Description of changes

### DIFF
--- a/.changes/8e87c6dd-5138-4b79-a0c2-255ecf31898b.json
+++ b/.changes/8e87c6dd-5138-4b79-a0c2-255ecf31898b.json
@@ -1,0 +1,6 @@
+{
+    "id": "8e87c6dd-5138-4b79-a0c2-255ecf31898b",
+    "type": "misc",
+    "description": "Upgrade to latest versions of OkHttp, Okio, Kotlin",
+    "requiresMinorVersionBump": true
+}
--- a/.changes/8e87c6dd-5138-4b79-a0c2-255ecf31898b.json
+++ b/.changes/8e87c6dd-5138-4b79-a0c2-255ecf31898b.json
@@ -1,6 +1,6 @@
 {
     "id": "8e87c6dd-5138-4b79-a0c2-255ecf31898b",
     "type": "misc",
-    "description": "Upgrade to latest versions of OkHttp, Okio, Kotlin",
+    "description": "⚠️ **IMPORTANT**: Upgrade to latest versions of OkHttp, Okio, Kotlin",
     "requiresMinorVersionBump": true
 }
--- a/breaking-change-discussion-draft.md
+++ b/breaking-change-discussion-draft.md
@@ -1,0 +1,21 @@
+An upcoming release of the **AWS SDK for Kotlin** will upgrade our dependency on OkHttp to **v5.0.0-alpha.14**. This version contains a number of breaking changes and earlier versions of the SDK are not compatible with it.
+
+# Release date
+
+This feature will ship with the **v1.2.0** release planned for **4/22/2024**.
+
+# What's changing
+
+The Kotlin SDK will be upgrading its dependency on OkHttp to v5.0.0-alpha.14. If you don't directly depend on OkHttp alongside the AWS SDK for Kotlin, this change should not affect you.
+
+If you _do_ have direct dependencies on OkHttp and the AWS SDK for Kotlin, you will need to upgrade to OkHttp **v5.0.0-alpha.14** along with your upgrade to AWS SDK for Kotlin **v1.2.0**.
+
+# How to migrate
+
+1. Upgrade your AWS SDK for Kotlin dependency to **v.1.2.0**.
+2. Upgrade your OkHttp dependency to **v5.0.0-alpha.14**.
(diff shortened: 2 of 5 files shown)


runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
```diff
@@ -48,6 +50,8 @@ public class OkHttpEngine(
 
         val engineRequest = request.toOkHttpRequest(context, callContext, metrics)
         val engineCall = client.newCall(engineRequest)
+
+        @OptIn(ExperimentalCoroutinesApi::class)
         val engineResponse = mapOkHttpExceptions { engineCall.executeAsync() }
```
ianbotsf, line 55: From [OkHttp's changelog](https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-500-alpha13):
> Breaking: Tag unstable new APIs as `@ExperimentalOkHttpApi`. We intend to release OkHttp 5.0 without stabilizing these new APIs first.
> 
> Do not use these experimental APIs in modules that may be executed using a version of OkHttp different from the version that the module was compiled with. Do not use them in published libraries. Do not use them if you aren't willing to track changes to them.

They're explicitly calling this out now as an unstable API which may change after 5.0 is fully released. We should carefully talk over the alternatives here.

ianbotsf (changes requested):
We should merge this into **v1.2**, not **main**.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
lauzadis: I think `ExperimentalOkHttpApi` is different from `ExperimentalCoroutinesApi`

ianbotsf: 🤦‍♂️Oh you're right, I stopped reading at "Experimental". My mistake.

.changes/8e87c6dd-5138-4b79-a0c2-255ecf31898b.json
```diff
@@ -0,0 +1,6 @@
+{
+    "id": "8e87c6dd-5138-4b79-a0c2-255ecf31898b",
+    "type": "misc",
+    "description": "Upgrade to latest versions of OkHttp, Okio, Kotlin",
```
ianbotsf, line 4: **Nit**: Prefix with `⚠️ **IMPORTANT**: `. Our **kat** tool does this automatically but if we manually tweak the file we have to manually prefix the description.

ianbotsf, line 4: **Question**: Is there value in creating a breaking change discussion for this?

lauzadis: I don't really think so. The OkHttp breaking changes are broad and we can't easily write a migration guide for each possible case.

0marperez: I think it's worth mentioning it and maybe not getting into too much detail about it. Like Matas said the changes are broad. This will break customers not using the latest ok http version?

0marperez (approved):

breaking-change-discussion-draft.md
```diff
@@ -0,0 +1,23 @@
+An upcoming release of the **AWS SDK for Kotlin** will upgrade our dependency on OkHttp to **v5.0.0-alpha.14**. This version contains a number of breaking changes and earlier versions of the SDK are not compatible with it.
+
+# Release date
+
+This feature will ship with the **v1.2.0** release planned for **4/22/2024**.
+
+# What's changing
+
+The Kotlin SDK will be upgrading its dependency on OkHttp to v5.0.0-alpha.14. If you don't directly depend on OkHttp alongside the AWS SDK for Kotlin, this change should not affect you.
```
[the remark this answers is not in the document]
lauzadis: You're right, I didn't cover this possibility. I put up a new commit with some changes to the description / mitigation steps.

ianbotsf (approved):

PR #1077, opened 2024-04-24
 feat: artifact size metrics

0marperez:
## Issue \#
N/A

## Description of changes
Added artifact size metrics plugin and misc changes to support the addition.

Note: The plugin currently runs because of the database being bootstrapped in my personal AWS account.


ianbotsf (approved):
See comments on https://github.com/awslabs/aws-crt-kotlin/pull/102

tests/codegen/nullability-tests/build.gradle.kts
```diff
@@ -23,6 +23,10 @@ tasks.generateSmithyProjections {
     smithyBuildConfigs.set(files("smithy-build.json"))
 }
 
+tasks.kotlinSourcesJar {
+    dependsOn(tasks.generateSmithyProjections)
+}
```
lauzadis, line 28: question: You said these `dependsOn` additions were required to fix your `ShapeExt.kt:35:80 Unresolved reference.` errors? That seems odd...

build.gradle.kts
```diff
@@ -24,6 +24,13 @@ plugins {
     // since build-plugins also has <some> version in its dependency closure
     alias(libs.plugins.kotlin.multiplatform) apply false
     alias(libs.plugins.kotlin.jvm) apply false
+    alias(libs.plugins.aws.kotlin.repo.tools.artifactsizemetrics)
+}
+
+artifactSizeMetrics {
+    artifactPrefixes = setOf(":codegen", ":runtime")
```
lauzadis, line 31: question: Do we care about the codegen artifact sizes? This is not something that customers are affected by

lauzadis (approved):

tests/codegen/nullability-tests/build.gradle.kts
0marperez: The `ShapeExt.kt:35:80 Unresolved reference` errors were fixed by: https://github.com/smithy-lang/smithy-kotlin/blob/423dc01680127b6552f140716aa3429b64edc1af/.github/workflows/artifact-size-metrics.yml#L21-L26

These `dependsOn` were fixes for these types of errors that I started getting afterwards:
```
A problem was found with the configuration of task ':tests:codegen:nullability-tests:generateSmithyProjections' (type 'SmithyBuildTask').
  - Gradle detected a problem with the following location: '/.../smithy-kotlin/tests/codegen/nullability-tests/build/smithyprojections/nullability-tests'.
    
    Reason: Task ':tests:codegen:nullability-tests:kotlinSourcesJar' uses this output of task ':tests:codegen:nullability-tests:generateSmithyProjections' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
```

build.gradle.kts
0marperez: I think we can get rid of them 👍

tests/codegen/nullability-tests/build.gradle.kts
lauzadis: Ah ok. Was that a warning or an error? We've had them for a long time if I recall correctly but they were never errors. Maybe on a newer version of Gradle?

0marperez: Yeah, they were errors. I'm not exactly sure what would changed. I was getting them in my machine as well

ianbotsf (approved):

PR #1086, opened 2024-05-01
fix: sparse struct serializer

0marperez:
## Issue \#
N/A

## Description of changes
Fix rendering of sparse structure serializers

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/SerializeStructGenerator.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/SerializeStructGenerator.kt
@@ -461,15 +461,10 @@ open class SerializeStructGenerator(
         val (keyName, valueName) = keyValueNames(nestingLevel)
         val keyValue = keyValue(keyShape, keyName)
 
-        when (isSparse) {
-            true -> {
-                writer.withBlock("$containerName$listMemberName.forEach { ($keyName, $valueName) ->", "}") {
-                    writer.withBlock("if ($valueName != null) {", "} else entry($keyValue, null as String?)") {
-                        writer.write("entry($keyValue, $valueName$enumPostfix)")
-                    }
-                }
+        writer.withBlock("$containerName$listMemberName.forEach { ($keyName, $valueName) ->", "}") {
+            writer.wrapBlockIf(isSparse, "if ($valueName != null) {", "} else entry($keyValue, null as String?)") {
+                writer.write("entry($keyValue, $valueName$enumPostfix)")
             }
-            false -> writer.write("$containerName$listMemberName.forEach { ($keyName, $valueName) -> entry($keyValue, $valueName$enumPostfix) }")
         }
     }
 
@@ -485,19 +480,9 @@ open class SerializeStructGenerator(
         val (keyName, valueName) = keyValueNames(nestingLevel)
         val keyValue = keyValue(keyShape, keyName)
 
-        when (isSparse) {
-            true -> {
-                writer.withBlock("$containerName$listMemberName.forEach { ($keyName, $valueName) ->", "}") {
-                    writer.withBlock("if ($valueName != null) {", "} else entry($keyValue, null as String?)") {
-                        writer.write("entry($keyValue, $valueName.#T())", RuntimeTypes.Core.Text.Encoding.encodeBase64String)
-                    }
-                }
-            }
-            false -> {
-                writer.write(
-                    "$containerName$listMemberName.forEach { ($keyName, $valueName) -> entry($keyValue, $valueName.#T()) }",
-                    RuntimeTypes.Core.Text.Encoding.encodeBase64String,
-                )
+        writer.withBlock("$containerName$listMemberName.forEach { ($keyName, $valueName) ->", "}") {
+            writer.wrapBlockIf(isSparse, "if ($valueName != null) {", "} else entry($keyValue, null as String?)") {
+                writer.write("entry($keyValue, $valueName.#T())", RuntimeTypes.Core.Text.Encoding.encodeBase64String)
             }
         }
     }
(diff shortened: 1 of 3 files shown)


ianbotsf (changes requested):
**Correctness**: Needs unit tests

ianbotsf (approved):
Looks good!

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/SerializeStructGenerator.kt
```diff
@@ -447,13 +455,22 @@ open class SerializeStructGenerator(
...
         val containerName = if (nestingLevel == 0) "input." else ""
         val enumPostfix = if (elementShape.isEnum) ".value" else ""
         val (keyName, valueName) = keyValueNames(nestingLevel)
         val keyValue = keyValue(keyShape, keyName)
 
-        writer.write("$containerName$listMemberName.forEach { ($keyName, $valueName) -> entry($keyValue, $valueName$enumPostfix) }")
+        when (isSparse) {
+            true -> {
+                writer.withBlock("$containerName$listMemberName.forEach { ($keyName, $valueName) ->", "}") {
+                    writer.withBlock("if (value != null) {", "} else entry(key, null as String?)") {
```
lauzadis, line 467: correctness: should `value` be `$valueName` and `key` be `$keyName`?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/SerializeStructGenerator.kt
```diff
@@ -463,15 +480,26 @@ open class SerializeStructGenerator(
...
         val keyValue = keyValue(keyShape, keyName)
 
-        writer.write(
-            "$containerName$listMemberName.forEach { ($keyName, $valueName) -> entry($keyValue, $valueName.#T()) }",
-            RuntimeTypes.Core.Text.Encoding.encodeBase64String,
-        )
+        when (isSparse) {
+            true -> {
+                writer.withBlock("$containerName$listMemberName.forEach { ($keyName, $valueName) ->", "}") {
+                    writer.withBlock("if (value != null) {", "} else entry(key, null as String?)") {
```
lauzadis, line 491: Same comment here about `keyName`, `valueName`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/SerializeStructGenerator.kt
```diff
@@ -463,15 +480,26 @@ open class SerializeStructGenerator(
...
+                    }
+                }
+            }
+            false -> {
+                writer.write(
+                    "$containerName$listMemberName.forEach { ($keyName, $valueName) -> entry($keyValue, $valueName.#T()) }",
+                    RuntimeTypes.Core.Text.Encoding.encodeBase64String,
+                )
+            }
+        }
```
lauzadis, line 502: possible simplification using `wrapBlockIf`:

```kt
writer.withBlock("$containerName$listMemberName.forEach { ($keyName, $valueName) ->", "}") {
    writer.wrapBlockIf(isSparse, "if ($valueName != null) {", "} else entry($keyName, null as String?)") {
        writer.write("entry($keyValue, $valueName.#T())", RuntimeTypes.Core.Text.Encoding.encodeBase64String)
    } 
}
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/SerializeStructGenerator.kt
lauzadis, line 491: question: Here we are assuming the value will always be a string (`null as String?`), I think that works because serializers will handle a `null` entry of any type the same way?

lauzadis (changes requested):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/SerializeStructGenerator.kt
0marperez: It's a good simplification but it would cause codegen changes to the serializers that are not sparse when not necessary

lauzadis: I'm not seeing the codegen change, unless you are you talking about the newline?

Non-sparse currently:
```kt
$containerName$#listMemberName.forEach { ($keyName, $valueName) -> entry($keyValue, $valueName.encodeBase64String()) }
```

Non-sparse with my suggestio:
```kt
$containerName$listMemberName.forEach { ($keyName, $valueName) ->
    entry($keyValue, $valueName.encodeBase64String())
}
```

0marperez: Yeah I'm talking about the new line

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/SerializeStructGenerator.kt
0marperez: >question: Here we are assuming the value will always be a string (null as String?), I think that works because serializers will handle a null entry of any type the same way?

Yes that's right, there should be a way to use the correct value type but it's not simple and the results would be the same

0marperez: Actually that's only true for the `JsonSerializer` because of the `XmlSerializer` we need to use `as String?`
https://github.com/smithy-lang/smithy-kotlin/blob/7d2012af6077ee1619b6ff66f73edba53081f3e0/runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/XmlSerializer.kt#L231-L263
`null.toString()` returns the string `"null"` according to the docs

lauzadis (approved):

PR #1096, opened 2024-05-23
feat: business metrics

0marperez:
## Issue \#

N/A

## Description of changes
This change tracks business metrics in the execution context of an operation. To do so:

- The `DefaultEndpointProviderGenerator` was modified to include endpoint business metrics in the endpoint attributes
- This change required modifying the `DefaultEndpointProviderTestGenerator` to be tolerant of business metrics attributes
- A section was declared in `EndpointResolverAdapterGenerator` to add endpoint business metrics to the execution context if business metrics exist in the endpoint attributes
- Business metrics are now added to the execution context directly in various locations
- Required runtime types to emit business metrics were added


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/DefaultEndpointProviderGenerator.kt
```diff
@@ -170,9 +185,13 @@ class DefaultEndpointProviderGenerator(
         withConditions(rule.conditions) {
             writer.withBlock("return #T(", ")", RuntimeTypes.SmithyClient.Endpoints.Endpoint) {
                 writeInline("#T.parse(", RuntimeTypes.Core.Net.Url.Url)
+                renderingEndpointUrl = true
                 renderExpression(rule.endpoint.url)
+                renderingEndpointUrl = false
```
ianbotsf, line 190: **Style**: Doing this kind of mutable state tracking via class-level fields is difficult to follow and error-prone. The visitor pattern allows using a return type to get information from within the tree but it looks like we currently use `ExpressionVisitor<Unit>` and skip returning anything.

My suggestion would be to refactor this to keep a context result in the visitor:

```kotlin
data class EndpointInfo(val params: MutableSet<String>) {
    companion object {
        val Empty = EndpointInfo(params = mutableSetOf())
    }

    operator fun plus(other: EndpointInfo) = EndpointInfo(
        params = this.params + other.params,
    )
}

class ExpressionGenerator(…): ExpressionVisitor<EndpointInfo>, … {
    …

    override fun visitRef(reference: Reference): EndpointInfo {
        val referenceName = reference.name.defaultName()
        val isParamReference = isParamRef(reference)

        if (isParamReference) {
            writer.writeInline("params.")
        }
        writer.writeInline(referenceName)

        return if (isParamReference) {
            EndpointInfo(params = mutableSetOf(referenceName))
        } else {
            EndpointInfo.Empty
        }
    }

    …
}
```

Then you can modify the expression renderer to return/use the information:

```
fun interface ExpressionRenderer {
    fun renderExpression(expr: Expression): EndpointInfo
}

class DefaultEndpointProviderGenerator(…): ExpressionRenderer {
    …

    override fun renderExpression(expr: Expression): EndpointInfo = expr.accept(expressionGenerator)

    private fun renderEndpointRule(rule: EndpointRule) {
        withConditions(rule.conditions) {
            writer.withBlock("return #T(", ")", RuntimeTypes.SmithyClient.Endpoints.Endpoint) {
                writeInline("#T.parse(", RuntimeTypes.Core.Net.Url.Url)
                val endpointInfo = renderExpression(rule.endpoint.url)
                write("),")

                val hasAccountIdBasedEndpoint = "accountId" in endpointInfo.params
                val hasServiceEndpointOverride = "endpoint" in endpointInfo.params
                val needAdditionalEndpointProperties = hasAccountIdBasedEndpoint || hasServiceEndpointOverride

                if (rule.endpoint.headers.isNotEmpty()) { … }

                if (rule.endpoint.properties.isNotEmpty() || needAdditionalEndpointProperties) {
                    withBlock("attributes = #T {", "},", RuntimeTypes.Core.Collections.attributesOf) {
                        rule.endpoint.properties.entries.forEach { (k, v) -> … }

                        if (hasAccountIdBasedEndpoint) {
                            writer.write("#T to true", RuntimeTypes.Core.BusinessMetrics.accountIdBasedEndPoint)
                        }
                        if (hasServiceEndpointOverride) {
                            writer.write("#T to true", RuntimeTypes.Core.BusinessMetrics.serviceEndpointOverride)
                        }
                    }
                }
            }
        }
    }

    …
}


```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/DefaultEndpointProviderTestGenerator.kt
```diff
@@ -61,6 +61,43 @@ class DefaultEndpointProviderTestGenerator(
...
+                        "if (actual.attributes.contains(#T)) updatedAttributes[#T] = true",
+                        RuntimeTypes.Core.BusinessMetrics.accountIdBasedEndPoint,
+                        RuntimeTypes.Core.BusinessMetrics.accountIdBasedEndPoint,
+                    )
+                    writer.write(
+                        "val newExpected = #T(expected.uri, expected.headers, updatedAttributes)",
+                        RuntimeTypes.SmithyClient.Endpoints.Endpoint,
+                    )
+                    writer.write("assertEquals(newExpected, actual)")
+                }
```
ianbotsf, line 97: **Correctness**: This appears to be setting the _expected_ attributes based on the _actual_ attributes, which is backwards. If the actual attributes erroneously omitted, for example, account ID then this test wouldn't catch it.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionInterceptor.kt
```diff
@@ -42,7 +44,9 @@ public class RequestCompressionInterceptor(
         }
 
         return if (algorithm != null && (context.protocolRequest.body.isStreaming || payloadSizeBytes?.let { it >= compressionThresholdBytes } == true)) {
-            algorithm.compressRequest(context.protocolRequest)
+            val request = algorithm.compressRequest(context.protocolRequest)
+            context.executionContext.emitBusinessMetric(BusinessMetrics.GZIP_REQUEST_COMPRESSION)
+            request
```
ianbotsf, line 49: **Style**: You can avoid this kind of temporary-storage-for-control-flow pattern by using [`also`](https://kotlinlang.org/docs/scope-functions.html#also):

```kotlin
algorithm.compressRequest(context.protocolRequest).also {
    context.executionContext.emitBusinessMetric(BusinessMetrics.GZIP_REQUEST_COMPRESSION)
}
```

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/businessmetrics/BusinessMetricsUtils.kt
```diff
@@ -0,0 +1,55 @@
...
+ * If an endpoint is account ID based
+ */
+@InternalApi
+public val accountIdBasedEndPoint: AttributeKey<Boolean> = AttributeKey("aws.smithy.kotlin#AccountIdBasedEndpoint")
+
+/**
+ * If an endpoint is service endpoint override based
+ */
+@InternalApi
+public val serviceEndpointOverride: AttributeKey<Boolean> = AttributeKey("aws.smithy.kotlin#ServiceEndpointOverride")
```
ianbotsf, line 28: **Question**: I'm not sure I see the value of dedicated attribute keys for these. At the point we have access to the these attributes, we _could_ be calling `emitBusinessMetrics` already, right?

ianbotsf (dismissed):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/businessmetrics/BusinessMetricsUtils.kt
```diff
@@ -0,0 +1,72 @@
...
+ * Keeps track of all business metrics along an operations execution
+ */
+@InternalApi
+public val businessMetrics: AttributeKey<MutableSet<String>> = AttributeKey("aws.sdk.kotlin#BusinessMetrics")
+
+/**
+ * The account ID in an account ID based endpoint
+ */
+@InternalApi
+public val accountIdBasedEndPointAccountId: AttributeKey<String> = AttributeKey("aws.smithy.kotlin#AccountIdBasedEndpointAccountId")
```
lauzadis, line 22: naming: `AccountIdBasedEndpoint`

- No camel casing
- `EndPoint` -> `Endpoint`
- Remove second `AccountId`

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/businessmetrics/BusinessMetricsUtils.kt
```diff
@@ -0,0 +1,72 @@
...
+ */
+@InternalApi
+public fun ExecutionContext.containsBusinessMetric(metric: BusinessMetrics): Boolean =
+    (this.attributes.contains(businessMetrics)) && this.attributes[businessMetrics].contains(metric.identifier)
+
+/**
+ * All the valid business metrics along with their identifiers
+ */
+@InternalApi
+public enum class BusinessMetrics(public val identifier: String) {
```
lauzadis, line 63: question: Are these enum values correct? I see something different in the internal spec. "J" is `S3_EXPRESS_BUCKET`, "L" is `GZIP_REQUEST_COMPRESSION`, etc.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/businessmetrics/BusinessMetricsUtils.kt
```diff
@@ -0,0 +1,72 @@
...
+import aws.smithy.kotlin.runtime.InternalApi
+import aws.smithy.kotlin.runtime.collections.AttributeKey
+import aws.smithy.kotlin.runtime.collections.get
+import aws.smithy.kotlin.runtime.operation.ExecutionContext
+
+/**
+ * Keeps track of all business metrics along an operations execution
+ */
+@InternalApi
+public val businessMetrics: AttributeKey<MutableSet<String>> = AttributeKey("aws.sdk.kotlin#BusinessMetrics")
```
lauzadis, line 16: naming: AttributeKeys are not usually camel-cased, `BusinessMetrics`

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/businessmetrics/BusinessMetricsUtils.kt
```diff
@@ -0,0 +1,72 @@
...
+ * The account ID in an account ID based endpoint
+ */
+@InternalApi
+public val accountIdBasedEndPointAccountId: AttributeKey<String> = AttributeKey("aws.smithy.kotlin#AccountIdBasedEndpointAccountId")
+
+/**
+ * If an endpoint is service endpoint override based
+ */
+@InternalApi
+public val serviceEndpointOverride: AttributeKey<Boolean> = AttributeKey("aws.smithy.kotlin#ServiceEndpointOverride")
```
lauzadis, line 28: naming: `ServiceEndpointOverride`

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/businessmetrics/BusinessMetricsUtils.kt
lauzadis, line 63: naming: `BusinessMetric`:  Enum class names are usually not pluralized,, since an instance of this enum will only ever represent a single metric

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/businessmetrics/BusinessMetricsUtilsTest.kt
```diff
@@ -0,0 +1,32 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.businessmetrics
+
+import aws.smithy.kotlin.runtime.collections.get
+import aws.smithy.kotlin.runtime.operation.ExecutionContext
+import kotlin.test.Test
+import kotlin.test.assertTrue
+
+class BusinessMetricsUtilsTest {
```
lauzadis, line 12: correctness: This is missing tests for removing and containing business metrics

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/DefaultEndpointProviderGenerator.kt
```diff
@@ -185,7 +204,7 @@ class DefaultEndpointProviderGenerator(
                     }
                 }
 
-                if (rule.endpoint.properties.isNotEmpty()) {
+                if (rule.endpoint.properties.isNotEmpty() || needAdditionalEndpointProperties) {
```
lauzadis, line 207: suggestion: `rule.endpoint.properties` are mutable, so instead of creating this new boolean / tracking extra state, you can just add the `accountIdBasedEndpoint` and `serviceEndpointOverride` parameters to the endpoint params before entering this `if` statement.

Something like this:
```kt
rule.endpoint.properties[Identifier.of(RuntimeTypes.Core.BusinessMetrics.accountIdBasedEndPointAccountId.name)] = Literal.of("params.accountId")
rule.endpoint.properties[Identifier.of(RuntimeTypes.Core.BusinessMetrics.serviceEndpointOverride.name)] = Literal.booleanLiteral(true)

if (rule.endpoint.properties.isNotEmpty() { 
    ...
} 
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/DefaultEndpointProviderTestGenerator.kt
```diff
@@ -61,6 +61,43 @@ class DefaultEndpointProviderTestGenerator(
     fun render() {
         writer.addImport("*", namespace = "kotlin.test")
         writer.withBlock("public class #L {", "}", CLASS_NAME) {
+            writer.write("")
+            writer.withBlock(
+                "private fun expectEqualEndpoints(expected: #T, actual: #T) {",
+                "}",
+                RuntimeTypes.SmithyClient.Endpoints.Endpoint,
+                RuntimeTypes.SmithyClient.Endpoints.Endpoint,
```
lauzadis, line 69: simplification: You can remove the duplicated `RuntimeTypes.SmithyClient.Endpoints.Endpoint` by replacing `#T` with `#1T`

Same for all other lines where you duplicate the `#T`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/DefaultEndpointProviderTestGenerator.kt
```diff
@@ -61,6 +61,43 @@ class DefaultEndpointProviderTestGenerator(
...
+            ) {
+                // Remove ONLY business metrics endpoint attributes
```
lauzadis, line 71: question: Was this modification required to get endpoint tests passing? It would be nice if we didn't have to customize the endpoint tests based on business metrics

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/middleware/RetryMiddleware.kt
```diff
@@ -47,6 +51,10 @@ internal class RetryMiddleware<I, O>(
                 withSpan<RetryMiddleware<*, *>, _>("Attempt-$attempt") {
                     if (attempt > 1) {
                         coroutineContext.debug<RetryMiddleware<*, *>> { "retrying request, attempt $attempt" }
+                        when (strategy::class) {
+                            StandardRetryStrategy::class -> modified.context.emitBusinessMetric(BusinessMetrics.RETRY_MODE_STANDARD)
+                            AdaptiveRetryStrategy::class -> modified.context.emitBusinessMetric(BusinessMetrics.RETRY_MODE_ADAPTIVE)
+                        }
```
lauzadis, line 57: correctness: this is only emitting metrics when a request is retried (first attempt failed), but it seems like the SEP wants us to emit the metrics any time the strategy is used (even on the first attempt).

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/InterceptorExecutor.kt
```diff
@@ -328,4 +336,22 @@ internal class InterceptorExecutor<I, O>(
             },
         )
     }
+
+    private fun updateBusinessMetrics(
```
lauzadis, line 340: suggestion: Instead of calling this function after almost every interceptor stage, can it just be added as a new interceptor? It can keep an internal `var request : HttpRequest` that is updated at each stage, and then you can use the `readBeforeX` interceptor hooks to compare the `modifiedRequest`

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/businessmetrics/BusinessMetricsUtils.kt
lauzadis, line 63: Some of these metrics are AWS-specific and don't belong in smithy-kotlin. We spent a lot of time making sure S3 Express was not mentioned in smithy-kotlin. 

Is there any way the AWS-specific enums can be relocated to aws-sdk-kotlin?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/DefaultEndpointProviderTestGenerator.kt
0marperez: Yes, the endpoint tests in the model don't include business metrics in the `expected`. To fix this the model would have to be updated

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/businessmetrics/BusinessMetricsUtils.kt
0marperez: I tried that too but it seems like that would require smithy kotlin have a dependency on the sdk. I think we want to avoid that

lauzadis: Yeah, smithy-kotlin can not depend on aws-sdk-kotlin at all, that's a circular dependency.

Fixing this will certainly involve some rearchitecture. You will probably need to make an `AwsBusinessMetrics` class which implements `BusinessMetrics` so it can be used interchangeably in all the functions / utilities you wrote. That would also probably mean moving away from `enum class`, we can discuss offline

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/businessmetrics/BusinessMetricsUtils.kt
0marperez: Do you mean in the `DefaultEndpointProviderGenerator`? If so I don't think we can, we only have access to the endpoint attributes but not to the execution context attributes

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/businessmetrics/BusinessMetricsUtils.kt
0marperez: Yeah, they're outdated now. Thanks

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/DefaultEndpointProviderGenerator.kt
lauzadis: @0marperez Were you able to take a look at this feedback, do you think it's possible to refactor? I also agree that the boolean toggling stands out and seems error-prone for future implementations

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/InterceptorExecutor.kt
lauzadis: Any thoughts on this suggestion? I think it would be cleaner, `InterceptorExecutor` shouldn't really be responsible for emitting / updating business metrics IMO

lauzadis (approved):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/businessmetrics/BusinessMetricsUtils.kt
```diff
@@ -0,0 +1,77 @@
...
+ */
+public interface BusinessMetric {
+    public val identifier: String
+}
+
+/**
+ * Generic business metrics
+ */
+@InternalApi
+public enum class SmithyBusinessMetric(public override val identifier: String) : BusinessMetric {
```
lauzadis, line 70: We support some features that aren't listed here. Should we add TODOs to add them in the future?  
- `WAITER("B")`
- `PAGINATOR("C")`

0marperez: Yeah, those features were added to the spec after we started the implementation. I think it's a good idea

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/InterceptorExecutor.kt
lauzadis: Currently we're only checking if the URL changed after the `modifyX` interceptors are executed. We could instead add a new interceptor right after in the `readY` phase to check if the URL had changed.

It's not a one-way decision so I'm fine to leave it as it is and refactor later if needed.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/endpoints/DefaultEndpointProviderGenerator.kt
0marperez: I ran into some difficulties while trying to implement this and decided to leave it as is since the comment is marked as `style`. I think it's possible so I'll give it another go

PR #1099, opened 2024-06-05
fix: silently ignore empty proxy host values

### LINKED ISSUE #1098: Check for empty proxy hosts
### Describe the bug

We're seeing the following issue in our logs which I've tracked down to `EnvironmentProxySelector`

`Caused by: aws.smithy.kotlin.runtime.ClientException: Could not parse http.proxyHost="", http.proxyPort="0" into a valid proxy URL`

### Expected behavior

Ideally the `EnvironmentProxySelector` would check the length of the proxy host before attempting to use it, like `DefaultProxySelector` does [here](https://github.com/frohoff/jdk8u-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/share/classes/sun/net/spi/DefaultProxySelector.java#L225)

### Current behavior

If SQS detects an value for `proxy_host`, even if it's empty, it will attempt to use it and the connection will fail.

### Steps to Reproduce

I'm not entirely sure how this is being set in the wild, but we've been able to reproduce it by setting the property before initialising the SQS client

`System.setProperty("http.proxyHost", "")`

### Possible Solution

Check the length of the host value before using the proxy settings

### Context

_No response_

### Smithy-Kotlin version

1.25

### Platform (JVM/JS/Native)

JVM

Android 8/9/10

ianbotsf:
## Issue \#

Closes #1098 

## Description of changes

This changes adds handling for empty/blank proxy host values sourced from environment variables and system properties. The previous behavior was to throw an exception because blank strings couldn't be parsed into hosts.

### DIFF
--- a/runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/engine/EnvironmentProxySelectorTest.kt
+++ b/runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/engine/EnvironmentProxySelectorTest.kt
@@ -49,6 +49,10 @@ class EnvironmentProxySelectorTest {
     private val tests = listOf(
         // no props
         TestCase(ProxyConfig.Direct),
+        TestCase(ProxyConfig.Direct, "http://aws.amazon.com", env = mapOf("http_proxy" to "")),
+        TestCase(ProxyConfig.Direct, "http://aws.amazon.com", env = mapOf("http_proxy" to " ")),
+        TestCase(ProxyConfig.Direct, "http://aws.amazon.com", props = mapOf("http.proxyHost" to "")),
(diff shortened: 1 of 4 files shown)


runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/engine/EnvironmentProxySelectorTest.kt
```diff
@@ -49,6 +49,10 @@ class EnvironmentProxySelectorTest {
     private val tests = listOf(
         // no props
```
lauzadis, line 50: For completeness should we also test a blank string for the `http.nonProxyHosts` property and `NO_PROXY` environment variable?

lauzadis (approved):

PR #1101, opened 2024-06-14
fix: paginators not respecting start key

0marperez:
## Issue \#

aws-sdk-kotlin#1330

## Description of changes

Check if start key is exists before setting it to null

### DIFF
--- a/.changes/15fb7dc6-dee9-4f40-adba-a48183d4f432.json
+++ b/.changes/15fb7dc6-dee9-4f40-adba-a48183d4f432.json
@@ -1,0 +1,8 @@
+{
+    "id": "15fb7dc6-dee9-4f40-adba-a48183d4f432",
+    "type": "bugfix",
+    "description": "Paginators not respecting start key",
+    "issues": [
+        "awslabs/aws-sdk-kotlin#1330"
+    ]
+}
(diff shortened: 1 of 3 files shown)


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
```diff
@@ -148,7 +148,7 @@ class PaginatorGenerator : KotlinIntegration {
                     write("")
                     withBlock("while (hasNextPage) {", "}") {
                         withBlock("val req = initialRequest.copy {", "}") {
-                            write("this.$markerLiteral = cursor")
+                            write("this.$markerLiteral = this.$markerLiteral ?: cursor")
```
lauzadis, line 151: question / correctness: Won't this logic **always** use the `initialRequest.$markerLiteral` if it's not null? How does the cursor get updated to fetch the second page in that case?

I was expecting the fix to be changing L146 to:

```kt
write("var cursor: #F = initialRequest.#L", cursorSymbol, markerLiteral)
```

This will be `null` when the `initialRequest.$markerLiteral` is not set, preserving existing behavior

.changes/15fb7dc6-dee9-4f40-adba-a48183d4f432.json
```diff
@@ -0,0 +1,8 @@
+{
+    "id": "15fb7dc6-dee9-4f40-adba-a48183d4f432",
+    "type": "bugfix",
+    "description": "Paginators not respecting start key",
```
lauzadis, line 4: nit: usually the changelog description should finish the sentence "This PR will..." 

"Fix paginators not respecting start key"

lauzadis (approved):

PR #1108, opened 2024-06-26
misc: upgrade to ktlint v1.3.0

lauzadis:
## Description of changes


runtime/auth/aws-signing-tests/common/src/aws/smithy/kotlin/runtime/auth/awssigning/tests/AwsChunkedTestBase.kt
```diff
@@ -55,10 +55,6 @@ fun AwsChunkedReaderFactory.create(
 abstract class AwsChunkedTestBase(
     val factory: AwsChunkedReaderFactory,
 ) : HasSigner {
-    val CHUNK_SIGNATURE_REGEX = Regex("chunk-signature=[a-zA-Z0-9]{64}") // alphanumeric, length of 64
-    val CHUNK_SIZE_REGEX = Regex("[0-9a-f]+;chunk-signature=") // hexadecimal, any length, immediately followed by the chunk signature
-    val UNSIGNED_CHUNK_SIZE_REGEX = Regex("[0-9a-f]+\r\n")
```
0marperez, line 60: question: could we have kept these values?

0marperez (approved):

runtime/auth/aws-signing-tests/common/src/aws/smithy/kotlin/runtime/auth/awssigning/tests/AwsChunkedTestBase.kt
lauzadis: Yes we could have, but since they were only used once, I didn't really see the point of keeping them. I do think it helps readability so let me replace them. They just need to be made top-level `private val` instead of class members, because they are using SCREAMING_SNAKE_CASE

0marperez: Screaming snake case😆 thank you

PR #1113, opened 2024-07-03
fix: outdated business metric ids

0marperez:
## Issue \#

N/A

## Description of changes

The name of these metrics changed, so this PR updates them


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/businessmetrics/BusinessMetricsUtils.kt
```diff
@@ -74,7 +74,7 @@ public enum class SmithyBusinessMetric(public override val identifier: String) :
     RETRY_MODE_ADAPTIVE("F"),
     GZIP_REQUEST_COMPRESSION("L"),
     PROTOCOL_RPC_V2_CBOR("M"),
-    SERVICE_ENDPOINT_OVERRIDE("N"),
-    ACCOUNT_ID_BASED_ENDPOINT("O"),
-    SIGV4A_SIGNING("P"),
+    ENDPOINT_OVERRIDE("N"),
+    ACCOUNT_ID_ENDPOINT("O"),
```
lauzadis, line 78: Can we also emit the new `ACCOUNT_ID_MODE_PREFERRED`, `ACCOUNT_ID_MODE_DISABLED`, `ACCOUNT_ID_MODE_REQUIRED` metrics?

0marperez: Yeah, we can add that

lauzadis (approved):

ianbotsf (commented):
**Question**: This is an binary-compat breaking change and I see the **acknowledge-api-break** tag applied to the PR. Why do we think this is safe to release? Isn't this precisely the kind of API change that can bite customers if they're using specific versions of **smithy-kotlin** different from **aws-sdk-kotlin**?

PR #1121, opened 2024-07-11
feat: add more conversion methods for JVM streams

ianbotsf:
## Issue \#

Closes https://github.com/awslabs/aws-sdk-kotlin/issues/1352

## Description of changes

This change adds new methods for converting to/from `ByteStream` and JVM streams:
* `InputStream.asByteStream`: Wrap an `InputStream` as a `ByteStream`
* `ByteStream.fromInputStream`: Alias for `InputStream.asByteStream`
* `ByteStream.writeToOutputStream`: Writes the contents of a `ByteStream` to an `OutputStream`

### DIFF
--- a/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/ByteStreamJVM.kt
+++ b/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/ByteStreamJVM.kt
@@ -11,6 +11,7 @@ import kotlinx.coroutines.withContext
 import java.io.ByteArrayInputStream
 import java.io.File
 import java.io.InputStream
+import java.io.OutputStream
 import java.nio.file.Path
 import kotlin.io.use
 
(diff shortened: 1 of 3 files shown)


runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/ByteStreamJVM.kt
```diff
@@ -97,3 +98,49 @@ public fun ByteStream.toInputStream(): InputStream = when (this) {
     is ByteStream.ChannelStream -> readFrom().toInputStream()
     is ByteStream.SourceStream -> readFrom().buffer().inputStream()
 }
+
+/**
+ * Create a [ByteStream.SourceStream] that reads from the given [InputStream]
+ * @param inputStream The [InputStream] from which to create a [ByteStream]
+ * @param contentLength If specified, indicates how many bytes remain in the input stream. Defaults to `null`.
+ */
+public fun ByteStream.Companion.fromInputStream(
```
0marperez, line 107: question: just curious, why an alias for`InputStream.asByteStream`?

0marperez (approved):

lauzadis (approved):
After this PR I think the only operation we're missing is converting an `OutputStream` to a `ByteStream`. Is that something you think would be useful, should we add it for the sake of completion?

ianbotsf:
> After this PR I think the only operation we're missing is converting an `OutputStream` to a `ByteStream`. Is that something you think would be useful, should we add it for the sake of completion?

`OutputStream`s are meant for writing to, which makes them a natural counterpart to Okio's `Sink` or our `SdkSink` components. Using them as a `Source` may be possible but would be a slight inversion of the typical control flow. `ByteStream`s typically _read from a source_ as opposed to _being written to by a source_ so the use case is not super clear here. I can imagine users could use machinery like [`PipedOutputStream`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/PipedOutputStream.html)/[`PipedInputStream`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/PipedInputStream.html) to bridge into something that _could_ be a `ByteStream` source but even there the semantics are a little unclear—how does the `ByteStream` know when the `OutputStream` is "complete"? If it _does_ know, does a hypothetical operation invocation simply block until the output stream is fully written?

I think we should wait for feature requests or user examples to put something concrete in place for `OutputStream`→`ByteStream` conversion.

PR #1123, opened 2024-07-15
misc(docs): only run accessibility scripts after the sidebar has fully loaded

lauzadis:
## Description of changes

### DIFF
--- a/docs/dokka-presets/scripts/accessibility.js
+++ b/docs/dokka-presets/scripts/accessibility.js
@@ -1,13 +1,13 @@
 /**
  * Check for elements with a navButton class, which indicates the sidebar has finished loading.
  */
-function dispatchNavigationLoadedEvent() {
-    if (document.querySelectorAll('.navButton').length > 0) {
-        window.dispatchEvent(new Event('navigationLoaded'));
-    } else {
-        setTimeout(dispatchNavigationLoadedEvent, 100);
+async function dispatchNavigationLoadedEvent() {
+    while (!document.querySelectorAll('.navButton').length > 0) {
+        await new Promise(resolve => setTimeout(resolve, 100));
     }
+    window.dispatchEvent(new Event('navigationLoaded'));
 }
+
 document.addEventListener('DOMContentLoaded', dispatchNavigationLoadedEvent);
 if (document.readyState === "interactive" || document.readyState === "complete" ) { dispatchNavigationLoadedEvent() }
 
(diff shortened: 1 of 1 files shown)


docs/dokka-presets/scripts/accessibility.js
```diff
@@ -1,3 +1,16 @@
...
+ * Check for elements with a navButton class, which indicates the sidebar has finished loading.
+ */
+function dispatchNavigationLoadedEvent() {
+    if (document.querySelectorAll('.navButton').length > 0) {
+        window.dispatchEvent(new Event('navigationLoaded'));
+    } else {
+        setTimeout(dispatchNavigationLoadedEvent, 100);
+    }
+}
+document.addEventListener('DOMContentLoaded', dispatchNavigationLoadedEvent);
```
0marperez, line 11: nit: spacing

docs/dokka-presets/scripts/accessibility.js
```diff
@@ -1,3 +1,16 @@
+/**
+ * Check for elements with a navButton class, which indicates the sidebar has finished loading.
+ */
+function dispatchNavigationLoadedEvent() {
+    if (document.querySelectorAll('.navButton').length > 0) {
+        window.dispatchEvent(new Event('navigationLoaded'));
+    } else {
+        setTimeout(dispatchNavigationLoadedEvent, 100);
```
0marperez, line 8: question: this is essentially a while loop, does it make more sense as one? also could an infinite loop ever happen here?

0marperez (approved):

docs/dokka-presets/scripts/accessibility.js
lauzadis: Sure it could be converted to a while loop. It's possible an infinite loop could happen, that means the sidebar never loaded. I'm not worried about this because it only means the accessibility fixes won't be applied to the (unloaded) sidebar

ianbotsf: JavaScript delays are asynchronous and non-blocking. A self-sustaining polling mechanism like this is effectively the best way to wait for a condition.

ianbotsf (approved):

0marperez (approved):

docs/dokka-presets/scripts/accessibility.js
lauzadis: @ianbotsf I pushed some code right after your comment, I think this accomplishes the same thing but it's a little simpler to read?

```
async function dispatchNavigationLoadedEvent() {
    while (!document.querySelectorAll('.navButton').length > 0) {
        await new Promise(resolve => setTimeout(resolve, 100));
    }
    window.dispatchEvent(new Event('navigationLoaded'));
}
```

ianbotsf: Oh yes I suppose that works. 😅 I have no opinions about the relative readability.

PR #1141, opened 2024-08-23
feat: smoke tests trait

0marperez:
## Description of changes

Adds support for smoke tests trait using an integration.

Companion PR: https://github.com/awslabs/aws-sdk-kotlin/pull/1388


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsIntegration.kt
```diff
@@ -0,0 +1,39 @@
...
+import software.amazon.smithy.kotlin.codegen.core.CodegenContext
+import software.amazon.smithy.kotlin.codegen.core.KotlinDelegator
+import software.amazon.smithy.kotlin.codegen.integration.KotlinIntegration
+import software.amazon.smithy.kotlin.codegen.model.hasTrait
+import software.amazon.smithy.kotlin.codegen.utils.operations
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.smoketests.traits.SmokeTestsTrait
+
+/**
+ * Renders smoke test runner for a service if any of the operations has the smoke test trait.
```
lauzadis, line 13: nit: "if any of the operations have the [SmokeTestsTrait]"

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/utils/Model.kt
```diff
@@ -0,0 +1,16 @@
...
+import software.amazon.smithy.model.knowledge.TopDownIndex
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.model.shapes.ShapeId
+import software.amazon.smithy.utils.SmithyInternalApi
+
+/**
+ * Syntactic sugar for getting a services operations
+ */
+@SmithyInternalApi
+fun Model.operations(service: ShapeId): Set<OperationShape> {
```
lauzadis, line 13: suggestion: Make this an extension property (`val`) instead of `fun`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsIntegration.kt
```diff
@@ -0,0 +1,39 @@
...
+ */
+class SmokeTestsIntegration : KotlinIntegration {
+    override fun enabledForService(model: Model, settings: KotlinSettings): Boolean =
+        model.operations(settings.service).any { it.hasTrait<SmokeTestsTrait>() } && !smokeTestDenyList.contains(settings.sdkId)
```
lauzadis, line 17: simplification for second half of the `&&`: `settings.sdkId !in smokeTestDenyList`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsIntegration.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    override fun writeAdditionalFiles(ctx: CodegenContext, delegator: KotlinDelegator) =
+        delegator.useFileWriter("SmokeTests.kt", "smoketests", "./jvm-src/main/java/") { writer ->
+            SmokeTestsRunnerGenerator(
+                writer,
+                ctx.symbolProvider.toSymbol(ctx.model.expectShape(ctx.settings.service)),
+                ctx.model.operations(ctx.settings.service).filter { it.hasTrait<SmokeTestsTrait>() },
+                ctx.model,
+                ctx.symbolProvider,
+                ctx.settings.sdkId,
```
lauzadis, line 27: I see `ctx.` repeated a lot, have you considered just passing the entire CodegenContext instead of unwrapping it here?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsIntegration.kt
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/**
+ * SDK ID's of services that model smoke tests incorrectly
+ */
+val smokeTestDenyList = setOf(
+    "Application Auto Scaling",
+    "SWF",
+    "WAFV2",
+)
```
lauzadis, line 39: Can you include GitHub issues (if they exist) filed against these services for them to fix their smoke tests?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
```diff
@@ -0,0 +1,174 @@
...
+class SmokeTestsRunnerGenerator(
+    private val writer: KotlinWriter,
+    private val service: Symbol,
+    private val operations: List<OperationShape>,
+    private val model: Model,
+    private val symbolProvider: SymbolProvider,
+    private val sdkId: String,
+) {
+    internal fun render() {
+        writer.write("import kotlin.system.exitProcess")
```
lauzadis, line 29: You should not need to import anything yourself, using the `#T` symbol should add the imports automatically

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
```diff
@@ -0,0 +1,174 @@
...
+        writer.emptyLine()
+        writer.write("private var exitCode = 0")
+        writer.write("private val regionOverride = System.getenv(\"AWS_SMOKE_TEST_REGION\")")
+        writer.write("private val skipTags = System.getenv(\"AWS_SMOKE_TEST_SKIP_TAGS\")?.let { it.split(\",\").map { it.trim() }.toSet() }")
```
lauzadis, line 33: This is missing `AWS_SMOKE_TEST_SERVICE_IDS`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
```diff
@@ -0,0 +1,174 @@
...
+    private val operations: List<OperationShape>,
+    private val model: Model,
+    private val symbolProvider: SymbolProvider,
+    private val sdkId: String,
+) {
+    internal fun render() {
+        writer.write("import kotlin.system.exitProcess")
+        writer.emptyLine()
+        writer.write("private var exitCode = 0")
+        writer.write("private val regionOverride = System.getenv(\"AWS_SMOKE_TEST_REGION\")")
```
lauzadis, line 32: suggestion: Use `#S` to avoid having to manage the quotes yourself

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
```diff
@@ -0,0 +1,174 @@
...
+        operations.forEach { operation ->
+            operation.getTrait<SmokeTestsTrait>()?.testCases?.forEach { testCase ->
+                renderFunction(operation, testCase)
+                writer.emptyLine()
+            }
+        }
+    }
+
+    private fun renderFunction(operation: OperationShape, testCase: SmokeTestCase) {
+        writer.withBlock("private suspend fun ${testCase.id.toCamelCase()}() {", "}") {
```
lauzadis, line 61: suggestion: I see `testCase.id.toCamelCase()` repeated a lot, you should consider making an extension like `SmokeTestCase.fnName` to deduplicate it

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
```diff
@@ -0,0 +1,174 @@
...
+            write("val tags = setOf<String>(${testCase.tags.joinToString(",") { it.dq()} })")
+            write("if (skipTags != null && tags.any { it in skipTags }) return")
```
lauzadis, line 63: You could make this always non-null when you create the set. It can be an empty set by default

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
```diff
@@ -0,0 +1,174 @@
...
+            if (testCase.vendorParams.isPresent) {
+                testCase.vendorParams.get().members.forEach { vendorParam ->
+                    if (vendorParam.key.value == "region") {
+                        write("region = regionOverride ?: #L", vendorParam.value.render())
+                    } else {
+                        write("#L = #L", vendorParam.key.value.toCamelCase(), vendorParam.value.render())
+                    }
+                }
+            } else {
+                write("region = regionOverride")
```
lauzadis, line 86: What happens when the `AWS_SMOKE_TEST_REGION` is not set?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
```diff
@@ -0,0 +1,174 @@
...
+            }
+            val expectingSpecificError = testCase.expectation.failure.getOrNull()?.errorId?.getOrNull() != null
+            write("interceptors.add(#T($expectingSpecificError))", RuntimeTypes.HttpClient.Interceptors.SmokeTestsInterceptor)
+        }
+        checkVendorParamsAreCorrect(testCase)
+    }
+
+    /**
+     * Smithy IDL doesn't check that vendor params are found in vendor params shape so we have to check here.
+     */
```
lauzadis, line 96: Should this be considered an "Invalid model" and added to the deny list? We shouldn't have to be checking this ourselves

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
```diff
@@ -0,0 +1,174 @@
...
+                    "Smithy smoke test \"${testCase.id}\" contains invalid vendor param \"$vendorParam\", it was not found in vendor params shape \"${testCase.vendorParamsShape}\""
+                }
+            }
+        }
+    }
+
+    private fun renderOperation(operation: OperationShape, testCase: SmokeTestCase) {
+        val operationSymbol = symbolProvider.toSymbol(model.getShape(operation.input.get()).get())
+
+        writer.addImport(operationSymbol)
```
lauzadis, line 114: Avoid manually adding imports if possible

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
```diff
@@ -0,0 +1,174 @@
...
+        )
+        writer.write("if (!success) exitCode = 1")
+    }
+
+    /**
+     * Tries to get the specific exception required in the failure criterion of a test.
+     * If no specific exception is required we default to the generic smoke tests failure exception.
+     *
+     * Some smoke tests model exceptions not found in the model, in that case we default to the generic smoke tests
+     * failure exception.
```
lauzadis, line 151: I think these should be considered invalid models and added to the deny list until fixed

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/util/Node.kt
```diff
@@ -0,0 +1,31 @@
...
+import software.amazon.smithy.model.node.Node
+import software.amazon.smithy.model.node.NullNode
+import software.amazon.smithy.model.node.NumberNode
+import software.amazon.smithy.model.node.ObjectNode
+import software.amazon.smithy.model.node.StringNode
+
+/**
+ * Renders a [Node] into String format for codegen.
+ */
+fun Node.render(): String = when (this) {
```
lauzadis, line 15: naming: render usually means you are making a `write(...)` call, which does not happen here

codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGeneratorTest.kt
```diff
@@ -0,0 +1,200 @@
...
+import software.amazon.smithy.kotlin.codegen.test.*
+import software.amazon.smithy.kotlin.codegen.utils.operations
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.smoketests.traits.SmokeTestsTrait
+import kotlin.test.Test
+
+class SmokeTestsRunnerGeneratorTest {
+    private val moneySign = "$"
+
+    private fun codegen(model: Model): String {
```
lauzadis, line 13: naming: generateSmokeTests

codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGeneratorTest.kt
```diff
@@ -0,0 +1,200 @@
...
+            codegenCtx.settings.sdkId,
+        ).render()
+        return writer.toString()
+    }
+
+    @Test
+    fun codegenTest() {
+        val model =
+            """
+            ${moneySign}version: "2"
```
lauzadis, line 32: Other places we do this use `${'$'}` which seems more readable

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsIntegration.kt
lauzadis: Also, this should be defined in aws-sdk-kotlin

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsIntegration.kt
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Renders smoke test runner for a service if any of the operations has the smoke test trait.
+ */
+class SmokeTestsIntegration : KotlinIntegration {
+    override fun enabledForService(model: Model, settings: KotlinSettings): Boolean =
+        model.operations(settings.service).any { it.hasTrait<SmokeTestsTrait>() } && !smokeTestDenyList.contains(settings.sdkId)
+
+    override fun writeAdditionalFiles(ctx: CodegenContext, delegator: KotlinDelegator) =
+        delegator.useFileWriter("SmokeTests.kt", "smoketests", "./jvm-src/main/java/") { writer ->
```
lauzadis, line 20: correctness: package ideally should be `aws.sdk.kotlin.services.$SERVICE.smoketests`, but we should not reference aws-sdk-kotlin here. Can this be made configurable through the integration somehow?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
0marperez: In that case `regionOverride ` will be set to `null`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
0marperez: I had some issues here doing that. The `exitProcess` function is not available in common, only JVM and Native. See: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.system/exit-process.html.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/traits/FailedResponseTrait.kt
```diff
@@ -0,0 +1,14 @@
...
+import software.amazon.smithy.model.node.ObjectNode
+import software.amazon.smithy.model.shapes.ShapeId
+import software.amazon.smithy.model.traits.AnnotationTrait
+
+/**
+ * Indicates the annotated service should always return a failed response.
+ */
+class FailedResponseTrait(node: ObjectNode) : AnnotationTrait(ID, node) {
+    companion object {
+        val ID: ShapeId = ShapeId.from("com.test#failedResponseTrait")
```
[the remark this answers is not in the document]
lauzadis: I think `FailedResponseTrait` and `SuccessResponseTrait` can better go together into a single file `SmokeTestTraits.kt`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsIntegration.kt
```diff
@@ -0,0 +1,30 @@
...
+ */
+class SmokeTestsIntegration : KotlinIntegration {
+    override fun enabledForService(model: Model, settings: KotlinSettings): Boolean =
+        model.operations(settings.service).any { it.hasTrait<SmokeTestsTrait>() }
+
+    override fun writeAdditionalFiles(ctx: CodegenContext, delegator: KotlinDelegator) =
+        delegator.useFileWriter(
+            "SmokeTests.kt",
+            "${ctx.settings.pkg.name}.smoketests",
+            "./jvm-src/test/java/",
```
lauzadis, line 23: Is `jvm-src` a standard sourceset location?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
```diff
@@ -0,0 +1,216 @@
...
+    // Generator config
+    private val model = ctx.model
+    private val sdkId = ctx.settings.sdkId
+    private val symbolProvider = ctx.symbolProvider
+    private val service = symbolProvider.toSymbol(model.expectShape(ctx.settings.service))
+    private val operations = ctx.model.operations(ctx.settings.service).filter { it.hasTrait<SmokeTestsTrait>() }
+
+    // Test config
+    private val hasSuccessResponseTrait = ctx.model.expectShape<ServiceShape>(ctx.settings.service).hasTrait(SuccessResponseTrait.ID)
+    private val hasFailedResponseTrait = ctx.model.expectShape<ServiceShape>(ctx.settings.service).hasTrait(FailedResponseTrait.ID)
```
lauzadis, line 43: Generally I prefer to just grab everything from the `ctx` instead of storing in local values, but since you've made `private val model = ctx.model`, you should not use `ctx.model` anywhere

`ctx.model` -> `model`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
```diff
@@ -0,0 +1,216 @@
...
+    init {
+        check(!(hasSuccessResponseTrait && hasFailedResponseTrait)) {
+            "A service can't have both the success response trait and the failed response trait."
+        }
+    }
```
lauzadis, line 48: Do we also need to check that `hasSuccessResponseTrait || hasFailedResponseTrait` (a service has exactly one of the traits)?

lauzadis (approved):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
0marperez: No because the trait is only used for the E2E tests. During code generation for actual real life services neither of the traits is necessary

0marperez: Yeah, I don't like it either. I can't think of any other reasonable way to replace the `httpClient` without high coupling. 

> If we're just using these traits to configure the httpClient, can it be configured through the smoke test vendor params instead?

We would have to add some custom logic for that to work, so the coupling would still be there.

.github/workflows/continuous-integration.yml
```diff
@@ -131,5 +131,6 @@ jobs:
         sed -i "s/smithy-kotlin-runtime-version = .*$/smithy-kotlin-runtime-version = \"$SMITHY_KOTLIN_RUNTIME_VERSION\"/" gradle/libs.versions.toml
         sed -i "s/smithy-kotlin-codegen-version = .*$/smithy-kotlin-codegen-version = \"$SMITHY_KOTLIN_CODEGEN_VERSION\"/" gradle/libs.versions.toml
         ./gradlew --parallel publishToMavenLocal
+        ./gradlew build
```
lauzadis, line 134: What is the reason for adding `./gradlew build` here? Don't the `test` tasks depend on `build` already?

0marperez: I think so, the `build` task depends on the `stageServices` task but it doesn't seem to be running when the `build` task is being called by the `test` task.

I added a dependency on `stageServices` to the `test` tasks and that should allow us to get rid of this ^

ianbotsf (changes requested):

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/SmokeTestsInterceptor.kt
```diff
@@ -0,0 +1,29 @@
...
+public class SmokeTestsInterceptor : HttpInterceptor {
+    override fun readBeforeDeserialization(context: ProtocolResponseInterceptorContext<Any, HttpRequest, HttpResponse>) {
+        val status = context.protocolResponse.status.value
+        when (status) {
+            in 400..599 -> throw SmokeTestsFailureException()
+            in 200..299 -> throw SmokeTestsSuccessException()
+            else -> throw SmokeTestsUnexpectedException()
+        }
+    }
+}
```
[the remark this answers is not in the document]
0marperez: That logic is [here](https://github.com/smithy-lang/smithy-kotlin/blob/524c8cae019627fceabd4b16ff8229b406e73aa7/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt#L125-L127) in the smoke test runner generator

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
```diff
@@ -0,0 +1,216 @@
...
+                            RuntimeTypes.Http.Response.HttpResponse,
+                            RuntimeTypes.Http.StatusCode,
+                            RuntimeTypes.Http.Headers,
+                            RuntimeTypes.Http.HttpBody,
+                        )
+                        write("val now = #T.now()", RuntimeTypes.Core.Instant)
+                        write("#T(request, resp, now, now)", RuntimeTypes.Http.HttpCall)
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
0marperez: This logic is supposed to be used during testing only. The test client is only used if the service has those custom traits that were created (`FailedResponseTrait` & `SuccessResponseTrait`)

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
```diff
@@ -0,0 +1,216 @@
...
+            writer.emptyLine()
+            writer.write("private var exitCode = 0")
+            writer.write("private val regionOverride = System.getenv(#S)", "AWS_SMOKE_TEST_REGION")
+            writer.write("private val skipTags = System.getenv(#S)?.let { it.split(\",\").map { it.trim() }.toSet() } ?: emptySet()", SKIP_TAGS)
+            writer.write("private val serviceFilter = System.getenv(#S)?.let { it.split(\",\").map { it.trim() }.toSet() }", SERVICE_FILTER)
+            writer.emptyLine()
+            writer.withBlock("public suspend fun main() {", "}") {
+                renderFunctionCalls()
+                write("exitProcess(exitCode)")
+            }
```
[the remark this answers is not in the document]
0marperez: This is part of the design document, we're supposed to return a non-zero exit code when any of the tests don't pass. Just returning the code doesn't set the exit code AFAIK.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsRunnerGenerator.kt
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    internal fun render() {
+        writer.declareSection(SmokeTestsRunner) {
+            writer.write("import kotlin.system.exitProcess")
+            writer.emptyLine()
+            writer.write("private var exitCode = 0")
+            writer.write("private val regionOverride = System.getenv(#S)", "AWS_SMOKE_TEST_REGION")
+            writer.write("private val skipTags = System.getenv(#S)?.let { it.split(\",\").map { it.trim() }.toSet() } ?: emptySet()", SKIP_TAGS)
+            writer.write("private val serviceFilter = System.getenv(#S)?.let { it.split(\",\").map { it.trim() }.toSet() }", SERVICE_FILTER)
```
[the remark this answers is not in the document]
0marperez: It's important to know if the user didn't set this because this is an include filter. It the default is an empty set then no smoke tests will run. We could change the default to an empty set and then check if it's that instead of null

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/traits/SmokeTestTraits.kt
```diff
@@ -0,0 +1,23 @@
...
+}
+
+/**
+ * Indicates the annotated service should always return a success response.
+ */
+class SuccessResponseTrait(node: ObjectNode) : AnnotationTrait(ID, node) {
+    companion object {
+        val ID: ShapeId = ShapeId.from("smithy.kotlin.traits#successResponseTrait")
+    }
+}
```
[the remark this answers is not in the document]
0marperez: The traits are applied [here](https://github.com/awslabs/aws-sdk-kotlin/tree/smoke-tests-trait/tests/codegen/smoke-tests/src/test/resources) to the either of those models for testing purposes

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsIntegration.kt
0marperez: Main reason is that there's no common code for `exitProcess` 
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.system/exit-process.html

I think I can wrap this in our own KMP function, thoughts?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinWriter.kt
```diff
@@ -198,6 +198,8 @@ class KotlinWriter(
             )
         }
 
+    fun emptyLine(): KotlinWriter = this.write("")
```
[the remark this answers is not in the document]
0marperez: I'll get rid of `emptyLine()`

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/smoketests/SmokeTestsFunctionsJVM.kt
```diff
@@ -0,0 +1,6 @@
+package aws.smithy.kotlin.runtime.smoketests
+
+import kotlin.system.exitProcess
+
+public actual fun exitProcess(status: Int): Nothing = exitProcess(status)
+public actual fun getEnv(name: String): String? = System.getenv(name)
```
ianbotsf, line 6: **Question**: This seems like [platform functionality](https://github.com/smithy-lang/smithy-kotlin/blob/main/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Platform.kt) to me—getting environment variables is already something we do cross-platform and process control seems like a natural addition. Is there any reason we can't reuse `PlatformProvider` for env vars and add support for exiting the process with a return code?

ianbotsf (approved):

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/smoketests/SmokeTestsFunctionsJVM.kt
0marperez: Thanks for calling this out, I almost reinvented the wheel here

PR #1143, opened 2024-08-23
misc: Downgrade clock skew exception stack trace to `debug` level

lauzadis:
## Description of changes

### DIFF
--- a/runtime/protocol/aws-protocol-core/common/src/aws/smithy/kotlin/runtime/awsprotocol/ClockSkewInterceptor.kt
+++ b/runtime/protocol/aws-protocol-core/common/src/aws/smithy/kotlin/runtime/awsprotocol/ClockSkewInterceptor.kt
@@ -89,7 +89,8 @@ public class ClockSkewInterceptor : HttpInterceptor {
             try {
                 Instant.fromRfc5322(it)
             } catch (e: ParseException) {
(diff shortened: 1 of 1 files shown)


runtime/protocol/aws-protocol-core/common/src/aws/smithy/kotlin/runtime/awsprotocol/ClockSkewInterceptor.kt
```diff
@@ -89,7 +89,8 @@ public class ClockSkewInterceptor : HttpInterceptor {
             try {
                 Instant.fromRfc5322(it)
             } catch (e: ParseException) {
-                logger.warn(e) { "Service returned malformed \"Date\" header value \"$it\", skipping skew calculation" }
+                logger.warn { "Service returned malformed \"Date\" header value \"$it\", skipping skew calculation" }
+                logger.debug(e) { e.message.toString() }
```
ianbotsf, line 93: **Nit**: Logging `e.message` is unnecessary since the logger should already include it. I'd just give a bit more context about where the error was encountered and leave it at that:

```kotlin
logger.debug(e) { "Error while parsing \"Date\" header from service response" }
logger.warn { "Service returned malformed \"Date\" header value \"$it\", skipping skew calculation" }
```

ianbotsf (approved):
Minor nit—fix and ship!

PR #1150, opened 2024-09-12
feat: OkHttp4Engine

lauzadis:
## Description of changes


runtime/protocol/http-client-engines/http-client-engine-okhttp/api/http-client-engine-okhttp.api
```diff
@@ -14,7 +65,7 @@ public final class aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine$Com
 
 public final class aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngineConfig : aws/smithy/kotlin/runtime/http/engine/HttpClientEngineConfigImpl {
 	public static final field Companion Laws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngineConfig$Companion;
-	public synthetic fun <init> (Laws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngineConfig$Builder;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Laws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngineConfig$Builder;)V
```
lauzadis, line 68: Moving the constructor from `internal` to `@InternalApi public` results in this "API break", I think it's safe in practice

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/MetricsInterceptor.kt
```diff
@@ -28,7 +30,7 @@ internal object MetricsInterceptor : Interceptor {
         }
 
         val originalResponse = chain.proceed(request)
-        val response = if (originalResponse.body.contentLength() != 0L) {
+        val response = if (originalResponse.body?.contentLength() != 0L) {
             originalResponse.newBuilder()
```
ianbotsf, line 34: **Question**: Why is the null-safe operator now necessary? And what does a null `body` imply?

I believe if `body` _is_ actually null then we'll encounter an NPE on L35 where we dereference `body.instrument`...

runtime/protocol/http-client-engines/http-client-engine-okhttp4/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        jvmTest {
+            dependencies {
+                implementation(project(":runtime:testing"))
+                // use otel testing capabilities
+                implementation(project(":runtime:observability:telemetry-provider-otel"))
+                implementation(libs.opentelemetry.sdk.testing)
+            }
+        }
```
ianbotsf, line 32: **Question**: Why do we have these dependencies for `commonTest` and `jvmTest`? I can't see a test source set of any kind in this module.

ianbotsf (changes requested):
This looks pretty good but we should add this new engine to [our HTTP benchmark suite](https://github.com/smithy-lang/smithy-kotlin/tree/main/tests/benchmarks/http-benchmarks) and remeasure each engine.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/MetricsInterceptor.kt
lauzadis: [`Response.body` is nullable in OkHttp3 4.x](https://github.com/square/okhttp/blob/4984568367caaf359b82c452bd28b5e192824d1c/okhttp/src/main/kotlin/okhttp3/Response.kt#L78) but not in 5.x, so I've added null-safe operators everywhere we use a body. 

Good catch that we would still dereference it on L35, I will fix that. In my testing I don't think the body _is_ ever null but it's safer to treat it like so

runtime/protocol/http-client-engines/http-client-engine-okhttp4/build.gradle.kts
lauzadis: They are left over from my copy-paste of the OkHttp `build.gradle.kts`, will remove

lauzadis:
> we should add this new engine to [our HTTP benchmark suite](https://github.com/smithy-lang/smithy-kotlin/tree/main/tests/benchmarks/http-benchmarks?rgh-link-date=2024-09-13T15%3A18%3A28Z) and remeasure each engine

Sure, I forgot we had those! They sure are dusty now...
> Baseline `0.14.0-SNAPSHOT` on EC2 **[m5.4xlarge](https://aws.amazon.com/ec2/instance-types/m5/)**

runtime/protocol/http-client-engines/http-client-engine-okhttp/api/http-client-engine-okhttp.api
ianbotsf: This is definitely a breaking change. If an SDK client built against the latest engine attempts to call the constructor and the _old_ engine is actually on the classpath, users will see a `NoSuchMethodError`. Method signatures used by callers are resolved when the caller is built. This change is _source-safe_ but not _binary-safe_.

Can we leave the previous `internal` constructor and add a new factory method instead?

lauzadis: Thanks for catching that and reminding me. Actually it seems like this change was not necessary after all. `OkHttpEngine4` only uses the public constructor here.

runtime/protocol/http-client-engines/http-client-engine-okhttp4/README.md
```diff
@@ -0,0 +1,89 @@
...
+import aws.sdk.kotlin.services.s3.*
+import aws.smithy.kotlin.runtime.http.engine.okhttp4.OkHttp4Engine
+import kotlinx.coroutines.runBlocking
+
+fun main() = runBlocking {
+    S3Client.fromEnvironment {
+        httpClient = OkHttp4Engine()
+    }.use {
+        // use the client!
+    }
```
ianbotsf, line 44: **Nit**: This usage pattern doesn't correctly manage the lifetime of the engine because it's instantiated inline without a `use` block. I suggest we document either of:

* Using the DSL factory function:

  ```kotlin
  S3Client.fromEnvironment {
      httpClient(OkHttp4Engine)
  }.use { s3 ->
      // use the client!
  }
  ```

* Using a `use` block:

  ```kotlin
  OkHttp4Engine().use { okHttp4 ->
      S3Client.fromEnvironment {
          httpClient = okHttp4
      }.use { s3 ->
          // use the client!
      }
  }
  ```

tests/benchmarks/http-benchmarks/jvm/src/aws/smithy/kotlin/benchmarks/http/HttpEngineBenchmarks.kt
```diff
@@ -54,7 +57,7 @@ private val largeData = ByteArray(MB_PER_THROUGHPUT_OP * 1024 * 1024)
 @State(Scope.Benchmark)
 @OutputTimeUnit(TimeUnit.SECONDS)
 open class HttpEngineBenchmarks {
-    @Param(OKHTTP_ENGINE, CRT_ENGINE)
+    @Param(OKHTTP_ENGINE, CRT_ENGINE, OKHTTP4_ENGINE)
     var httpClientName: String = ""
```
ianbotsf, line 61: **Correctness**: Thanks for adding the new engine to the benchmark suite. Can you please run the benchmarks and update [the **README**](https://github.com/smithy-lang/smithy-kotlin/blob/84a601373ce9584541b885feef1eaa5434e0f04a/tests/benchmarks/http-benchmarks/README.md) with the results?

Feel free to use whatever instance type and JDK are convenient.

ianbotsf (changes requested):

ianbotsf (approved):
Perfect!

runtime/protocol/http-client-engines/http-client-engine-okhttp4/README.md
```diff
@@ -0,0 +1,91 @@
...
+	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:702)
+Caused by: java.lang.ClassNotFoundException: okhttp3.coroutines.ExecuteAsyncKt
+	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
+	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
+	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:525)
+	... 9 more
+```
+
+It likely means you failed to configure the SDK client to use the `OkHttpEngine4`. 
+Please double-check all of your SDK client configurations to ensure `httpClient = OkHttpEngine4()` is configured,
```
0marperez, line 90: nit: this is still using the old syntax

0marperez (approved):

runtime/protocol/http-client-engines/http-client-engine-okhttp4/README.md
```diff
@@ -0,0 +1,91 @@
+# OkHttp4 Engine
```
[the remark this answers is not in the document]
lauzadis: No it's just a README for the module. I was not planning to announce it in the SDK, users can start using this immediately once smithy-kotlin is released, the SDK does not need a version bump.

runtime/protocol/http-client-engines/http-client-engine-okhttp4/README.md
lauzadis: Yes, it's still a valid syntax, I don't think everyone will be using the .use { } style I recommended above, it's just an example.

PR #1155, opened 2024-09-24
fix: paginator generator nullability bug caused by documents

0marperez:
## Issue \#
The paginator generator is not code-generating nullable literals for item paginators of lists that have values that are nullable. It will only code-generate nullability if the list is marked as [sparse](https://smithy.io/2.0/spec/type-refinement-traits.html#sparse-trait) and not check if the list has values that are nullable.

Lists are not supposed to contain nullable items unless marked as `sparse`, so this is non-modeled behavior that we were not looking for but is necessary because [documents](https://smithy.io/2.0/spec/simple-types.html#document) are always nullable in smithy kotlin.

## Description of changes
The paginator generator now checks if the collection is made of nullable symbols and code generates nullability

### DIFF
--- a/.changes/f71f083b-6e5f-4a3c-9069-464b1f9f6d36.json
+++ b/.changes/f71f083b-6e5f-4a3c-9069-464b1f9f6d36.json
@@ -1,0 +1,5 @@
+{
+    "id": "f71f083b-6e5f-4a3c-9069-464b1f9f6d36",
+    "type": "bugfix",
+    "description": "Fix paginator generator nullability unreliablity when collecting flows of collection types like map or list"
+}
--- a/.changes/f71f083b-6e5f-4a3c-9069-464b1f9f6d36.json
+++ b/.changes/f71f083b-6e5f-4a3c-9069-464b1f9f6d36.json
@@ -1,5 +1,5 @@
 {
     "id": "f71f083b-6e5f-4a3c-9069-464b1f9f6d36",
     "type": "bugfix",
-    "description": "Fix paginator generator `List<Document>` nullability"
+    "description": "Fix paginator generator `List<*>` nullability"
 }
--- a/.changes/f71f083b-6e5f-4a3c-9069-464b1f9f6d36.json
+++ b/.changes/f71f083b-6e5f-4a3c-9069-464b1f9f6d36.json
@@ -1,5 +1,5 @@
 {
     "id": "f71f083b-6e5f-4a3c-9069-464b1f9f6d36",
     "type": "bugfix",
-    "description": "Fix paginator generator `List<*>` nullability"
+    "description": "Fix paginator generator `List<*>` & `Map.Entry<String, *>` nullability"
 }
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
@@ -274,7 +274,7 @@ private fun getItemDescriptorOrNull(paginationInfo: PaginationInfo, ctx: Codegen
     val (collectionLiteral, targetMember) = when (itemMember) {
         is MapShape -> {
             val symbol = ctx.symbolProvider.toSymbol(itemMember)
-            val entryExpression = symbol.expectProperty(SymbolProperty.ENTRY_EXPRESSION) as String + if (isSparse) "?" else ""
+            val entryExpression = symbol.expectProperty(SymbolProperty.ENTRY_EXPRESSION) as String
             entryExpression to itemMember
         }
         is CollectionShape -> {
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
@@ -272,18 +272,21 @@ private fun getItemDescriptorOrNull(paginationInfo: PaginationInfo, ctx: Codegen
     val itemMember = ctx.model.expectShape(itemMemberId)
     val isSparse = itemMember.isSparse
     val (collectionLiteral, targetMember) = when (itemMember) {
-        is MapShape ->
-            ctx.symbolProvider.toSymbol(itemMember)
-                .expectProperty(SymbolProperty.ENTRY_EXPRESSION) as String to itemMember
-        is CollectionShape ->
-            ctx.symbolProvider.toSymbol(ctx.model.expectShape(itemMember.member.target)).name to ctx.model.expectShape(
-                itemMember.member.target,
-            )
+        is MapShape -> {
+            val literal = ctx.symbolProvider.toSymbol(itemMember)
+                .expectProperty(SymbolProperty.ENTRY_EXPRESSION) as String + if (isSparse) "?" else ""
+            literal to itemMember
+        }
+        is CollectionShape -> {
+            val symbol = ctx.symbolProvider.toSymbol(ctx.model.expectShape(itemMember.member.target))
+            val literal = symbol.name + if (symbol.isNullable || isSparse) "?" else ""
+            literal to ctx.model.expectShape(itemMember.member.target)
+        }
         else -> error("Unexpected shape type ${itemMember.type}")
     }
 
     return ItemDescriptor(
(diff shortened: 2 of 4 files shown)


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
```diff
@@ -272,18 +272,21 @@ private fun getItemDescriptorOrNull(paginationInfo: PaginationInfo, ctx: Codegen
...
-                .expectProperty(SymbolProperty.ENTRY_EXPRESSION) as String to itemMember
-        is CollectionShape ->
-            ctx.symbolProvider.toSymbol(ctx.model.expectShape(itemMember.member.target)).name to ctx.model.expectShape(
-                itemMember.member.target,
-            )
+        is MapShape -> {
+            val literal = ctx.symbolProvider.toSymbol(itemMember)
+                .expectProperty(SymbolProperty.ENTRY_EXPRESSION) as String + if (isSparse) "?" else ""
+            literal to itemMember
+        }
```
ianbotsf, line 279: **Question**: Don't we also need to check if the symbol is nullable alongside the map being sparse?

ianbotsf (changes requested):
**Correctness**: Missing test coverage. Add unit tests as a minimum and preferably [codegen tests](https://github.com/smithy-lang/smithy-kotlin/tree/main/tests/codegen/paginator-tests) as well.

0marperez:
Yeah, this was my mistake I marked as ready for review before adding the tests!

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
0marperez: No, the nullability of documents in this case is handled by the `KotlinSymbolProvider`
https://github.com/smithy-lang/smithy-kotlin/blob/main/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt#L168-L179

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
```diff
@@ -272,18 +272,21 @@ private fun getItemDescriptorOrNull(paginationInfo: PaginationInfo, ctx: Codegen
...
+        is CollectionShape -> {
+            val symbol = ctx.symbolProvider.toSymbol(ctx.model.expectShape(itemMember.member.target))
```
lauzadis, line 281: simplification/duplication: `ctx.model.expectShape(itemMember.member.target)` is used twice, refactor to local value

.changes/f71f083b-6e5f-4a3c-9069-464b1f9f6d36.json
```diff
@@ -0,0 +1,5 @@
+{
+    "id": "f71f083b-6e5f-4a3c-9069-464b1f9f6d36",
+    "type": "bugfix",
+    "description": "Fix paginator generator `List<Document>` nullability"
```
lauzadis, line 4: nit/clarification: This change does not only affect `List<Document>`, right?

codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGeneratorTest.kt
```diff
@@ -762,4 +762,308 @@ class PaginatorGeneratorTest {
...
+                @httpQuery("MaxItems")
+                MaxItems: Integer
+            }
+            
+            structure ListFunctionsResponse {
+                Functions: FunctionConfigurationList,
+                NextMarker: String
+            }
+            
+            map FunctionConfigurationList {
```
lauzadis, line 957: naming: This is a map

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
lauzadis: Can you clarify a little bit? I'm confused why MapShape only checks `isSparse` while CollectionShape checks `symbol.isNullable || isSparse`

0marperez: Yeah, for `MapShape` the symbol provider provides the symbol via `toSymbol` and then we use the `ENTRY_EXPRESSION` system property from the symbol as the literal, and never look at the type that the map targets like in collections. 

The `ENTRY_EXPRESSION` system property is set in the code I linked above. It uses `val valueType` which uses `val valueSuffix` to add the `?` is the value is nullable.

.changes/f71f083b-6e5f-4a3c-9069-464b1f9f6d36.json
0marperez: It _should_ only affect `List<Document>` types because they're the only type of list that can contain nullables without the `sparse` trait being set on the list. Looking at [KotlinSymbolProvider](https://github.com/smithy-lang/smithy-kotlin/blob/main/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt), document is the only symbol that always returns `asNullable()`

lauzadis (approved):

codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGeneratorTest.kt
```diff
@@ -762,4 +762,308 @@ class PaginatorGeneratorTest {
 
         actual.shouldContainOnlyOnceWithDiff(expected)
     }
+
+    @Test
+    fun testRenderPaginatorWithDocumentList() {
```
ianbotsf, line 767: **Nit**: Let add a comment to each linking the issue that spawned them.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
ianbotsf: I don't think that's what we want. Your new unit test `testRenderPaginatorWithSparseDocumentMap` shows the problem: It takes a sparse map shape with `Document` values and codegens a paginator that returns `Flow<Map.Entry<String, Document?>?>`.

Note that there are _two_ nullability modifiers:
* one on `Document` because `Document` is nullable, which is correct 
* one on `Map.Entry<String, Document?>` because the map is sparse, which is _incorrect_

A sparse map does not have missing _entries_, only missing _values_. I believe the logic needs to coalesce the `Document` nullability and the map sparseness into the same spot. In other words, I believe the `testRenderPaginatorWithSparseDocumentMap` test should be expecting a `Flow<Map.Entry<String, Document?>>`.

codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGeneratorTest.kt
0marperez: I'll create one, we don't have an external issue for this

ianbotsf: Oh maybe not required if there wasn't an external issue...that's just busywork.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGenerator.kt
0marperez: I noticed that too, we've been code-generating them like that so I wasn't 100% sure if it was wrong. The fix is in the PR now.

ianbotsf (approved):

PR #1174, opened 2024-11-06
misc: Set up Kotlin/Native build

lauzadis:
## Description of changes


0marperez (approved):

runtime/protocol/http-client/native/test/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionInterceptorTestNative.kt
```diff
@@ -5,5 +5,5 @@
 package aws.smithy.kotlin.runtime.http.interceptors
 
 internal actual fun decompressGzipBytes(bytes: ByteArray): ByteArray {
-    TODO("Not yet implemented")
+    TODO("Not yet implemented. Can we write a pure Kotlin implementation?")
```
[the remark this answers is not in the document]
lauzadis: This is not leftover. It's something we should try to do while implementing Native GZIP / before launching Kotlin/Native support

runtime/protocol/http-client-engines/http-client-engine-crt/jvmAndNative/test/aws/smithy/kotlin/runtime/http/engine/crt/AsyncStressTest.kt
```diff
@@ -25,6 +25,7 @@ import kotlinx.coroutines.yield
 import kotlin.test.Test
 import kotlin.time.Duration.Companion.seconds
 
+// FIXME This test implements [TestWithLocalServer] which is JVM-only.
 class AsyncStressTest : TestWithLocalServer() {
```
ianbotsf, line 29: **Question**: Are we tracking these K/N-specific FIXMEs/TODOs somehow? I'm worried that some may slip through the cracks as we fill out Native support.

runtime/protocol/http-client/native/test/aws/smithy/kotlin/runtime/http/interceptors/RequestCompressionInterceptorTestNative.kt
```diff
@@ -5,5 +5,5 @@
...
 }
```
ianbotsf, line 9: **Comment**: We _can_ but we probably _shouldn't_ for performance and correctness reasons. We should look for existing platform implementations where possible. For Native that probably means binding to something like zlib or CRT.

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/time/InstantJVM.kt
```diff
@@ -26,6 +26,7 @@ import kotlin.time.toKotlinDuration
 import java.time.Duration as jtDuration
 import java.time.Instant as jtInstant
 
+// FIXME Consider making this multiplatform (`common`) using kotlinx.datetime
```
ianbotsf, line 29: **Comment**: Early in the SDK design, we rejected **kotlinx-datetime** as a backing impl because of concerns about support and long-term stability. These concerns might be addressed once [**kotlinx-datetime** is promoted to Beta](https://youtrack.jetbrains.com/issue/KT-64578/Promote-kotlinx-datetime-to-Beta) or once [`Instant` and `Clock` are moved to stdlib](https://youtrack.jetbrains.com/issue/KT-72480/Move-Instant-and-Clock-from-kotlinx-datetime-to-stdlib).

settings.gradle.kts
```diff
@@ -26,6 +26,15 @@ dependencyResolutionManagement {
...
 
+// Set up aws-crt-kotlin as a composite build
+val localAwsCrtKotlin = file("../aws-crt-kotlin")
+if (localAwsCrtKotlin.exists()) {
+    println("Including aws-crt-kotlin as a composite build")
+    includeBuild(localAwsCrtKotlin)
+} else {
+    println("Could not find aws-crt-kotlin in a sibling directory, not including it")
+}
+
```
ianbotsf, line 37: **Suggestion**: This seems brittle. I'd rather see a more flexible solution [like we have for **aws-sdk-kotlin**](https://github.com/awslabs/aws-sdk-kotlin/blob/26fe2120ae46fa25bc32bce4e5658b472b610db3/settings.gradle.kts#L113-L153), ideally extracted into the repo tools if that's possible.

ianbotsf, line 37: **Question**: This sets up a composite build with **aws-crt-kotlin** for **smithy-kotlin** but what if I'm building **aws-sdk-kotlin** and I'm already setting `compositeProjects=../smithy-kotlin,../aws-crt-kotlin`? Does the CRT project get built twice? Does it fail?

runtime/protocol/http-client-engines/test-suite/jvm/src/aws/smithy/kotlin/runtime/http/test/suite/Downloads.kt
```diff
@@ -53,6 +53,7 @@ internal fun Application.downloadTests() {
                 call.response.header("x-bar", "bar")
             }
 
+            // FIXME This is the only thing keeping this directory in `jvm`, could be `jvmAndNative`
             get("/gzipped") {
                 val uncompressed = ByteArray(1024) { it.toByte() }
                 val compressed = ByteArrayOutputStream().use { baStream ->
```
ianbotsf, line 59: **Question**: Don't we already know we need KMP GZip support and already have `expect`/`actual` declarations for it? Can we use those here and then move this to the **jvmAndNative** source set? Or do other blockers remain?

ianbotsf (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,195 @@
...
+
+      - name: Setup build
+        uses: ./smithy-kotlin/.github/actions/setup-build
+
+      - name: Build and Test on Windows JVM
+        working-directory: ./smithy-kotlin
+        run: |
+          ./gradlew apiCheck
+          ./gradlew -P"aws.sdk.kotlin.crt.disableCrossCompile"=true build
+          ./gradlew -P"aws.sdk.kotlin.crt.disableCrossCompile"=true allTests
```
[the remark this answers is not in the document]
lauzadis: I forgot exactly why, but it is necessary on Windows, I'll remove them and see what fails (or if the issue has been fixed since then)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,195 @@
...
+
+      - name: Build and Test on Apple platforms
+        working-directory: ./smithy-kotlin
+        shell: bash
+        run: |
+          # FIXME K2. Re-enable warnings as errors after this warning is removed: https://youtrack.jetbrains.com/issue/KT-68532
+          # echo "kotlinWarningsAsErrors=true" >> $GITHUB_WORKSPACE/local.properties              
+          ./gradlew apiCheck
+          ./gradlew -Paws.sdk.kotlin.crt.disableCrossCompile=true build
+          ./gradlew -Paws.sdk.kotlin.crt.disableCrossCompile=true allTests
```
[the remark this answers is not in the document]
lauzadis: Yes I think you're right, this was a leftover copy-paste where only `allTests` was running

gradle/libs.versions.toml
```diff
@@ -2,7 +2,7 @@
 kotlin-version = "2.0.21"
 dokka-version = "1.9.10"
 
-aws-kotlin-repo-tools-version = "0.4.13"
+aws-kotlin-repo-tools-version = "0.4.15-kn"
```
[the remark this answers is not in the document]
lauzadis: The main difference so far is all the targets which get enabled: https://github.com/awslabs/aws-kotlin-repo-tools/compare/main...kn-main

I think we _might_ be able to move this to **main** and drop the parallel versioning strategy.


This parallel release strategy is something that's been planned from the beginning. I can't say for sure but there may be more changes forthcoming which will require us to have this split.

runtime/protocol/http-client-engines/test-suite/jvm/src/aws/smithy/kotlin/runtime/http/test/suite/Downloads.kt
lauzadis: Maybe we can use our `expect`/`actual` declarations, I'll take a look. Right now this is using `java.util.zip.GZIPOutputStream` which is definitely JVM-only

settings.gradle.kts
lauzadis: What's brittle about it besides supporting a `local.properties` file? Does anyone even use that configuration option? I didn't see any use of it in our CI / other tooling

lauzadis: I don't configure `compositeProjects` at all and my **aws-sdk-kotlin** build gets resolved correctly with all three projects included

runtime/protocol/http-client-engines/http-client-engine-crt/jvmAndNative/test/aws/smithy/kotlin/runtime/http/engine/crt/AsyncStressTest.kt
lauzadis: Not currently. I was imagining we'd spot them during the **kn-main** -> **main** merge

runtime/protocol/http-client-engines/test-suite/jvm/src/aws/smithy/kotlin/runtime/http/test/suite/Downloads.kt
lauzadis: Moved to `jvmAndNative` by using our existing `GzipSdkSource` class

.github/workflows/ci.yml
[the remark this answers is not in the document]
lauzadis:
```
* What went wrong:
Task '.sdk.kotlin.crt.disableCrossCompile=true' not found in root project 'smithy-kotlin' and its subprojects.
```

Issue #1176, opened 2024-11-07
OkHttp4 Support Broken Due to Idle Connection PR


Rostifar:
### Describe the bug

A recent [PR](https://github.com/smithy-lang/smithy-kotlin/pull/1171/files) was merged, which requires [ConnectionListener](https://github.com/smithy-lang/smithy-kotlin/pull/1171/files#diff-75872e76cfe5cafdc473fe3016292828fea83e9d314d8605202bcfa92a230b73R103) support to function correctly. 

ConnectionListener was introduced in [OkHttp 5.x](https://square.github.io/okhttp/changelogs/changelog/#version-500-alpha12), so this change will break OkHttp 4.x support. 

### Regression Issue

- [X] Select this option if this issue appears to be a regression.

### Expected behavior

Building an [OkHttpEngine4](https://github.com/smithy-lang/smithy-kotlin/tree/main/runtime/protocol/http-client-engines/http-client-engine-okhttp4) instance should not require any OkHttp 5.x concepts.

### Current behavior

Instantiating `OkHttpEngine4` will lead to a `ClassNotFoundException` exception for `ConnectionListener` when `config.buildClient(...)` is [called](https://github.com/smithy-lang/smithy-kotlin/blob/03cdfb45610b3b0a98fb481a24f598b9f5726ea0/runtime/protocol/http-client-engines/http-client-engine-okhttp4/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp4/OkHttp4Engine.kt#L48).  

### Steps to Reproduce

Configure a OkHttp4Engine in accordance with [these](https://github.com/smithy-lang/smithy-kotlin/tree/main/runtime/protocol/http-client-engines/http-client-engine-okhttp4) instructions:
```
import aws.smithy.kotlin.runtime.http.engine.okhttp4.OkHttp4Engine
val client = OkHttp4Engine()
```
Use http-client-engine-okhttp4 v1.3.20 to reproduce. 

### Smithy-Kotlin version

aws.smithy.kotlin:http-client-engine-okhttp4:1.3.20

### Platform (JVM/JS/Native)

JVM

### Operating system and version

Ubuntu 20.04

0marperez:
Hello, thanks for reporting this issue. We have a PR up now that should fix it.

PR #1178, opened 2024-11-11
fix: http call waiting for idle connection monitor to finish

0marperez:
## Issue \#

Original issue that sparked idle connection monitoring: https://github.com/awslabs/aws-sdk-kotlin/issues/1214
OkHttp4 was broken by idle connection monitoring: https://github.com/smithy-lang/smithy-kotlin/issues/1176

## Description of changes

- Launching the connection monitors using the call context was causing the call to wait on the connection monitor coroutines. This fix creates a coroutine scope dedicated to connection monitors. This new scope will be cancelled/joined when the HTTP engine being used is closed.
- Noticed `shutdown` was missing from `OkHttp4Engine` so added it.
- The refactoring of `OkHttpEngineConfig.buildClient` fixes https://github.com/smithy-lang/smithy-kotlin/issues/1176 by not including `connectionListener` as part of `OkHttp4Engine`

Calling `S3` `getObject` twice went from taking ~6 sec to ~370ms
 
### DIFF
--- a/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/ConnectionIdleMonitor.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/ConnectionIdleMonitor.kt
@@ -5,18 +5,10 @@
 package aws.smithy.kotlin.runtime.http.engine.okhttp
 
 import aws.smithy.kotlin.runtime.telemetry.logging.logger
-import kotlinx.coroutines.CoroutineName
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.Job
-import kotlinx.coroutines.cancelAndJoin
-import kotlinx.coroutines.isActive
-import kotlinx.coroutines.launch
-import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.*
 import okhttp3.Call
 import okhttp3.Connection
 import okhttp3.ConnectionListener
-import okhttp3.ExperimentalOkHttpApi
 import okhttp3.internal.closeQuietly
 import okio.EOFException
 import okio.buffer
@@ -24,13 +16,19 @@ import okio.source
 import java.net.SocketException
 import java.net.SocketTimeoutException
 import java.util.concurrent.ConcurrentHashMap
+import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.coroutineContext
 import kotlin.time.Duration
 import kotlin.time.measureTime
 
-@OptIn(ExperimentalOkHttpApi::class)
 internal class ConnectionIdleMonitor(val pollInterval: Duration) : ConnectionListener() {
+    private val monitorScope = CoroutineScope(Dispatchers.IO + SupervisorJob())
     private val monitors = ConcurrentHashMap<Connection, Job>()
+
+    fun close(): Unit = runBlocking {
+        monitors.values.forEach { it.cancelAndJoin() }
+        monitorScope.cancel()
+    }
 
     private fun Call.callContext() =
         request()
@@ -58,12 +56,11 @@ internal class ConnectionIdleMonitor(val pollInterval: Duration) : ConnectionLis
 
     override fun connectionReleased(connection: Connection, call: Call) {
         val connId = System.identityHashCode(connection)
-        val context = call.callContext()
-        val scope = CoroutineScope(context)
-        val monitor = scope.launch(CoroutineName("okhttp-conn-monitor-for-$connId")) {
-            doMonitor(connection)
+        val callContext = call.callContext()
+        val monitor = monitorScope.launch(CoroutineName("okhttp-conn-monitor-for-$connId")) {
+            doMonitor(connection, callContext)
         }
-        context.logger<ConnectionIdleMonitor>().trace { "Launched coroutine $monitor to monitor $connection" }
+        callContext.logger<ConnectionIdleMonitor>().trace { "Launched coroutine $monitor to monitor $connection" }
 
         // Non-locking map access is okay here because this code will only execute synchronously as part of a
         // `connectionReleased` event and will be complete before any future `connectionAcquired` event could fire for
@@ -71,8 +68,8 @@ internal class ConnectionIdleMonitor(val pollInterval: Duration) : ConnectionLis
         monitors[connection] = monitor
     }
 
-    private suspend fun doMonitor(conn: Connection) {
-        val logger = coroutineContext.logger<ConnectionIdleMonitor>()
+    private suspend fun doMonitor(conn: Connection, callContext: CoroutineContext) {
+        val logger = callContext.logger<ConnectionIdleMonitor>()
 
         val socket = conn.socket()
         val source = try {
@@ -111,6 +108,7 @@ internal class ConnectionIdleMonitor(val pollInterval: Duration) : ConnectionLis
                 logger.trace { "Attempting to reset soTimeout..." }
                 try {
                     conn.socket().soTimeout = oldTimeout
+                    logger.trace { "SoTimeout reset." }
                 } catch (e: Throwable) {
                     logger.warn(e) { "Failed to reset socket timeout on $conn. Connection may be unstable now." }
                 }
--- a/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/ConnectionIdleMonitor.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/ConnectionIdleMonitor.kt
@@ -9,6 +9,7 @@ import kotlinx.coroutines.*
 import okhttp3.Call
 import okhttp3.Connection
 import okhttp3.ConnectionListener
+import okhttp3.ExperimentalOkHttpApi
 import okhttp3.internal.closeQuietly
 import okio.EOFException
 import okio.buffer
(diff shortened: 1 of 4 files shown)


runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/ConnectionIdleMonitor.kt
```diff
@@ -5,33 +5,31 @@
...
-@OptIn(ExperimentalOkHttpApi::class)
 internal class ConnectionIdleMonitor(val pollInterval: Duration) : ConnectionListener() {
+    private val monitorScope = CoroutineScope(Dispatchers.IO + SupervisorJob())
     private val monitors = ConcurrentHashMap<Connection, Job>()
 
+    fun close(): Unit = runBlocking {
+        monitors.values.forEach { it.cancelAndJoin() }
+        monitorScope.cancel()
+    }
+
```
ianbotsf, line 32: **Question**: Cancelling/joining child jobs sequentially in a loop is time-consuming. Moreover, the docs say that cancelling a parent job (even a `SupervisorJob`) should automatically cancel all child jobs. Does this code accomplish the same thing?

```kotlin
fun close(): Unit = runBlocking {
    monitorScope.cancel()
}
```

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
```diff
@@ -44,15 +44,15 @@ public class OkHttpEngine(
     }
 
     private val metrics = HttpClientMetrics(TELEMETRY_SCOPE, config.telemetryProvider)
-    private val client = config.buildClient(metrics)
+    private val connectionIdleMonitor = if (config.connectionIdlePollingInterval != null) ConnectionIdleMonitor(config.connectionIdlePollingInterval) else null
```
ianbotsf, line 47: **Nit**:

```kotlin
private val connectionIdleMonitor = config.connectionIdlePollingInterval?.let { ConnectionIdleMonitor(it) }
```

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
```diff
@@ -99,22 +99,13 @@ public fun OkHttpEngineConfig.buildClient(metrics: HttpClientMetrics): OkHttpCli
...
         // use our own pool configured with the timeout settings taken from config
-        @OptIn(ExperimentalOkHttpApi::class)
         val pool = ConnectionPool(
             maxIdleConnections = 5, // The default from the no-arg ConnectionPool() constructor
             keepAliveDuration = config.connectionIdleTimeout.inWholeMilliseconds,
             TimeUnit.MILLISECONDS,
-            connectionListener = connectionListener,
         )
-        connectionPool(pool)
+        connectionPool(poolOverride ?: pool)
```
ianbotsf, line 108: **Nit**: With this change we're now constructing a `ConnectionPool` we'll just throw away if `poolOverride` is set. We should only construct the OkHttp4-safe pool _if_ `poolOverride` is null.

ianbotsf (changes requested):

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/ConnectionIdleMonitor.kt
0marperez: The difference is that we're not waiting for the child jobs to actually finish cancelling (join). So we could be leaving some resources behind. 

I think it should improve performance if we just cancel and don't wait for a join. It shouldn't be an issue unless someone is using a lot of clients.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/ConnectionIdleMonitor.kt
```diff
@@ -5,33 +5,31 @@
...
 import java.net.SocketTimeoutException
 import java.util.concurrent.ConcurrentHashMap
+import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.coroutineContext
 import kotlin.time.Duration
 import kotlin.time.measureTime
 
-@OptIn(ExperimentalOkHttpApi::class)
 internal class ConnectionIdleMonitor(val pollInterval: Duration) : ConnectionListener() {
+    private val monitorScope = CoroutineScope(Dispatchers.IO + SupervisorJob())
```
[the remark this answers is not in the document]
0marperez: I think so, let me test this.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/ConnectionIdleMonitor.kt
0marperez: I tested this and time to cancel the connection idle monitor(s) went from ~200ms to ~600us. Now that I think of it, if someone is opening and closing clients very often and with a lot of connections per client then not waiting for the connection monitors to join could be a bigger issue.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/ConnectionIdleMonitor.kt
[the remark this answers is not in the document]
0marperez: I'm seeing some issues trying to use the HTTP engine scope. We try to wait for the HTTP engine job to [complete](https://github.com/smithy-lang/smithy-kotlin/blob/03cdfb45610b3b0a98fb481a24f598b9f5726ea0/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/HttpClientEngine.kt#L55) before calling [shutdown](https://github.com/smithy-lang/smithy-kotlin/blob/03cdfb45610b3b0a98fb481a24f598b9f5726ea0/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/HttpClientEngine.kt#L57). And that make us wait on the connection idle monitor to just fail it seems like, since it's running in a while loop.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/ConnectionIdleMonitor.kt
```diff
@@ -111,6 +109,7 @@ internal class ConnectionIdleMonitor(val pollInterval: Duration) : ConnectionLis
                 logger.trace { "Attempting to reset soTimeout..." }
                 try {
                     conn.socket().soTimeout = oldTimeout
+                    logger.trace { "SoTimeout reset." }
```
lauzadis, line 112: nit: `soTimeout`

runtime/protocol/http-client-engines/http-client-engine-okhttp4/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp4/OkHttp4Engine.kt
```diff
@@ -70,6 +70,12 @@ public class OkHttp4Engine(
             }
         }
     }
+
+    override fun shutdown() {
+        client.connectionPool.evictAll()
+        client.dispatcher.executorService.shutdown()
+        metrics.close()
+    }
```
lauzadis, line 78: Nice catch

lauzadis (approved):

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/ConnectionIdleMonitor.kt
ianbotsf: Ah I see, you're right that could be an issue. So then how about:

```kotlin
fun close(): Unit = runBlocking {
    monitorScope.cancelAndJoin()
}
```

0marperez: Unfortunately there's no `cancelAndJoin()` for `CoroutineScope`, only `Job`.

ianbotsf: Oh jeez you're right. Well the [`CoroutineScope.cancel()` method implementation](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/common/src/CoroutineScope.kt#L276-L285) pretty much just verifies the scope _contains_ a job and then calls `cancel` on that job. We can do something similar here:

```kotlin
val monitorJob = requireNotNull(monitorScope.coroutineContext[Job]) { "<some error msg>" }
monitorJob.cancelAndJoin()
```

ianbotsf (approved):
Approved pending `cancelAndJoin` feedback.

PR #1182, opened 2024-11-18
kn: checksum bindings

lauzadis:
## Description of changes

### DIFF
--- a/.github/scripts/run-container-test.py
+++ b/.github/scripts/run-container-test.py
@@ -1,0 +1,152 @@
+#!/usr/bin/env python3
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Run precompiled Kotlin/Native test binaries in a Docker container for a specific Linux distribution and architecture.
+
+This requires Docker multiarch support, see https://docs.docker.com/build/building/multi-platform/ and https://github.com/multiarch/qemu-user-static
+In GitHub we use a provided action for this: https://github.com/docker/setup-qemu-action
+
+Locally you would need to run one of:
+
+`docker run --rm --privileged multiarch/qemu-user-static --reset -p yes --credential yes`
+
+OR
+
+`docker run --privileged --rm tonistiigi/binfmt --install all`
+"""
+
+import argparse
+import os
+import subprocess
+import shlex
+import shutil
+
+VERBOSE = False
+
+DISTRO_TO_IMAGE_NAME = {
+    "ubuntu-22.04": "public.ecr.aws/lts/ubuntu:22.04_stable",
+    "al2023": "public.ecr.aws/amazonlinux/amazonlinux:2023",
+    "al2": "public.ecr.aws/amazonlinux/amazonlinux:2"
+}
+
+DOCKER_PLATFORM_BY_ARCH = {
+    "x64": "linux/amd64",
+    "arm64": "linux/arm64"
+}
+
+
+def vprint(message):
+    global VERBOSE
+    if VERBOSE:
+        print(message)
+
+
+def running_in_github_action():
+    """
+    Test if currently running in a GitHub action or running locally
+    :return: True if running in GH, False otherwise
+    """
+    return "GITHUB_WORKFLOW" in os.environ
+
+
+def shell(command, cwd=None, check=True, capture_output=False):
+    """
+    Run a command
+    :param command: command to run
+    :param cwd: the current working directory to change to before executing the command
+    :param check: flag indicating if the status code should be checked. When true an exception will be
+    thrown if the command exits with a non-zero exit status.
+    :returns: the subprocess CompletedProcess output
+    """
+    vprint(f"running `{command}`")
+    return subprocess.run(command, shell=True, check=check, cwd=cwd, capture_output=capture_output)
+
+
+def oci_executable():
+    """
+    Attempt to find the OCI container executor used to build and run docker containers
+    """
+    oci_exe = os.environ.get('OCI_EXE')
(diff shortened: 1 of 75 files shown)


.github/scripts/run-container-test.py
```diff
@@ -0,0 +1,152 @@
...
+    path_to_exe = f'./linux{opts.arch.capitalize()}/debugTest/test.kexe'
+
+    cmd = [
+        oci_exe,
+        'run',
+        '--rm',
+        f'-v{test_bin_dir}:/test',
+    ]
+    if not opts.no_system_certs:
+        cmd.append(f'-v/etc/ssl:/etc/ssl')
```
ianbotsf, line 106: **Nit**: `f` is unnecessary since we're not formatting any parameters in this string

runtime/protocol/http-test/jvmAndNative/src/aws/smithy/kotlin/runtime/httptest/TestWithLocalServer.kt
```diff
@@ -19,7 +18,7 @@ import kotlin.time.Duration.Companion.seconds
  * mocking an HTTP client engine is difficult.
  */
 public abstract class TestWithLocalServer {
-    protected val serverPort: Int = ServerSocket(0).use { it.localPort }
+    protected val serverPort: Int = 54734
```
ianbotsf, line 21: **Question**: This will conflict if any other process (including a parallel build of this package in another workspace) is using port 54734. Is there a reason why it's necessary to use a static port?

runtime/protocol/http-test/jvmAndNative/src/aws/smithy/kotlin/runtime/httptest/TestWithLocalServer.kt
```diff
@@ -33,32 +32,18 @@ public abstract class TestWithLocalServer {
...
-                    println("test server listening on: $testHost:$serverPort")
                     break
                 } catch (cause: Throwable) {
                     if (attempt >= 10) throw cause
-                    Thread.sleep(250L * attempt)
+                    delay(250L * attempt)
                 }
             } while (true)
-
-            ensureServerRunning()
```
ianbotsf, line 44: **Question**: Isn't `server.start()` asynchronous? Won't removing `ensureServerRunning()` mean that it's possible to return from this method before the server is _actually_ running?

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/hashing/Crc32Test.kt
```diff
@@ -20,7 +20,6 @@ class Crc32Test {
         val input = "foobar"
 
         crc.update(input.encodeToByteArray(), 0, input.length)
-        assertEquals(2666930069U, crc.digestValue()) // checksum of "foobar"
         assertEquals("nvYflQ==", crc.digest().encodeBase64String())
```
ianbotsf, line 23: **Question**: Why are we moving the `digestValue` checks from these tests?

lauzadis: Our current JVM implementation of `digestValue()` does not reset the checksum and these tests were depending on that faulty behavior (calling `digestValue()` then `digest()` immediately after).

I've clarified the behavior in KDocs and also updated our JVM implementation. Let me know if you think we should keep the existing (arguably broken) behavior. KotlinNative / CRT will likely be unable to match that behavior though.

runtime/protocol/http-test/jvmAndNative/src/aws/smithy/kotlin/runtime/httptest/TestWithLocalServer.kt
lauzadis: If we don't configure a port, [it will default to `80`](https://api.ktor.io/older/2.2.4/ktor-server/ktor-server-host-common/io.ktor.server.engine/embedded-server.html), we could make it a randomly selected port if you think that will be safer?

ianbotsf: Random ports are marginally safer but still could conflict with an open port. I think we can still bind to an available address using Ktor's low-level KMP socket interfaces:

```kotlin
protected val serverPort: Int = runBlocking {
    SelectorManager(coroutineContext).use { selector ->
        aSocket(selector)
            .tcp()
            .bind()
            .use { (it.localAddress as InetSocketAddress).port }
    }
}
```

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/hashing/Crc32Test.kt
ianbotsf: Oh I see. If this was already broken behavior then you're probably right.

runtime/protocol/http-test/jvmAndNative/src/aws/smithy/kotlin/runtime/httptest/TestWithLocalServer.kt
lauzadis: Good point, I refactored this to use ktor's monitoring API. I was having trouble with a KMP solution to `ensureServerRunning` by making a request.

runtime/protocol/http-test/jvmAndNative/src/aws/smithy/kotlin/runtime/httptest/TestWithLocalServer.kt
lauzadis: Nice, this works great

Issue #1189, opened 2024-11-26
java.lang.NoClassDefFoundError: Failed resolution of: Ljava/time/Duration


lon9man:
### Describe the bug

got error

```
D  Delegated worker class com.appname.data.worker.upload.UploadWorker threw exception in startWork.
java.lang.NoClassDefFoundError: Failed resolution of: Ljava/time/Duration;
at aws.smithy.kotlin.runtime.http.engine.okhttp.OkHttpEngineKt.buildClientFromConfig(OkHttpEngine.kt:98)
at aws.smithy.kotlin.runtime.http.engine.okhttp.OkHttpEngineKt.buildClientWithConnectionListener(OkHttpEngine.kt:159)
at aws.smithy.kotlin.runtime.http.engine.okhttp.OkHttpEngineKt.access$buildClientWithConnectionListener(OkHttpEngine.kt:1)
at aws.smithy.kotlin.runtime.http.engine.okhttp.OkHttpEngine.<init>(OkHttpEngine.kt:48)
at aws.smithy.kotlin.runtime.http.engine.okhttp.OkHttpEngine$Companion.invoke(OkHttpEngine.kt:41)
at aws.smithy.kotlin.runtime.http.engine.DefaultHttpEngineJVMKt.newDefaultHttpEngine(DefaultHttpEngineJVM.kt:14)
at aws.smithy.kotlin.runtime.http.engine.DefaultHttpEngineKt.DefaultHttpEngine(DefaultHttpEngine.kt:24)
at aws.smithy.kotlin.runtime.http.engine.HttpEngineConfigImpl$BuilderImpl$engineConstructor$1.invoke(HttpEngineConfigImpl.kt:20)
at aws.smithy.kotlin.runtime.http.engine.HttpEngineConfigImpl$BuilderImpl$engineConstructor$1.invoke(HttpEngineConfigImpl.kt:20)
at aws.smithy.kotlin.runtime.http.engine.HttpEngineConfigImpl$BuilderImpl.engineSupplier$lambda$2(HttpEngineConfigImpl.kt:21)
at aws.smithy.kotlin.runtime.http.engine.HttpEngineConfigImpl$BuilderImpl.$r8$lambda$FxDGT4mi-_cG_55NQVyO8SbpKzQ(HttpEngineConfigImpl.kt)
[...]

PR #1190, opened 2024-11-27
fix: remove usage of `java.time.Duration` from `OkHttpEngineConfig.buildClientFromConfig`

lauzadis:
## Description of changes


ianbotsf:
Discussed offline as a possible solution to https://github.com/smithy-lang/smithy-kotlin/issues/1189 but, as noted there, it would not address the usage of `Duration` in our `Instant` implementation, which is unavoidable. Making this change alone wouldn't materially affect the compatibility of the Kotlin SDK with earlier Android versions so we should skip this PR for now.

PR #1191, opened 2024-11-27
feat: support default checksums

0marperez:
## Description of changes


runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptor.kt
```diff
@@ -33,57 +34,72 @@ internal val CHECKSUM_HEADER_VALIDATION_PRIORITY_LIST: List<String> = listOf(
...
             .firstOrNull { context.protocolResponse.headers.contains(it) } ?: run {
             logger.warn { "User requested checksum validation, but the response headers did not contain any valid checksums" }
             return context.protocolResponse
         }
+        val serviceChecksumValue = context.protocolResponse.headers[checksumHeader]!!
+
+        if (serviceChecksumValue.isCompositeChecksum()) {
+            logger.debug { "Service returned composite checksum. Skipping validation." }
+            return context.protocolResponse
+        }
```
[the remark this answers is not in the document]
0marperez: It's currently being reviewed, it has the required amount of approvals so I added it

runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptorTest.kt
```diff
@@ -126,23 +133,6 @@ class FlexibleChecksumsRequestInterceptorTest {
...
-
-            val op = newTestOperation<Unit, Unit>(req, Unit)
-            op.context[HttpOperationContext.ChecksumAlgorithm] = checksumAlgorithmName
-            op.interceptors.add(FlexibleChecksumsRequestInterceptor<Unit>())
-
-            op.roundTrip(client, Unit)
-            val call = op.context.attributes[HttpOperationContext.HttpCallList].first()
-            assertEquals(expectedChecksumValue, call.request.headers["x-amz-checksum-$checksumAlgorithmName"])
-        }
-    }
```
ianbotsf, line 144: **Question**: Do we still use `HttpOperationContext.ChecksumAlgorithm` or can it be deprecated?

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptor.kt
```diff
@@ -33,57 +34,72 @@ internal val CHECKSUM_HEADER_VALIDATION_PRIORITY_LIST: List<String> = listOf(
...
-        // let the user know which checksum will be validated
         logger.debug { "Validating checksum from $checksumHeader" }
         context.executionContext[ChecksumHeaderValidated] = checksumHeader
 
-        val checksumAlgorithm = checksumHeader.removePrefix("x-amz-checksum-").toHashFunction() ?: throw ClientException("could not parse checksum algorithm from header $checksumHeader")
+        val checksumAlgorithm = checksumHeader
+            .removePrefix("x-amz-checksum-")
+            .toHashFunction() ?: throw ClientException("Could not parse checksum algorithm from header $checksumHeader")
+
+        if (context.protocolResponse.body is HttpBody.Bytes) {
```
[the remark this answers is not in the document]
0marperez: `toHashingBody` doesn't support `Bytes` bodies. I decided to calculate the checksum in memory instead of adding support for `Bytes` to `toHashingBody`. It should be safe if the request is not being streamed but I think you're right that we're not following the spec

runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptorTest.kt
```diff
@@ -163,29 +168,4 @@ class FlexibleChecksumsResponseInterceptorTest {
...
-            append(responseChecksumHeaderName, "incorrect-checksum-would-throw-if-validated")
-        }
-
-        val client = getMockClient(response, responseHeaders)
-
-        val output = op.roundTrip(client, TestInput("input"))
-        output.body.readAll()
-
-        assertNull(op.context.getOrNull(ChecksumHeaderValidated))
-    }
```
ianbotsf, line 190: **Question**: Why did we delete this test?

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/HttpChecksumClientConfig.kt
```diff
@@ -0,0 +1,40 @@
...
+         */
+        public var responseChecksumValidation: HttpChecksumConfigOption?
+    }
+}
+
+public enum class HttpChecksumConfigOption {
+    /**
+     * SDK will create/validate checksum if the service marks it as required or if this is set.
+     */
+    WHEN_SUPPORTED,
```
ianbotsf, line 34: **Nit**: "or if this is set" → "or if the service offers optional checksums"

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptor.kt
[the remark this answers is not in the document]
lauzadis: > `toHashingBody` doesn't support `Bytes` bodies
Does it need to? The previous implementation never seemed to need it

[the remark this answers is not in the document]
0marperez: I think it does, the spec has some unit tests with non-streaming bodies

runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptorTest.kt
0marperez: This should be covered under ChecksumConfigTest > ResponseChecksumValidation but I added it back

runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptorTest.kt
0marperez: I think it's still used for MD5 checksums in the [httpChecksumRequiredTrait](https://smithy.io/2.0/spec/http-bindings.html#httpchecksumrequired-trait)

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
```diff
@@ -7,102 +7,124 @@ package aws.smithy.kotlin.runtime.http.interceptors
...
         } else {
-            return super.modifyBeforeSigning(context)
+            checksumAlgorithm.update(
+                request.body.readAll() ?: byteArrayOf(),
+            )
+            request.headers[checksumHeader.toString()] = checksumAlgorithm.digest().encodeBase64String()
         }
+
+        context.executionContext.emitBusinessMetric(checksumAlgorithm.toBusinessMetric())
+        request.headers.removeAllChecksumHeadersExcept(checksumHeader.toString())
```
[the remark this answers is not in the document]
0marperez: It can be used given two different cases. The first is for when the checksum is provided via a header, we call it, then return. The second time is for when we computed the checksum ourselves, we call it, then return.

runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptorTest.kt
```diff
@@ -163,29 +168,4 @@ class FlexibleChecksumsResponseInterceptorTest {
 
         op.roundTrip(client, TestInput("input"))
     }
-
-    @Test
-    fun testSkipsValidationWhenDisabled() = runTest {
```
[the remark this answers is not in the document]
0marperez: This should be covered under `ChecksumConfigTest` > `ResponseChecksumValidation` but I added it back

runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptorTest.kt
```diff
@@ -126,23 +133,6 @@ class FlexibleChecksumsRequestInterceptorTest {
         assertEquals(0, call.request.headers.getNumChecksumHeaders())
     }
 
-    @Test
-    fun itSetsChecksumHeaderViaExecutionContext() = runTest {
```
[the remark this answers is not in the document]
0marperez: We're no longer setting the checksum header via execution context

runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptorTest.kt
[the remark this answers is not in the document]
0marperez: The request test cases are already [here](https://github.com/awslabs/aws-sdk-kotlin/blob/flexible-checksums/tests/codegen/checksums/src/commonTest/kotlin/aws/sdk/kotlin/tests/codegen/checksums/ChecksumConfigTest.kt). I think the response validation unit tests need one more to be exhaustive.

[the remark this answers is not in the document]
0marperez: The request test cases are already [here](https://github.com/awslabs/aws-sdk-kotlin/blob/flexible-checksums/tests/codegen/checksums/src/commonTest/kotlin/aws/sdk/kotlin/tests/codegen/checksums/ChecksumConfigTest.kt). I think the response validation unit tests need one more to be exhaustive.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptor.kt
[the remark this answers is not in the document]
0marperez: Discussed offline and we'll be removing support for `HttpBody.Bytes` from `FlexibleChecksumsRequestInterceptor` because HTTP bodies are never bytes. [Source](https://github.com/smithy-lang/smithy-kotlin/blob/main/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpUtils.kt#L104-L120)

runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptorTest.kt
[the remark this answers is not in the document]
ianbotsf: Those tests in **ChecksumConfigTest.kt** aren't validating interceptor behavior like the tests in _this_ file do.

ianbotsf (changes requested):

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptor.kt
```diff
@@ -135,3 +148,13 @@ private fun validateAndThrow(expected: String, actual: String) {
...
+
+/**
+ * Verifies if a checksum is composite.
+ * Composite checksums are used for multipart uploads.
+ */
+private fun String.isCompositeChecksum(): Boolean {
+    // Ends with "-#" where "#" is a number
+    val regex = Regex("-(\\d)+$")
+    return regex.containsMatchIn(this)
+}
```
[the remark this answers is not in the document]
ianbotsf: After reading more of the **aws-sdk-kotlin** PR and re-reading this interceptor, I can see now that this composite checksum usage is gated by a constructor flag. This _works_ but it still seems too specific to a non-**smithy-kotlin** concern to me. I'd be interested in seeing if this logic can be completely removed from **smithy-kotlin** in favor of something more additive in **aws-sdk-kotlin**.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptor.kt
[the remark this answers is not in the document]
ianbotsf: I think removing support for `HttpBody.Bytes` is a bad idea. Our _current_ implementations of OkHttp and CRT engines do not return bytes, that's true, but future implementations or other engines wrapped by users _may_. In addition, interceptors may be used to rewrite response bodies by streaming them into memory, processing them, and then rewriting the body in a new form (possibly `HttpBody.Bytes`). I think we need to ensure each type of `HttpBody` is handled correctly.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
```diff
@@ -7,102 +7,146 @@ package aws.smithy.kotlin.runtime.http.interceptors
...
+        }
 
-                deferredChecksum.complete(checksum)
-            } else {
-                logger.debug { "Calculating checksum asynchronously" }
-                req.body = req.body
-                    .toHashingBody(checksumAlgorithm, req.body.contentLength)
-                    .toCompletingBody(deferredChecksum)
-            }
+        context.executionContext.emitBusinessMetric(checksumAlgorithm.toBusinessMetric())
```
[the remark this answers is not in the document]
0marperez: We're emitting those metrics in the SDK side. [Here](https://github.com/awslabs/aws-sdk-kotlin/blob/5328641644fc341b07e20a12ff11ca5095428bc0/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/flexiblechecksums/FlexibleChecksumsRequest.kt#L54-L69) and [here](https://github.com/awslabs/aws-sdk-kotlin/blob/5328641644fc341b07e20a12ff11ca5095428bc0/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/flexiblechecksums/FlexibleChecksumsResponse.kt#L55-L70).

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptor.kt
```diff
@@ -32,74 +31,72 @@ internal val CHECKSUM_HEADER_VALIDATION_PRIORITY_LIST: List<String> = listOf(
...
- * @param responseValidationRequired Flag indicating if the checksum validation is mandatory.
+ * @param responseValidationRequired Model sourced flag indicating if the checksum validation is mandatory.
  * @param responseChecksumValidation Configuration option that determines when checksum validation should be done.
  */
 @InternalApi
-public class FlexibleChecksumsResponseInterceptor(
+public class FlexibleChecksumsResponseInterceptor<I>(
     private val responseValidationRequired: Boolean,
     private val responseChecksumValidation: HttpChecksumConfigOption?,
+    private val serviceUsesCompositeChecksums: Boolean,
```
lauzadis, line 49: question: Why do we need this extra flag to tell if the operation **might** send back a composite checksum? Why can't we just rely on `serviceChecksumValue.isCompositeChecksum()`?

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
```diff
@@ -36,65 +36,85 @@ import kotlin.coroutines.coroutineContext
...
+    requestChecksumAlgorithm: String?,
 ) : AbstractChecksumInterceptor() {
-    private val forcedToCalculateChecksum = requestChecksumRequired || requestChecksumCalculation == HttpChecksumConfigOption.WHEN_SUPPORTED
-    private val checksumHeader = StringBuilder("x-amz-checksum-")
-    private val defaultChecksumAlgorithm = lazy { Crc32() }
-    private val defaultChecksumAlgorithmHeaderPostfix = "crc32"
 
-    private val checksumAlgorithm = userSelectedChecksumAlgorithm?.let {
-        val hashFunction = userSelectedChecksumAlgorithm.toHashFunction()
+    // FIXME: Remove in next minor version bump
```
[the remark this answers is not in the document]
0marperez: Sure, SDK-KT-385 seems to be more for SDK changes. I can create a ticket for the smith kotlin version bump since it seems like we'll do a minor version bump for the SDK soon and one for smithy kotlin later at a different time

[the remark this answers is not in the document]
lauzadis: I'm pretty sure we will bump both at the same time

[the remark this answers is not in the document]
0marperez: Okay, in that case I'll just fix the `FIXME`s since this should go into a `v1.4` branch. I have to update the PR.

runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptorTest.kt
0marperez: Update on this, it's no longer used anywhere so we can deprecate/remove it

codegen/protocol-tests/model/error-correction-tests.smithy
```diff
@@ -60,7 +64,7 @@ structure TestStruct {
     @required
     nestedListValue: NestedList
 
-    @required
+    // @required
```
lauzadis, line 67: Were these `@required` removed to fix some protocol tests? I'd rather leave them disabled than change our test models

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpProtocolClientGenerator.kt
```diff
@@ -338,21 +337,14 @@ open class HttpProtocolClientGenerator(
 
     /**
      * Render optionally installing Md5ChecksumMiddleware.
-     * The Md5 middleware will only be installed if the operation requires a checksum and the user has not opted-in to flexible checksums.
+     * The Md5 middleware will only be installed if the operation requires a checksum.
      */
```
lauzadis, line 341: question: Why is this installed for every operation now? Doesn't CRC32 get used by default?

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
```diff
@@ -172,6 +172,11 @@ public class AwsSigningConfig(builder: Builder) {
      */
     public val logRequest: Boolean = builder.logRequest
 
+    /**
+     * Determines the checksum to add to the canonical request query parameters before signing.
```
lauzadis, line 176: docs: KDocs should be more descriptive about the contents of the `Pair`, what is first, second?

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
```diff
@@ -7,102 +7,135 @@ package aws.smithy.kotlin.runtime.http.interceptors
...
+ * Handles request checksums for operations with the [HttpChecksumTrait] applied.
  *
- * If the checksum will be sent as a header, calculate the checksum.
+ * If a user supplies a checksum via an HTTP header no calculation will be done. The exception is MD5, if a user
+ * supplies an MD5 checksum header it will be ignored.
  *
- * Otherwise, if it will be sent as a trailing header, calculate the checksum as asynchronously as the body is streamed.
- * In this case, a [LazyAsyncValue] will be added to the execution context which allows the trailing checksum to be sent
- * after the entire body has been streamed.
+ * If the request configuration and model requires checksum calculation:
```
lauzadis, line 34: These KDocs don't describe what happens if the request already has a checksum configured (not a checksum _algorithm_, the checksum itself)

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
```diff
@@ -7,102 +7,135 @@ package aws.smithy.kotlin.runtime.http.interceptors
...
+        checksumAlgorithmNameInitializer: ((I) -> String?)? = null,
+    ) : this(
+        false,
+        HttpChecksumConfigOption.WHEN_REQUIRED,
+        null,
+    )
+
+    private val checksumHeader = buildString {
+        append("x-amz-checksum-")
+        append(requestChecksumAlgorithm?.lowercase() ?: "crc32")
```
lauzadis, line 79: style: "crc32" could be lifted to a `const val SDK_DEFAULT_CHECKSUM_ALGORITHM` for easier understanding

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
```diff
@@ -7,102 +7,135 @@ package aws.smithy.kotlin.runtime.http.interceptors
...
+            request.headers.removeAllChecksumHeadersExcept(it)
+            return request.build()
         }
 
-        val req = context.protocolRequest.toBuilder()
-
-        check(context.protocolRequest.body !is HttpBody.Empty) {
-            "Can't calculate the checksum of an empty body"
+        if (checksumAlgorithm == null) {
+            logger.debug { "A checksum algorithm isn't selected and checksum calculation isn't required, skipping checksum calculation" }
```
lauzadis, line 103: We could also fall into this case when a user _does_ select a checksum algorithm but they also configure `WHEN_REQUIRED`. 

I think a better log would be:
"A checksum algorithm isn't selected **or** checksum calculation isn't required, skipping checksum calculation"

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptor.kt
```diff
@@ -31,59 +32,84 @@ internal val CHECKSUM_HEADER_VALIDATION_PRIORITY_LIST: List<String> = listOf(
...
-
-        // Wrap the response body in a hashing body
-        return context.protocolResponse.copy(
-            body = context.protocolResponse.body
-                .toHashingBody(checksumAlgorithm, context.protocolResponse.body.contentLength)
-                .toChecksumValidatingBody(context.protocolResponse.headers[checksumHeader]!!),
-        )
+        val checksumAlgorithm = checksumHeader
+            .removePrefix("x-amz-checksum-")
+            .toHashFunction() ?: throw ClientException("Could not parse checksum algorithm from header $checksumHeader")
```
lauzadis, line 84: Should use `toHashFunctionOrThrow` here

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptor.kt
```diff
@@ -31,59 +32,84 @@ internal val CHECKSUM_HEADER_VALIDATION_PRIORITY_LIST: List<String> = listOf(
...
+            is HttpBody.SourceContent, is HttpBody.ChannelContent -> {
+                logger.debug { "Validating checksum during deserialization from $checksumHeader" }
+
+                return context.protocolResponse.copy(
+                    body = context.protocolResponse.body
+                        .toHashingBody(checksumAlgorithm, context.protocolResponse.body.contentLength)
+                        .toChecksumValidatingBody(serviceChecksumValue),
+                )
+            }
+            else -> throw IllegalStateException("Http body type '$bodyType' is not supported for flexible checksums.")
```
lauzadis, line 111: nit: capitalize HTTP or make it `HttpBody`

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/Exceptions.kt
```diff
@@ -174,3 +175,9 @@ public open class ServiceException : SdkBaseException {
 
     override val sdkErrorMetadata: ServiceErrorMetadata = ServiceErrorMetadata()
 }
+
+/**
+ * Runtime accessible version of the [IllegalStateException]
+ */
+@InternalApi
+public class IllegalStateException(message: String) : IllegalStateException(message)
```
lauzadis, line 183: What's the purpose of this, why is not accessible from runtime already?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/text/Text.kt
```diff
@@ -6,9 +6,13 @@
 package aws.smithy.kotlin.runtime.text
 
 import aws.smithy.kotlin.runtime.InternalApi
+import java.util.*
```
lauzadis, line 9: correctness: You can't use `java` imports in a common source set

codegen/protocol-tests/model/error-correction-tests.smithy
0marperez: The required trait isn't supposed to be there, a smithy validator throw an error if it is, even if I disable the protocol test

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
0marperez: I think this case is covered under:
> If a user supplies a checksum via an HTTP header no calculation will be done.
 Is there something else you want to see added?

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
```diff
@@ -5,131 +5,155 @@
...
-            }
 
-            req.trailingHeaders.append(headerName, deferredChecksum)
-            return req.build()
-        } else {
-            return super.modifyBeforeSigning(context)
+                request.headers.append("x-amz-trailer", checksumHeader)
+                request.trailingHeaders.append(checksumHeader, deferredChecksum)
+                context.executionContext.emitBusinessMetric(checksumAlgorithm.toBusinessMetric())
+            } else { // Delegate checksum calculation to super class
```
lauzadis, line 87: > // Delegate checksum calculation to super class

This comment is a little misleading, it's not actually going to delegate to the super class, it's going to use the `calculateChecksum` and `applyChecksum` functions you defined below.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
```diff
@@ -139,75 +163,50 @@ public class FlexibleChecksumsRequestInterceptor<I>(
...
-    ) : SdkSource by hashingSource {
-        override fun read(sink: SdkBuffer, limit: Long): Long = hashingSource.read(sink, limit)
-            .also {
-                if (it == -1L) {
-                    deferred.complete(hashingSource.digest().encodeBase64String())
+    private fun HttpRequest.userProvidedChecksumHeader(logger: Logger) = headers
+        .names()
+        .firstOrNull {
+            it.startsWith("x-amz-checksum-", ignoreCase = true) &&
+                !it.equals("x-amz-checksum-md5", ignoreCase = true).also { itEqualsMd5 ->
```
lauzadis, line 187: naming: `itEqualsMd5` -> `isMd5` or just use `it`

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
```diff
@@ -139,75 +163,50 @@ public class FlexibleChecksumsRequestInterceptor<I>(
...
+        is Sha1 -> SmithyBusinessMetric.FLEXIBLE_CHECKSUMS_REQ_SHA1
+        is Sha256 -> SmithyBusinessMetric.FLEXIBLE_CHECKSUMS_REQ_SHA256
+        else -> throw IllegalStateException("Checksum was calculated using an unsupported hash function: ${this::class.simpleName}")
     }
 
     /**
-     * Compute the rolling hash of an [SdkByteReadChannel] using [hashFunction], reading up-to [bufferSize] bytes into memory
-     * @return a ByteArray of the hash function's digest
+     * Removes all checksum headers except specified header
      */
```
lauzadis, line 207: nit/docs: previous KDocs for `removeAllChecksumHeadersExcept` seemed better:

```
    /**
     * Removes all checksum headers except [headerName]
     * @param headerName the checksum header name to keep
     */
```

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptor.kt
```diff
@@ -31,60 +32,85 @@ internal val CHECKSUM_HEADER_VALIDATION_PRIORITY_LIST: List<String> = listOf(
...
+
+                validateAndThrow(
+                    serviceChecksumValue,
+                    sdkChecksumValue,
+                )
+
+                return context.protocolResponse
+            }
+            is HttpBody.SourceContent, is HttpBody.ChannelContent -> {
+                logger.debug { "Validating checksum during deserialization from $checksumHeader" }
```
lauzadis, line 98: misleading: It's validated **after** deserialization, after the user is done reading the entire body.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HttpBody.kt
```diff
@@ -191,6 +193,49 @@ public fun HttpBody.toHashingBody(
     else -> throw ClientException("HttpBody type is not supported")
 }
 
+/**
+ * Convert an [HttpBody] with an underlying [HashingSource] or [HashingByteReadChannel]
+ * to a [CompletingSource] or [CompletingByteReadChannel], respectively.
+ */
+@InternalApi
+public fun HttpBody.toCompletingBody(deferred: CompletableDeferred<String>): HttpBody = when (this) {
```
lauzadis, line 201: These used to be `private`, now they are `@InternalApi public`, meaning they are included in our API dumps and part of backwards compatibility. Was this move necessary?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/hashing/HashFunction.kt
```diff
@@ -70,3 +71,60 @@ public fun String.toHashFunction(): HashFunction? = when (this.lowercase()) {
...
+@InternalApi
+public fun String.toHashFunctionOrThrow(): HashFunction =
+    toHashFunction() ?: throw ClientException("Checksum algorithm is not supported: $this")
+
+/**
+ * @return If the [HashFunction] is supported by flexible checksums
+ */
+@InternalApi
+public val HashFunction.isSupportedForFlexibleChecksums: Boolean get() =
+    algorithmsSupportedForFlexibleChecksums.contains(this::class.simpleName)
```
lauzadis, line 87: correctness: Relying on `this::class.simpleName` seems unsafe. 

Better solution:
```
@InternalApi
public val HashFunction.isSupportedForFlexibleChecksums: Boolean 
    get() = when (this) {
        is Crc32, is Crc32c, is Sha1, is Sha256 -> true
        else -> false
    }
```

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Concurrency.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import aws.smithy.kotlin.runtime.InternalApi
+import kotlinx.coroutines.runBlocking
+
+@InternalApi
+public fun runBlocking(block: suspend () -> Unit) {
+    runBlocking {
+        block()
+    }
+}
```
lauzadis, line 11: question: What is this used for?

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/HttpChecksumClientConfig.kt
```diff
@@ -0,0 +1,40 @@
+package aws.smithy.kotlin.runtime.client.config
```
lauzadis, line 1: correctness: missing license at the top. same applies to `Concurrency.kt`

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/HttpChecksumClientConfig.kt
```diff
@@ -0,0 +1,40 @@
...
+        public var requestChecksumCalculation: HttpChecksumConfigOption?
+
+        /**
+         * Configures response checksum validation
+         */
+        public var responseChecksumValidation: HttpChecksumConfigOption?
+    }
+}
+
+public enum class HttpChecksumConfigOption {
```
lauzadis, line 30: docs: missing KDocs

lauzadis (dismissed):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/hashing/HashFunction.kt
0marperez: Yeah agree, that's what we had before but we would have to keep track of what checksum algorithms are supported for flexible checksums in two places. In `HashFunction.isSupportedForFlexibleChecksums` & `algorithmsSupportedForFlexibleChecksums`. I made a tradeoff here, it seems more risky to possibly miss updating what algorithms we support in both locations than relying on the class name. Plus we have unit tests that should lessen some of our concerns here

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/HttpChecksumClientConfig.kt
lauzadis, line 30: correctness: Right now request and response checksum configs both use this enum. In the future, their config options might diverge, which will force us to make an API break. I'd recommend splitting this into `HttpChecksumRequestConfigOption` and `HttpChecksumResponseConfigOption` now to get ahead of that.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HttpBody.kt
0marperez: Oops, I moved this here because I was going to add support for streaming checksums in the `HttpCheksumsRequired` interceptor

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/ShapeExt.kt
```diff
@@ -32,7 +33,7 @@ import software.amazon.smithy.rulesengine.traits.EndpointTestsTrait
  * shape's closure for example)
  */
 @Suppress("EXTENSION_SHADOWED_BY_MEMBER")
-inline fun <reified T : Shape> Model.shapes(): List<T> = shapes(T::class.java).toList()
+inline fun <reified T : Shape> Model.shapes(): List<T> = shapes(T::class.java).kotlinToList()
```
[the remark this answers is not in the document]
0marperez: There was a refactor in AWS SDK Kotlin to make the codegen tests KMP, which led to the JVM CI checks failing because we've been using JVM 17. I set the JVM version in the tests to 1.8 and the `toList` function used above is only available since JVM `16`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/checksums/HttpChecksumRequiredIntegration.kt
```diff
@@ -0,0 +1,66 @@
...
+import software.amazon.smithy.kotlin.codegen.integration.KotlinIntegration
+import software.amazon.smithy.kotlin.codegen.model.hasTrait
+import software.amazon.smithy.kotlin.codegen.rendering.protocol.ProtocolGenerator
+import software.amazon.smithy.kotlin.codegen.rendering.protocol.ProtocolMiddleware
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.model.traits.HttpChecksumRequiredTrait
+
+/**
+ * Handles the `httpChecksumRequired` trait.
```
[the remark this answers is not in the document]
0marperez: The "meat" of the integration is the middleware and both of them have their own Kdocs. It feels like docs overkill to also add a summary of what each middleware is doing here

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
```diff
@@ -5,131 +5,155 @@
...
-        }
-
-        return req.build()
+        val request = context.protocolRequest.toBuilder()
+        val checksumAlgorithm = checksumAlgorithm(
+            requestChecksumRequired,
+            requestChecksumCalculation,
+            requestChecksumAlgorithm,
+            context,
+        )!!
```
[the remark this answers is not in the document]
0marperez: No, `applyChecksum` is used by `super.modifyBeforeSigning(context)` which is only called when the `checksumAlgorithm` isn't null

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Concurrency.kt
0marperez: It's so we can have `runBlocking` as a runtime type, it's used in the presigner generator

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/HttpChecksumClientConfig.kt
0marperez: Thanks, I think we need to adjust our Ktlint for be more stringent when it comes to license headers

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/HttpChecksumClientConfig.kt
0marperez: It seems like just speculation for now, I say we only separate them when that time comes, changing the type of this config option would be a breaking change for all SDKs. I think it's more likely that we would make additive changes

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/hashing/HashFunction.kt
lauzadis: The only other use of `algorithmsSupportedForFlexibleChecksums` is in a log message. I'd rather remove the log message than keep this class reflection

runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/HttpChecksumClientConfig.kt
lauzadis: It's not speculative, there was some discussion during the spec review about adding more options. 

The spec calls out `RequestChecksumCalculation` and `ResponseChecksumValidation` as separate config types, so adding a new config option to either of those wouldn't be a breaking change for SDKs which implement it as the spec says.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Concurrency.kt
lauzadis: You can just add a new `runBlocking` type under the existing `KotlinxCoroutines` runtime type

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/ShapeExt.kt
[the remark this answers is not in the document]
lauzadis: Ok that makes sense. You should be able to use `shapes(T::class.java).collect(Collectors.toList())` instead of hacking around with the Kotlin `toList`

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/AbstractChecksumInterceptor.kt
```diff
@@ -7,18 +7,33 @@ package aws.smithy.kotlin.runtime.http.interceptors
...
     public abstract suspend fun calculateChecksum(context: ProtocolRequestInterceptorContext<Any, HttpRequest>): String?
 
     public abstract fun applyChecksum(context: ProtocolRequestInterceptorContext<Any, HttpRequest>, checksum: String): HttpRequest
 }
+
+/**
+ * @return The default checksum algorithm name, null if default checksums are disabled.
+ */
+internal fun defaultChecksumAlgorithmName(context: ProtocolRequestInterceptorContext<Any, HttpRequest>): String? =
+    context.executionContext.getOrNull(HttpOperationContext.DefaultChecksumAlgorithm)
```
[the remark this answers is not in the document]
0marperez: Where would be better?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/hashing/HashFunction.kt
```diff
@@ -70,3 +71,60 @@ public fun String.toHashFunction(): HashFunction? = when (this.lowercase()) {
...
+    val prefix = "x-amz-checksum-"
+    return when (checksumAlgorithm.lowercase()) {
+        "crc32" -> prefix + "crc32"
+        "crc32c" -> prefix + "crc32c"
+        "sha1" -> prefix + "sha1"
+        "sha256" -> prefix + "sha256"
+        "md5" -> "Content-MD5"
+        else -> throw ClientException("Checksum algorithm is not supported: ${checksumAlgorithm::class.simpleName}")
+    }
+}
```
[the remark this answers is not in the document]
0marperez: Do you mean why we have a function for each one or why isn't the logic shared? Maybe something else?

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
0marperez: What do you think would be a better comment?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/hashing/HashFunction.kt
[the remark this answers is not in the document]
lauzadis: I meant both: why do we need to have a function for each one, and if we _do_ need a function for each one, why isn't the logic shared?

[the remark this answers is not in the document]
0marperez: We don't strictly need to have a function for each one but it saves us from doing `String` -> `HashFunction` when applying the checksum in the `HttpChecksumRequiredInterceptor`. It was just an oversight not to share the logic, working on a fix...

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/checksums/HttpChecksumRequiredIntegration.kt
```diff
@@ -0,0 +1,66 @@
...
+    override val order: Byte = -2 // Before S3 Express (possibly) changes the default (-1) and before calculating checksum (0)
+
+    override fun isEnabledFor(ctx: ProtocolGenerator.GenerationContext, op: OperationShape): Boolean =
+        op.hasTrait<HttpChecksumRequiredTrait>() && !op.hasTrait<HttpChecksumTrait>()
+
+    override fun render(ctx: ProtocolGenerator.GenerationContext, op: OperationShape, writer: KotlinWriter) {
+        writer.write(
+            "op.context[#T.DefaultChecksumAlgorithm] = #S",
+            RuntimeTypes.HttpClient.Operation.HttpOperationContext,
+            "MD5",
```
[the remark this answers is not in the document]
0marperez: We could, I think we would have to initialize the `HashFunction` in the context and its seems slightly worse than what we have right now. Thoughts?

ianbotsf (changes requested):

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/HttpChecksumRequiredInterceptor.kt
```diff
@@ -0,0 +1,54 @@
...
+
+        return when (val body = context.protocolRequest.body) {
+            is HttpBody.Bytes -> {
+                checksumAlgorithm.update(
+                    body.readAll() ?: byteArrayOf(),
+                )
+                checksumAlgorithm.digest().encodeBase64String()
+            }
+            else -> null // TODO: Support other body types
+        }
```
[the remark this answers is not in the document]
0marperez: Yeah, we can use those types. A small refactor is needed and I think deadlines shifting left me uncertain if I could get this functionality supported on time. I think we might as well do it now since the refactor will be a breaking change

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt
```diff
@@ -7,129 +7,154 @@ package aws.smithy.kotlin.runtime.http.interceptors
...
                 val channel = req.body.toSdkByteReadChannel()!!
                 channel.rollingHash(checksumAlgorithm).encodeBase64String()
             }
             else -> {
-                val bodyBytes = req.body.readAll()!!
-                req.body = bodyBytes.toHttpBody()
+                val bodyBytes = req.body.readAll() ?: byteArrayOf()
+                if (req.body.isOneShot) req.body = bodyBytes.toHttpBody()
                 bodyBytes.hash(checksumAlgorithm).encodeBase64String()
             }
```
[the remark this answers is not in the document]
0marperez: Yeah, it seemed like we could improve performance a bit here but I hadn't considered what you said. I don't have a strong inclination towards keeping the optimization fwiw

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/AbstractChecksumInterceptor.kt
```diff
@@ -7,18 +7,33 @@ package aws.smithy.kotlin.runtime.http.interceptors
...
-        cachedChecksum ?: calculateChecksum(context).also { cachedChecksum = it }
-        return cachedChecksum?.let { applyChecksum(context, it) } ?: context.protocolRequest
+        cachedChecksum = cachedChecksum ?: calculateChecksum(context)
+
+        return if (cachedChecksum != null) {
+            applyChecksum(context, cachedChecksum!!)
+        } else {
+            context.protocolRequest
+        }
     }
```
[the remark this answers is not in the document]
0marperez: It was originally added for performance reasons. There's a discussion about it [here](https://github.com/smithy-lang/smithy-kotlin/pull/1033#discussion_r1490070241). In your example someone/something would have to modify the underlying file in a file stream between retries. I'm not sure how often that happens for our customers but I'm open to getting rid of it.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/HttpChecksumRequiredInterceptor.kt
```diff
@@ -12,43 +12,87 @@ import aws.smithy.kotlin.runtime.http.*
...
-            is HttpBody.Bytes -> {
-                checksumAlgorithm.update(
-                    body.readAll() ?: byteArrayOf(),
-                )
-                checksumAlgorithm.digest().encodeBase64String()
+        return when {
+            req.body.contentLength == null && !req.body.isOneShot -> {
+                val channel = req.body.toSdkByteReadChannel()!!
+                channel.rollingHash(checksumAlgorithm).encodeBase64String()
+            }
```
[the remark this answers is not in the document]
0marperez: I'm assuming it's a requirement for `toSdkByteReadChannel` and `rollingHash`. This code is originally from [here](https://github.com/smithy-lang/smithy-kotlin/blob/0bba3084ebb3160abe2a21fc86f4223c7bf37c23/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsRequestInterceptor.kt#L108)

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/FlexibleChecksumsResponseInterceptor.kt
```diff
@@ -110,7 +110,7 @@ public open class FlexibleChecksumsResponseInterceptor(
     /**
      * Additional check on the checksum itself to see if it should be validated
      */
-    public open fun ignoreChecksum(checksum: String): Boolean = false
+    public open fun ignoreChecksum(checksum: String, logger: Logger): Boolean = false
```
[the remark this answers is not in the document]
0marperez: Good suggestion, let me do this before I merge

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/HttpChecksumRequiredInterceptor.kt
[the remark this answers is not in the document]
ianbotsf: Ah, this was discussed in https://github.com/smithy-lang/smithy-kotlin/pull/772#discussion_r1067479535. That helps add some context.

PR #1197, opened 2024-12-17
misc: enhance support for replayable instances of `InputStream`

lauzadis:
## Description of changes


runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/ByteStreamJVM.kt
```diff
@@ -114,11 +114,25 @@ public fun ByteStream.Companion.fromInputStream(
...
+                return object : SdkSource by source() {
+                    // no-op close
+                    override fun close() { }
+                }
+            }
+
+            return source()
+        }
     }
 }
```
ianbotsf, line 137: **Question**: Is this a behavioral change wrt closing? It seems like before we returned the raw underlying `SdkSource` before which, if closed, would close the underlying `InputStream`. Now it looks like that won't happen...why is that necessary?

lauzadis: Yes it is. Without this change the underlying `InputStream` is closed after hashing (because we call a `.use { }` on the `SdkSource`) and subsequent attempts to read the body (e.g. during transmission) fail with `IOException: Stream closed`

0marperez (approved):

PR #1200, opened 2024-12-19
feat: Kotlin/Native implementation of BigDecimal and BigInteger

ianbotsf:
## Issue \#

(none)

## Description of changes

This change implements Kotlin/Native support for `BigInteger`/`BigDecimal` by way of the [IonSpin **kotlin-multiplatform-bignum**](https://github.com/ionspin/kotlin-multiplatform-bignum) library. This change adds a workaround for building **smithy-kotlin** on Amazon Linux 2 because of linking issues with zlib (see **[runtime/build.gradle.kts](https://github.com/smithy-lang/smithy-kotlin/compare/kn-main...kn-bignums?expand=1#diff-23f01691d6dba0dd32c4e07f9490975ea4bfe9c0a29a6a004fd8b51c7a952c1a)**).

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,6 +2,7 @@ kotlin.code.style=official
 kotlin.incremental.js=true
 kotlin.incremental.multiplatform=true
 kotlin.mpp.stability.nowarn=true
+kotlin.native.binary.sourceInfoType=libbacktrace
 kotlin.native.ignoreDisabledTargets=true
 
 # atomicfu
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,7 +2,7 @@ kotlin.code.style=official
 kotlin.incremental.js=true
 kotlin.incremental.multiplatform=true
 kotlin.mpp.stability.nowarn=true
-kotlin.native.binary.sourceInfoType=libbacktrace
+# kotlin.native.binary.sourceInfoType=libbacktrace
 kotlin.native.ignoreDisabledTargets=true
 
 # atomicfu
--- a/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigDecimalJVM.kt
+++ b/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigDecimalJVM.kt
@@ -4,33 +4,52 @@
  */
 package aws.smithy.kotlin.runtime.content
 
-public actual class BigDecimal actual constructor(public val value: String) :
+import java.math.BigDecimal as JvmBigDecimal
+
+public actual class BigDecimal private constructor(private val delegate: JvmBigDecimal) :
     Number(),
     Comparable<BigDecimal> {
-    private val delegate = java.math.BigDecimal(value)
 
-    public actual constructor(mantissa: BigInteger, exponent: Int) : this(
-        java.math.BigDecimal(
-            java.math.BigInteger(mantissa.toString()),
-            exponent,
-        ).toPlainString(),
-    )
+    private companion object {
+        /**
+         * Returns a new or existing [BigDecimal] wrapper for the given delegate [value]
+         * @param value The delegate value to wrap
+         * @param left A candidate wrapper which may already contain [value]
+         * @param right A candidate wrapper which may already contain [value]
+         */
+        fun coalesceOrCreate(value: JvmBigDecimal, left: BigDecimal, right: BigDecimal): BigDecimal = when (value) {
+            left.delegate -> left
+            right.delegate -> right
+            else -> BigDecimal(value)
+        }
+    }
+
+    public actual constructor(value: String) : this(JvmBigDecimal(value))
+    public actual constructor(mantissa: BigInteger, exponent: Int) : this(JvmBigDecimal(mantissa.delegate, exponent))
 
     public actual fun toPlainString(): String = delegate.toPlainString()
-    actual override fun toByte(): Byte = delegate.toByte()
-    actual override fun toDouble(): Double = delegate.toDouble()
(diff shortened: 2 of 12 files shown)


gradle.properties
```diff
@@ -2,6 +2,7 @@ kotlin.code.style=official
 kotlin.incremental.js=true
 kotlin.incremental.multiplatform=true
 kotlin.mpp.stability.nowarn=true
+kotlin.native.binary.sourceInfoType=libbacktrace
```
lauzadis, line 5: Were you able to get better stack traces with this enabled? I remember you said it didn't help, so I'd rather leave it off if it's not working as expected

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigDecimalJVM.kt
```diff
@@ -4,33 +4,52 @@
...
+    public actual override fun toByte(): Byte = delegate.toByte()
+    public actual override fun toDouble(): Double = delegate.toDouble()
+    public actual override fun toFloat(): Float = delegate.toFloat()
+    public actual override fun toInt(): Int = delegate.toInt()
+    public actual override fun toLong(): Long = delegate.toLong()
+    public actual override fun toShort(): Short = delegate.toShort()
 
-    actual override fun equals(other: Any?): Boolean = other is BigDecimal && delegate == other.delegate
+    public actual override fun equals(other: Any?): Boolean = other is BigDecimal && delegate == other.delegate
+    public actual override fun hashCode(): Int = 31 + delegate.hashCode()
```
lauzadis, line 40: Usually 31 is multiplied in these hashCode implementations, is this intentional addition? Is there any reason we can't just take the delegate's hash code?

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigDecimalJVM.kt
```diff
@@ -4,33 +4,52 @@
...
+        get() = BigInteger(delegate.unscaledValue())
 
     public actual val exponent: Int
         get() = delegate.scale()
 
+    public actual operator fun plus(other: BigDecimal): BigDecimal =
+        coalesceOrCreate(delegate + other.delegate, this, other)
+
+    public actual operator fun minus(other: BigDecimal): BigDecimal =
+        coalesceOrCreate(delegate - other.delegate, this, other)
```
lauzadis, line 52: Nice 👍

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigIntegerJVM.kt
```diff
@@ -4,25 +4,47 @@
...
     public actual override fun toShort(): Short = delegate.toShort()
     public actual override fun toInt(): Int = delegate.toInt()
     public actual override fun toFloat(): Float = delegate.toFloat()
     public actual override fun toDouble(): Double = delegate.toDouble()
-    public actual override fun toString(): String = delegate.toString()
-    public actual override fun hashCode(): Int = delegate.hashCode()
+    public actual override fun toString(): String = toString(10)
+    public actual fun toString(radix: Int): String = delegate.toString(radix)
+
+    public actual override fun hashCode(): Int = 17 + delegate.hashCode()
```
lauzadis, line 39: Same question about adding 17, can we just take the delegate's hash code?

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/content/BigDecimalNative.kt
```diff
@@ -4,46 +4,53 @@
...
+         * @param value The delegate value to wrap
+         * @param left A candidate wrapper which may already contain [value]
+         * @param right A candidate wrapper which may already contain [value]
+         */
+        fun coalesceOrCreate(value: IonSpinBigDecimal, left: BigDecimal, right: BigDecimal): BigDecimal = when (value) {
+            left.delegate -> left
+            right.delegate -> right
+            else -> BigDecimal(value)
+        }
     }
```
lauzadis, line 25: I feel like this `coalesceOrCreate` might be a candidate for a generic function

lauzadis (approved):

0marperez (approved):
Nothing to add beyond what Matas said

gradle.properties
ianbotsf: Good point, will remove. I eventually discovered that full stack traces _are_ available in the build artifacts (e.g., **build/reports/tests/linuxX64Test/index.html**) just not in the console output.

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigDecimalJVM.kt
ianbotsf: Yes, I intentionally used addition instead of multiplication although either should work. I wanted to avoid using the unmodified hash code from the delegate because that would bin them the same if they were both added to a hashed container (e.g., `Set`). That's unlikely but also trivial to avoid.

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/content/BigDecimalNative.kt
ianbotsf: That'd be nice but our `BigInteger` and `BigDecimal` definitions don't have a common interface which defines `val delegate` as a public member. I'd rather not add one either since the delegate is an internal implementation detail.

PR #1212, opened 2025-01-09
feat: add `AuthTokenGenerator`

lauzadis:
## Description of changes

### DIFF
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AuthTokenGenerator.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AuthTokenGenerator.kt
@@ -30,12 +30,10 @@ public class AuthTokenGenerator(
     public suspend fun generateAuthToken(endpoint: Url, region: String, expiration: Duration): String {
         val req = HttpRequest(HttpMethod.GET, endpoint)
 
-        val serv = service
-
         val config = AwsSigningConfig {
             credentials = credentialsProvider.resolve()
             this.region = region
-            service = serv
+            service = this@AuthTokenGenerator.service
             signingDate = clock.now()
             expiresAfter = expiration
             signatureType = AwsSignatureType.HTTP_REQUEST_VIA_QUERY_PARAMS
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AuthTokenGenerator.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AuthTokenGenerator.kt
@@ -16,9 +16,6 @@ import kotlin.time.Duration
 import kotlin.time.Duration.Companion.minutes
 import kotlin.time.Duration.Companion.seconds
 
-// The default expiration value to use for [Credentials] when none is provided.
-private val DEFAULT_CREDENTIALS_EXPIRATION = 10.minutes
-
 /**
  * Generates an authentication token, which is a SigV4-signed URL with the HTTP scheme removed.
  * @param service The name of the service the token is being generated for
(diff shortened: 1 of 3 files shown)


runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AuthTokenGenerator.kt
```diff
@@ -0,0 +1,63 @@
...
+
+        val req = HttpRequest(HttpMethod.GET, endpoint)
+
+        val creds = credentials.value
+        val serv = service
+
+        val config = AwsSigningConfig {
+            credentials = creds
+            this.region = region
+            service = serv
```
0marperez, line 55: Nit: You can do `this.service = service` to avoid creating `serv`. Same for credentials: `this.credentials = credentials.value`

lauzadis: No it's not possible, I tried already. It's because those `service` and `credentials` are class parameters not function parameters

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AuthTokenGenerator.kt
```diff
@@ -0,0 +1,63 @@
...
+import aws.smithy.kotlin.runtime.http.request.HttpRequest
+import aws.smithy.kotlin.runtime.net.url.Url
+import aws.smithy.kotlin.runtime.time.Clock
+import aws.smithy.kotlin.runtime.util.ExpiringValue
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.minutes
+import kotlin.time.Duration.Companion.seconds
+
+// The default expiration value to use for [Credentials] when none is provided.
+private val DEFAULT_CREDENTIALS_EXPIRATION = 10.minutes
```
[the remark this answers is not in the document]
lauzadis: Yeah it can be, good point

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AuthTokenGenerator.kt
```diff
@@ -0,0 +1,63 @@
...
+) {
+    private lateinit var credentials: ExpiringValue<Credentials>
+
+    private fun Url.trimScheme(): String = toString().removePrefix(scheme.protocolName).removePrefix("://")
+
+    public suspend fun generateAuthToken(endpoint: Url, region: String, expiration: Duration): String {
+        if (!::credentials.isInitialized || (credentials.expiresAt - clock.now()).absoluteValue <= credentialsRefreshBuffer) {
+            val resolved = credentialsProvider.resolve()
+            credentials = ExpiringValue(resolved, resolved.expiration ?: (clock.now() + DEFAULT_CREDENTIALS_EXPIRATION))
+        }
```
ianbotsf, line 45: **Question**: Why duplicate this logic from our caching layer? Why not simply call the `resolve` method on the provider like we do in regular signing?

ianbotsf (changes requested):
Missing unit tests

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AuthTokenGenerator.kt
lauzadis: Shouldn't we try to cache credentials instead of resolving them for every call to `generateAuthToken`?

lauzadis: I guess the default chain will cache credentials, but not if a user passes in a custom credentials provider.

ianbotsf: Yes, people who use a custom creds provider without caching will face the same behaviors in regular requests and in auth token generation, not all of which are necessarily negative. If someone intentionally _wants_ fresh credentials every time or they have their _own_ caching mechanism, we'll bypass that by adding in a new caching layer. We've got a clean, easy way for users to wrap their custom provider chains/impls in our caching provider which should be good enough.

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AuthTokenGenerator.kt
lauzadis: I thought about it some more and was able to get this to work: `service = this@AuthTokenGenerator.service`

ianbotsf (approved):

PR #1214, opened 2025-01-09
kn: implement `SdkBuffer`

lauzadis:
## Description of changes


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/io/SdkSource.kt
```diff
@@ -43,16 +51,42 @@ public interface SdkSource : Closeable {
  * Consume the [SdkSource] and pull the entire contents into memory as a [ByteArray].
  */
 @InternalApi
-public expect suspend fun SdkSource.readToByteArray(): ByteArray
+public suspend fun SdkSource.readToByteArray(): ByteArray = withContext(SdkDispatchers.IO) {
+    use { it.buffer().readByteArray() }
+}
```
lauzadis, line 56: note: the only reason I could see these were `expect/actual` is because of `Dispatchers.IO`, but we already have an `expect/actual` version of that called `SdkDispatchers`, so I've moved these to **common** and deleted the JVM/Native implementations

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/io/internal/SdkDispatchersNative.kt
```diff
@@ -6,12 +6,14 @@ package aws.smithy.kotlin.runtime.io.internal
...
 
 @InternalApi
 public actual object SdkDispatchers {
     /**
      * The CoroutineDispatcher that is designed for offloading blocking IO tasks to a shared pool of threads.
      */
     public actual val IO: CoroutineDispatcher
-        get() = TODO("Not yet implemented")
+        get() = Dispatchers.IO
 }
```
lauzadis, line 19: note: we made this `expect/actual` because Native didn't have Dispatchers.IO at the time. It does now, so we _could_ remove SdkDispatchers... but I think we'll need it again in the future when we support platforms like JS and WASM.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/io/BuffereredSourceAdapter.kt
```diff
@@ -32,6 +32,17 @@ internal expect class BufferedSourceAdapter(source: okio.BufferedSource) : SdkBu
...
+/**
+ * Used to wrap calls to Okio, catching Okio exceptions (e.g. okio.EOFException) and throwing our own (e.g. aws.smithy.kotlin.runtime.io.EOFException).
+ */
+internal inline fun <T> SdkBufferedSource.wrapOkio(block: SdkBufferedSource.() -> T): T = try {
+    block()
+} catch (e: okio.EOFException) {
+    throw EOFException("Okio operation failed", e)
+} catch (e: okio.IOException) {
+    throw IOException("Okio operation failed", e)
+}
```
ianbotsf, line 44: **Suggestion**: Consider adding the Okio exception message to the wrapped exception message:

```kotlin
throw EOFException("Okio operation failed: ${e.message}", e)
```

Different loggers and exception dumpers _may_ print an entire exception including the cause but some won't and users may only have a top-level exception message to work with in some contexts.

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/io/ExceptionsNative.kt
```diff
@@ -12,6 +12,7 @@ public actual open class IOException actual constructor(
     public actual constructor(message: String?) : this(message, null)
 }
 
-public actual open class EOFException actual constructor(message: String?) : IOException(message) {
-    public actual constructor() : this(null)
+public actual open class EOFException actual constructor(message: String?, cause: Throwable?) : IOException(message) {
+    public actual constructor() : this(null, null)
+    public actual constructor(message: String?) : this(message, null)
 }
```
ianbotsf, line 18: **Correctness**: Pass the constructor `cause` parameter to the superclass constructor:

```kotlin
class EOFException actual constructor(message: String?, cause: Throwable?) : IOException(message, cause)
```

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/io/SdkBufferNative.kt
```diff
@@ -4,157 +4,102 @@
...
-    actual override fun readShort(): Short {
-        TODO("Not yet implemented")
-    }
+    actual override fun flush(): Unit = commonFlush()
 
-    actual override fun readShortLe(): Short {
-        TODO("Not yet implemented")
+    actual override fun emit() {
+        wrapOkio { inner.emit() }
     }
```
ianbotsf, line 64: **Question**: Same as JVM implementation? Can this be moved to **common**?

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/io/SdkBufferNative.kt
```diff
@@ -4,157 +4,102 @@
...
-    }
+    override fun hashCode(): Int = inner.hashCode()
 
-    actual override fun write(source: SdkSource, byteCount: Long) {
-        TODO("Not yet implemented")
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is SdkBuffer) return false
+        return inner == other.inner
     }
```
ianbotsf, line 34: **Question**: Same as JVM implementation? Can this be moved to **common**?

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/io/internal/SdkDispatchersNative.kt
ianbotsf: Unless JetBrains implements dispatchers on those platforms first! 😁

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/io/SdkBufferNative.kt
```diff
@@ -4,157 +4,102 @@
  */
 package aws.smithy.kotlin.runtime.io
 
-import okio.Buffer
+import aws.smithy.kotlin.runtime.io.internal.*
 
 public actual class SdkBuffer :
```
ianbotsf, line 9: **Question**: Do we still _need_ platform-specific implementations of SDK buffer? Seems like basically everything is commonized or commonizable at this point.

ianbotsf (changes requested):

runtime/auth/aws-signing-default/common/src/aws/smithy/kotlin/runtime/auth/awssigning/Canonicalizer.kt
```diff
@@ -20,6 +20,7 @@ import aws.smithy.kotlin.runtime.text.encoding.Encodable
 import aws.smithy.kotlin.runtime.text.encoding.PercentEncoding
 import aws.smithy.kotlin.runtime.text.encoding.encodeToHex
 import aws.smithy.kotlin.runtime.time.TimestampFormat
+import kotlinx.coroutines.IO
 import kotlinx.coroutines.withContext
```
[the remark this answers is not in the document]
lauzadis: I was working in this file earlier, this import is unused now and I'm surprised ktlint didn't complain

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/io/SdkBufferNative.kt
lauzadis: The only difference I see is JVM has inputStream and outputStream methods. I'll see if this can be made **common**.

I think at some point we were considering using CRT for Native I/O. Okio is multiplatform so I was thinking we should just use that?

ianbotsf: Agreed. Their IO model matches ours better than CRT iirc.

PR #1223, opened 2025-01-21
fix: address various failing protocol tests

lauzadis:
- Fixes a failing protocol test by allowing serialization of empty headers. Related to https://github.com/smithy-lang/smithy/pull/2433 and https://github.com/smithy-lang/smithy/pull/2403.
- Fixes a few failing protocol tests by updating our handling of default blob values to decode the Base64 string
- Address another failing protocol test by ~~merging an upstream fix to make Base64 string length a multiple of 4 (aka padded)~~ adding support for unpadded Base64 inputs

## Description of changes

This change is required to comply with Smithy protocol tests which assert that empty header values are sent.

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
@@ -248,6 +248,11 @@ class StructureGenerator(
                     } else {
                         memberSymbol
                     }
+
+                    if (builderMemberSymbol.shape is BlobShape && builderMemberSymbol.isNotNullable) {
+                        writer.addImport(RuntimeTypes.Core.Text.Encoding.decodeBase64)
+                    }
+
                     write("public var #L: #E", memberName, builderMemberSymbol)
                 }
                 write("")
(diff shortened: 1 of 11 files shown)


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
```diff
@@ -248,6 +248,11 @@ class StructureGenerator(
                     } else {
                         memberSymbol
                     }
+
+                    if (builderMemberSymbol.shape is BlobShape && builderMemberSymbol.isNotNullable) {
+                        writer.addImport(RuntimeTypes.Core.Text.Encoding.decodeBase64)
+                    }
```
lauzadis, line 254: decodeBase64 is used in `KotlinSymbolProvider`, but we don't have access to the `KotlinWriter` there, so the import gets added here if necessary. Any other ideas to work around this?

ianbotsf: Correct, we don't have a `KotlinWriter` (nor should we) in `KotlinSymbolProvider`. But we're still providing symbols and [symbols can declare additional imports](https://github.com/smithy-lang/smithy-kotlin/blob/0f8db4452966822949bbe6616e3642b2063cfb63/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/SymbolBuilder.kt#L43-L60). We should be setting up those references before returning from the symbol provider.

This probably means a bit of a light refactor in the symbol provider since it's mostly slinging text strings around and wrapping it into a symbol right before returning. But we probably need better awareness of non-textual symbol info in places like `getDefaultValue`.

lauzadis: I did a minimal refactor to give `getDefaultValue` access to the Symbol being built, calling `addReferences` where appropriate

ianbotsf (approved):

PR #1228, opened 2025-01-29
kn: gzip compression

lauzadis:
- Implements gzip compression using zlib bindings (`GzipCompressor`)
- Deduplicates cross-module definitions of `decompressGzipBytes`
    - Required making this function `@InternalApi public`, which I think is a fine tradeoff 	

## Description of changes


runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/compression/GzipCompressor.kt
```diff
@@ -0,0 +1,128 @@
...
+        val result = outputBuffer.take(count).toByteArray()
+        repeat(count) { outputBuffer.removeAt(0) }
+        return result
+    }
+
+    /**
+     * Close the compressor, clean up all resources, and return the terminal sequence of bytes
+     * that represent the end of the gzip compression.
+     */
+    fun close(): ByteArray {
```
ianbotsf, line 96: **Suggestion**: This isn't a `close` method in the traditional `Closeable` sense. I recommend calling it something different like or `complete` or `finish` **–or–** implementing a `flush` method and changing this method to just track closure and freeing resources (and then implementing `Closeable`).

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/compression/GzipNative.kt
```diff
@@ -13,12 +17,35 @@ import aws.smithy.kotlin.runtime.content.ByteStream
...
+                stream
+            } else {
+                val compressed = GzipCompressor().apply {
+                    update(sourceBytes)
+                }.close()
+
+                object : ByteStream.Buffer() {
+                    override fun bytes(): ByteArray = compressed
+                }
+            }
```
ianbotsf, line 48: **Style**:

```kotlin
GzipCompressor()
    .apply { update(sourceBytes) }
    .close()
    .let(::ByteArrayContent)
```

Regarding that last call, it seems like a gap that we don't have `asByteStream` extension functions for `String` and `ByteArray`. We do have [companion object methods](https://github.com/smithy-lang/smithy-kotlin/blob/b7e50e3c66a073e5696b363aae29823284beef09/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/content/ByteStream.kt#L72-L82) but ideally I think we'd also have similar methods to [what we have for some JVM-specific types](https://github.com/smithy-lang/smithy-kotlin/blob/b7e50e3c66a073e5696b363aae29823284beef09/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/ByteStreamJVM.kt#L25-L62).

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/io/GzipByteReadChannelNative.kt
```diff
@@ -5,26 +5,60 @@
...
+                compressor.update(input)
+            }
+        }
+
+        // If still no data is available and the channel is closed, we've hit EOF. Close the compressor and write the remaining bytes
+        if (compressor.availableForRead == 0 && channel.isClosedForRead) {
+            val terminationBytes = compressor.close()
+            sink.write(terminationBytes)
+            return terminationBytes.size.toLong()
+        }
```
ianbotsf, line 51: **Correctness**: Is it _always_ true that `availableForRead == 0` after `compressor.update(...)` means EOF? [Zlib's manual](https://www.zlib.net/manual.html) seems to imply that calls to `deflate` might result in data being internally buffered but not necessarily producing output.

ianbotsf (changes requested):

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/compression/GzipCompressor.kt
```diff
@@ -0,0 +1,128 @@
...
+                if (deflateResult != Z_STREAM_END && deflateResult != Z_OK) {
+                    throw RuntimeException("Deflate failed during finish: $deflateResult")
+                }
+
+                val bytesWritten = BUFFER_SIZE - stream.avail_out.toInt()
+                outputBuffer.addAll(buffer.take(bytesWritten))
+
+                finished = deflateResult == Z_STREAM_END
+                outputPin.unpin()
+            }
```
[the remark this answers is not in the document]
lauzadis: Not directly, because the data is still inside the `stream`, not in the internal buffer we manage. It could possibly be replaced with `update(byteArrayOf())` followed by `consume(availableForRead)`

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/compression/GzipCompressor.kt
```diff
@@ -0,0 +1,128 @@
...
+import kotlinx.cinterop.*
+import platform.zlib.*
+
+/**
+ * Streaming-style gzip compressor, implemented using zlib bindings
+ */
+internal class GzipCompressor {
+    companion object {
+        internal const val BUFFER_SIZE = 16384
+    }
```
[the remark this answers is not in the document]
lauzadis: There's no reason, I just chose this number. I don't think it matters too much, it just controls how often we call `deflate` (whether the input passed to `update` can fit in single `deflate` call or not)

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/compression/GzipCompressor.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    private val buffer = ByteArray(BUFFER_SIZE)
+    private val stream = nativeHeap.alloc<z_stream>()
+    private val outputBuffer = ArrayList<Byte>()
+    private val _isClosed = atomic(false)
+
+    internal val isClosed
+        get() = _isClosed.value
```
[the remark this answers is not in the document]
lauzadis: Sure, it's not necessary

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/compression/GzipCompressor.kt
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Streaming-style gzip compressor, implemented using zlib bindings
+ */
+internal class GzipCompressor {
+    companion object {
+        internal const val BUFFER_SIZE = 16384
+    }
+
+    private val buffer = ByteArray(BUFFER_SIZE)
```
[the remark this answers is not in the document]
lauzadis: This doesn't exactly buffer the user's input, it buffers the compression data out of `deflate`

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/io/GzipByteReadChannelNative.kt
lauzadis: No it doesn't, but if the underlying channel is closed (second half of the condition), then it's safe to flush the compressor (preparing all the internally-buffered data) and write the remaining bytes

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/compression/GzipCompressor.kt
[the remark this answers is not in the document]
lauzadis: Actually an empty `update` won't work, because we need to specify `Z_FINISH` rather than `Z_NO_FLUSH` to ensure the entire contents are flushed out

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/compression/GzipNative.kt
lauzadis: We do have ByteStream.fromBytes

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/compression/GzipCompressor.kt
```diff
@@ -0,0 +1,128 @@
...
+    private val stream = nativeHeap.alloc<z_stream>()
+    private val outputBuffer = ArrayList<Byte>()
```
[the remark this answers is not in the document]
lauzadis: Refactored to use `SdkByteChannel`

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/compression/GzipCompressor.kt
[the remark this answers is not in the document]
lauzadis: Refactored to remove this class-level val, now it's instantiated only where it's needed

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/compression/GzipNative.kt
lauzadis: I totally missed that you already linked `ByteStream.fromBytes`... I added those methods you suggested

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/compression/GzipCompressor.kt
```diff
@@ -4,68 +4,74 @@
...
     }
 
-    private val buffer = ByteArray(BUFFER_SIZE)
     private val stream = nativeHeap.alloc<z_stream>()
-    private val outputBuffer = ArrayList<Byte>()
-    private val _isClosed = atomic(false)
-
-    internal val isClosed
-        get() = _isClosed.value
+    private val outputBuffer = SdkByteChannel()
```
ianbotsf, line 30: **Question**: Do we have to use a channel instead of a source/sink here? This makes all of the data methods `suspend` but I cannot see why they need to be, conceptually.

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/compression/GzipNative.kt
```diff
@@ -34,16 +38,12 @@ public actual class Gzip : CompressionAlgorithm {
...
             } else {
-                val compressed = GzipCompressor().apply {
-                    update(sourceBytes)
-                }.close()
-
-                object : ByteStream.Buffer() {
-                    override fun bytes(): ByteArray = compressed
+                runBlocking {
+                    GzipSdkSource(bytes.source()).readToByteArray().asByteStream()
                 }
```
ianbotsf, line 47: **Question**: Why do we need to invoke a `suspend` method here? We're pulling this all into memory and blocking until it's done so it seems like we can avoid `runBlocking`/`suspend` altogether...

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/compression/GzipCompressor.kt
lauzadis: We can't use an `SdkSource` or an `SdkSink` because those only support either `read` or `write` but not both. I'll try `SdkBuffer`

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/compression/GzipNative.kt
lauzadis: Related to https://github.com/smithy-lang/smithy-kotlin/pull/1228#discussion_r1937662633

ianbotsf (approved):

PR #1233, opened 2025-02-04
kn: enable many `@IgnoreNative` tests

lauzadis:
## Description of changes

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1235, opened 2025-02-05
fix: correctly check equality for CaseInsensitiveMap

ianbotsf:
## Issue \#

(none)

## Description of changes

Fix for https://github.com/smithy-lang/smithy-kotlin/pull/1233#discussion_r1943384793


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/collections/CaseInsensitiveMap.kt
```diff
@@ -36,11 +36,11 @@ internal class CaseInsensitiveMap<Value> : MutableMap<String, Value> {
 
     override val entries: MutableSet<MutableMap.MutableEntry<String, Value>>
         get() = impl.entries.map {
-            Entry(it.key.s, it.value)
+            Entry(it.key.normalized, it.value)
         }.toMutableSet()
 
     override val keys: MutableSet<String>
-        get() = impl.keys.map { it.s }.toMutableSet()
+        get() = impl.keys.map { it.normalized }.toMutableSet()
```
lauzadis, line 43: These changes mean that tests like the following will fail:
```
    @Test
    fun testSimpleKeyOp() {
        val map = CaseInsensitiveMap<String>()
        map["A"] = "apple"
        assertTrue(map.keys.contains("A"))
    }
```

Is that expected from a case insensitive map? I think the keys should remain unmodified, they should only be case-insensitive for equality operations.

Maybe both `map.keys.contains("A")` and `map.keys.contains("a")` should be true.

ianbotsf: Oh yes, they should both be true. Revision incoming...

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/collections/CaseInsensitiveMutableStringSet.kt
```diff
@@ -0,0 +1,40 @@
...
+    override fun containsAll(elements: Collection<String>) = elements.all { it in this }
+    override fun equals(other: Any?) = other is CaseInsensitiveMutableStringSet && delegate == other.delegate
+    override fun hashCode() = delegate.hashCode()
+    override fun isEmpty() = delegate.isEmpty()
+    override fun remove(element: String) = delegate.remove(element.toInsensitive())
+    override val size: Int get() = delegate.size
+    override fun toString() = delegate.toString()
+
+    override fun addAll(elements: Collection<String>) =
+        elements.fold(false) { modified, item -> add(item) || modified }
```
lauzadis, line 20: Could be simplified to `elements.any { add(it) }`, same applies for `removeAll`

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/collections/CaseInsensitiveString.kt
```diff
@@ -0,0 +1,11 @@
+package aws.smithy.kotlin.runtime.collections
+
+internal class CaseInsensitiveString(val original: String) {
```
lauzadis, line 3: Missing tests for CaseInsensitiveString. It is tested indirectly through the other classes though

lauzadis (approved):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/collections/CaseInsensitiveMutableStringSet.kt
ianbotsf: No, [`any` short-circuits once it finds a match](https://github.com/JetBrains/kotlin/blob/c6f337283d59fcede75954eebaa589ad1b479aea/libraries/stdlib/common/src/generated/_Collections.kt#L1749-L1758). If we used that, we'd only ever add at most a single item.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/collections/CaseInsensitiveString.kt
ianbotsf: Tested indirectly but you're right, it'd be better to have dedicated tests. Will add!

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/collections/CaseInsensitiveMutableStringSet.kt
lauzadis: Oh ok. I updated the implementation before posting this comment and all the tests still passed, but it's probably just the way they're written

Issue #1129, opened 2024-07-25
Codegen doesn't properly handle unions with a member that has the same name as the union


ianbotsf:
### Describe the bug

Given a model:

```smithy
union Foo {
    foo: Boolean
}
```

The generated union code does not take into account the shadowed name:

```kotlin
public sealed class Foo {
    public data class Foo(val value: Boolean)

    public fun asFoo(): Boolean = (this as Foo.Foo).value // Compilation error: `Foo` (the data class) has no member `Foo`
}
```

The problem is likely in [`UnionGenerator`](https://github.com/smithy-lang/smithy-kotlin/blob/main/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/UnionGenerator.kt)'s handling of symbols.

### Expected behavior

Code should be generated correctly for a union that has a member of the same name

### Current behavior

Code is generated incorrectly and fails to compile

### Steps to Reproduce

See above model

### Smithy-Kotlin version

1.2.18

### Platform (JVM/JS/Native)

JVM

### Operating system and version

(n/a)

PR #1241, opened 2025-02-13
misc: test union member name same as union

0marperez:
## Issue \#

closes https://github.com/smithy-lang/smithy-kotlin/issues/1129

## Description of changes

Adds a test to make sure we don't regress.
See: https://github.com/awslabs/aws-sdk-kotlin/pull/1526

### DIFF
--- a/tests/compile/src/test/kotlin/software/amazon/smithy/kotlin/codegen/SmithySdkTest.kt
+++ b/tests/compile/src/test/kotlin/software/amazon/smithy/kotlin/codegen/SmithySdkTest.kt
@@ -227,7 +227,34 @@ class SmithySdkTest {
 
     @Test
     fun `it compiles models with unions with members that have the same name as the union`() {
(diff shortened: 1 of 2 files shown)


tests/compile/src/test/kotlin/software/amazon/smithy/kotlin/codegen/SmithySdkTest.kt
```diff
@@ -224,6 +224,44 @@ class SmithySdkTest {
 
         assertEquals(KotlinCompilation.ExitCode.OK, compilationResult.exitCode, compileOutputStream.toString())
     }
+
+    @Test
+    fun `it compiles models with unions with members that have the same name as the union`() {
```
lauzadis, line 229: docs: add a comment with the original GitHub issue that inspired this test case

ianbotsf (approved):

PR #1245, opened 2025-02-20
fix: account ID based endpoints

0marperez:
## Issue \#

N/A

## Description of changes

Adds support for Kotlin maps as JMESPath objects to the [`keys` function](https://jmespath.org/specification.html#keys)

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
@@ -36,7 +36,7 @@ Specifically it's the keys function
 Smithy spec: https://smithy.io/2.0/additional-specs/rules-engine/parameters.html#smithy-rules-operationcontextparams-trait
 JMESPath spec: https://jmespath.org/specification.html#keys
 
-TODO: Proactively support JMESPath objects as Kotlin maps throughout thr entire visitor if more instances of this behavior start popping up
+TODO: Test relevant uses of JMESPath to determine if we should support JMESPath objects as Kotlin maps throughout the entire visitor
  */
 
 /**
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
@@ -19,13 +19,11 @@ import software.amazon.smithy.kotlin.codegen.model.isEnum
 import software.amazon.smithy.kotlin.codegen.model.targetOrSelf
 import software.amazon.smithy.kotlin.codegen.model.traits.OperationInput
 import software.amazon.smithy.kotlin.codegen.model.traits.OperationOutput
+import software.amazon.smithy.kotlin.codegen.utils.doubleQuote
 import software.amazon.smithy.kotlin.codegen.utils.dq
 import software.amazon.smithy.kotlin.codegen.utils.getOrNull
 import software.amazon.smithy.kotlin.codegen.utils.toCamelCase
-import software.amazon.smithy.model.shapes.ListShape
-import software.amazon.smithy.model.shapes.MapShape
-import software.amazon.smithy.model.shapes.MemberShape
-import software.amazon.smithy.model.shapes.Shape
+import software.amazon.smithy.model.shapes.*
 
 private val suffixSequence = sequenceOf("") + generateSequence(2) { it + 1 }.map(Int::toString) // "", "2", "3", etc.
 
@@ -526,11 +524,20 @@ class KotlinJmespathExpressionVisitor(
             ".$expr"
         }
 
-    private fun VisitedExpression.getKeys(): String {
-        val keys = this.shape?.targetOrSelf(ctx.model)?.allMembers
-            ?.keys?.joinToString(", ", "listOf(", ")") { "\"$it\"" }
-        return keys ?: "listOf<String>()"
-    }
+    // TODO: Support maps as objects throughout the visitor
+    // This visitor assumes JMESPath 'objects' will be classes. DDB assumes JMESPath 'objects' will be maps
+    private fun VisitedExpression.getKeys(): String =
+        if (this.shape?.targetOrSelf(ctx.model)?.type == ShapeType.MAP) {
+            "${this.identifier}?.keys?.toList()"
+        } else {
+            this
+                .shape
+                ?.targetOrSelf(ctx.model)
+                ?.allMembers
+                ?.keys
+                ?.joinToString(", ", "listOf(", ")") { it.doubleQuote() }
+                ?: "listOf<String>()"
+        }
 
     private fun VisitedExpression.getValues(): String {
         val values = this.shape?.targetOrSelf(ctx.model)?.allMembers?.keys
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
@@ -27,6 +27,18 @@ import software.amazon.smithy.model.shapes.*
 
 private val suffixSequence = sequenceOf("") + generateSequence(2) { it + 1 }.map(Int::toString) // "", "2", "3", etc.
 
+/*
+JMESPath has the concept of objects.
+This visitor assumes JMESPath objects will be Kotlin objects/classes.
+The smithy spec contains an instance where it's assumed a JMESPath object will be a Kotlin map.
+
+Specifically it's the keys function
+Smithy spec: https://smithy.io/2.0/additional-specs/rules-engine/parameters.html#smithy-rules-operationcontextparams-trait
+JMESPath spec: https://jmespath.org/specification.html#keys
+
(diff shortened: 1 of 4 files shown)


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -526,11 +524,20 @@ class KotlinJmespathExpressionVisitor(
             ".$expr"
         }
 
-    private fun VisitedExpression.getKeys(): String {
-        val keys = this.shape?.targetOrSelf(ctx.model)?.allMembers
-            ?.keys?.joinToString(", ", "listOf(", ")") { "\"$it\"" }
-        return keys ?: "listOf<String>()"
-    }
+    // TODO: Support maps as objects throughout the visitor
+    // This visitor assumes JMESPath 'objects' will be classes. DDB assumes JMESPath 'objects' will be maps
```
lauzadis, line 528: Should this be a FIXME instead of a TODO? Do we have a backlog item we can link here (SDK-KT-x)?

lauzadis: Remove note about DDB

tests/codegen/waiter-tests/src/test/kotlin/com/test/FunctionKeysTest.kt
```diff
@@ -22,9 +23,9 @@ class FunctionKeysTest {
...
 
     @Test
-    fun testKeysFunctionPrimitivesIntegerEquals() = successTest(
+    fun testKeysFunctionMapStringEquals() = successTest(
         GetFunctionKeysEqualsRequest { name = "test" },
-        WaitersTestClient::waitUntilKeysFunctionPrimitivesIntegerEquals,
-        GetFunctionKeysEqualsResponse { primitives = EntityPrimitives { } },
+        WaitersTestClient::waitUntilKeysFunctionMapStringEquals,
+        GetFunctionKeysEqualsResponse { maps = EntityMaps { strings = mapOf("key" to "value") } },
     )
```
ianbotsf, line 30: **Question**: Why did we modify this test case instead of creating a new one? Is the previous behavior no longer valid?

tests/codegen/waiter-tests/model/function-keys.smithy
```diff
@@ -18,14 +18,14 @@ use smithy.waiters#waitable
             }
         ]
     },
-    KeysFunctionPrimitivesIntegerEquals: {
```
lauzadis, line 21: Why was this integer primitives test removed?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -526,11 +524,20 @@ class KotlinJmespathExpressionVisitor(
...
+    private fun VisitedExpression.getKeys(): String =
+        if (this.shape?.targetOrSelf(ctx.model)?.type == ShapeType.MAP) {
+            "${this.identifier}?.keys?.toList()"
+        } else {
```
lauzadis, line 532: What happens if the map has keys that aren't strings?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
ianbotsf, line 528: **Question**: What does this comment mean? It seems like this visitor now assumes objects can be classes _or_ maps.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -526,11 +524,20 @@ class KotlinJmespathExpressionVisitor(
...
 
-    private fun VisitedExpression.getKeys(): String {
-        val keys = this.shape?.targetOrSelf(ctx.model)?.allMembers
-            ?.keys?.joinToString(", ", "listOf(", ")") { "\"$it\"" }
-        return keys ?: "listOf<String>()"
-    }
+    // TODO: Support maps as objects throughout the visitor
+    // This visitor assumes JMESPath 'objects' will be classes. DDB assumes JMESPath 'objects' will be maps
+    private fun VisitedExpression.getKeys(): String =
+        if (this.shape?.targetOrSelf(ctx.model)?.type == ShapeType.MAP) {
```
lauzadis, line 530: style: unnecessary use of `this`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
0marperez: Not sure if it should be a `FIXME` because the JMESPath visitor has been assuming classes > maps so far and we have had no apparent problems. Other services use JMESPath but this is the first instance I've seen of maps being used as objects. It might be a DDB only quirk or an indication that we took the wrong approach to objects

0marperez: The `keys` function assumes objects can be classes or maps. The rest of the visitor does not. Hence the comment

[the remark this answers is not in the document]
0marperez: It would be close to a re-write of the whole visitor. A decent amount of work

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
0marperez: Good catch, thanks

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
lauzadis: Our internal spec says:
> [Keys function](https://jmespath.org/specification.html#keys) - return a list of the keys in a map. This is the only supported function but is required for binding to key values.

Do we need to support classes?

lauzadis: Either way it sounds like we need a backlog item for this, to at least investigate the discrepancy

0marperez: I think we should keep

>     // This visitor assumes JMESPath 'objects' will be classes. DDB assumes JMESPath 'objects' will be maps

But change it to mention the discrepancy between the smithy spec and the JMESPath spec for this function

lauzadis: I just think every TODO should have an associated backlog task, otherwise it never gets done

ianbotsf: No I believe we _should_ support classes and it's an oversight that we didn't support it before. I was just wondering what else is required. Since nothing seems to be required for this `getKeys` function anymore I think we should move the `TODO` up to the class level or wherever's more appropriate for the scope.

0marperez: Updated the class with newer comments, let me know what you think

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/waiters/KotlinJmespathExpressionVisitor.kt
```diff
@@ -19,16 +19,26 @@ import software.amazon.smithy.kotlin.codegen.model.isEnum
...
+/*
+JMESPath has the concept of objects.
+This visitor assumes JMESPath objects will be Kotlin objects/classes.
+The smithy spec contains an instance where it's assumed a JMESPath object will be a Kotlin map.
+
+Specifically it's the keys function
+Smithy spec: https://smithy.io/2.0/additional-specs/rules-engine/parameters.html#smithy-rules-operationcontextparams-trait
+JMESPath spec: https://jmespath.org/specification.html#keys
+
+TODO: Proactively support JMESPath objects as Kotlin maps throughout thr entire visitor if more instances of this behavior start popping up
```
lauzadis, line 39: "Proactively" and "if more instances of this behavior start popping up" are inconsistent. To be proactive we would need to identify any gaps and fix them now

lauzadis (approved):

PR #1246, opened 2025-02-20
feat: Kotlin implementation of SigV4a signing

lauzadis:
## Description of changes


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/hashing/Ecdsa.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.hashing
+
+/**
+ * ECDSA on the SECP256R1 curve.
+ */
+public expect fun ecdsasecp256r1(key: ByteArray, message: ByteArray): ByteArray
```
ianbotsf, line 10: **Style**: ECSDA and Secp256r1 are distinct words so we should uppercase that `s` (e.g., `ecdsaSecp256r1`)

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/hashing/EcdsaJVMTest.kt
```diff
@@ -0,0 +1,84 @@
...
+package aws.smithy.kotlin.runtime.hashing
+
+import org.junit.jupiter.api.Assertions.assertFalse
+import org.junit.jupiter.api.Assertions.assertTrue
+import java.security.*
+import java.security.interfaces.*
+import java.security.spec.*
+import kotlin.test.Test
+
+class EcdsaJVMTest {
```
ianbotsf, line 14: **Question**: Can we move these tests into **common** instead of **jvm**?

runtime/auth/aws-signing-default/common/test/aws/smithy/kotlin/runtime/auth/awssigning/SigV4SignatureCalculatorTest.kt
```diff
@@ -14,7 +14,7 @@ import kotlinx.coroutines.test.runTest
 import kotlin.test.Test
 import kotlin.test.assertEquals
 
-class DefaultSignatureCalculatorTest {
+class SigV4SignatureCalculatorTest {
     // Test adapted from https://docs.aws.amazon.com/general/latest/gr/sigv4-calculate-signature.html
     @Test
     fun testCalculate() {
```
ianbotsf, line 20: **Nit**: We should add tests for SigV4a signature calculation with explicit inputs/outputs.

runtime/auth/aws-signing-default/common/src/aws/smithy/kotlin/runtime/auth/awssigning/SigV4aSignatureCalculator.kt
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * Forms the fixed input string used for ECDSA private key derivation
+     * The final output looks like:
+     * 0x00000001 || "AWS4-ECDSA-P256-SHA256" || 0x00 || AccessKeyId || counter || 0x00000100
+     */
+    private fun fixedInputString(accessKeyId: String, counter: Byte): ByteArray =
+        byteArrayOf(0x00, 0x00, 0x00, 0x01) +
+            "AWS4-ECDSA-P256-SHA256".encodeToByteArray() +
```
ianbotsf, line 80: **Nit**: Could reuse `AwsSigningAlgorithm.SIGV4_ASYMMETRIC.signingName`

ianbotsf (changes requested):

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/hashing/EcdsaJVMTest.kt
lauzadis: Not easily, due to the use of `java.security.*`. Given the FIXME written in `EcdsaNative.kt` (which we discussed / agreed upon in standup), we are unlikely to ever implement this on Native, so I think it's fine to leave as JVM-only.

lauzadis: Specifically, that we'll be using `CrtAwsSigner` on Native instead of `DefaultAwsSigner`. This `ecdsaSecp256r1` will not be used on Native in that case

runtime/auth/aws-signing-default/common/test/aws/smithy/kotlin/runtime/auth/awssigning/SigV4SignatureCalculatorTest.kt
lauzadis: java.security's ECDSA implementation is non-deterministic, it chooses a random `k` value with each call to `sign()`. We'd get different (yet still valid) signatures when providing a consistent input. This library doesn't allow overriding the random generator which would allow us to provide a fixed "random" input for tests. [The top recommendation seems to be switching to BouncyCastle](https://stackoverflow.com/questions/19904548/calculating-ecdsa-signature-in-java-as-per-an-rfc-test-vector) and overriding the implementation of `SecureRandom` used. 

Would you be fine with the tests asserting on everything _except_ the signature value?

lauzadis: We spoke offline and agreed to assert on everything except the signature value. Added a test suite for "string to sign" with test cases copied from CRT.

runtime/auth/aws-signing-default/common/test/aws/smithy/kotlin/runtime/auth/awssigning/SigV4aSignatureCalculatorTest.kt
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Tests for [SigV4aSignatureCalculator]. Currently only tests forming the string-to-sign.
+ *
+ * TODO Add tests against header-signature.txt when java.security implements RFC 6979 / deterministic ECDSA.
+ * https://bugs.openjdk.org/browse/JDK-8239382
+ */
+class SigV4aSignatureCalculatorTest {
+    @Test
+    fun testStringToSign() = TESTS.forEach { testId ->
+        runTest {
```
0marperez, line 75: Suggestion: Logging which test is running or adding a message to the `assertEquals` in case the assertion fails

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
```diff
@@ -24,6 +24,17 @@ public enum class AwsSigningAlgorithm {
...
+    ;
+
+    /**
+     * The name of this algorithm to use when signing requests.
+     */
+    public val signingName: String
+        get() = when (this) {
+            SIGV4 -> "AWS4-HMAC-SHA256"
+            SIGV4_ASYMMETRIC -> "AWS4-ECDSA-P256-SHA256"
+        }
```
ianbotsf, line 37: **Style**: Embed as a constructor parameter

```kotlin
public enum class AwsSigningAlgorithm(public val signingName: String) {
    SIGV4("AWS4-HMAC-SHA256"),
    SIGV4_ASYMMETRIC("AWS4-ECDSA-P256-SHA256"),
}
```

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/hashing/EcdsaNative.kt
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * ECDSA on the SECP256R1 curve.
+ */
+// FIXME Implement using aws-c-cal: https://github.com/awslabs/aws-c-cal/blob/main/include/aws/cal/ecc.h
+/**
+ * Will need to be implemented and exposed in aws-crt-kotlin.
+ * Or maybe we can _only_ offer the CRT signer on Native?
+ * Will require updating DefaultAwsSigner to be expect/actual and set to CrtSigner on Native.
+ */
+public actual fun ecdsaSecp256r1(key: ByteArray, message: ByteArray): ByteArray = TODO("Not yet implemented")
```
ianbotsf, line 16: **Nit**: The `FIXME` ends and a new KDocs block opens but it seems like the content of it is more related to the `FIXME` than documentation about what the function does. Restructure to just continue the `FIXME`:

```kotlin
/**
 */
// FIXME Implement using aws-c-cal: https://github.com/awslabs/aws-c-cal/blob/main/include/aws/cal/ecc.h
//  Will need to be implemented and exposed in aws-crt-kotlin.
//  Or maybe we can _only_ offer the CRT signer on Native?
//  Will require updating DefaultAwsSigner to be expect/actual and set to CrtSigner on Native.
public actual fun ecdsaSecp256r1(key: ByteArray, message: ByteArray): ByteArray = TODO("Not yet implemented")
```

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/hashing/EcdsaJVMTest.kt
ianbotsf: Got it, thanks for the clarification.

PR #1249, opened 2025-02-26
fix: correctly codegen paginators for types which require fully-qualified names

ianbotsf:
## Issue \#

(none)

## Description of changes

The codegen for paginators does not correctly handle the case of item types whose names collide with other types. Specifically, the function signature for the items paginator is formed without taking into account existing imports. This was discovered in a test model in which the pagination items were of type `Flow` which caused a paginator like this:

```kotlin
public fun Flow<FooResponse>.flows(): Flow<Flow> = ...
```

This PR improves the logic which forms item descriptors in paginator generators and will fully qualify type names which collide with existing imports, leading to this:

```kotlin
public fun Flow<FooResponse>.flows(): Flow<com.foo.Flow> = ...
```

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGeneratorTest.kt
+++ b/codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGeneratorTest.kt
@@ -264,7 +264,7 @@ class PaginatorGeneratorTest {
     }
 
     @Test
-    fun testRenderPaginatorWithItemRequiringFqName() {
+    fun testRenderPaginatorWithItemRequiringFullName() {
         val testModelWithItems = """
             namespace com.test
             
--- a/codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGeneratorTest.kt
+++ b/codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGeneratorTest.kt
@@ -264,6 +264,130 @@ class PaginatorGeneratorTest {
     }
 
     @Test
+    fun testRenderPaginatorWithItemRequiringFqName() {
+        val testModelWithItems = """
+            namespace com.test
+            
+            use aws.protocols#restJson1
+            
+            service FlowService {
+                operations: [ListFlows]
+            }
+            
+            @paginated(
+                inputToken: "Marker",
+                outputToken: "NextMarker",
+                pageSize: "MaxItems",
+                items: "Flows"
+            )
+            @readonly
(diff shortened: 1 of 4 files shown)


codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGeneratorTest.kt
```diff
@@ -263,6 +263,130 @@ class PaginatorGeneratorTest {
...
+             *
+             * When this operation is called, a [kotlinx.coroutines.Flow] is created. Flows are lazy (cold) so no service
+             * calls are made until the flow is collected. This also means there is no guarantee that the request is valid
+             * until then. Once you start collecting the flow, the SDK will lazily load response pages by making service
+             * calls until there are no pages left or the flow is cancelled. If there are errors in your request, you will
+             * see the failures only after you start collection.
+             * @param initialRequest A [ListFlowsRequest] to start pagination
+             * @return A [kotlinx.coroutines.flow.Flow] that can collect [ListFlowsResponse]
+             */
+            public fun TestClient.listFlowsPaginated(initialRequest: ListFlowsRequest = ListFlowsRequest { }): kotlinx.coroutines.flow.Flow<ListFlowsResponse> =
```
lauzadis, line 343: question: These will only be fully qualified `kotlinx.coroutines.flow.Flow` when there's an import collision, right?

lauzadis (approved):

0marperez (approved):

codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGeneratorTest.kt
ianbotsf: Yes, although typically `kotlinx.coroutines.flow.Flow` will be referenced _first_ in our codegen and so a conflicting symbol like `com.foo.Flow` would actually be the one which gets the fully-qualified name in code. This test case shows it the other way around because the very first paginator is the one which has the collision and the model shape is referenced before `kotlinx.coroutines.flow.Flow`.

codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/PaginatorGeneratorTest.kt
```diff
@@ -263,6 +263,130 @@ class PaginatorGeneratorTest {
         actual.shouldContainOnlyOnceWithDiff(expectedImports)
     }
 
+    @Test
+    fun testRenderPaginatorWithItemRequiringFqName() {
```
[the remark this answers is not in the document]
ianbotsf: How about just `Full` instead?

```kotlin
@Test
fun testRenderPaginatorWithItemRequiringFullName() {
```

PR #1257, opened 2025-03-20
kn/misc: enable various ignored tests

lauzadis:
## Description of changes


runtime/auth/aws-signing-default/jvm/src/aws/smithy/kotlin/runtime/auth/awssigning/DefaultAwsSignerJVM.kt
```diff
@@ -0,0 +1,121 @@
...
+    DefaultAwsSignerBuilder().apply(block).build()
+
+/** A builder class for creating instances of [AwsSigner] using the default implementation */
+public class DefaultAwsSignerBuilder {
+    public var telemetryProvider: TelemetryProvider? = null
+
+    public fun build(): AwsSigner = DefaultAwsSignerImpl(
+        telemetryProvider = telemetryProvider,
+    )
+}
```
ianbotsf, line 30: **Opinion**: It now feels weird that this factory function and builder type exist only in JVM when the named value exists in all source sets.

runtime/auth/aws-signing-default/jvm/src/aws/smithy/kotlin/runtime/auth/awssigning/DefaultAwsSignerJVM.kt
```diff
@@ -0,0 +1,121 @@
...
+import aws.smithy.kotlin.runtime.ExperimentalApi
+import aws.smithy.kotlin.runtime.http.Headers
+import aws.smithy.kotlin.runtime.http.request.HttpRequest
+import aws.smithy.kotlin.runtime.telemetry.TelemetryProvider
+import aws.smithy.kotlin.runtime.telemetry.logging.logger
+import kotlin.coroutines.coroutineContext
+
+/** The default implementation of [AwsSigner] */
+public actual val DefaultAwsSigner: AwsSigner
+    get() = DefaultAwsSignerImpl()
```
ianbotsf, line 16: **Question**: Why is a new signer instantiated on every call? Before it was initialized only once and then reused everywhere.

runtime/auth/aws-signing-default/native/src/aws/smithy/kotlin/runtime/auth/awssigning/DefaultAwsSignerNative.kt
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.auth.awssigning
+
+import aws.smithy.kotlin.runtime.auth.awssigning.crt.CrtAwsSigner
+
+/** The default implementation of [AwsSigner] */
+public actual val DefaultAwsSigner: AwsSigner
+    get() = CrtAwsSigner
```
ianbotsf, line 11: **Question**: Do we have to use a getter for this or can we just set the value one time?

```kotlin
public actual val DefaultAwsSigner: AwsSigner = CrtAwsSigner
```

runtime/crt-util/native/src/aws/smithy/kotlin/runtime/crt/ReadChannelBodyStreamNative.kt
```diff
@@ -7,4 +7,8 @@ package aws.smithy.kotlin.runtime.crt
 import aws.sdk.kotlin.crt.io.MutableBuffer
 import aws.smithy.kotlin.runtime.io.SdkBuffer
 
-internal actual fun transferRequestBody(outgoing: SdkBuffer, dest: MutableBuffer): Int = TODO("Not yet implemented")
+internal actual fun transferRequestBody(outgoing: SdkBuffer, dest: MutableBuffer): Int {
+    val length = minOf(outgoing.size, dest.writeRemaining.toLong())
+    if (length <= 0) return 0
+    return dest.write(outgoing.readByteArray(length))
+}
```
ianbotsf, line 14: **Question**: Why is this different than the JVM (formerly common) implementation? Why won't simply `outgoing.read(dest.buffer)` work?

runtime/protocol/http-client-engines/http-client-engine-crt/jvmAndNative/test/aws/smithy/kotlin/runtime/http/engine/crt/SdkStreamResponseHandlerTest.kt
```diff
@@ -153,9 +153,9 @@ class SdkStreamResponseHandlerTest {
         assertEquals(data, respChan.readToBuffer().readUtf8())
     }
 
-    @IgnoreNative // FIXME Re-enable after Kotlin/Native Implementation.
     @Test
     fun testStreamError() = runTest {
+        CRT.initRuntime()
```
ianbotsf, line 158: **Question**: This shouldn't properly be here, right? Does this need a `FIXME` because some underlying component is not properly initializing the CRT?

runtime/protocol/http-client/common/test/aws/smithy/kotlin/runtime/http/operation/HttpInterceptorTest.kt
```diff
@@ -261,7 +259,7 @@ class HttpInterceptorTest {
             op.roundTrip(client, Unit)
         }
 
-        val cause = assertNotNull(ex.cause)
+        val cause = assertNotNull(ex.cause ?: ex)
```
ianbotsf, line 262: **Question**: Why is there a fallback in this test? Why would the exception have a cause in one run/target but not another?

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/hashing/EcdsaNative.kt
```diff
@@ -4,10 +4,7 @@
...
 package aws.smithy.kotlin.runtime.hashing
 
-// FIXME Implement using aws-c-cal: https://github.com/awslabs/aws-c-cal/blob/main/include/aws/cal/ecc.h
-// Will need to be implemented and exposed in aws-crt-kotlin. Or maybe we can _only_ offer the CRT signer on Native?
-// Will require updating DefaultAwsSigner to be expect/actual and set to CrtSigner on Native.
 /**
  * ECDSA on the SECP256R1 curve.
  */
-public actual fun ecdsaSecp256r1(key: ByteArray, message: ByteArray): ByteArray = TODO("Not yet implemented")
+public actual fun ecdsaSecp256r1(key: ByteArray, message: ByteArray): ByteArray = error("This function should not be invoked on Native, which uses the CrtAwsSigner.")
```
ianbotsf, line 10: **Nit**: It feels weird to have an `actual` function that we'll never implement and should never call. That, to me, indicates that our `expect` declaration is on the wrong component and should be higher up.

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/hashing/EcdsaNative.kt
```diff
@@ -4,10 +4,7 @@
  */
 package aws.smithy.kotlin.runtime.hashing
 
-// FIXME Implement using aws-c-cal: https://github.com/awslabs/aws-c-cal/blob/main/include/aws/cal/ecc.h
-// Will need to be implemented and exposed in aws-crt-kotlin. Or maybe we can _only_ offer the CRT signer on Native?
-// Will require updating DefaultAwsSigner to be expect/actual and set to CrtSigner on Native.
```
ianbotsf, line 9: **Question**: Did we go down the road of binding to CRT's crypto primitives and abandon it for some reason? Or did we just decide to use the whole CRT signer for a more important reason?

runtime/auth/aws-signing-default/jvm/src/aws/smithy/kotlin/runtime/auth/awssigning/DefaultAwsSignerJVM.kt
lauzadis: Will be refactored, same applies for CrtAwsSigner

runtime/crt-util/native/src/aws/smithy/kotlin/runtime/crt/ReadChannelBodyStreamNative.kt
lauzadis: `dest.buffer` exists only on JVM

runtime/protocol/http-client-engines/http-client-engine-crt/jvmAndNative/test/aws/smithy/kotlin/runtime/http/engine/crt/SdkStreamResponseHandlerTest.kt
lauzadis: I could not decide which underlying component needs to initialize CRT in this test. This is testing the `SdkStreamResponseHandler` which is a callback machine for reading a CRT `HttpStream`. We're not creating a CRT engine or anything else which might initialize CRT here.

The only reason this `initRuntime` is needed is for the final assertion:
`ex.message.shouldContain("socket is closed.; crtErrorCode=$socketClosedEc")`

CRT needs to be initialized to return human-readable error codes, otherwise it returns "unknown". This assertion is on an exception message which we typically avoid doing, so I could remove that assertion / remove `CRT.init`?

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/hashing/EcdsaNative.kt
lauzadis: No, I didn't even start binding to CRT for this. We previously discussed using CrtAwsSigner as the default on Native so that's where I went.

It sounds like you've changed your mind on this, we can talk about it and decide what to do

runtime/protocol/http-client-engines/http-client-engine-crt/jvmAndNative/test/aws/smithy/kotlin/runtime/http/engine/crt/SdkStreamResponseHandlerTest.kt
ianbotsf: Yes removing the `CRT.initRuntime()` is probably the simplest thing to do for this test. But I still worry we have our CRT initialization calls in so many disparate places and that we may still be missing the calls in other places.

runtime/auth/aws-signing-default/jvm/src/aws/smithy/kotlin/runtime/auth/awssigning/DefaultAwsSignerJVM.kt
ianbotsf: Discussed separately and agreed that this API disparity is fine for now and can be reexamined if we ever see problems or get feedback.

ianbotsf (approved):

PR #1266, opened 2025-04-09
misc: prepare codegen packages for internal import

lauzadis:
## Description of changes

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -30,7 +30,7 @@ kotlinx-serialization-version = "1.7.3"
 docker-java-version = "3.4.0"
 ktor-version = "3.1.1"
 kaml-version = "0.55.0"
-jsoup-version = "1.18.1"
+jsoup-version = "1.19.1"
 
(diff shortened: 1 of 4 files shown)


gradle/libs.versions.toml
```diff
@@ -30,7 +30,7 @@ kotlinx-serialization-version = "1.7.3"
 docker-java-version = "3.4.0"
 ktor-version = "3.1.1"
 kaml-version = "0.55.0"
-jsoup-version = "1.18.1"
+jsoup-version = "1.19.1"
```
ianbotsf, line 33: **Question**: Why was this change necessary?

ianbotsf (approved):

gradle/libs.versions.toml
lauzadis: Ideally we would just remove this dependency, I took a quick attempt but it's deeply entangled in our documentation preprocessor. There's no alternative API offered by Kotlin / another one of our existing dependencies

PR #1277, opened 2025-04-22
feat: Add `authSchemePreference` configuration

lauzadis:
## Description of changes


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt
```diff
@@ -174,6 +174,17 @@ object RuntimeConfigProperty {
...
     }
+
+    val AuthSchemePreference = ConfigProperty {
+        name = "authSchemePreference"
+        symbol = KotlinTypes.Collections.list(RuntimeTypes.Auth.Identity.AuthSchemeId, isNullable = true)
+        baseClass = RuntimeTypes.Auth.HttpAuth.HttpAuthConfig
+        useNestedBuilderBaseClass()
+
+        documentation = """
+            The ordered preference of [AuthScheme] that this client will use.
```
0marperez, line 185: Suggestion: The documentation looks good. I think it would help to be more explicit about the preference list not being a hard override of auth schemes. Some people might get confused when the auth scheme(s) they prefer isn't being used.

0marperez (approved):

runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/ReprioritizeAuthOptions.kt
```diff
@@ -0,0 +1,29 @@
...
+ * Re-prioritize a resolved list of auth options based on a user's preference list
+ */
+@InternalApi
+public fun reprioritizeAuthOptions(authSchemePreference: List<AuthSchemeId>, authOptions: List<AuthOption>): List<AuthOption> {
+    // add preferred candidates first
+    val preferredAuthOptions = authSchemePreference.mapNotNull { preferredSchemeId ->
+        val preferredSchemeName = preferredSchemeId.id.substringAfter('#')
+        authOptions.singleOrNull {
+            it.schemeId.id.substringAfter('#') == preferredSchemeName
+        }
```
ianbotsf, line 22: **Question**: Why are we dropping the prefix for this comparison?

ianbotsf (approved):

runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/ReprioritizeAuthOptions.kt
lauzadis: It's the behavior we decided to do in the SEP:
> The value stored in these keys is a comma-separated list of case-sensitive auth scheme names. An auth scheme name is a fully qualified auth scheme ID with the namespace prefix trimmed. For example, the auth scheme with ID `aws.auth#sigv4` is named `sigv4`.

lauzadis: This allows users to configure `sigv4` instead of `aws.auth#sigv4`, `httpBearerAuth` instead of `smithy.api#httpBearerAuth`, etc.

PR #1292, opened 2025-06-04
misc: enable ignored kn tests

0marperez:
## Description of changes

Currently PR is being used for CI testing. NOT ready for review yet.

### DIFF
--- a/runtime/protocol/http-client-engines/http-client-engine-crt/jvmAndNative/test/aws/smithy/kotlin/runtime/http/engine/crt/AsyncStressTest.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-crt/jvmAndNative/test/aws/smithy/kotlin/runtime/http/engine/crt/AsyncStressTest.kt
@@ -5,7 +5,6 @@
 
 package aws.smithy.kotlin.runtime.http.engine.crt
 
-import aws.smithy.kotlin.runtime.IgnoreNative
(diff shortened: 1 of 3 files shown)


runtime/protocol/http-client-engines/http-client-engine-crt/jvmAndNative/test/aws/smithy/kotlin/runtime/http/engine/crt/AsyncStressTest.kt
```diff
@@ -55,13 +53,14 @@ class AsyncStressTest : TestWithLocalServer() {
             }
         }
 
-        withTimeout(10.seconds) {
+        withTimeout(1000.seconds) {
```
ianbotsf, line 56: **Question**: 16 minutes and 40 seconds is a very long timeout. Shouldn't we be able to handle 1000 calls in 10 seconds? How long is this test actually taking?

0marperez: 1000 is just for testing, I'll bring it back to 10 when this is ready for review. I'm currently just testing runs of the KN tests in CI, I forgot to say so in the description. Thanks for taking a preemptive look at it though.

PR #1293, opened 2025-06-06
feat: add protocol tests runner

sugmanue:
## Description of changes

Adds a protocol tests runner, a new mechanism to run the tests and report the results in JSON format detached from the test phase of a regular build.

This change adds new modules to the project:

* `codegen/smithy-kotlin-protocol-tests-codegen` builds on top of `smithy-kotlin-codegen` to generate a client and methods for each defined protocol test that generates a JSON result of  detached from the test phase of the build.
* `tests/protocol-tests-utils` implements utility classes that are used to run the protocol tests such a the enums and JSON writer used in the reports
* `tests/protocol-tests-utils` implements the protocol tests runner, that identifies all the clients with protocol tests and for each
  1. Generate the protocol tests
  2. Build the generated protocol tests
  3. Execute the protocol tests
  4. Collect the output of the service into the final result

### Protocol Tests Runner

This is an example of how to run the protocol tests runner. The runner is a single, fat jar, that can be run directly on the command line. It takes as an argument the JARs containing the protocol tests definitions, for instance:

```
java -jar tests/protocol-tests-runner/build/libs/protocol-tests-runner-all.jar  ~/.m2/repository/software/amazon/smithy/smithy-protocol-tests/1.58.1-SNAPSHOT/smithy-protocol-tests-1.58.1-SNAPSHOT.jar  ~/.m2/repository/software/amazon/smithy/smithy-aws-protocol-tests/1.54.0/smithy-aws-protocol-tests-1.54.0.jar
```

At the end it creates in the same directory a file `results.json` that contains all the results from running the protocol tests. 

A gist of the resulting JSON file can be found [here](https://gist.github.com/sugmanue/6d7101f080ce85d6cff3aca99b8847e7)

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
@@ -119,7 +119,6 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Unit>() {
         writers.flushWriters()
     }
 
-
     fun generateShapes() {
         logger.info("Generating Kotlin client for service ${settings.service}")
 
@@ -159,17 +158,15 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Unit>() {
         integrations.forEach { it.writeAdditionalFiles(baseGenerationContext, writers) }
     }
 
-    fun generationContext(): ProtocolGenerator.GenerationContext {
-        return ProtocolGenerator.GenerationContext(
-            settings,
-            model,
-            service,
-            symbolProvider,
-            integrations,
-            protocolGenerator!!.protocol,
-            writers,
-        )
-    }
+    fun generationContext(): ProtocolGenerator.GenerationContext = ProtocolGenerator.GenerationContext(
+        settings,
+        model,
+        service,
+        symbolProvider,
+        integrations,
+        protocolGenerator!!.protocol,
+        writers,
+    )
 
     override fun getDefault(shape: Shape?) { }
 
--- a/codegen/smithy-kotlin-protocol-tests-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/pt/KotlinProtocolTestCodegenPlugin.kt
+++ b/codegen/smithy-kotlin-protocol-tests-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/pt/KotlinProtocolTestCodegenPlugin.kt
@@ -14,22 +14,28 @@ import software.amazon.smithy.kotlin.codegen.rendering.protocol.TestMemberDelta
 import software.amazon.smithy.kotlin.codegen.rendering.protocol.pt.ProtocolTestGenerator
 import software.amazon.smithy.kotlin.codegen.rendering.writeGradleBuild
 
+// We redefine the kotlin-test and smithy-tests dependencies since for this use case
+// we need them in the implementation scope instead of just in the test scope.
+val KOTLIN_TEST_RT = KotlinDependency(
+    GradleConfiguration.Implementation,
+    "kotlin.test",
+    "org.jetbrains.kotlin",
+    "kotlin-test",
+    KOTLIN_COMPILER_VERSION,
+)
+val SMITHY_TEST_RT = KotlinDependency(
+    GradleConfiguration.Implementation,
+    KotlinDependency.SMITHY_TEST.namespace,
+    KotlinDependency.SMITHY_TEST.group,
+    KotlinDependency.SMITHY_TEST.artifact,
+    KotlinDependency.SMITHY_TEST.version,
+)
+
 /**
  * Plugin to trigger Kotlin protocol tests code generation. This plugin also generates the client and the
  * request/response shapes.
  */
 public class KotlinProtocolTestCodegenPlugin : SmithyBuildPlugin {
-
-    // We redefine the kotlin-test and smithy-tests dependencies since for this use case
-    // we need them in the implementation scope instead of just in the test scope.
-    val KOTLIN_TEST_RT = KotlinDependency(
-        GradleConfiguration.Implementation, "kotlin.test", "org.jetbrains.kotlin", "kotlin-test",
-        KOTLIN_COMPILER_VERSION,
-    )
-    val SMITHY_TEST_RT = KotlinDependency(
-        GradleConfiguration.Implementation, KotlinDependency.SMITHY_TEST.namespace,
-        KotlinDependency.SMITHY_TEST.group, KotlinDependency.SMITHY_TEST.artifact, KotlinDependency.SMITHY_TEST.version,
-    )
 
     override fun getName(): String = "kotlin-protocol-tests-codegen"
 
@@ -39,7 +45,7 @@ public class KotlinProtocolTestCodegenPlugin : SmithyBuildPlugin {
         codegen.generateShapes()
 
         // Generate the protocol tests
-        val ctx = codegen.generationContext();
+        val ctx = codegen.generationContext()
         val requestTestBuilder = ProtocolTestRequestGenerator.Builder()
         val responseTestBuilder = ProtocolTestResponseGenerator.Builder()
         val errorTestBuilder = ProtocolTestErrorGenerator.Builder()
@@ -54,7 +60,7 @@ public class KotlinProtocolTestCodegenPlugin : SmithyBuildPlugin {
             ignoredTests,
         ).generateProtocolTests()
 
-        val writers = ctx.delegator;
+        val writers = ctx.delegator
         writers.finalize()
         val settings = ctx.settings
 
--- a/codegen/smithy-kotlin-protocol-tests-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/pt/ProtocolTestErrorGenerator.kt
+++ b/codegen/smithy-kotlin-protocol-tests-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/pt/ProtocolTestErrorGenerator.kt
@@ -13,8 +13,7 @@ import software.amazon.smithy.protocoltests.traits.HttpResponseTestCase
 /**
  * Renders a class with all the defined protocol tests for responses with errors.
  */
-open class ProtocolTestErrorGenerator protected constructor(builder: HttpProtocolUnitTestErrorGenerator.Builder) :
-    HttpProtocolUnitTestErrorGenerator(builder) {
+open class ProtocolTestErrorGenerator protected constructor(builder: HttpProtocolUnitTestErrorGenerator.Builder) : HttpProtocolUnitTestErrorGenerator(builder) {
 
     /**
      * Render a test class and unit tests for the specified [testCases]
--- a/codegen/smithy-kotlin-protocol-tests-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/pt/ProtocolTestsGenerator.kt
+++ b/codegen/smithy-kotlin-protocol-tests-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/pt/ProtocolTestsGenerator.kt
@@ -39,7 +39,6 @@ class ProtocolTestGenerator(
 ) {
     private val logger = Logger.getLogger(javaClass.name)
 
-
     /**
      * Generates the API HTTP protocol tests defined in the smithy model.
      */
@@ -74,7 +73,7 @@ class ProtocolTestGenerator(
             ?.filter(::isTestCaseAllowedForRunMode)
 
         if (requestTests?.isEmpty() != false) {
-            return;
+            return
         }
         requestTests.let { testCases ->
             val testOperationName = operation.id.name.replaceFirstChar { c -> c.uppercaseChar() }
@@ -103,7 +102,7 @@ class ProtocolTestGenerator(
             ?.filter(::isTestCaseAllowedForRunMode)
 
         if (responseTests?.isEmpty() != false) {
-            return;
+            return
         }
         responseTests.let { testCases ->
             val testOperationName = operation.id.name.replaceFirstChar { c -> c.uppercaseChar() }
@@ -170,7 +169,7 @@ class ProtocolTestGenerator(
         ctx.delegator.useFileWriter(testFilename, ctx.settings.pkg.name + ".pt") { writer ->
             logger.fine("Generating protocol test runner")
             writer.openBlock("public fun main() {")
-                .write("val results = ArrayList<#T>()", ProtocolTestsUtils.TestResult);
+                .write("val results = ArrayList<#T>()", ProtocolTestsUtils.TestResult)
             for (testClass in classes) {
                 writer.write("#L(results).runAll()", testClass)
             }
--- a/tests/protocol-tests-runner/src/main/kotlin/software/amazon/smithy/kotlin/protocolTests/Generation.kt
+++ b/tests/protocol-tests-runner/src/main/kotlin/software/amazon/smithy/kotlin/protocolTests/Generation.kt
@@ -1,6 +1,5 @@
 package software.amazon.smithy.kotlin.protocolTests
 
-import jdk.internal.net.http.common.Log.errors
 import software.amazon.smithy.build.FileManifest
 import software.amazon.smithy.build.PluginContext
 import software.amazon.smithy.kotlin.codegen.pt.KotlinProtocolTestCodegenPlugin
@@ -29,14 +28,12 @@ fun generateProtocolTests(model: Model, service: ServiceShape, codegenPath: Path
  * Creates the `Node` instance that represents the configuration used to run the
  * `KotlinProtocolTestCodegenPlugin` plugin.
  */
-private fun settingsForService(service: ServiceShape): ObjectNode {
-    return ObjectNode.builder()
-        .withMember("service", service.id.toString())
-        .withMember("package", packageSettings(service))
-        .withMember("build", buildSettings())
-        .withMember("api", apiSettings())
-        .build()
-}
+private fun settingsForService(service: ServiceShape): ObjectNode = ObjectNode.builder()
+    .withMember("service", service.id.toString())
+    .withMember("package", packageSettings(service))
+    .withMember("build", buildSettings())
+    .withMember("api", apiSettings())
+    .build()
 
 /**
  * Crates the `Node` instance that defines the package settings.
@@ -52,23 +49,19 @@ private fun packageSettings(service: ServiceShape): ObjectNode {
 /**
  * Creates the `Node` instance that defines the build settings.
  */
-private fun buildSettings(): ObjectNode {
-    return ObjectNode.builder()
-        .withMember("generateFullProject", true)
-        .withMember(
-            "optInAnnotations",
-            ArrayNode.arrayNode(StringNode.from("aws.smithy.kotlin.runtime.InternalApi")),
-        )
-        .withMember("rootProject", true)
-        .build()
-}
+private fun buildSettings(): ObjectNode = ObjectNode.builder()
+    .withMember("generateFullProject", true)
+    .withMember(
+        "optInAnnotations",
+        ArrayNode.arrayNode(StringNode.from("aws.smithy.kotlin.runtime.InternalApi")),
+    )
+    .withMember("rootProject", true)
+    .build()
 
 /**
  * Creates the `Node` instance that defines the api settings.
  */
-private fun apiSettings(): ObjectNode {
-    return ObjectNode.builder()
-        // By default, this value is WHEN_DIFFERENT, which is incorrect for all protocols
-        .withMember("defaultValueSerializationMode", "always")
-        .build()
-}
+private fun apiSettings(): ObjectNode = ObjectNode.builder()
+    // By default, this value is WHEN_DIFFERENT, which is incorrect for all protocols
+    .withMember("defaultValueSerializationMode", "always")
+    .build()
--- a/tests/protocol-tests-runner/src/main/kotlin/software/amazon/smithy/kotlin/protocolTests/Runner.kt
+++ b/tests/protocol-tests-runner/src/main/kotlin/software/amazon/smithy/kotlin/protocolTests/Runner.kt
@@ -49,7 +49,7 @@ fun main(args: Array<String>) {
                 // Build the client and protocol tests
                 buildProtocolTests(codegenPath)
                 // Run the protocol tests and get the results
-                val results = runProtocolTests(codegenPath);
+                val results = runProtocolTests(codegenPath)
                 // Add the result to the final report
                 jsonWriter.writeEncodedValue(results)
             }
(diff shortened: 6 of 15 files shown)


.changes/e9fa30cd-1aa7-4c5d-b01a-384078b8863a.json
```diff
@@ -0,0 +1,5 @@
+{
+    "id": "e9fa30cd-1aa7-4c5d-b01a-384078b8863a",
+    "type": "feature",
+    "description": "Added a protocol tests runner, a new mechanism to run the tests and report the results in JSON format detached from the test phase of a regular build."
+}
```
ianbotsf, line 5: **Nit**: This update isn't user-facing so wouldn't typically need a changelog message. You can delete this and add the **no-changelog** label to this PR.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinDependency.kt
```diff
@@ -38,6 +38,7 @@ private fun getDefaultRuntimeVersion(): String {
 const val RUNTIME_GROUP: String = "aws.smithy.kotlin"
 val RUNTIME_VERSION: String = System.getProperty("smithy.kotlin.codegen.clientRuntimeVersion", getDefaultRuntimeVersion())
 val KOTLIN_COMPILER_VERSION: String = System.getProperty("smithy.kotlin.codegen.kotlinCompilerVersion", "2.1.0")
+val SHADOW_JAR_VERSION: String = System.getProperty("com.github.johnrengelman.shadowVersion", "8.1.1")
```
ianbotsf, line 41: **Nit**: We need to stop embedding hardcoded versions into our codegen files and should be auto-discovering them instead, similar to how we do with `clientRuntimeVersion` above. I don't expect you to implement that, though. Please leave a `// FIXME` on these two lines mentioning that we should detect versions.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/GradleGenerator.kt
```diff
@@ -53,6 +54,12 @@ fun writeGradleBuild(
                 }
             },
         )
+        if (enableApplication) {
+            indent()
+            write("application")
+            write("id(\"com.github.johnrengelman.shadow\") version #S", SHADOW_JAR_VERSION)
+            dedent()
+        }
```
ianbotsf, line 62: **Style**: Prefer Smithy placeholders or [triple-quotes](https://kotlinlang.org/docs/strings.html#multiline-strings) to avoid escaping strings which may contain quotes.

```kotlin
// either
write("id(#S) version #S", "com.github.johnrengelman.shadow", SHADOW_JAR_VERSION)

// or
write("""id("com.github.johnrengelman.shadow") version #S""", SHADOW_JAR_VERSION)
```

 Applies in other spots as well.

codegen/smithy-kotlin-protocol-tests-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/pt/ProtocolTestErrorGenerator.kt
```diff
@@ -0,0 +1,72 @@
...
+ * Renders a class with all the defined protocol tests for responses with errors.
+ */
+open class ProtocolTestErrorGenerator protected constructor(builder: HttpProtocolUnitTestErrorGenerator.Builder) : HttpProtocolUnitTestErrorGenerator(builder) {
+
+    /**
+     * Render a test class and unit tests for the specified [testCases]
+     */
+    override fun renderTestClass(testClassName: String) {
+        writer.addImport("${ctx.settings.pkg.name}", "*")
+
```
ianbotsf, line 23: **Nit**: We should not be adding more star imports to our codegen. Let's remove this and use `#T` placeholders to handle automatic importing.

codegen/smithy-kotlin-protocol-tests-codegen/build.gradle.kts
```diff
@@ -0,0 +1,103 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import aws.sdk.kotlin.gradle.dsl.configurePublishing
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    alias(libs.plugins.kotlin.jvm)
```
ianbotsf, line 10: **Question**: Does this _need_ to be JVM only? Or can these tests be run in Kotlin/Native as well?

codegen/smithy-kotlin-protocol-tests-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/pt/KotlinProtocolTestCodegenPlugin.kt
```diff
@@ -0,0 +1,79 @@
...
+    "kotlin-test",
+    KOTLIN_COMPILER_VERSION,
+)
+val SMITHY_TEST_RT = KotlinDependency(
+    GradleConfiguration.Implementation,
+    KotlinDependency.SMITHY_TEST.namespace,
+    KotlinDependency.SMITHY_TEST.group,
+    KotlinDependency.SMITHY_TEST.artifact,
+    KotlinDependency.SMITHY_TEST.version,
+)
```
ianbotsf, line 32: **Nit**: Could be `private`.

ianbotsf, line 32: **Question**: What does "RT" mean in these names?

ianbotsf, line 32: **Nit**: Prefer the generated `copy` method that every `data class` has:

```kotlin
val KOTLIN_TEST_RT = KOTLIN_TEST.copy(config = GradleConfiguration.Implementation)
val SMITHY_TEST_RT = SMITHY_TEST.copy(config = GradleConfiguration.Implementation)
```

This saves us from duplicating the other values in code and potentially diverging.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
```diff
@@ -144,7 +142,9 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Unit>() {
             logger.info("[${service.id}] Generating auth scheme provider for protocol $protocol")
             generateAuthSchemeProvider(ctx)
         }
+    }
 
+    fun generateNonShapes() {
```
ianbotsf, line 147: **Style**: The name of this method is not very clear because it's not obvious what "non-shapes" are. Can we come up with a more descriptive name that will make more sense in calling code? (e.g., `generateIntegrations`, `generateAdditionalFiles`, etc.)

codegen/smithy-kotlin-protocol-tests-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/pt/KotlinProtocolTestCodegenPlugin.kt
```diff
@@ -0,0 +1,79 @@
...
+        val ignoredTests = TestMemberDelta(
+            setOf(),
+        )
+        ProtocolTestGenerator(
+            ctx,
+            requestTestBuilder,
+            responseTestBuilder,
+            errorTestBuilder,
+            ignoredTests,
+        ).generateProtocolTests()
```
ianbotsf, line 61: **Question**: Doesn't `generateShapes` already generate the protocol tests? Why are we doing it twice?

codegen/smithy-kotlin-protocol-tests-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/pt/KotlinProtocolTestCodegenPlugin.kt
```diff
@@ -0,0 +1,79 @@
...
+            val dependencies = writers.dependencies
+                .mapNotNull { it.properties["dependency"] as? KotlinDependency }
+                .distinct()
+            val newDependencies = ArrayList<KotlinDependency>()
+            newDependencies.addAll(dependencies)
+            newDependencies.add(KOTLIN_TEST_RT)
+            newDependencies.add(SMITHY_TEST_RT)
+            writeGradleBuild(settings, writers.fileManifest, newDependencies, enableApplication = true)
+        }
+        writers.flushWriters()
```
ianbotsf, line 77: **Correctness**: All of this seems like the responsibility of the `CodegenVisitor` and not the plugin directly. I think the architecture of this split may not be grouping concerns properly.

I suggest making `CodegenVisitor` more flexible and passing arguments to its constructor for the normal and special cases here. Alternatively, we could make a new visitor class for this runner use case, which inherits from the existing `CodegenVisitor` or from a refactored abstract parent class.

codegen/smithy-kotlin-protocol-tests-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/pt/ProtocolTestErrorGenerator.kt
```diff
@@ -0,0 +1,72 @@
...
+                    renderTestFunctionCall(test)
+                }
+            }
+            .closeBlock("}")
+            .call {
+                for (test in testCases) {
+                    renderTestFunction(test)
+                }
+            }
+            .closeBlock("}")
```
ianbotsf, line 38: **Style**: Prefer the Kotlin extensions to `CodeWriter` such as [`withBlock`](https://github.com/smithy-lang/smithy-kotlin/blob/da208cdcb21e56d96e7c9e9c1765ac42a157520d/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/AbstractCodeWriterExt.kt#L13-L36):

```kotlin
writer.withBlock("internal class $testClassName(val results: MutableList<TestResult>) {", "}") {
    withBlock("public fun runAll() {", "}") {
        for (test in testCases) {
            renderTestFunctionCall(test)
        }
    }

    for (test in testCases) {
        renderTestFunction(test)
    }
}
```

Applies in other places as well.

codegen/smithy-kotlin-protocol-tests-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/pt/ProtocolTestsGenerator.kt
```diff
@@ -0,0 +1,202 @@
...
+        .write("val sw = java.io.StringWriter()")
+        .write("ex.printStackTrace(java.io.PrintWriter(sw))")
+        .write("res.log = sw.toString()")
+        .closeAndOpenBlock("} catch (ex: Exception) {")
+        .write("res.result = #T.ERRORED", ProtocolTestsUtils.Result)
+        .write("val sw = java.io.StringWriter()")
+        .write("ex.printStackTrace(java.io.PrintWriter(sw))")
+        .write("res.log = sw.toString()")
+        .closeBlock("}")
+}
```
ianbotsf, line 200: **Suggestion**: These prelude and epilogue functions (kudos on the names! 🤩) feel like they shouldn't be generating the same code over and over for each test. Instead it'd be cleaner if test classes had a `runTest` method which wrapped the test body in an arbitrary block.

Also, we can clean up the `catch` cases a bit by combining them into a single `catch` (they only differ in the `Result` value) and by using [the `stackTraceToString` method](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/stack-trace-to-string.html).

Example generated code:

```kotlin
internal class DefaultFooTests {
    private fun `testFoo1`() = runTest {
        // test case 1
    }

    private fun `testFoo2`() = runTest {
        // test case 2
    }

    private fun runTest(block: () -> Unit) {
        // The default implementation is just to invoke the block
        block()
    }
}

internal class ProtocolTestRunnerFooTests {
    private fun `testFoo1`() = runTest {
        // test case 1
    }

    private fun `testFoo2`() = runTest {
        // test case 2
    }

    private fun runTest(block: () -> Unit) {
        // For the runner we need to capture each failure and handle it so tests keep running
        try {
            block()
        } catch (ex: Exception) {
            res.result = if (ex is AssertionError) Result.FAILED else Result.ERRORED
            res.log = ex.stackTraceToString()
        }
    }
}
```

This structure of generated code would lend itself nicely to a codegen hierarchy which has a base `renderRunTest` method which is overridden in your implementation.

codegen/smithy-kotlin-protocol-tests-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/pt/ProtocolTestsRuntimeTypes.kt
```diff
@@ -0,0 +1,24 @@
...
+import software.amazon.smithy.kotlin.codegen.core.KotlinDependency
+import software.amazon.smithy.kotlin.codegen.core.RuntimeTypePackage
+
+val PROTOCOL_TEST_RUNTIME_UTILS = KotlinDependency(
+    GradleConfiguration.Implementation,
+    "software.amazon.smithy.kotlin.protocolTests.utils",
+    "software.amazon.smithy.kotlin",
+    "protocol-tests-utils",
+    "0.34.17-SNAPSHOT",
+)
```
ianbotsf, line 17: **Correctness**: Runtime utils shouldn't be using the codegen version, it should be using the runtime version (i.e., [`RUNTIME_VERSION` defined in **KotlinDependency.kt**](https://github.com/smithy-lang/smithy-kotlin/blob/da208cdcb21e56d96e7c9e9c1765ac42a157520d/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinDependency.kt#L39)).

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/GradleGenerator.kt
```diff
@@ -74,6 +81,13 @@ fun writeGradleBuild(
         )
     }
 
+    if (enableApplication) {
+        writer.write("")
+        writer.openBlock("application {")
+        writer.write("mainClass.set(\"${settings.pkg.name}.pt.RunnerKt\")")
+        writer.closeBlock("}")
+    }
```
ianbotsf, line 89: **Question**: Why do the generated projects need the `application` plugin if **protocol-tests-runner** already has it statically configured?

codegen/smithy-kotlin-protocol-tests-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/pt/KotlinProtocolTestCodegenPlugin.kt
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package software.amazon.smithy.kotlin.codegen.pt
+
```
ianbotsf, line 6: **Style**: We should use a more descriptive namespace than `pt`. I suggest `tests.protocol`.

tests/protocol-tests-runner/src/main/kotlin/software/amazon/smithy/kotlin/protocolTests/Generation.kt
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Crates the `Node` instance that defines the package settings.
+ */
+private fun packageSettings(service: ServiceShape): ObjectNode {
+    val version = service.version ?: "1.0"
+    return ObjectNode.builder()
+        .withMember("version", version)
+        .withMember("name", "smithy.protocolTests")
+        .build()
+}
```
ianbotsf, line 47: **Question**: Shouldn't we be using IDL version 2.0 by default?

tests/protocol-tests-runner/src/main/kotlin/software/amazon/smithy/kotlin/protocolTests/Runner.kt
```diff
@@ -0,0 +1,114 @@
...
+    // requests or for responses.
+    TopDownIndex.of(model).getContainedOperations(service).forEach { operation ->
+        if (operation.hasTrait(HttpResponseTestsTrait.ID) ||
+            operation.hasTrait(HttpRequestTestsTrait.ID)
+        ) {
+            return true
+        }
+    }
+    return false
+}
```
ianbotsf, line 83: **Style**: Can be simplified:

```kotlin
private fun hasStandardProtocolTests(service: ServiceShape, model: Model): Boolean {
    val isAwsServiceTest = service
        .getTrait<TagsTrait>()
        ?.values
        // We add a tag to protocol tests for specific AWS services
        // that require customizations to run properly. Those
        // not standard protocol tests are filtered out here.
        ?.contains("aws-service-test")
        ?: false

    return 
        !isAwsServiceTest &&
        TopDownIndex
            .of(model)
            .getContainedOperations(service)
            .any {
                it.hasTrait<HttpResponseTestsTrait>() ||
                it.hasTrait<HttpRequestTestsTrait>()
            }
    }
}
```

tests/protocol-tests-runner/src/main/kotlin/software/amazon/smithy/kotlin/protocolTests/Runner.kt
```diff
@@ -0,0 +1,114 @@
...
+ * to generate an uber JAR that we can the call directly without having to setup
+ * all the classpath elements.
+ */
+private fun buildProtocolTests(codegenPath: Path) {
+    ProcessBuilder("gradle", "build", "-x", "test")
+        .directory(codegenPath.toFile())
+        .inheritIO()
+        .start()
+        .waitFor()
+}
```
ianbotsf, line 97: **Correctness**: This code assumes that Gradle will be installed on the `$PATH` in the current environment. IDK about other SDK developers but I never _install_ Gradle on my machines, I just use the Gradle wrapper that comes with each repository. On my machine and anyone else's who doesn't install Gradle, this code will fail.

codegen/smithy-kotlin-protocol-tests-codegen/build.gradle.kts
```diff
@@ -0,0 +1,103 @@
...
+publishing {
+    publications {
+        create<MavenPublication>("codegen") {
+            from(components["java"])
+            artifact(sourcesJar)
+        }
+    }
+}
+
+configurePublishing("smithy-kotlin", "smithy-lang")
```
ianbotsf, line 103: **Question**: Why do we need to publish this project? Will we need to fetch/run it from Maven Central in some cases?

Applies in other places too.

tests/protocol-tests-utils/src/main/kotlin/software/amazon/smithy/kotlin/protocolTests/utils/JsonWriter.kt
```diff
@@ -0,0 +1,192 @@
+package software.amazon.smithy.kotlin.protocolTests.utils
+
+import java.io.Closeable
+import java.io.Writer
+
+/**
+ * A JSON simple, streaming, JSON writer.
+ */
+class JsonWriter(private val writer: Writer) : Closeable {
```
ianbotsf, line 9: **Style**: We should replace this custom writer with **kotlinx-serialization-json**.

You can use [`buildJsonObject`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/build-json-object.html) to create a JSON document instance and then either [`encodeToString`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json/encode-to-string.html) or [`encodeToStream`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/encode-to-stream.html) to write out the final JSON body:

```kotlin
val jsonDoc = buildJsonObject {
    put("foo", "bar")
    putJsonObject("baz") {
        put("a", "Apple")
        put("b", "Banana")
    }
    
    maybeNullQux?.let { put("qux", it) }
}

val jsonText = Json.encodeToString(jsonDoc)
/*
    Resulting JSON:
    {
      "foo": "bar",
      "baz": {
        "a": "Apple",
        "b": "Banana"
      },
      "qux": "turns out it wasn't null!"
    }
*/
```

tests/protocol-tests-utils/src/main/kotlin/software/amazon/smithy/kotlin/protocolTests/utils/Utils.kt
```diff
@@ -0,0 +1,61 @@
...
+    REQUEST("request"),
+    RESPONSE("response"),
+}
+
+data class TestResult(
+    val testId: String,
+    val testType: TestType,
+    var result: Result = Result.PASSED,
+    var log: String? = null,
+)
```
ianbotsf, line 22: **Style**:

This can also be replaced with **kotlinx-serialziation** by marking the classes as serializable:

```kotlin
@Serializable
enum class Result(val value: String) {
    @SerialName("passed") PASSED,
    @SerialName("failed") FAILED,
    @SerialName("errored") ERRORED,
    @SerialName("skipped") SKIPPED,
}

@Serializable
enum class TestType(val value: String) {
    @SerialName("request") REQUEST,
    @SerialName("response") RESPONSE,
}

@Serializable
data class TestResult(
    val testId: String,
    val testType: TestType,
    var result: Result = Result.PASSED,
    var log: String? = null,
)
```

Now the `Json` serializer knows how to read/write these classes we can mix it up with your hand-written (i.e., unmodeled) results container via [`encodeToJsonElement`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/encode-to-json-element.html):

```kotlin
fun writeResults(serviceId: String, protocolId: String, results: List<TestResult>) {
    val jsonDoc = buildJsonObject {
        put("service", serviceId)
        put("protocol", protocolId)
        putJsonArray("results") {
            for (result in results) {
                add(Json.encodeToJsonElement(result))
            }
        }
    }
    
    println(Json.encodeToString(jsonDoc))
}
```

...but wait, there's more! 😃 By actually modelling your results container as a data class, you can achieve ultimate JSON zen:

```kotlin
@Serializable
data class ProtocolTestResults(service: String, protocol: String, results: List<TestResult>)

fun writeResults(serviceId: String, protocolId: String, results: List<TestResult>) {
    val resultsObj = ProtocolTestResults(serviceId, protocolId, results)
    println(Json.encodeToString(resultsObj))
}
```

ianbotsf (changes requested):
Looks like a great start. Thanks for contributing this @sugmanue!

PR #1294, opened 2025-06-09
fix: extra byte read from chunk transfer

xinsong-cui:
## Description of changes

### DIFF
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsChunkedSource.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsChunkedSource.kt
@@ -40,6 +40,9 @@ public class AwsChunkedSource(
         trailingHeaders,
     )
 
+    // The number of bytes transferred without chunk metadata
+    public var contentBytesTransferred: Long = 0L
+
     override fun read(sink: SdkBuffer, limit: Long): Long {
         require(limit >= 0L) { "Invalid limit ($limit) must be >= 0L" }
         // COROUTINE SAFETY: runBlocking is allowed here because SdkSource is a synchronous blocking interface
@@ -47,7 +50,11 @@ public class AwsChunkedSource(
             chunkReader.ensureValidChunk()
         }
         if (!isChunkValid) return -1L
-        return chunkReader.chunk.read(sink, limit)
+
+        val actualBytesTransferred = chunkReader.chunk.read(sink, limit)
+        contentBytesTransferred = actualBytesTransferred - chunkReader.chunkMetadataBytes
+
+        return actualBytesTransferred
     }
 
     override fun close() {
--- a/runtime/auth/aws-signing-common/api/aws-signing-common.api
+++ b/runtime/auth/aws-signing-common/api/aws-signing-common.api
@@ -23,7 +23,9 @@ public final class aws/smithy/kotlin/runtime/auth/awssigning/AwsChunkedSource : 
 	public fun <init> (Laws/smithy/kotlin/runtime/io/SdkSource;Laws/smithy/kotlin/runtime/auth/awssigning/AwsSigner;Laws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig;[BLaws/smithy/kotlin/runtime/http/DeferredHeaders;)V
 	public synthetic fun <init> (Laws/smithy/kotlin/runtime/io/SdkSource;Laws/smithy/kotlin/runtime/auth/awssigning/AwsSigner;Laws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig;[BLaws/smithy/kotlin/runtime/http/DeferredHeaders;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public fun close ()V
+	public final fun getContentBytesTransferred ()J
 	public fun read (Laws/smithy/kotlin/runtime/io/SdkBuffer;J)J
+	public final fun setContentBytesTransferred (J)V
 }
 
 public final class aws/smithy/kotlin/runtime/auth/awssigning/AwsSignatureType : java/lang/Enum {
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/internal/AwsChunkedReader.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/internal/AwsChunkedReader.kt
@@ -58,6 +58,13 @@ internal class AwsChunkedReader(
     internal var hasLastChunkBeenSent: Boolean = false
 
     /**
+     * Tracks the number of bytes used for chunk metadata.
+     * This includes chunk headers, terminators (CRLF), and trailers.
+     * Used to calculate the actual payload bytes transferred by subtracting metadata bytes from the total bytes read.
+     */
+    internal var chunkMetadataBytes: Long = 0L
+
+    /**
      * Ensures that the internal [chunk] is valid for reading. If it's not valid, try to load the next chunk. Note that
      * this function will suspend until the whole chunk has been loaded.
      *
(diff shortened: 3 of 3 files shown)


runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsChunkedSource.kt
```diff
@@ -40,14 +40,24 @@ public class AwsChunkedSource(
...
         val isChunkValid = runBlocking {
             chunkReader.ensureValidChunk()
         }
         if (!isChunkValid) return -1L
-        return chunkReader.chunk.read(sink, limit)
+
+        val totalBytesTransferred = chunkReader.chunk.read(sink, limit)
+        contentBytesTransferred = totalBytesTransferred - chunkReader.chunkMetadataBytes
+
+        return totalBytesTransferred
```
ianbotsf: **Question**: I feel pretty strongly that `AwsChunkedSource.read` should return the number of bytes which were _read_ from the delegate source rather than the number of bytes _written_ to the destination sink. Can we just `return totalBytesTransferred - chunkReader.chunkMetadataBytes` instead of storing it in a property? What do we _need_ the total bytes transferred for (at least, as the output from `read`)?

xinsong-cui: We still need to return totalBytes for some of our tests: https://github.com/smithy-lang/smithy-kotlin/blob/444ff8a71ad8a6658264c44796eda736aa05405c/runtime/auth/aws-signing-tests/common/src/aws/smithy/kotlin/runtime/auth/awssigning/tests/AwsChunkedTestBase.kt#L445
If we not return the `totalBytesTransferred`, those tests will fail.

lauzadis: I'm pretty sure we can find a way around that, maybe by exposing the total bytes as some kind of test utility?

xinsong-cui:
Manually added **no-changelog** label as changelog check somehow not passing

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsChunkedSource.kt
```diff
@@ -40,14 +40,36 @@ public class AwsChunkedSource(
...
+    /**
+     * Tracks the total bytes transferred, including chunk metadata.
+     */
+    private var totalBytesTransferred: Long = 0L
+
+    /**
+     * Gets the total bytes transferred for testing purposes only.
+     */
+    @InternalApi
+    public fun getTotalBytesTransferred(): Long = totalBytesTransferred
```
lauzadis: I don't think adding a public API "for testing purposes only" is the best. Can we refactor this to make it internal-only, or otherwise refactor the tests?

lauzadis (changes requested):

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsChunkedSource.kt
xinsong-cui: We cannot make it internal-only, unless refactor all chunk related tests from **aws-signing-tests** to **aws-signing-common** but that doesn't feel right.

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsChunkedByteReadChannel.kt
```diff
@@ -45,10 +45,32 @@ public class AwsChunkedByteReadChannel(
...
+    /**
+     * Tracks the total bytes transferred, including chunk metadata.
+     */
+    private var totalBytesTransferred: Long = 0L
+
+    /**
+     * Gets the total bytes transferred for testing purposes only.
+     */
+    @InternalApi
+    public fun getTotalBytesTransferred(): Long = totalBytesTransferred
```
lauzadis: Why does this need to be a function? Why can't we make the `var totalBytesTransferred` above `@InternalApi public`

.changes/f643c64f-aa0b-43cc-821c-33abf9f93ee8.json
lauzadis, line 1: This is a breaking change, right?

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsChunkedByteReadChannel.kt
```diff
@@ -45,10 +45,32 @@ public class AwsChunkedByteReadChannel(
...
+
     override suspend fun read(sink: SdkBuffer, limit: Long): Long {
         require(limit >= 0L) { "Invalid limit ($limit) must be >= 0L" }
-        if (!chunkReader.ensureValidChunk()) return -1L
-        return chunkReader.chunk.read(sink, limit)
+
+        // Reset metadata bytes counter
+        chunkReader.chunkMetadataBytes = 0
```
lauzadis: Why can't the chunkReader reset its own metadata bytes when needed? I don't really like consumers (AwsChunkedByteReadChannel) having to keep track of the internal state of the reader

xinsong-cui: Yeah I agree, moved to chunkReader

.changes/f643c64f-aa0b-43cc-821c-33abf9f93ee8.json
xinsong-cui: Sorry I don't get it, why this is a breaking change?

lauzadis: We are removing the chunked overhead from the reported bytes. If users are relying on the previous number, they will be broken. It's a breaking _behavior_ change, not an API break. We can discuss offline

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsChunkedSource.kt
ianbotsf: Do we need to track the total underlying bytes read in _this_ class? The `delegate` used to create the `AwsChunkedSource` knows how many raw bytes are being read so that feels like a more natural place to track it.

I wonder whether our tests could wrap the delegates in some abstraction that accumulates byte counts _before_ passing them to `AwsChunkedSource`. Something like this:

```kotlin
class CountingSource(delegate: SdkSource) : SdkSourceObserver(delegate) {
    var bytesRead: Long = 0L // track how many bytes are read from this source
        private set // `private set` so that it's only mutable internally

    override fun observe(data: ByteArray, offset: Int, length: Int) {
        // this function is called by the `read` method for each chunk of data
        bytesRead += length.toLong()
    }
}
```

Then each test makes a `CountingSource` and asserts on the total length:

```kotlin
val source = // wherever the existing sources already come from
val coutingSource = CountingSource(source)
val chunked = AwsChunkedSource(countingSource, ...)

// Process the chunked source

assertEquals(EXPECTED_BYTE_COUNT, countingSource.bytesRead)
```

xinsong-cui: I gave it a try but doest work well. 
I might be wrong but wasn't the `countingSource.BytesRead` still count the bytes without metadata? 
We are getting bytes size (without metadata) from chunk source [read function](https://github.com/smithy-lang/smithy-kotlin/blob/main/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/io/internal/Observers.kt#L26), then calling observe and pass that bytes

Another thing is we don't have an observer for `SdkByteChannel` yet.

xinsong-cui:
close for #1386

PR #1298, opened 2025-06-19
fix: revert to Dokka1 module paths

lauzadis:
Preserve Dokka v1's paths, which are just the module name, rather than the full path to the module. 
 
https://kotlinlang.org/docs/dokka-migration.html#revert-to-the-dgp-v1-directory-behavior

## Description of changes

### DIFF
--- a/runtime/auth/aws-credentials/build.gradle.kts
+++ b/runtime/auth/aws-credentials/build.gradle.kts
@@ -30,3 +30,7 @@ kotlin {
         }
     }
 }
+
+dokka {
+    modulePath = "aws-credentials"
+}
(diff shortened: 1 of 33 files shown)


runtime/auth/aws-credentials/build.gradle.kts
```diff
@@ -30,3 +30,7 @@ kotlin {
         }
     }
 }
+
+dokka {
+    modulePath = "aws-credentials"
+}
```
ianbotsf, line 36: **Question**: Could we replace all of these per-module declarations with a top-level block instead?

```kotlin
subprojects {
    dokka {
        modulePath = this@subprojects.name
    }
}
```

lauzadis: Possibly, thanks, I'll take a look at the simplification

lauzadis: Here is a PR with the simplification, thanks! https://github.com/smithy-lang/smithy-kotlin/pull/1299

PR #1306, opened 2025-06-27
Service sdk setup

luigi617:
Setting up the service sdk.

## Description of changes

- `serviceExecute` function for generating service.
- Include dependencies for Ktor.
- Gradle build file modification for running the service.
- `ServiceStubGenerator` class for generation. It includes only main file, simple static routing, and simple standard cbor serde.


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/ServiceStubGenerator.kt
```diff
@@ -0,0 +1,82 @@
...
+        renderPerOperationHandlers()
+    }
+
+    // Writes `Main.kt` that boots the embedded Ktor service.
+    private fun renderMainFile() {
+        delegator.useFileWriter("Main.kt", packageName) { writer ->
+
+            writer.addImport(RuntimeTypes.KtorServerCore.embeddedServer)
+            writer.addImport(RuntimeTypes.KtorServerNetty.Netty)
+            writer.dependencies.addAll(KotlinDependency.KTOR_LOGGING_BACKEND.dependencies)
```
lauzadis: I think this is fine for now, but you might be missing an abstraction that will allow `ServiceStubGenerator` to generate non-Ktor services in the future. We might need some `ServiceStubSettings` that the user can configure to choose between Ktor and other options, as well as configure Ktor generation

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/ServiceStubGenerator.kt
```diff
@@ -0,0 +1,82 @@
...
+
+            writer.openBlock("public fun main(): Unit {")
+                .openBlock("embeddedServer(Netty, port = 8080) {")
```
lauzadis: style: instead of adding the `embeddedServer` and `Netty` imports (L25-26) manually, you can reference it using a `#T` symbol, and the import gets added automatically

```
.openBlock("#T(#T, port = 8080) {", RuntimeTypes.KtorServerCore.embeddedServer, RuntimeTypes.KtorServerCore.Netty)
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/ServiceStubGenerator.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    // Writes `Main.kt` that boots the embedded Ktor service.
+    private fun renderMainFile() {
+        delegator.useFileWriter("Main.kt", packageName) { writer ->
+
+            writer.addImport(RuntimeTypes.KtorServerCore.embeddedServer)
+            writer.addImport(RuntimeTypes.KtorServerNetty.Netty)
+            writer.dependencies.addAll(KotlinDependency.KTOR_LOGGING_BACKEND.dependencies)
+
+            writer.openBlock("public fun main(): Unit {")
```
lauzadis: style: use `withBlock` to make this more readable:
```
writer.withBlock("public fun main(): Unit {", "}") {
    withBlock("embeddedServer(Netty, port = 8080) {", "}.start(wait = true)") {
        write(...) // write here
    }
}
// no closeBlock needed
```

lauzadis (approved):
This looks like a great start! 

In general I don't know how closely we want the server codegen to be tied to our existing client codegen (`CodegenVisitor.kt`, `GradleGenerator.kt`, `RuntimeTypes.kt`). I think reusing it is fine for now, but we should be ready to separate them out if there becomes a conflict. Curious what others think

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/KotlinCodegenPlugin.kt
```diff
@@ -17,7 +17,15 @@ import software.amazon.smithy.model.Model
 class KotlinCodegenPlugin : SmithyBuildPlugin {
     override fun getName(): String = "kotlin-codegen"
 
-    override fun execute(context: PluginContext?) = CodegenVisitor(context ?: error("context was null")).execute()
+    override fun execute(context: PluginContext?) {
+        val buildType = context?.settings?.getStringMemberOrDefault("type", "client")
+        println("Kotlin codegen plugin triggered with build type: $buildType")
```
ianbotsf: **Nit**: We should use a `Logger` here instead of `println`. Loggers offer users more choices for how output is filtered (e.g., only show warnings/errors but not info/debug) and where logs go (e.g., the console, a file, a logging service like CloudWatch, etc.).

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/RuntimeTypes.kt
```diff
@@ -484,4 +484,22 @@ object RuntimeTypes {
...
+        val embeddedServer = symbol("embeddedServer", "engine")
+        val application = symbol("Application", "application")
+        val install = symbol("install", "application")
+    }
+    object KtorServerNetty : RuntimeTypePackage(KotlinDependency.KTOR_SERVER_NETTY) {
+        val Netty = symbol("Netty")
+    }
+    object KtorServerContentNegotiation : RuntimeTypePackage(KotlinDependency.KTOR_SERVER_CONTENT_NEGOTIATION) {
+        val all = symbol("*")
+    }
```
ianbotsf: **Style**: Missing blank lines between object definitions:

```kotlin
object KtorServerCore : RuntimeTypePackage(KotlinDependency.KTOR_SERVER_CORE) {
    val routingAll = symbol("*", "routing")
    val responseAll = symbol("*", "response")
    val embeddedServer = symbol("embeddedServer", "engine")
    val application = symbol("Application", "application")
    val install = symbol("install", "application")
}

object KtorServerNetty : RuntimeTypePackage(KotlinDependency.KTOR_SERVER_NETTY) {
    val Netty = symbol("Netty")
}

object KtorServerContentNegotiation : RuntimeTypePackage(KotlinDependency.KTOR_SERVER_CONTENT_NEGOTIATION) {
    val all = symbol("*")
}
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/RuntimeTypes.kt
```diff
@@ -484,4 +484,22 @@ object RuntimeTypes {
 
         val sign = symbol("sign")
     }
+
+    object KtorServerCore : RuntimeTypePackage(KotlinDependency.KTOR_SERVER_CORE) {
+        val routingAll = symbol("*", "routing")
+        val responseAll = symbol("*", "response")
```
ianbotsf: **Nit**: We typically avoid importing `*` because it can cause unexpected naming collisions later. Do we know which declarations _specifically_ we need from `routing` and `response`? Can we import those by name instead?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinDependency.kt
```diff
@@ -38,6 +38,9 @@ private fun getDefaultRuntimeVersion(): String {
 const val RUNTIME_GROUP: String = "aws.smithy.kotlin"
 val RUNTIME_VERSION: String = System.getProperty("smithy.kotlin.codegen.clientRuntimeVersion", getDefaultRuntimeVersion())
 val KOTLIN_COMPILER_VERSION: String = System.getProperty("smithy.kotlin.codegen.kotlinCompilerVersion", "2.1.0")
+val KTOR_VERSION: String = System.getProperty("smithy.kotlin.codegen.ktorVersion", "3.1.3")
+val COROUTINES_VERSION: String = System.getProperty("smithy.kotlin.codegen.coroutinesVersion", "1.10.2")
+val KTOR_LOGGING_BACKEND_VERSION: String = System.getProperty("smithy.kotlin.codegen.ktorLoggingBackendVersion", "1.4.14")
```
ianbotsf, line 43: **Nit**: We _shouldn't_ be hardcoding version numbers into our codegen because we _should_ be setting them dynamically based on the Gradle library versions. Unfortunately, we don't have an easy way to do this right now so the versions might have to stay hardcoded for a while. So that we don't forget about them, please add a `// FIXME` comment before these lines with a short description of what needs to be fixed.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
```diff
@@ -160,6 +160,49 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Unit>() {
...
+            writeGradleBuild(settings, fileManifest, dependencies)
+        }
+
+        // write files defined by integrations
+        integrations.forEach { it.writeAdditionalFiles(baseGenerationContext, writers) }
+        logger.info("Generating endpoint tests done")
+
+        writers.flushWriters()
+    }
+
```
ianbotsf: **Style**: A lot of this appears to be copy/pasted from `execute()`. Rather than duplicating so much code, I wonder if it might be more efficient to simply have _one_ `execute()` method that handles both clients _and_ servers. The things that work the same both ways can be left as-is. The things that work differently can be wrapped in `if`/`else` logic.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/KotlinCodegenPlugin.kt
```diff
@@ -17,7 +17,15 @@ import software.amazon.smithy.model.Model
...
+        if (buildType == "service") {
+            CodegenVisitor(context ?: error("context was null")).serviceExecute()
+        } else {
+            CodegenVisitor(context ?: error("context was null")).execute()
+        }
+    }
```
ianbotsf: **Style**: Getting the `type` is probably better done inside `KotlinSettings` where we parse all of the other settings from JSON. Additionally, `type` should be an `enum class` rather than a `String`.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/KotlinSettings.kt
```diff
@@ -196,25 +196,28 @@ data class BuildSettings(
     val generateDefaultBuildFiles: Boolean = true,
     val optInAnnotations: List<String>? = null,
     val generateMultiplatformProject: Boolean = false,
+    val enableApplications: Boolean = false,
 ) {
```
ianbotsf: **Style**: When we add new parameters to public declarations, we should generally update the [**KDocs**](https://kotlinlang.org/docs/kotlin-doc.html) too. Please add a new `@param enableApplications` item to the docs above.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/GradleGenerator.kt
```diff
@@ -53,6 +54,9 @@ fun writeGradleBuild(
                 }
             },
         )
+        if (enableApplications) {
+            write("application")
+        }
```
ianbotsf: **Question**: Why is this parameter called `enableApplications` (plural) if it only adds a single `application` call?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/ServiceStubGenerator.kt
```diff
@@ -0,0 +1,82 @@
...
+            writer.addImport(RuntimeTypes.KtorServerCore.responseAll)
+            writer.addImport(RuntimeTypes.KtorServerCore.routingAll)
+
+            writer.openBlock("internal fun Application.configureRouting(): Unit {")
+                .openBlock("routing {")
+                .openBlock("get(\"/\") {")
+                .write("call.respondText(\"hello world\")")
+                .closeBlock("}")
+                .closeBlock("}")
+                .closeBlock("}")
```
ianbotsf: **Style**: Escaped quotes in strings can be difficult to read. Favor using Smithy's formatters to improve the legibility of this code:

```kotlin
.openBlock("get(#S) {", "")
.write("call.respondText(#S)", "hello world")
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/KotlinCodegenPlugin.kt
```diff
@@ -17,7 +17,9 @@ import software.amazon.smithy.model.Model
 class KotlinCodegenPlugin : SmithyBuildPlugin {
     override fun getName(): String = "kotlin-codegen"
 
-    override fun execute(context: PluginContext?) = CodegenVisitor(context ?: error("context was null")).execute()
+    override fun execute(context: PluginContext?) {
+        CodegenVisitor(context ?: error("context was null")).execute()
+    }
```
ianbotsf: **Nit**: This doesn't need to change anymore and could be restored to its original code.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
```diff
@@ -114,13 +114,20 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Unit>() {
...
+            logger.info("Generating Kotlin client for service ${settings.service}")
+        }
 
         logger.info("Walking shapes from ${settings.service} to find shapes to generate")
         val modelWithoutTraits = ModelTransformer.create().getModelWithoutTraitShapes(model)
         val serviceShapes = Walker(modelWithoutTraits).walkShapes(service)
         serviceShapes.forEach { it.accept(this) }
 
+        val serviceStubGenerator = ServiceStubGenerator(settings, writers)
+
```
ianbotsf: **Nit**: We only need this `serviceStubGenerator` if we're building a service project so move the declaration/initialization down into the `if` block below:

```kotlin
if (generateServiceProject) {
    val serviceStubGenerator = ServiceStubGenerator(settings, writers)
    serviceStubGenerator.render()
}
```

PR #1314, opened 2025-07-02
Service sdk cbor serde

luigi617:
- Service supports now CBOR serde.
- Change `HttpBindingProtocol` to support service serde generation.
- Add hardcoded routing for testing.
- Modify `requestBindings` in `HttpBindingResolver` so that it accepts more general input type.

## Description of changes

- Now the service supports CBOR.


codegen/smithy-aws-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/aws/protocols/RpcV2Cbor.kt
```diff
@@ -103,12 +103,16 @@ class RpcV2Cbor : AwsHttpBindingProtocolGenerator() {
         if (!op.hasHttpBody(ctx)) return
 
         // payload member(s)
-        val requestBindings = resolver.requestBindings(op)
-        val httpPayload = requestBindings.firstOrNull { it.location == HttpBinding.Location.PAYLOAD }
+        val binding = if (ctx.settings.build.generateServiceProject) {
```
ianbotsf: **Nit**: `binding` → `bindings`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
```diff
@@ -143,14 +143,17 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Unit>() {
             logger.info("[${service.id}] Generating service client for protocol $protocol")
             generateProtocolClient(ctx)
 
-            logger.info("[${service.id}] Generating endpoint provider for protocol $protocol")
-            generateEndpointsSources(ctx)
+            if (!ctx.settings.build.generateServiceProject) {
```
ianbotsf: **Nit**: `ctx.settings.build.generateServiceProject` is already available locally as `generateServiceProject`.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
```diff
@@ -143,14 +143,17 @@ class CodegenVisitor(context: PluginContext) : ShapeVisitor.Default<Unit>() {
...
+                logger.info("[${service.id}] Generating auth scheme provider for protocol $protocol")
+                generateAuthSchemeProvider(ctx)
+            }
         }
+
         if (generateServiceProject) {
-            val serviceStubGenerator = ServiceStubGenerator(settings, writers)
+            val serviceStubGenerator = ServiceStubGenerator(settings, writers, serviceShapes)
             serviceStubGenerator.render()
         }
```
ianbotsf: **Nit**: Generally we try to avoid dealing with _all_ the shapes in a service, particularly when passing them to a generator (e.g., `ServiceStubGenerator`). It's necessary here in the `CodegenVisitor` because we need to actually _visit_ every shape. But in general, we try to pass limited info to the generators so that they contain more of their own filtering logic.

In this case I recommend passing a `ProtocolGenerator.GenerationContext` to the `ServiceStubGenerator` instead of `serviceShapes`. That way the generator can derive its own operations list:

```kotlin
class ServiceStubGenerator(
    private val ctx: ProtocolGenerator.GenerationContext,
    private val delegator: KotlinDelegator,
) {
    private val operations = TopDownIndex
        .of(ctx.model)
        .getContainedOperations(service)
        .sortedBy { it.defaultName() }

    ...
}
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt
```diff
@@ -332,6 +332,9 @@ class KotlinSymbolProvider(private val model: Model, private val settings: Kotli
     }
 
     override fun serviceShape(shape: ServiceShape): Symbol {
+        if (settings.build.generateServiceProject) {
+            return RuntimeTypes.Core.Content.Document.asNullable()
+        }
         val serviceName = clientName(settings.sdkId)
         return createSymbolBuilder(shape, "${serviceName}Client")
             .namespace(rootNamespace, ".")
```
ianbotsf: **Question**: What is the `Document` type used to represent in the server-side case?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/RuntimeTypes.kt
```diff
@@ -496,12 +495,23 @@ object RuntimeTypes {
...
 
     object KtorServerNetty : RuntimeTypePackage(KotlinDependency.KTOR_SERVER_NETTY) {
         val Netty = symbol("Netty")
     }
 
+    object KtorServerHTTP : RuntimeTypePackage(KotlinDependency.KTOR_SERVER_HTTP) {
+        val ContentType = symbol("ContentType")
+        val HttpStatusCode = symbol("HttpStatusCode")
+        val Cbor = symbol("Cbor", "ContentType.Application")
+    }
```
ianbotsf: **Nit**: `KtorServerHTTP` → `KtorServerHttp`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/ServiceStubGenerator.kt
```diff
@@ -65,15 +68,49 @@ class ServiceStubGenerator(
...
+            writer.addImport(RuntimeTypes.KtorServerRouting.requestRespondBytes)
+            writer.addImport(RuntimeTypes.KtorServerHTTP.ContentType)
+
+            serviceShapes.filter { it.type == ShapeType.OPERATION }
+                .forEach { shape ->
+                    writer.addImport("com.example.server.serde", "${shape.id.name}OperationDeserializer")
+                    writer.addImport("com.example.server.serde", "${shape.id.name}OperationSerializer")
+                    writer.addImport("com.example.server.model", "${shape.id.name}Request")
+                    writer.addImport("com.example.server.model", "${shape.id.name}Response")
+                }
```
ianbotsf: **Question**: Generally we should not use manual imports for codegen. We should use `#T` formatters and `Symbol` references to handle imports automatically. Will these be cleaned up in your routing PR? Or should these be addressed now?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/CodegenVisitor.kt
luigi617: I considered passing `ProtocolGenerator.GenerationContext`, but I want the service to be generated even when users don't specify any protocols, e.g. in the case of an empty service without operations. That's why `ServiceStubGenerator` is defined outside the scope where `generateProtocolClient` is called. Do you have any suggestions for handling this scenario?

ianbotsf: Oh that's a cool idea. OK, well what about passing just a `CodegenContext` instead? Really, anything we have in scope that contains the model, symbol provider, and settings should do the trick. Maybe `baseGenerationContext`?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt
luigi617: What do you mean `Document` Type? I think the document generated based on `serviceShape` is not really useful for the service. In my opinion, service shape defined in smithy file is used only for selecting which protocol to use (correct me if I'm wrong), and I will use this information when generating `Routing.kt`. The code here is used for not generating the file not used by the service.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/ServiceStubGenerator.kt
luigi617: Yes! I've already cleaned up all in my routing PR, which will be my next PR :)

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt
ianbotsf: I mean this line specifically:

```kotlin
return RuntimeTypes.Core.Content.Document.asNullable()
```

Why is `Document` the right type to return for a service shape in the `generateServiceProject` case?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinSymbolProvider.kt
```diff
@@ -332,6 +333,10 @@ class KotlinSymbolProvider(private val model: Model, private val settings: Kotli
     }
 
     override fun serviceShape(shape: ServiceShape): Symbol {
+        if (settings.build.generateServiceProject) {
+            // Intentionally not generating a class for the service
```
0marperez: Correctness: `"Intentionally not generating a *client symbol* for the service"`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/CborParserGenerator.kt
```diff
@@ -24,12 +24,17 @@ class CborParserGenerator(
         op: OperationShape,
         members: List<MemberShape>,
     ): Symbol {
-        val outputSymbol = ctx.symbolProvider.toSymbol(ctx.model.expectShape(op.outputShape))
+        val operationShape = if (ctx.settings.build.generateServiceProject) {
```
0marperez: Suggestion: `operationShape` -> `deserializedShape`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/CborParserGenerator.kt
```diff
@@ -91,8 +96,12 @@ class CborParserGenerator(
         writer: KotlinWriter,
     ) {
         writer.write("val deserializer = #T(payload)", RuntimeTypes.Serde.SerdeCbor.CborDeserializer)
-
-        val shape = ctx.model.expectShape(op.output.get())
+        val operationData = if (ctx.settings.build.generateServiceProject) {
```
0marperez: Suggestion: `operationData` -> `deserializationTarget`, same goes for other uses if you decide to change it.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/CborSerializerGenerator.kt
```diff
@@ -20,8 +20,13 @@ class CborSerializerGenerator(
     private val protocolGenerator: ProtocolGenerator,
 ) : StructuredDataSerializerGenerator {
     override fun operationSerializer(ctx: ProtocolGenerator.GenerationContext, op: OperationShape, members: List<MemberShape>): Symbol {
-        val input = op.input.get().let { ctx.model.expectShape(it) }
-        val symbol = ctx.symbolProvider.toSymbol(input)
+        val operationData = if (ctx.settings.build.generateServiceProject) {
```
0marperez: Suggestion: `operationData` -> `serializationTarget`, same goes for other uses if you decide to change it.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt
```diff
@@ -693,11 +745,68 @@ abstract class HttpBindingProtocolGenerator : ProtocolGenerator {
...
+            }
+            // Render client side error correction for `@required` members.
+            // NOTE: nested members bound via the document/payload will be handled by the deserializer for the relevant
+            // content type. All other members (e.g. bound via REST semantics) will be corrected here.
+            .write("builder.correctErrors()")
+            .write("return builder.build()")
+            .closeBlock("}")
+    }
+
+    private fun renderServiceHttpDeserialize(
```
0marperez, line 765: Question: We didn't need a `renderServiceHttpSerialize`?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt
```diff
@@ -693,11 +745,68 @@ abstract class HttpBindingProtocolGenerator : ProtocolGenerator {
...
+                bindings.firstOrNull { it.location == HttpBinding.Location.PREFIX_HEADERS }
+                    ?.let {
+                        renderDeserializePrefixHeaders(ctx, it, writer)
+                    }
+            }
+            .write("")
+            .call {
+                // FIXME: event stream is not in the scope of service generation for now.
+                if (!ctx.settings.build.generateServiceProject && op != null && op.isOutputEventStream(ctx.model)) {
+                    deserializeViaEventStream(ctx, op, writer)
```
0marperez, line 803: Suggestion: I would just add a [TODO](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-t-o-d-o.html) with a reason here unless it interferes with your testing

Nit: `!ctx.settings.build.generateServiceProject` isn't necessary since this is only called if `generateServiceProject` is true.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt
luigi617: Deserialization inherits from a class that isn't suitable for the service, whereas I can use serialization directly to get the response. I believe the issue was due to a type inconsistency

ianbotsf (approved):

PR #1302, opened 2025-06-20
fix: always serialize defaults

ianbotsf:
## Issue \#

Resolves https://github.com/awslabs/aws-sdk-kotlin/issues/1608

## Description of changes

This change switches the default serialization mode from `WHEN_DIFFERENT` (meaning only if the value has been changed from the default) to `ALWAYS` (meaning...well, always). This is the current [recommendation from Smithy](https://smithy.io/2.0/spec/type-refinement-traits.html#default-value-serialization) and will offer more deterministic communication with services.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1315, opened 2025-07-07
fix: always serialize defaults

ianbotsf:
## Issue \#

Resolves https://github.com/awslabs/aws-sdk-kotlin/issues/1608

## Description of changes

This change is identical to #1302 but has been rebased onto **v1.5-main** since it is a behavioral breaking change.

This change switches the default serialization mode from `WHEN_DIFFERENT` (meaning only if the value has been changed from the default) to `ALWAYS` (meaning...well, always). This is the current [recommendation from Smithy](https://smithy.io/2.0/spec/type-refinement-traits.html#default-value-serialization) and will offer more deterministic communication with services.

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpStringValuesMapSerializerTest.kt
+++ b/codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpStringValuesMapSerializerTest.kt
@@ -53,7 +53,8 @@ class HttpStringValuesMapSerializerTest {
 
     @Test
     fun `it handles primitive header shapes when different mode`() {
-        val contents = getTestContents(defaultModel, "com.test#PrimitiveShapesOperation", HttpBinding.Location.HEADER)
+        val settings = defaultModel.defaultSettings(defaultValueSerializationMode = DefaultValueSerializationMode.WHEN_DIFFERENT)
+        val contents = getTestContents(defaultModel, "com.test#PrimitiveShapesOperation", HttpBinding.Location.HEADER, settings)
         contents.assertBalancedBracesAndParens()
 
         val expectedContents = """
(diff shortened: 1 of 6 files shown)


codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpStringValuesMapSerializerTest.kt
```diff
@@ -53,7 +53,8 @@ class HttpStringValuesMapSerializerTest {
 
     @Test
     fun `it handles primitive header shapes when different mode`() {
-        val contents = getTestContents(defaultModel, "com.test#PrimitiveShapesOperation", HttpBinding.Location.HEADER)
+        val settings = defaultModel.defaultSettings(defaultValueSerializationMode = DefaultValueSerializationMode.WHEN_DIFFERENT)
+        val contents = getTestContents(defaultModel, "com.test#PrimitiveShapesOperation", HttpBinding.Location.HEADER, settings)
```
lauzadis, line 57: I still think we should have tests for the new behavior

https://github.com/smithy-lang/smithy-kotlin/pull/1302#discussion_r2159672896

ianbotsf: We already had/have tests which explicitly set `ALWAYS`. My changes in this PR make all of the implicit `WHEN_DIFFERENT` values into explicit ones. Are there specific tests you'd like to see added/expanded?

lauzadis: Oh right, I totally missed the names of the tests here. These are perfectly fine

lauzadis (approved):

PR #1319, opened 2025-07-10
Service sdk error

luigi617:
- Error handler
- Support Malformed CBOR input/output (during serialization and deserialization) error
- Support content type mismatch error

## Description of changes


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/ServiceStubGenerator.kt
```diff
@@ -192,6 +192,7 @@ class ServiceStubGenerator(
 
     private fun renderKTORServiceFramework(writer: KotlinWriter) {
         writer.addImport(RuntimeTypes.KtorServerNetty.Netty)
+        writer.addImport("${ctx.settings.pkg.name}.plugins", "configureErrorHandling")
         writer.addImport(ctx.settings.pkg.name, "configureRouting")
         writer.addImport("${ctx.settings.pkg.name}.utils", "configureLogging")
         writer.addImport("${ctx.settings.pkg.name}.config", "ServiceFrameworkConfig")
```
ianbotsf: **Question**: Why are there still so many manual imports in this codegen? I thought I understood [these were being removed in a previous PR](https://github.com/smithy-lang/smithy-kotlin/pull/1314#discussion_r2190642646).

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/ServiceStubGenerator.kt
ianbotsf, line 1: **Style**: In general it feels like this generator has gotten too big. We should look for ways to split apart the functionality by delegating to new subgenerators and moving code out of codegen to runtime wherever possible,

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/ServiceStubGenerator.kt
luigi617: These are imports of class/functions that are generated. Is there a better way to import them?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/ServiceStubGenerator.kt
luigi617: Sure I will handle it

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/ServiceStubGenerator.kt
ianbotsf: Generally, we'd declare a symbol to represent the declaration:

```kotlin
val configureErrorHandling = buildSymbol {
    name = "configureErrorHandling"
    namespace = "${ctx.settings.pkg.name}.plugins"
}
```

And then reference that symbol when we need to implement the declaration:

```kotlin
writer.withBlock(
    "internal fun #T.#L() {",
    "}",
    RuntimeTypes.KtorServerCore.Application,
    configureErrorHandling,
) {
    // ...
}
```

As well as where we invoke the declaration:

```kotlin
writer.write("#T()", configureErrorHandling)
```

Doing this allows us to avoid unnecessary imports and maintain cleaner codegen.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/ServiceStubGenerator.kt
luigi617: I shortened a bit the generator. I will try to delegate some functionality to new subgenerators in next PR

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/ServiceStubConfigurations.kt
```diff
@@ -1,5 +1,38 @@
...
+    JSON("JSON"),
+    ;
+
+    override fun toString(): String = value
+
+    companion object {
+        fun fromValue(value: String): ContentType = when (value.uppercase()) {
+            "CBOR" -> CBOR
+            "JSON" -> JSON
+            else -> throw IllegalArgumentException("$value is not a valid ContentType value, expected $JSON or $CBOR")
```
ianbotsf: **Style**: Don't replicate each enum value in the error message since that'll likely get out of date when new entries are added. Use `ContentType.entries` instead. Additionally, there's no reason to duplicate the enum values (e.g., `"CBOR"`) when you could just iterate over them instead:

```kotlin
fun fromValue(value: String): ContentType = ContentType
    .entries
    .firstOrNull { it.name.equals(value, ignoreCase = true) }
    ?: throw IllegalArgumentException("$value is not a validContentType value, expected one of ${ContentType.entries}")
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/ServiceStubConfigurations.kt
```diff
@@ -1,5 +1,38 @@
...
+        }
+
+        fun fromServiceShape(shape: ServiceShape): ContentType {
+            if (shape.getTrait<Rpcv2CborTrait>() != null) {
+                return CBOR
+            } else if (shape.getTrait<RestJson1Trait>() != null) {
+                return JSON
+            }
+            throw IllegalArgumentException("service shape does not a valid protocol")
+        }
```
ianbotsf: **Style**: This is a more natural fit for a `when` block (with no subject). Also, the extension function [`hasTrait<T>`](https://github.com/smithy-lang/smithy-kotlin/blob/444ff8a71ad8a6658264c44796eda736aa05405c/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/ShapeExt.kt#L70-L73) is simpler than comparing `getTrait<T>` to `null`:

```kotlin
fun fromServiceShape(shape: ServiceShape): ContentType = when {
    shape.hasTrait<Rpcv2CborTrait>() -> CBOR
    shape.hasTrait<RestJson1Trait>() -> JSON
    else -> throw IllegalArgumentException("service shape does not a valid protocol")
}
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/ServiceTypes.kt
```diff
@@ -0,0 +1,55 @@
...
+        namespace = "$pkgName.config"
+    }
+
+    val serviceFrameworkConfig = buildSymbol {
+        name = "ServiceFrameworkConfig"
+        namespace = "$pkgName.config"
+    }
+
+    val ktorServiceFramework = buildSymbol {
+        name = "KTORServiceFramework"
```
ianbotsf: **Nit**: `KTOR` → `Ktor`

ianbotsf (approved):
Looking pretty good, only minor issues left from me.

PR #1330, opened 2025-07-16
Service sdk testing

luigi617:
- Serializer will serialize only the data and not create a HttpRequest.
- Change JAVA version to 17
- Enable CIO, Jetty-jakarta engine (and Netty which was already supported)
- Accept type checker
- Creation of service generator tests
- Unit tests will be refined in the next PR
## Description of changes


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinDependency.kt
```diff
@@ -38,7 +38,7 @@ private fun getDefaultRuntimeVersion(): String {
 const val RUNTIME_GROUP: String = "aws.smithy.kotlin"
 val RUNTIME_VERSION: String = System.getProperty("smithy.kotlin.codegen.clientRuntimeVersion", getDefaultRuntimeVersion())
 val KOTLIN_COMPILER_VERSION: String = System.getProperty("smithy.kotlin.codegen.kotlinCompilerVersion", "2.2.0")
-val KTOR_VERSION: String = System.getProperty("smithy.kotlin.codegen.ktorVersion", "3.1.3")
+val KTOR_VERSION: String = System.getProperty("smithy.kotlin.codegen.ktorVersion", "3.2.2")
```
ianbotsf, line 41: **Correctness**: This doesn't match the version specified in **gradle/libs.versions.toml**:

```toml
ktor-version = "3.1.1"
```

We should unify these versions so they're the same.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/RuntimeTypes.kt
```diff
@@ -582,4 +602,8 @@ object RuntimeTypes {
     object KotlinxJsonSerde : RuntimeTypePackage(KotlinDependency.KOTLINX_JSON_SERDE) {
         val Json = symbol("Json")
     }
+
+    object Duration {
+        val seconds = "kotlin.time.Duration.Companion.seconds".toSymbol()
+    }
```
ianbotsf: **Nit**: We already have [`KotlinTypes.Time`](https://github.com/smithy-lang/smithy-kotlin/blob/b5d177cb678f771c2f50c16fa1695e36fb38fe10/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/lang/KotlinTypes.kt#L116-L119) for symbol declarations like this. We should move `seconds` there.

codegen/smithy-kotlin-codegen-testutils/build.gradle.kts
0marperez, line 1: Question: Why were these changes made? I think we need to have compatibility set to 1.8

See: https://github.com/smithy-lang/smithy-kotlin/pull/1297

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt
```diff
@@ -182,27 +182,47 @@ abstract class HttpBindingProtocolGenerator : ProtocolGenerator {
...
-                        .write("val builder = #T()", RuntimeTypes.Http.Request.HttpRequestBuilder)
-                        .call {
-                            renderHttpSerialize(ctx, op, writer)
-                        }
-                        .write("return builder")
-                        .closeBlock("}")
-                }
-                .closeBlock("}")
+            // FIXME: this works only for Cbor protocol now
+            if (ctx.settings.build.generateServiceProject) {
```
0marperez, line 189: Suggestion: You can add a check to throw an exception if CBOR isn't used

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/serde/CborSerializerGenerator.kt
```diff
@@ -27,12 +27,22 @@ class CborSerializerGenerator(
...
+        } else {
+            return op.bodySerializer(ctx.settings) { writer ->
+                addNestedDocumentSerializers(ctx, op, writer)
+                writer.withBlock("private fun #L(context: #T, input: #T): #T {", "}", op.bodySerializerName(), RuntimeTypes.Core.ExecutionContext, serializationSymbol, RuntimeTypes.Http.HttpBody) {
+                    call {
+                        renderSerializeOperationBody(ctx, op, members, writer)
+                    }
                 }
             }
         }
```
ianbotsf: **Nit**: It appears these two cases only differ by a single symbol reference and they could be combined into a smaller block of code:

```kotlin
val serializerResultSymbol = when {
    ctx.settings.build.generateServiceProject -> KotlinTypes.ByteArray
    else -> RuntimeTypes.Http.HttpBody
}

return op.bodySerializer(ctx.settings) { writer ->
    addNestedDocumentSerializers(ctx, op, writer)
    writer.withBlock("private fun #L(context: #T, input: #T): #T {", "}", op.bodySerializerName(), RuntimeTypes.Core.ExecutionContext, serializationSymbol, serializerResultSymbol) {
        call {
            renderSerializeOperationBody(ctx, op, members, writer)
        }
    }
}
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/KtorStubGenerator.kt
```diff
@@ -32,25 +32,53 @@ internal class KtorStubGenerator(
...
+            write("#T()", ServiceTypes(pkgName).configureAuthentication)
+            write("#T()", ServiceTypes(pkgName).configureRouting)
+            write("#T()", ServiceTypes(pkgName).configureLogging)
+        }
+            .write("")
+        writer.withBlock("internal class KtorServiceFramework() : ServiceFramework {", "}") {
             write("private var engine: #T<*, *>? = null", RuntimeTypes.KtorServerCore.EmbeddedServerType)
+            write("")
+            write("@Suppress(#S)", "UNCHECKED_CAST")
+            write("private val engineFactory = #T.engine.toEngineFactory()", ServiceTypes(pkgName).serviceFrameworkConfig)
```
ianbotsf: **Question**: Why is `@Suppress("UNCHECKED_CAST")` necessary? This statement doesn't appear to perform any casting (i.e., no `as` keyword).

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/KtorStubGenerator.kt
```diff
@@ -302,11 +339,7 @@ internal class KtorStubGenerator(
                 RuntimeTypes.KtorServerCore.applicationCall,
                 RuntimeTypes.KtorServerRouting.responseRespond,
             ) {
-                write(
-                    "bytes = response.body.#T()?.decodeToString() ?:  #S,",
-                    RuntimeTypes.Http.readAll,
-                    "{}",
-                )
+                write("bytes = response.decodeToString(),")
```
0marperez: Question: Why are we setting the bytes parameter to a string now?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/KtorStubGenerator.kt
```diff
@@ -464,6 +505,66 @@ internal class KtorStubGenerator(
...
+                    withBlock("onCall { call ->", "}") {
+                        write("val accepted = call.request.acceptedContentTypes()")
+                        write("if (accepted.isEmpty()) return@onCall")
+                        write("")
+                        write("val isOk = accepted.any { candidate -> allowed.any { candidate.match(it) } }")
+
+                        withBlock("if (!isOk) {", "}") {
+                            withBlock("throw #T(", ")", ServiceTypes(pkgName).errorEnvelope) {
+                                write("#T.NotAcceptable.value, ", RuntimeTypes.KtorServerHttp.HttpStatusCode)
+                                write("#S", "Supported Accept(s): \${allowed.joinToString()}")
```
0marperez: Nit: `"Not acceptable content type found: '${accepted}'. Accepted content types: \${allowed.joinToString()}"`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/KtorStubGenerator.kt
```diff
@@ -464,6 +505,66 @@ internal class KtorStubGenerator(
...
+                    RuntimeTypes.KtorServerHttp.parseAndSortHeader,
+                    RuntimeTypes.KtorServerHttp.ContentType,
+                )
+            }
+
+            writer.withBlock("public class AcceptTypeGuardConfig {", "}") {
+                write("public var allow: List<#T> = emptyList()", RuntimeTypes.KtorServerHttp.ContentType)
+                write("")
+                withBlock("public fun json(): Unit {", "}") {
+                    write("allow = listOf(#T)", RuntimeTypes.KtorServerHttp.Json)
```
0marperez, line 529: Question: Should we be overriding or appending to the list?

tests/codegen/service-codegen-tests/src/main/kotlin/com/test/DefaultServiceGeneratorTest.kt
```diff
@@ -0,0 +1,108 @@
...
+import software.amazon.smithy.build.PluginContext
+import software.amazon.smithy.kotlin.codegen.KotlinCodegenPlugin
+import software.amazon.smithy.model.loader.ModelAssembler
+import software.amazon.smithy.model.node.Node
+import software.amazon.smithy.model.node.ObjectNode
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
+
+fun main() {
```
ianbotsf: **Question**: Our test projects don't normally need `main()` methods. Why couldn't this work be done in a `@Before` method or even manually written and checked into the repository?

tests/codegen/service-codegen-tests/src/test/kotlin/com/test/ServiceGeneratorTest.kt
```diff
@@ -0,0 +1,551 @@
...
+        assertTrue(projectDir.resolve("src/main/kotlin/$packagePath/utils/Logging.kt").exists())
+        assertTrue(projectDir.resolve("src/main/kotlin/$packagePath/auth/Authentication.kt").exists())
+        assertTrue(projectDir.resolve("src/main/kotlin/$packagePath/auth/Validation.kt").exists())
+
+        assertTrue(projectDir.resolve("src/main/kotlin/$packagePath/model/PostTestRequest.kt").exists())
+        assertTrue(projectDir.resolve("src/main/kotlin/$packagePath/model/PostTestResponse.kt").exists())
+        assertTrue(projectDir.resolve("src/main/kotlin/$packagePath/serde/PostTestOperationSerializer.kt").exists())
+        assertTrue(projectDir.resolve("src/main/kotlin/$packagePath/serde/PostTestOperationDeserializer.kt").exists())
+        assertTrue(projectDir.resolve("src/main/kotlin/$packagePath/operations/PostTestOperation.kt").exists())
+    }
```
ianbotsf: **Style**: These are the sort of assertions which would exist in codegen unit tests, not functional tests like these. I think they should be removed from this file because the rest of this class is not actually testing that the correct files were written in the correct places—it's testing that a generated server works correctly.

Consider adding codegen unit tests where checks like this might be better suited, although note that our codegen unit tests typically test a single file each (e.g., `MainTest`, `RoutingTest`, `ServiceFrameworkConfigTest`, etc.).

tests/codegen/service-codegen-tests/model/service-generator-test.smithy
```diff
@@ -0,0 +1,70 @@
+$version: "1.0"
```
0marperez: Suggestion: Are you able to use `2.0`? It's preferable since it's a newer version of smithy lang

tests/codegen/service-codegen-tests/src/test/kotlin/com/test/ServiceGeneratorTest.kt
```diff
@@ -0,0 +1,551 @@
...
+            PostTestResponse.serializer(),
+            response.body(),
+        )
+
+        assertEquals("Hello world!", body.output1)
+        assertEquals(input2 + 1, body.output2)
+    }
+
+    @Test
+    @OptIn(ExperimentalPathApi::class, ExperimentalSerializationApi::class)
```
0marperez: Suggestion: I think you should be able to opt in to `ExperimentalPathApi` and `ExperimentalSerializationApi` at the project level using `build.gradle.kts` instead of adding this to each test

tests/codegen/service-codegen-tests/src/test/kotlin/com/test/ServiceGeneratorTest.kt
```diff
@@ -0,0 +1,551 @@
...
+        )
+        assertIs<HttpResponse<ByteArray>>(response)
+        assertEquals(405, response.statusCode(), "Expected 405")
+    }
+
+    @Test
+    @OptIn(ExperimentalPathApi::class, ExperimentalSerializationApi::class)
+    fun `checks request body limit`() {
+        val cbor = Cbor { }
+        val overLimitPayload = "x".repeat(10 * 1024 * 1024 + 1)
```
0marperez: Nit: `10 * 1024 * 1024` could be a constant or val

0marperez (changes requested):
Overall looks good! Just had some small suggestions, nit's and questions but the Java 17 changes would cause us to regress on the behavior introduced by https://github.com/smithy-lang/smithy-kotlin/pull/1297. That's the only blocking thing for me.

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/KtorStubGenerator.kt
luigi617: There was a issue as engine factories have different types and I was getting some errors, but I tested and I don't actually need it.

tests/codegen/service-codegen-tests/src/main/kotlin/com/test/DefaultServiceGeneratorTest.kt
luigi617: This function is used to generate the service. In `gradle.build.kts`, you can see that the test depends on `runServiceGenerator` which will call this funciton. The reason why not having this in `@Before` method is that I want to separate the file for generating service and testing it. I don't understand what you meant by "manually written and checked into the repository".

tests/codegen/service-codegen-tests/src/test/kotlin/com/test/ServiceGeneratorTest.kt
luigi617: Are you suggesting to have these assert in another test file?

For the second part, I will divide the test into different files, maybe in the next PR. I did not want to start the service for each part of the code as every time it starts a service will make the whole test slower.

codegen/smithy-kotlin-codegen-testutils/build.gradle.kts
luigi617: It was needed before for my unit tests, but since I moved the tests outside, I'll change them back

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/KtorStubGenerator.kt
luigi617: It is for JSON, I fix it anyways

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/KtorStubGenerator.kt
luigi617: Since each service accept only one protocol, I think overriding should be ok

0marperez (approved):

PR #1338, opened 2025-07-22
Service sdk constraints

luigi617:
- Add constraints: required, length, range, uniqueItems, pattern.
- Add unit tests for each constraints.
- Minor fix.

## Description of changes


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/contraints/AbstractConstraintTrait.kt
```diff
@@ -0,0 +1,5 @@
+package software.amazon.smithy.kotlin.codegen.service.contraints
+
+internal abstract class AbstractConstraintTrait {
```
0marperez: `AbstractConstraintTrait` can be interpreted as referring to an actual trait instead of the generator/renderer. The name `AbstractConstraintTraitGenerator` avoids confusion and is more accurate

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/contraints/LengthConstraint.kt
```diff
@@ -0,0 +1,25 @@
...
+import software.amazon.smithy.kotlin.codegen.core.KotlinWriter
+import software.amazon.smithy.kotlin.codegen.core.withBlock
+import software.amazon.smithy.kotlin.codegen.service.ServiceTypes
+import software.amazon.smithy.model.traits.LengthTrait
+
+internal class LengthConstraint(val memberPrefix: String, val memberName: String, val trait: LengthTrait, val pkgName: String, val writer: KotlinWriter) : AbstractConstraintTrait() {
+    override fun render() {
+        val min = trait.min.orElse(null)
+        val max = trait.max.orElse(null)
+        writer.write("if (${memberPrefix}$memberName == null) { return }")
```
0marperez: Assign `${memberPrefix}$memberName` to a `val` so you don't have to re-write it so much.
```kt
val member = memberPrefix + memberName
writer.write("if ($member == null) { return }")
```

The same goes for every other `AbstractConstraintTrait`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/contraints/LengthConstraint.kt
```diff
@@ -0,0 +1,25 @@
...
+        writer.withBlock("require(${memberPrefix}$memberName is List<*> || ${memberPrefix}$memberName is Map<*, *> || ${memberPrefix}$memberName is String || ${memberPrefix}$memberName is java.sql.Blob) {", "}") {
```
0marperez: Blobs are represented as [ByteArray's](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-byte-array/) not `java.sql.Blob`'s

Check this out: https://github.com/smithy-lang/smithy-kotlin/blob/ee5f4269378c1b016df2ea9afc24f855d099d08e/docs/design/kotlin-smithy-sdk.md?plain=1#L32

0marperez: [Smithy IDL](https://github.com/smithy-lang/smithy) is supposed to be checking if the trait is being applied correctly. I'm not sure if you verified this already but we should be able to remove this type of check if it's not necessary

The same applies to other `AbstractConstraintTrait`'s

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/contraints/ConstraintUtilsGenerator.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    private fun renderLengthTraitUtils(writer: KotlinWriter) {
+        writer.withBlock("internal fun sizeOf(value: Any?): Long = when (value) {", "}") {
+            write("is Collection<*> -> value.size.toLong()")
+            write("is List<*> -> value.size.toLong()")
+            write("is Array<*> -> value.size.toLong()")
+            write("is Map<*, *> -> value.size.toLong()")
+            write("is String -> value.codePointCount(0, value.length).toLong()")
+            write("is ByteArray -> value.size.toLong()")
+            write("is java.sql.Blob -> value.length()")
```
0marperez: Same comment about using `ByteArray`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/contraints/ConstraintUtilsGenerator.kt
```diff
@@ -0,0 +1,76 @@
...
+                .call { renderUniqueItemsTraitUtils(writer) }
+        }
+    }
+
+    private fun renderLengthTraitUtils(writer: KotlinWriter) {
+        writer.withBlock("internal fun sizeOf(value: Any?): Long = when (value) {", "}") {
+            write("is Collection<*> -> value.size.toLong()")
+            write("is List<*> -> value.size.toLong()")
+            write("is Array<*> -> value.size.toLong()")
+            write("is Map<*, *> -> value.size.toLong()")
```
0marperez: You can combine conditions into a single result like [this](https://github.com/smithy-lang/smithy-kotlin/blob/ee5f4269378c1b016df2ea9afc24f855d099d08e/docs/design/event-streams.md?plain=1#L270-L274) to avoid duplicate code

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/contraints/ConstraintUtilsGenerator.kt
```diff
@@ -0,0 +1,76 @@
...
+
+internal class ConstraintUtilsGenerator(
+    val ctx: GenerationContext,
+    val delegator: KotlinDelegator,
+) {
+    val pkgName = ctx.settings.pkg.name
+
+    fun render() {
+        delegator.useFileWriter("utils.kt", "$pkgName.constraints") { writer ->
+            writer.call { renderLengthTraitUtils(writer) }
```
0marperez: Can you call `function(writer)` and achieve the same results as `writer.call { function(writer) }`? It's easier to follow

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/contraints/LengthConstraint.kt
0marperez: You can move this check to line 37-38 of the `ConstraintGenerator` so it's not repeated on every `AbstractConstraintTrait` except for the required trait.
```kt
writer.write("if ($memberPrefix$memberName == null) { return }")
```

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/KtorStubGenerator.kt
```diff
@@ -271,7 +277,17 @@ internal class KtorStubGenerator(
...
                                                 )
                                             }
+                                            write(
+                                                "try { check${shape.id.name}RequestConstraint(requestObj) } catch (ex: Exception) { throw #T(ex?.message ?: #S, ex) }",
+                                                RuntimeTypes.KtorServerCore.BadRequestException,
+                                                "Error while validating constraints",
+                                            )
                                             write("val responseObj = handle${shape.id.name}Request(requestObj)")
+                                            write(
+                                                "try { check${shape.id.name}ResponseConstraint(responseObj) } catch (ex: Exception) { throw #T(ex?.message ?: #S, ex) }",
```
0marperez: There shouldn't be any response constraints, we would be deviating from the [docs](https://smithy.io/2.0/spec/constraint-traits.html#uniqueitems-trait):

> Constraint traits SHOULD NOT be enforced when serializing shapes or when deserializing output. For example, **_when returning a response from a server to a client_**

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/contraints/ConstraintGenerator.kt
```diff
@@ -0,0 +1,74 @@
...
+    val outputMembers = outputShape.allMembers
+
+    val opName = operation.id.name
+    val pkgName = ctx.settings.pkg.name
+
+    fun render() {
+        renderRequestConstraintsValidation()
+        renderResponseConstraintsValidation()
+    }
+    private fun addConstaintValidation(prefix: String, memberShape: MemberShape, writer: KotlinWriter) {
```
0marperez: `addConstaintValidation` -> `generateConstaintValidations` just so we're clear that this is generating code

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/contraints/ConstraintGenerator.kt
```diff
@@ -0,0 +1,74 @@
...
+        val targetShape = ctx.model.expectShape(memberShape.target)
+
+        val memberName = memberShape.memberName
+        val memberTraits = memberShape.allTraits + targetShape.allTraits
+
+        for (memberTraitMap in memberTraits) {
```
0marperez: Things get a little confusing with naming here, `memberTraits` is actually member + target traits

0marperez: It seems like we're only using the `memberTraitMap`'s values. Can we iterate over them instead to simplify the code?

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/contraints/ConstraintGenerator.kt
```diff
@@ -0,0 +1,74 @@
...
+
+            writer.withBlock("public fun check${opName}RequestConstraint(data: ${opName}Request) {", "}") {
+                for (memberShape in inputMembers.values) {
+                    addConstaintValidation("data.", memberShape, writer)
+                }
+            }
+        }
+    }
+
+    private fun renderResponseConstraintsValidation() {
```
0marperez: Regarding my [other](https://github.com/smithy-lang/smithy-kotlin/pull/1338/files#r2229524657) comment, I think we should get rid of `renderResponseConstraintsValidation`

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/contraints/ConstraintGenerator.kt
```diff
@@ -0,0 +1,74 @@
...
+        for (memberTraitMap in memberTraits) {
+            val memberTrait = memberTraitMap.value
+            val traitGenerator = getTraitGeneratorFromTrait(prefix, memberName, memberTrait, pkgName, writer)
+            traitGenerator?.render()
+        }
+
+        for (member in targetShape.allMembers) {
+            val newMemberPrefix = "${targetShape.id.name}".replaceFirstChar { it.lowercase() }
+            writer.withBlock("if (${prefix}$memberName != null) {", "}") {
+                withBlock("for (${newMemberPrefix}${member.key} in ${prefix}$memberName) {", "}") {
```
0marperez: You don't need to use `${}` if you're not accessing something / modifying / writing code. You can just use `$` twice e.g. `$prefix$memberName`. It helps with readability

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/service/contraints/ConstraintUtilsGenerator.kt
luigi617: Although they seems to use all `.size.toLong()`, but their `size` property is different, so the compiler doesn't allow me to combine them

tests/codegen/service-codegen-tests/model/service-constraints-test.smithy
```diff
@@ -0,0 +1,174 @@
...
+    @length(min: 1, max: 2)
+    betweenLengthInput: MyMap
+}
+
+@output
+structure LengthConstraintTestOutput {}
+
+
+
+
```
0marperez: Spacing is inconsistent across the file and a little hard to follow. It's easier to read if we standardize the spacing to just 1 line between things and add comments to separate sections when needed

tests/codegen/service-codegen-tests/src/test/kotlin/com/test/ServiceConstraintsTest.kt
```diff
@@ -0,0 +1,571 @@
...
+    fun boot() {
+        proc = startService("netty", port, closeGracePeriodMillis, closeTimeoutMillis, requestBodyLimit, projectDir)
+        val ready = waitForPort(port, portListenerTimeout)
+        assertTrue(ready, "Service did not start within $portListenerTimeout s")
+    }
+
+    @AfterAll
+    fun shutdown() = cleanupService(proc)
+
+    /* Tests for checking constraint traits work */
```
0marperez: This comment should probably be placed at the class level as a Kdoc comment since all the tests in the class check constraint traits work

tests/codegen/service-codegen-tests/src/test/kotlin/com/test/ServiceConstraintsTest.kt
```diff
@@ -0,0 +1,571 @@
...
+        )
+        assertIs<HttpResponse<ByteArray>>(response)
+        assertEquals(400, response.statusCode(), "Expected 400")
+
+        val body = cbor.decodeFromByteArray(
+            ErrorResponse.serializer(),
+            response.body(),
+        )
+        assertEquals(400, body.code)
+        assertEquals("betweenLengthInput's size must be between 1 and 2", body.message)
```
0marperez: Some people might be confused about whether the size range is inclusive or exclusive, we should include that in the message to make it clear

tests/codegen/service-codegen-tests/src/test/kotlin/com/test/ServiceConstraintsTest.kt
```diff
@@ -0,0 +1,571 @@
...
+        )
+        assertIs<HttpResponse<ByteArray>>(response)
+        assertEquals(400, response.statusCode(), "Expected 400")
+
+        val body = cbor.decodeFromByteArray(
+            ErrorResponse.serializer(),
+            response.body(),
+        )
+        assertEquals(400, body.code)
+        assertEquals("betweenInput must be between 0 and 5", body.message)
```
0marperez: Same comment about inclusive/exclusive for range

tests/codegen/service-codegen-tests/src/test/kotlin/com/test/ServiceConstraintsTest.kt
```diff
@@ -0,0 +1,571 @@
...
+        )
+        assertIs<HttpResponse<ByteArray>>(response)
+        assertEquals(400, response.statusCode(), "Expected 400")
+
+        val body = cbor.decodeFromByteArray(
+            ErrorResponse.serializer(),
+            response.body(),
+        )
+        assertEquals(400, body.code)
+        assertEquals("member must have unique items", body.message)
```
0marperez: We should add what member we're referring to if possible. The error message says what the issue is but it's missing how to fix it

0marperez (commented):
Just a few small things and it should be ready to merge

0marperez (approved):

PR #1343, opened 2025-07-24
feat: add support for call timeout and attempt timeout

### LINKED ISSUE #1320: Support per-attempt timeout configuration
### Describe the feature

The Kotlin SDK does not presently support per-attempt timeouts, although other SDKs such as [Java v2](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/best-practices.html#bestpractice5) and [Rust](https://docs.aws.amazon.com/sdk-for-rust/latest/dg/timeouts.html) do. Design and implement new config parameter(s) to support this.

Consider also whether a dedicated request-level timeout setting would also be helpful. Users can implement their own guards using [**kotlinx.coroutines**'s `withTimeout`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/with-timeout.html) function but a native config param is likely more expected from other SDKs and allows more centralized control of timeouts.

### Is your feature request related to a problem?

Users currently have no way to limit attempt duration and can only control overall request duration via `withTimeout`.

### Proposed Solution

_No response_

### Describe alternative solutions or features you've considered

_No response_

### Acknowledge

### Smithy-Kotlin version

1.4.22

### Platform (JVM/JS/Native)

(any)

(any)

ianbotsf:
## Issue \#

Resolves #1320 

## Description of changes

This change adds two new parameters to all Smithy clients:
* `callTimeout` — enforces a limit on overall call duration, including any retries. Call timeouts are _not_ retried.
* `attemptTimeout` — enforces a limit on individual attempt durations (i.e., inside of a retry loop). Attempt timeouts are retryable.

Both of these new options default to `null`, maintaining the existing behavior of enforcing no client-level timeouts on either calls or attempts.

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt
@@ -190,24 +190,25 @@ object RuntimeConfigProperty {
     val AttemptTimeout = ConfigProperty {
         name = "attemptTimeout"
         symbol = KotlinTypes.Time.Duration.asNullable()
-        baseClass = RuntimeTypes.HttpClient.Config.TimeoutClientConfig
+        baseClass = RuntimeTypes.HttpClient.Config.TimeoutConfig
         useNestedBuilderBaseClass()
 
         documentation = """
             The maximum amount of time to wait for any single attempt of a request within the retry loop. By default,
-            the value is `null` indicating no timeout is enforced.
+            the value is `null` indicating no timeout is enforced. Attempt timeouts may be retried if allowed by the
+            current retry policy and retry capacity.
         """.trimIndent()
     }
 
     val CallTimeout = ConfigProperty {
         name = "callTimeout"
         symbol = KotlinTypes.Time.Duration.asNullable()
-        baseClass = RuntimeTypes.HttpClient.Config.TimeoutClientConfig
+        baseClass = RuntimeTypes.HttpClient.Config.TimeoutConfig
         useNestedBuilderBaseClass()
 
         documentation = """
             The maximum amount of time to wait for completion of a call, including any retries after the first attempt.
-            By default, the value is `null` indicating no timeout is enforced.
+            By default, the value is `null` indicating no timeout is enforced. Call timeouts are not retried.
         """.trimIndent()
     }
 }
--- a/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/operation/Timeouts.kt
+++ b/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/operation/Timeouts.kt
@@ -25,11 +25,25 @@ internal sealed interface TimeoutScope {
 /**
  * Indicates that a client-side configured timeout was exceeded (e.g., call timeout, attempt timeout, etc.)
  */
-public class ClientTimeoutException(message: String?, cause: Throwable?) : ClientException(message, cause) {
+public abstract class ClientTimeoutException(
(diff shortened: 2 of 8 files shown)


codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt
```diff
@@ -185,6 +186,30 @@ object RuntimeConfigProperty {
...
+        name = "attemptTimeout"
+        symbol = KotlinTypes.Time.Duration.asNullable()
+        baseClass = RuntimeTypes.HttpClient.Config.TimeoutClientConfig
+        useNestedBuilderBaseClass()
+
+        documentation = """
+            The maximum amount of time to wait for any single attempt of a request within the retry loop. By default,
+            the value is `null` indicating no timeout is enforced.
+        """.trimIndent()
+    }
```
lauzadis: Should we mention that these timeouts are subject to retries?

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/config/TimeoutClientConfig.kt
```diff
@@ -0,0 +1,37 @@
+package aws.smithy.kotlin.runtime.http.config
+
+import kotlin.time.Duration
+
+/**
+ * Defines optional timeout configuration for clients.
+ */
+public interface TimeoutClientConfig {
```
lauzadis: naming: `TimeoutConfig`?

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/operation/Timeouts.kt
```diff
@@ -0,0 +1,51 @@
...
+        override val description = "Attempt #$attempt"
+    }
+}
+
+/**
+ * Indicates that a client-side configured timeout was exceeded (e.g., call timeout, attempt timeout, etc.)
+ */
+public class ClientTimeoutException(message: String?, cause: Throwable?) : ClientException(message, cause) {
+    init {
+        sdkErrorMetadata.attributes[ErrorMetadata.Retryable] = true
```
lauzadis: Call timeouts should not be retried, right? I can see they aren't in your unit tests, but don't understand why we mark all ClientTimeoutExceptions as retryable when only attempt timeouts should be

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt
ianbotsf: Yes, I'll elaborate the docs.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/operation/Timeouts.kt
ianbotsf: Yeah, they're marked retryable but are only thrown _after_ the retry loop. I'll separate out different exceptions for calls vs attempts.

.changes/02bc36ae-952e-40ec-a0c2-38a8c8673508.json
```diff
@@ -0,0 +1,8 @@
+{
+    "id": "02bc36ae-952e-40ec-a0c2-38a8c8673508",
+    "type": "feature",
+    "description": "Enable configuration of timeouts for calls and attempt",
```
0marperez, line 4: Nit: `calls and attempt` -> `calls and attempts`

PR #1346, opened 2025-07-30
fix: skip unnecessary null checks in `equals()` methods for generated structures

ianbotsf:
## Issue \#

(none)

## Description of changes

`StructureGenerator`'s equality methods don't check for nullability of member symbols, leading to unnecessary null checks.

### DIFF
--- a/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
+++ b/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
@@ -6,7 +6,10 @@ package software.amazon.smithy.kotlin.codegen.rendering
 
 import software.amazon.smithy.codegen.core.CodegenException
 import software.amazon.smithy.codegen.core.Symbol
-import software.amazon.smithy.kotlin.codegen.core.*
+import software.amazon.smithy.kotlin.codegen.core.RenderingContext
+import software.amazon.smithy.kotlin.codegen.core.RuntimeTypes
+import software.amazon.smithy.kotlin.codegen.core.defaultName
+import software.amazon.smithy.kotlin.codegen.core.withBlock
 import software.amazon.smithy.kotlin.codegen.lang.KotlinTypes
 import software.amazon.smithy.kotlin.codegen.model.*
 import software.amazon.smithy.kotlin.codegen.rendering.serde.ClientErrorCorrection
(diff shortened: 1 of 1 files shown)


lauzadis (approved):

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
```diff
@@ -192,20 +195,31 @@ class StructureGenerator(
...
-                        .write("if (other.#1L == null) return false", memberName)
-                        .write("if (!#1L.contentEquals(other.#1L)) return false", memberName)
-                        .closeBlock("} else if (other.#1L != null) return false", memberName)
+                    if (memberSymbol.isNullable) {
+                        openBlock("if (#1L != null) {", memberName)
+                            .write("if (other.#1L == null) return false", memberName)
+                            .write("if (!#1L.contentEquals(other.#1L)) return false", memberName)
+                            .closeBlock("} else if (other.#1L != null) return false", memberName)
+                    } else {
+                        write("if (!#1L.contentEquals(other.#1L)) return false", memberName)
```
0marperez, line 206: Should we be checking the nullability of `other.#1L`? It's done in the block above

> .write("if (other.#1L == null) return false", memberName)

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
```diff
@@ -192,20 +195,31 @@ class StructureGenerator(
...
-                        .closeBlock("} else if (other.#1L != null) return false", memberName)
+                    if (memberSymbol.isNullable) {
+                        openBlock("if (#L != null) {", memberName)
+                            .write("if (other.#L == null) return false", memberName)
+                            .write("if (#1L.size != other.#1L.size) return false", memberName)
+                            .openBlock("for (i in #L.indices) {", memberName)
+                            .write("if (!#1L[i].contentEquals(other.#1L[i])) return false", memberName)
+                            .closeBlock("}")
+                            .closeBlock("} else if (other.#1L != null) return false", memberName)
+                    } else {
```
0marperez, line 217: Same comment as above

codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/StructureGenerator.kt
ianbotsf: No, we're checking equality between two instances of the same shape. If `this.foo` cannot be null then `other.foo` cannot be null either.

0marperez: Ah, got it!

0marperez (approved):

PR #1347, opened 2025-07-30
chore: add config for enabling Dependabot version updates

ianbotsf:
## Issue \#

(none)

## Description of changes

Enables [Dependabot version updates](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: gradle
+    directory: /
+    schedule:
+      interval: daily # means every _weekday_ (Monday through Friday)


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: gradle
+    directory: /
```
lauzadis, line 4: In the documentation I see:
> Use to define the location of the package manifests for each package manager

which makes me think it should be pointing to `/gradle`? But I also see this:
> For GitHub Actions, use the value /.

which I think means Dependabot would automatically upgrade our GitHub action dependencies.

Do we need both?

.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
...
+    schedule:
+      interval: daily # means every _weekday_ (Monday through Friday)
```
lauzadis, line 6: It would be nice to also trigger it on a release of aws-crt-kotlin or aws-kotlin-repo-tools. I don't think that's possible with this yaml configuration though

.github/dependabot.yml
ianbotsf: I don't think so. Technically, the package manifest for Gradle starts from **/settings.gradle.kt** (if it exists) and **/build.gradle.kts**. It typically spiders out from there and may include **/gradle/libs.versions.toml** but the root is still **/** I think. I believe this configuration would vary if we had a monorepo containing many independent Gradle projects in various subdirectories.

I found [an example in **kotlin-logging**](https://github.com/oshai/kotlin-logging/blob/b8c52216cde014c6019defa92fc0257924631479/.github/dependabot.yml#L3-L4) and they use **/** too.

.github/dependabot.yml
ianbotsf: The [options for `schedule`](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#schedule-) are pretty limited. It's either a simple interval like `daily` or a cron expression. I think auto-creating PRs on release of upstream software will require a GitHub action or backend tooling.

lauzadis: Yes I was thinking of a GitHub action that can trigger a Dependabot scan manually, there's a bit of discussion here: https://github.com/dependabot/dependabot-core/issues/2980

0marperez (approved):

PR #1348, opened 2025-07-30
feat: emit bearer business metric

xinsong-cui:
## Description of changes

### DIFF
--- a/runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/EnvironmentBearerTokenProvider.kt
+++ b/runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/EnvironmentBearerTokenProvider.kt
@@ -28,7 +28,7 @@ public class EnvironmentBearerTokenProvider(
     override suspend fun resolve(attributes: Attributes): BearerToken {
         val bearerToken = sysPropKey?.let { platform.getProperty(it) }
             ?: envKey?.let { platform.getenv(it) }
-            ?: error("environment variable is not set")
+            ?: error("neither system property $sysPropKey nor environment variable $envKey is set")
         return object : BearerToken {
             override val token: String = bearerToken
             override val attributes: Attributes = mutableAttributes().apply {
--- a/runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/EnvironmentBearerTokenProvider.kt
+++ b/runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/EnvironmentBearerTokenProvider.kt
@@ -14,21 +14,28 @@ import aws.smithy.kotlin.runtime.util.PlatformProvider
 /**
  * A [BearerTokenProvider] that extracts the bearer token from JVM system properties or environment variables.
  */
-public class EnvironmentBearerTokenProvider(
+public class EnvironmentBearerTokenProvider private constructor(
     private val sysPropKey: String?,
-    private val envKey: String?,
+    private val envKey: String,
     private val platform: PlatformProvider = PlatformProvider.System,
 ) : BearerTokenProvider {
-    @Deprecated("Use constructor with separate system property and environment variable keys")
+    @Deprecated("This constructor does not support a parameter for a system property key and will be removed in version 1.6.x")
     public constructor(
-        key: String,
+        envKey: String,
         platform: PlatformProvider = PlatformProvider.System,
-    ) : this(null, key, platform)
+    ) : this(null, envKey, platform)
+
+    public constructor(
+        sysPropKey: String,
+        envKey: String,
+        platform: PlatformProvider = PlatformProvider.System,
+        @Suppress("UNUSED_PARAMETER") dummy: Boolean = true,
+    ) : this(sysPropKey as String?, envKey, platform)
 
     override suspend fun resolve(attributes: Attributes): BearerToken {
-        val bearerToken = sysPropKey?.let { platform.getProperty(it) }
-            ?: envKey?.let { platform.getenv(it) }
-            ?: error("neither system property $sysPropKey nor environment variable $envKey is set")
+        val bearerToken = sysPropKey?.let(platform::getProperty) ?: platform.getenv(envKey)
+        if (bearerToken.isNullOrBlank()) throw IllegalStateException("""Missing values for system property "$sysPropKey" and environment variable "$envKey"""")
+
         return object : BearerToken {
             override val token: String = bearerToken
             override val attributes: Attributes = mutableAttributes().apply {
--- a/runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/EnvironmentBearerTokenProvider.kt
+++ b/runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/EnvironmentBearerTokenProvider.kt
@@ -9,18 +9,24 @@ import aws.smithy.kotlin.runtime.businessmetrics.emitBusinessMetric
 import aws.smithy.kotlin.runtime.collections.Attributes
 import aws.smithy.kotlin.runtime.collections.toMutableAttributes
 import aws.smithy.kotlin.runtime.time.Instant
+import aws.smithy.kotlin.runtime.util.PlatformProvider
 
 /**
- * A [BearerTokenProvider] that extracts the bearer token from the target environment variable.
+ * A [BearerTokenProvider] that extracts the bearer token from JVM system properties or environment variables.
  */
 public class EnvironmentBearerTokenProvider(
-    private val bearerToken: String,
+    private val key: String,
+    private val platform: PlatformProvider = PlatformProvider.System,
 ) : BearerTokenProvider {
-    override suspend fun resolve(attributes: Attributes): BearerToken = object : BearerToken {
-        override val token: String = bearerToken
-        override val attributes: Attributes = attributes.toMutableAttributes().apply {
-            emitBusinessMetric(SmithyBusinessMetric.BEARER_SERVICE_ENV_VARS)
+    override suspend fun resolve(attributes: Attributes): BearerToken {
+        val bearerToken = platform.getProperty(key) ?: platform.getenv(key)
+            ?: error("$key environment variable is not set")
+        return object : BearerToken {
+            override val token: String = bearerToken
+            override val attributes: Attributes = attributes.toMutableAttributes().apply {
+                emitBusinessMetric(SmithyBusinessMetric.BEARER_SERVICE_ENV_VARS)
+            }
+            override val expiration: Instant? = null
         }
-        override val expiration: Instant? = null
     }
 }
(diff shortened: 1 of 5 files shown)


runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/EnvironmentBearerTokenProvider.kt
```diff
@@ -4,25 +4,28 @@
...
+ * A [BearerTokenProvider] that extracts the bearer token from JVM system properties or environment variables.
  */
 public class EnvironmentBearerTokenProvider(
     private val key: String,
     private val platform: PlatformProvider = PlatformProvider.System,
 ) : BearerTokenProvider {
     override suspend fun resolve(attributes: Attributes): BearerToken {
-        val bearerToken = platform.getenv(key)
+        val bearerToken = platform.getProperty(key) ?: platform.getenv(key)
             ?: error("$key environment variable is not set")
```
ianbotsf: **Correctness**: The naming/casing convention for bearer token values is different between env vars and system properties. Environment variables are in the form `AWS_BEARER_TOKEN_<SERVICE>` whereas system properties are in the form `aws.bearerToken<SERVICE>`. See [`resolveEndpointUrl`](https://github.com/awslabs/aws-sdk-kotlin/blob/02de8c25e155a74281df31d2b6240fb261be23b7/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/endpoints/Resolvers.kt#L42-L74) for an example of how we handle a similar pattern elsewhere.

runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/EnvironmentBearerTokenProvider.kt
```diff
@@ -4,25 +4,28 @@
...
-
         return object : BearerToken {
             override val token: String = bearerToken
-            override val attributes: Attributes = emptyAttributes()
+            override val attributes: Attributes = attributes.toMutableAttributes().apply {
+                emitBusinessMetric(SmithyBusinessMetric.BEARER_SERVICE_ENV_VARS)
+            }
```
ianbotsf: **Correctness**: The attributes used to resolve credentials are different than the attributes returned with credentials. We should not be copying them into each other.

runtime/auth/http-auth/common/test/aws/smithy/kotlin/runtime/http/auth/EnvironmentBearerTokenProviderTest.kt
```diff
@@ -12,23 +11,29 @@ class EnvironmentBearerTokenProviderTest {
...
+                env = mutableMapOf("TEST_TOKEN" to "test-env-bearer-token"),
+                props = mutableMapOf("TEST_TOKEN" to "test-sys-props-bearer-token"),
+            ),
         )
 
         val token = provider.resolve()
 
-        assertEquals("test-bearer-token", token.token)
+        assertEquals("test-sys-props-bearer-token", token.token)
     }
```
ianbotsf: **Nit**: This test no longer verifies that env vars can be used to resolve a bearer token. We should add a new test for verifying system properties work and leave the existing test as-is.

runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/EnvironmentBearerTokenProvider.kt
xinsong-cui: Yeah, I am aware of this. I handle the key resolution in  in [FinalizeConfig](https://github.com/awslabs/aws-sdk-kotlin/pull/1658/files#diff-cf40f05ddf203416728fc2a12bb34c3667251a8d58d2bc21a7cda42df8a0b8b9R71-R90) before passing it to this function. So the ```resolve``` function should always sourcing the correct token.

This is to prevent API breaking, but I don't think an api change here could actually affect customers. If you prefer I can modify it to accept separate params for ```env``` and ```sysProps```

ianbotsf: The problem with handling it outside the provider is that you're reading the values twice and it introduces a bug where if a user sets a system property `AWS_BEARER_TOKEN_BEDROCK` you'll actually use that instead of `aws.bearerTokenBedrock`.

Modifying this constructor _would_ be a breaking change that definitely could affect users. Please add a new constructor, mark the old one as deprecated, and skip checking the system property if the old constructor was used and we have no property name to query.

runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/EnvironmentBearerTokenProvider.kt
```diff
@@ -4,25 +4,36 @@
...
+    @Deprecated("Use constructor with separate system property and environment variable keys")
+    public constructor(
+        key: String,
+        platform: PlatformProvider = PlatformProvider.System,
+    ) : this(null, key, platform)
 
+    override suspend fun resolve(attributes: Attributes): BearerToken {
+        val bearerToken = sysPropKey?.let { platform.getProperty(it) }
+            ?: envKey?.let { platform.getenv(it) }
+            ?: error("environment variable is not set")
```
0marperez: The error message needs to be updated, both the environment variable and system property are not set

runtime/auth/http-auth/common/test/aws/smithy/kotlin/runtime/http/auth/EnvironmentBearerTokenProviderTest.kt
```diff
@@ -9,27 +8,49 @@ import kotlin.test.assertFailsWith
...
+        assertEquals("test-env-bearer-token", token.token)
+    }
+
+    @Test
+    fun testResolveFromSysProps() = runTest {
+        val provider = EnvironmentBearerTokenProvider(
+            "TEST_SYS_PROPS_TOKEN",
+            "TEST_ENV_TOKEN",
+            TestPlatformProvider(
+                props = mutableMapOf("TEST_SYS_PROPS_TOKEN" to "test-sys-props-bearer-token"),
```
0marperez: We should add `env = mutableMapOf("TEST_ENV_TOKEN" to "test-env-bearer-token")` to the platform provider in this test to also check that the resolution priority is correct

0marperez (approved):

runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/EnvironmentBearerTokenProvider.kt
```diff
@@ -4,25 +4,36 @@
...
     private val platform: PlatformProvider = PlatformProvider.System,
 ) : BearerTokenProvider {
-    override suspend fun resolve(attributes: Attributes): BearerToken {
-        val bearerToken = platform.getenv(key)
-            ?: error("$key environment variable is not set")
+    @Deprecated("Use constructor with separate system property and environment variable keys")
+    public constructor(
+        key: String,
+        platform: PlatformProvider = PlatformProvider.System,
+    ) : this(null, key, platform)
```
ianbotsf: **Correctness**: This still isn't quite right. Long term, we don't want a public constructor that takes nullable keys—we want to support only a single constructor that takes non-nullable keys. We need a nullable system property key constructor for now to address backwards compatibility but we must also add the non-null constructor we _want_ to be used.

Also, don't we need to validate that the value we resolve is not blank?

```kotlin
public class EnvironmentBearerTokenProvider private constructor(
    private val sysPropKey: String?,
    private val envKey: String,
    private val platform: PlatformProvider = PlatformProvider.System,
) : BearerTokenProvider {
    @Deprecated("This constructor does not support a parameter for a system property key and will be removed in version 1.6.x")
    public constructor(
        envKey: String,
        platform: PlatformProvider = PlatformProvider.System,
    ) : this(null, key, platform)

    public constructor(
        sysPropKey: String,
        envKey: String,
        platform: PlatformProvider,
    ) : this(sysPropKey, envKey, platform)

    override suspend fun resolve(attributes: Attributes): BearerToken {
        val bearerToken = sysPropKey?.let(platform::getProperty) ?: platform.getenv(envKey)
        if (bearerToken.isBlank) throw ProviderConfigurationException("""Missing values for system property "$sysPropKey" and environment variable "$envKey"""")

        return object : BearerToken {
            override val token: String = bearerToken
            override val attributes: Attributes = emptyAttributes()
            override val attributes: Attributes = mutableAttributes().apply {
                emitBusinessMetric(SmithyBusinessMetric.BEARER_SERVICE_ENV_VARS)
            }
            override val expiration: Instant? = null
        }
    }
}
```

runtime/auth/http-auth/common/test/aws/smithy/kotlin/runtime/http/auth/EnvironmentBearerTokenProviderTest.kt
```diff
@@ -9,27 +8,65 @@ import kotlin.test.assertFailsWith
...
     @Test
-    fun testResolveWithValidToken() = runTest {
+    fun testResolveFromEnvVar() = runTest {
         val provider = EnvironmentBearerTokenProvider(
-            "TEST_TOKEN",
-            TestPlatformProvider(mutableMapOf("TEST_TOKEN" to "test-bearer-token")),
+            "TEST_SYS_PROPS_TOKEN",
+            "TEST_ENV_TOKEN",
+            TestPlatformProvider(
+                env = mutableMapOf("TEST_ENV_TOKEN" to "test-env-bearer-token"),
```
ianbotsf: **Nit**: None of these entries have to be mutable maps, they can use `mapOf` instead of `mutableMapOf`.

Applies to other test cases as well.

ianbotsf (changes requested):

runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/EnvironmentBearerTokenProvider.kt
xinsong-cui: Adding a constructor with non-null keys causes a JVM signature clash, seems like JVM treat `String` and `String?` as the same. Two options I can think of is to add a dummy parameter to constructor and remove it later or use a factory method. 
```
public constructor(
    sysPropKey: String,
    envKey: String,
    platform: PlatformProvider = PlatformProvider.System,
    @Suppress("UNUSED_PARAMETER") dummy: Boolean = true,
) : this(sysPropKey as String?, envKey, platform)
```
Do you think a dummy param is okay to use here?

ianbotsf: Oh right, forgot about JVM nullability. 🤦‍♂️

OK well I don't like the idea of the dummy parameter because it's part of our public API and is likely to cause confusion. One way to resolve this conundrum is to just use the empty string as a (temporary) replacement for null:

```kotlin
public class EnvironmentBearerTokenProvider(
    sysPropKey: String,
    envKey: String,
    platform: PlatformProvider,
) : BearerTokenProvider {
    @Deprecated("This constructor does not support a parameter for a system property key and will be removed in version 1.6.x")
    public constructor(
        envKey: String,
        platform: PlatformProvider = PlatformProvider.System,
    ) : this("", key, platform)

    override suspend fun resolve(attributes: Attributes): BearerToken {
        val bearerToken = sysPropKey.takeUnless(String::isBlank).let(platform::getProperty)
            ?: platform.getenv(envKey)
        if (bearerToken.isBlank) throw ProviderConfigurationException("""Missing values for system property "$sysPropKey" and environment variable "$envKey"""")

        return object : BearerToken {
            override val token: String = bearerToken
            override val attributes: Attributes = emptyAttributes()
            override val attributes: Attributes = mutableAttributes().apply {
                emitBusinessMetric(SmithyBusinessMetric.BEARER_SERVICE_ENV_VARS)
            }
            override val expiration: Instant? = null
        }
    }
}
```

PR #1386, opened 2025-08-14
fix: correctly return number of bytes read from chunked streams

### LINKED ISSUE #1285: Progress listening with chunked encoding reads more bytes transferred than file size
### Describe the bug

Amplify Android has been investigating an issue (https://github.com/aws-amplify/amplify-android/issues/2869) where the root cause is that when doing an S3 transfer with chunked encoding, we read the number of bytes transferred as larger than the actual file size itself. 

I understand that there's technically byte transfer overhead holistically when it comes to chunked encoding passing additional headers however the numbers don't really add up. e.g. for a 10,000,000 byte file, there's an additional 13,941 bytes that get transferred which seems way higher than the headers overhead size for chunked encoding. Additionally, other SDKs don't have this issue. I cross-referenced with Amplify Swift and even the legacy AWS SDK for Android's S3 transfer and their progress is always reported accurately -- for a 10,000,000 byte file, 10,000,000 bytes are reported as transferred. 

If we disable chunked encoding, the number of bytes that are reported is accurate.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected behavior

We understand that for situations where the length of the content is unknown, progress listening doesn't make sense -- we're not asking about that. We have a known file size, so we expect that we should know when the file has finished transferring. 

### Current behavior

If you read the number of bytes transferred for an S3 download/upload with chunked encoding enabled, the number of bytes transferred is larger than the file size.

### Steps to Reproduce

See our implementation of a progress listener interceptor:
https://github.com/aws-amplify/amplify-android/blob/e31b961e97c5896f200ff5939fa6d249832c0d27/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/ProgressListenerHttpInterceptor.kt#L53-L72

That gets added as an interceptor to the S3 client like this:
https://github.com/aws-amplify/amplify-android/blob/e31b961e97c5896f200ff5939fa6d249832c0d27/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/transfer/worker/SinglePartUploadWorker.kt#L45-L51

From there, just initiate an S3 download or upload and read the number of bytes and it'll read a higher number of bytes with encoding than the file size itself.

### Possible Solution

_No response_

### Context

_No response_

### Smithy-Kotlin version

1.3.23

### Platform (JVM/JS/Native)

Android

Android 11

ianbotsf:
## Issue \#

Fixes #1285 

## Description of changes

This change updates `AwsChunkedSource` and `AwsChunkedByteReadChannel` to return the actual number of bytes read from the underlying stream, _not_ the number of bytes written to the sink. It does this by adding read count tracking to `AwsChunkedReader` which must be read and reset by the calling source/channel. Also updates various tests to correctly assert the number of bytes read _and_ the number of bytes written.

This PR is an alternate implementation of #1294.


lauzadis (approved):

xinsong-cui (approved):

runtime/auth/aws-signing-tests/common/src/aws/smithy/kotlin/runtime/auth/awssigning/tests/AwsChunkedByteReadChannelTestBase.kt
```diff
@@ -47,11 +47,12 @@ abstract class AwsChunkedByteReadChannelTestBase : AwsChunkedTestBase(AwsChunked
         val sink = SdkBuffer()
 
         val bytesRead = awsChunked.readAll(sink)
-        writeJob.join()
+        // writeJob.join()
```
0marperez: Do we still need to join the write job?

.changes/7251f5e7-0e9a-4ea4-b4c7-30dad31f4622.json
```diff
@@ -0,0 +1,9 @@
+{
+    "id": "7251f5e7-0e9a-4ea4-b4c7-30dad31f4622",
+    "type": "bugfix",
+    "description": "⚠️ **IMPORTANT**: Correctly return number of bytes read from chunked streams",
+    "issues": [
+        "awslabs/smithy-kotlin#1285"
+    ],
+    "requiresMinorVersionBump": true
```
0marperez: Adding a note so we don't forget to remove `requiresMinorVersionBump`

runtime/auth/aws-signing-tests/common/src/aws/smithy/kotlin/runtime/auth/awssigning/tests/AwsChunkedByteReadChannelTestBase.kt
ianbotsf: It doesn't hurt but it's not strictly necessary. The `readAll` call _cannot_ return until all the data's been consumed which means the write job will have finished. I've uncommented that line just for a sanity check.

.changes/7251f5e7-0e9a-4ea4-b4c7-30dad31f4622.json
ianbotsf: Removed!

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/internal/StreamChunker.kt
```diff
@@ -0,0 +1,110 @@
...
+                hasLastChunkBeenSent = true
+                return 0 // no bytes read from source
+            }
+        } else {
+            val rc = input.read(inputBuffer, limit)
+            while (inputBuffer.size >= chunkSize) {
+                consumeAndWriteChunk(destination)
+            }
+            if (rc == -1L) {
+                check(input.eof) { "Got back -1 bytes but input adapter is not EOF" }
```
0marperez, line 57: When would the `rc` be `-1` but the input not `eof`?

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/internal/StreamChunker.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    private val inputBuffer = SdkBuffer()
+    private var hasLastChunkBeenSent = false
+
+    internal suspend inline fun readAndMaybeWrite(destination: SdkSink, limit: Long): Long {
+        if (input.eof) {
+            if (hasLastChunkBeenSent) {
+                return -1 // this stream is exhausted
+            } else {
+                // Write any remaining data as final chunk(s)
```
0marperez: I think we would only have one final chunk in the buffer at this point not multiple?

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/internal/StreamChunker.kt
ianbotsf: This is mostly a sanity check. It was useful while debugging the implementation but could provide a valuable fail-fast clue if a future bug involves streams which return `-1` bytes read but the stream is not yet closed/exhausted.

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/internal/StreamChunker.kt
ianbotsf: You're right, corrected the comment.

PR #1407, opened 2025-09-15
fix: overwrite `Content-Length` rather than appending

lauzadis:
We were appending to `Content-Length` rather than overwriting it. This means if content length was already set by the user, we send a list of content lengths, causing a signature mismatch.

## Description of changes

### DIFF
--- a/.changes/7e4f8c08-941c-4ea5-91b3-4c284746c453.json
+++ b/.changes/7e4f8c08-941c-4ea5-91b3-4c284746c453.json
@@ -1,5 +1,0 @@
-{
-    "id": "7e4f8c08-941c-4ea5-91b3-4c284746c453",
-    "type": "documentation",
-    "description": "Improve documentation for [`StandardRetryPolicy`](https://docs.aws.amazon.com/smithy-kotlin/api/latest/runtime-core/aws.smithy.kotlin.runtime.retries.policy/-standard-retry-policy/)"
-}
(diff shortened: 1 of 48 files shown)


runtime/protocol/http-client-engines/http-client-engine-crt/jvm/src/aws/smithy/kotlin/runtime/http/engine/crt/RequestUtil.kt
```diff
@@ -68,7 +68,7 @@ internal fun HttpRequest.toCrtRequest(callContext: CoroutineContext): aws.sdk.ko
     }
 
     val contentLength = body.contentLength?.takeIf { it >= 0 }?.toString() ?: headers[CONTENT_LENGTH_HEADER]
-    contentLength?.let { crtHeaders.append(CONTENT_LENGTH_HEADER, it) }
+    contentLength?.let { crtHeaders[CONTENT_LENGTH_HEADER] = it }
 
     return aws.sdk.kotlin.crt.http.HttpRequest(method.name, url.requestRelativePath, crtHeaders.build(), bodyStream)
```
ianbotsf, line 73: Can we add a new test for this? And also one that verifies the (presumably already correct?) behavior in OkHttp engine?

lauzadis: Added some tests

PR #1410, opened 2025-09-18
feat: minor version strategy rules

0marperez:
## Issue \#

N/A
## Description of changes

- Custom rule-set-provider with minor-version-strategy rule (deprecatedUntilVersion)
- GH workflow to run rule

### DIFF
--- a/.github/workflows/minor-version-bump.yml
+++ b/.github/workflows/minor-version-bump.yml
@@ -1,0 +1,12 @@
+name: Minor version bump check
+on:
+  pull_request:
+
+jobs:
+  minor-version-bump-check:
+    if: github.head_ref == '^v\d+\.\d+.*$' # Only runs on branches starting with vX.X (e.g., v1.2, v1.2.3, v1.2-main)
+    permissions: {}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Minor version bump check
+        uses: awslabs/aws-kotlin-repo-tools/.github/actions/minor-version-bump@main
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,6 +4,7 @@
  */
 import aws.sdk.kotlin.gradle.dsl.configureJReleaser
 import aws.sdk.kotlin.gradle.dsl.configureLinting
+import aws.sdk.kotlin.gradle.dsl.configureMinorVersionStrategyRules
 import aws.sdk.kotlin.gradle.util.typedProp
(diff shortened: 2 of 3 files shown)


.github/workflows/minor-version-bump.yml
```diff
@@ -0,0 +1,18 @@
+name: Minor version bump check
+on:
+  pull_request:
+
+jobs:
+  minor-version-bump-check:
+    permissions: {}
```
lauzadis, line 7: Do you need to specify this? Isn't the default "no permissions"?

.github/workflows/minor-version-bump.yml
```diff
@@ -0,0 +1,18 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check branch name
+        run: |
+          # Only continue on branches starting with vX.X (e.g. v1.2, v1.2.3, v1.2-main)
+          if [[ ! "$GITHUB_HEAD_REF" =~ ^v\.[0-9]+\.[0-9]+.*$ ]]; then
+            exit 0
+          fi
```
lauzadis: This whole step can be removed by using an `if:` configuration and checking `github.ref`

xinsong-cui (approved):

.github/workflows/minor-version-bump.yml
0marperez: It's best to specify it. The default is likely read and write, I ran into issues with this recently: https://github.com/smithy-lang/smithy-kotlin/security/code-scanning/93

We also do the same in other workflows for the same reason:
https://github.com/aws/aws-sdk-kotlin/blob/8716c234c59185fd09d819c65189ab6d4ede071d/.github/workflows/continuous-integration.yml#L7

.github/workflows/minor-version-bump.yml
0marperez: As far as I know we can't use regex in [conditionals](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-jobs-with-conditions) but I'll give it a try.

ianbotsf (approved):

.github/workflows/minor-version-bump.yml
lauzadis: If it's not simple to implement don't spend too much time getting it to work. It just looked like this could be replaced with an `if` on the step

PR #1415, opened 2025-09-19
misc: plannedRemoval annotation

0marperez:
## Issue \#

N/A

## Description of changes

Adds annotation to indicate when a deprecated API will be removed.

### DIFF
--- a/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/Annotations.kt
+++ b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/Annotations.kt
@@ -64,7 +64,7 @@ public annotation class SdkDsl
  * Any code using this API should migrate to the suggested replacement as soon as possible.
  */
 @RequiresOptIn(
-    message = "This API is deprecated and will be scheduled for removal in an upcoming version.",
+    message = "This API is deprecated and scheduled for removal in an upcoming version.",
     level = RequiresOptIn.Level.WARNING,
 )
 public annotation class PlannedRemoval(


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/Annotations.kt
```diff
@@ -58,3 +58,16 @@ public annotation class ExperimentalApi
  */
 @DslMarker
 public annotation class SdkDsl
+
+/**
+ * Indicates that the annotated API is deprecated and scheduled for removal in an upcoming minor version.
+ * Any code using this API should migrate to the suggested replacement as soon as possible.
+ */
+@RequiresOptIn(
+    message = "This API is deprecated and will be removed in an upcoming minor version.",
```
lauzadis: question: Can we use `major` and `minor` in this message to indicate _which_ minor version the API will be removed in?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/Annotations.kt
```diff
@@ -58,3 +58,16 @@ public annotation class ExperimentalApi
...
+    level = RequiresOptIn.Level.WARNING,
+)
+public annotation class DeprecatedUntilVersion(
```
lauzadis: `DeprecatedUntilVersion` makes it sound like we'll be _un_-deprecating this API after X version. What if we called it `DeprecatedPendingRemoval`?

lauzadis: Is it possible for this to extend / implement `Deprecated` so we don't need two annotations?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/Annotations.kt
0marperez: If we could inherit from annotation classes, maybe, but Kotlin doesn't allow us to do so. Unless you know a workaround.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/Annotations.kt
0marperez: Yeah, I agree. When I first read our doc about the annotation, it reads like the annotation is meant to do that. I'd like for @ianbotsf to chime in. I personally like `@UpcomingDeletion`, `@WillBeRemoved`, or something similar. We shouldn't need to include anything about deprecation because the `@Deprecated` annotation will also always be included.

This is a slightly bigger change than it seems by the way. I'll need to update the annotation regex and searches in a few places.

0marperez: Unless you know some workaround, we can't extend `@Deprecated`. Kotlin doesn't allow us to do that. We can "implement" deprecated again but that doesn't seem ideal. It doesn't seem like we can fully implement the same functionality ourselves.
https://github.com/JetBrains/kotlin/blob/2.2.20/libraries/stdlib/src/kotlin/Annotations.kt#L36

lauzadis: Yeah, if we need to keep `@Deprecated` then including `...Deprecated...` in the new annotation feels redundant. I like `@PendingRemoval`

ianbotsf: Agreed, we don't need use the word "deprecated" twice. I don't love "upcoming deletion" or "will be removed" because they sound like promises and might imply a definitive timeline to callers. "Pending removal" sounds better but I think I'd prefer `@PlannedRemoval`.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/Annotations.kt
lauzadis: nit: Hopefully this doesn't happen for a while, but we may make a _major_ version bump that deletes APIs, so this message should probably just say "... scheduled for removal in an upcoming ~~minor~~ version"

lauzadis (approved):

PR #1423, opened 2025-09-24
kn: merge to main

lauzadis:
## Description of changes


runtime/auth/aws-signing-default/api/aws-signing-default.api
```diff
@@ -5,7 +5,7 @@ public final class aws/smithy/kotlin/runtime/auth/awssigning/DefaultAwsSignerBui
 	public final fun setTelemetryProvider (Laws/smithy/kotlin/runtime/telemetry/TelemetryProvider;)V
 }
 
-public final class aws/smithy/kotlin/runtime/auth/awssigning/DefaultAwsSignerKt {
```
lauzadis, line 8: filename changed, but the classname and package is the same. Not a break

runtime/auth/aws-signing-crt/api/aws-signing-crt.api
```diff
@@ -1,4 +1,4 @@
-public final class aws/smithy/kotlin/runtime/auth/awssigning/crt/CrtAwsSigner : aws/smithy/kotlin/runtime/auth/awssigning/AwsSigner {
```
lauzadis, line 1: added `WithCrt`, not a break

runtime/runtime-core/api/runtime-core.api
```diff
@@ -866,12 +891,8 @@ public final class aws/smithy/kotlin/runtime/io/SdkBuffer : aws/smithy/kotlin/ru
 	public fun getBuffer ()Laws/smithy/kotlin/runtime/io/SdkBuffer;
 	public final fun getSize ()J
 	public fun hashCode ()I
-	public fun inputStream ()Ljava/io/InputStream;
-	public fun isOpen ()Z
-	public fun outputStream ()Ljava/io/OutputStream;
 	public fun peek ()Laws/smithy/kotlin/runtime/io/SdkBufferedSource;
 	public fun read (Laws/smithy/kotlin/runtime/io/SdkBuffer;J)J
-	public fun read (Ljava/nio/ByteBuffer;)I
```
lauzadis, line 879: These deletions are caused by changes in SdkBuffereredSink and SdkBufferedSource, see explanations below

runtime/runtime-core/api/runtime-core.api
```diff
@@ -1068,16 +1085,13 @@ public abstract interface class aws/smithy/kotlin/runtime/io/SdkSource : java/io
 	public abstract fun read (Laws/smithy/kotlin/runtime/io/SdkBuffer;J)J
 }
 
-public final class aws/smithy/kotlin/runtime/io/SdkSourceJVMKt {
```
lauzadis, line 1076: SdkSourceJvm was merged to create SdkSourceJvmAndNative, not a break

runtime/runtime-core/api/runtime-core.api
```diff
@@ -1068,16 +1085,13 @@ public abstract interface class aws/smithy/kotlin/runtime/io/SdkSource : java/io
...
+public final class aws/smithy/kotlin/runtime/io/SdkSourceKt {
+	public static final fun readFully (Laws/smithy/kotlin/runtime/io/SdkSource;Laws/smithy/kotlin/runtime/io/SdkBuffer;J)V
 	public static final fun readToByteArray (Laws/smithy/kotlin/runtime/io/SdkSource;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static final fun toSdkByteReadChannel (Laws/smithy/kotlin/runtime/io/SdkSource;Lkotlinx/coroutines/CoroutineScope;)Laws/smithy/kotlin/runtime/io/SdkByteReadChannel;
 	public static synthetic fun toSdkByteReadChannel$default (Laws/smithy/kotlin/runtime/io/SdkSource;Lkotlinx/coroutines/CoroutineScope;ILjava/lang/Object;)Laws/smithy/kotlin/runtime/io/SdkByteReadChannel;
 }
 
-public final class aws/smithy/kotlin/runtime/io/SdkSourceKt {
-	public static final fun readFully (Laws/smithy/kotlin/runtime/io/SdkSource;Laws/smithy/kotlin/runtime/io/SdkBuffer;J)V
-}
```
lauzadis, line 1084: Not a break, just relocated in the file. See L1088 in PR changes

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigIntegerJVM.kt
```diff
@@ -4,25 +4,47 @@
  */
 package aws.smithy.kotlin.runtime.content
 
-public actual class BigInteger actual constructor(public val value: String) :
```
lauzadis, line 7: removing `public val value: String` was a break

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigDecimalJVM.kt
```diff
@@ -4,33 +4,52 @@
  */
 package aws.smithy.kotlin.runtime.content
 
-public actual class BigDecimal actual constructor(public val value: String) :
```
lauzadis, line 7: removing `public val value: String` was a break

runtime/runtime-core/api/runtime-core.api
```diff
@@ -383,9 +391,11 @@ public final class aws/smithy/kotlin/runtime/content/BigDecimal : java/lang/Numb
 	public final fun floatValue ()F
 	public final fun getExponent ()I
 	public final fun getMantissa ()Laws/smithy/kotlin/runtime/content/BigInteger;
-	public final fun getValue ()Ljava/lang/String;
```
lauzadis, line 391: real break, see BigDecimalJVM

runtime/runtime-core/api/runtime-core.api
```diff
@@ -405,7 +416,6 @@ public final class aws/smithy/kotlin/runtime/content/BigInteger : java/lang/Numb
 	public final fun doubleValue ()D
 	public fun equals (Ljava/lang/Object;)Z
 	public final fun floatValue ()F
-	public final fun getValue ()Ljava/lang/String;
```
lauzadis, line 413: real break, see BigIntegerJVM

runtime/runtime-core/jvmAndNative/src/aws/smithy/kotlin/runtime/io/SdkBufferJvmAndNative.kt
```diff
@@ -74,48 +94,12 @@ public actual class SdkBuffer :
...
-    actual override fun flush(): Unit = commonFlush()
-
-    actual override fun emit() {
-        inner.emit()
-    }
     actual override fun close(): Unit = commonClose()
-    override fun isOpen(): Boolean = inner.isOpen
-
-    override fun inputStream(): InputStream = inner.inputStream()
-    override fun outputStream(): OutputStream = inner.outputStream()
```
lauzadis, line 120: These are now exposed as extension functions, see `runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/io/JavaIO.kt`

runtime/runtime-core/api/runtime-core.api
lauzadis: Fixed by adding `public val value: String = delegate.toString()` to JVM implementation

runtime/runtime-core/api/runtime-core.api
lauzadis: Fixed by adding `public val value: String = delegate.toString()` to JVM implementation

runtime/runtime-core/api/runtime-core.api
```diff
@@ -904,11 +924,9 @@ public final class aws/smithy/kotlin/runtime/io/SdkBuffer : aws/smithy/kotlin/ru
 	public fun writeUtf8 (Ljava/lang/String;II)V
 }
 
-public abstract interface class aws/smithy/kotlin/runtime/io/SdkBufferedSink : aws/smithy/kotlin/runtime/io/SdkSink, java/nio/channels/WritableByteChannel {
+public abstract interface class aws/smithy/kotlin/runtime/io/SdkBufferedSink : aws/smithy/kotlin/runtime/io/SdkSink {
 	public abstract fun emit ()V
-	public abstract fun flush ()V
 	public abstract fun getBuffer ()Laws/smithy/kotlin/runtime/io/SdkBuffer;
-	public abstract fun outputStream ()Ljava/io/OutputStream;
```
lauzadis, line 916: We renamed `SdkBufferedSinkJvm` to `SdkBufferedSinkJvmAndNative` which no longer implements `WritableByteChannel`. This removes the `outputStream()` API from the interface, but we now expose it in JavaIO as an extension function. Same package, so it's not a break?

runtime/runtime-core/api/runtime-core.api
```diff
@@ -929,10 +947,9 @@ public final class aws/smithy/kotlin/runtime/io/SdkBufferedSink$DefaultImpls {
 	public static synthetic fun writeUtf8$default (Laws/smithy/kotlin/runtime/io/SdkBufferedSink;Ljava/lang/String;IIILjava/lang/Object;)V
 }
 
-public abstract interface class aws/smithy/kotlin/runtime/io/SdkBufferedSource : aws/smithy/kotlin/runtime/io/SdkSource, java/nio/channels/ReadableByteChannel {
+public abstract interface class aws/smithy/kotlin/runtime/io/SdkBufferedSource : aws/smithy/kotlin/runtime/io/SdkSource {
 	public abstract fun exhausted ()Z
 	public abstract fun getBuffer ()Laws/smithy/kotlin/runtime/io/SdkBuffer;
-	public abstract fun inputStream ()Ljava/io/InputStream;
```
lauzadis, line 940: We renamed `SdkBufferedSourceJvm` to `SdkBufferedSourceJvmAndNative` which no longer implements `ReadableByteChannel`. This removes the `inputStream()` API from the interface, but we now expose it in JavaIO.kt as an extension function. Same package, so it's not a break?

runtime/runtime-core/api/runtime-core.api
ianbotsf: Definitely a break. The fact that this interface no longer inherits from `WritableByteChannel` means callers can no longer pass it into fields/args of that type. Additionally, extension functions must be imported to be used and this changes the import requirement.

lauzadis: Fixed by splitting this back into Jvm and Native sourcesets like we discussed. We can explore combining them again in a future minor version bump

PR #1437, opened 2025-10-09
feat: OkHttpEngine BYOC

lauzadis:
Adds a new constructor to OkHttpEngine which allows users to bring their own client, allowing them to customize behavior outside of what the SDK exposes.

## Issue \#

Closes https://github.com/aws/aws-sdk-kotlin/issues/1707

## Description of changes

### DIFF
--- a/runtime/build.gradle.kts
+++ b/runtime/build.gradle.kts
@@ -3,7 +3,9 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import aws.sdk.kotlin.gradle.dsl.configurePublishing
-import aws.sdk.kotlin.gradle.kmp.*
+import aws.sdk.kotlin.gradle.kmp.configureKmpTargets
+import aws.sdk.kotlin.gradle.kmp.kotlin
+import aws.sdk.kotlin.gradle.kmp.needsKmpConfigured
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
 import org.jetbrains.kotlin.gradle.targets.native.tasks.KotlinNativeSimulatorTest


runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
```diff
@@ -38,6 +38,27 @@ public class OkHttpEngine(
...
+            eventListenerFactory { call ->
+                EventListenerChain(
+                    listOf(
+                        HttpEngineEventListener(client.connectionPool, config.hostResolver, client.dispatcher, metrics, call),
+                    ),
+                )
+            }
+            addInterceptor(MetricsInterceptor)
+        }.build()
+    }
```
ianbotsf: **Correctness**: We shouldn't create an OkHttp client just to close it and replace it with the user's client. Let's refactor these constructors to avoid unnecessary client creation and to remove the need for `var` fields.

runtime/build.gradle.kts
```diff
@@ -143,7 +141,7 @@ subprojects {
     }
 }
 
-// configureIosSimulatorTasks()
+configureIosSimulatorTasks()
```
ianbotsf: **Question**: Why did this change?

lauzadis: I don't know why it was commented, but I don't think it should be. This is necessary for TLS tests to pass, it was originally added to aws-crt-kotlin and I applied it here too, but it might not be necessary since our CI has been passing without it. Want me to delete it?

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
lauzadis: refactored to initialize the client lazily. I think we still need one var since we can't edit the primary constructor

0marperez (approved):

runtime/build.gradle.kts
ianbotsf: If it's not necessary for CI and we don't need it for any other reason then yes, I think we should remove it.

runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpEngine.kt
```diff
@@ -39,6 +40,11 @@ public class OkHttpEngine(
 ) : HttpClientEngineBase("OkHttp") {
     public constructor() : this(OkHttpEngineConfig.Default)
 
+    private var userProvidedClient: OkHttpClient? = null
+    public constructor(client: OkHttpClient) : this(OkHttpEngineConfig.Default) {
+        userProvidedClient = client
+    }
+
```
ianbotsf: **Comment**: The refactor looks much better and addresses my main concern about unnecessary client instantiation. Thanks!

**Nit**:

> we can't edit the primary constructor

I don't believe a constructor's "primaryness" affects the public API. It's a source code term which indicates it's the one that appears on the `class` declaration line, as opposed to those which appear in the body. The compiler also enforces that _if_ a primary constructor exists then all secondary constructors must invoke it.

I believe this code could be further improved by moving the existing primary constructor to a secondary one and either **a)** making each secondary constructor fully set the initial state or **b)** adding a new `private` primary constructor to set state. As I noted above, my primary concern is addressed so I'll leave it up to you whether to continue iterating.

lauzadis: Refactored a bit more for option **b)**, what do you think?

ianbotsf: Looks great!

ianbotsf (approved):

PR #1440, opened 2025-10-13
chore: refresh Mac CI targets, enable parallel builds by default

ianbotsf:
## Issue \#

(none)

## Description of changes

This change enables parallelism by default in **gradle.properties**, reproducing the work which was done in #1438 and reverted in #1442.

This change also updates our macOS CI targets to:
* Drop macOS 13 (which [is deprecated](https://github.com/actions/runner-images?tab=readme-ov-file#available-images)) and macOS 14
* Switch to using large/xlarge images for better performance

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -99,7 +99,6 @@ jobs:
         os:
           - macos-15-large  # x64
           - macos-15-xlarge # ARM
-        instance: [1, 2, 3, 4, 5, 6, 7, 8]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout sources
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -111,6 +113,17 @@ jobs:
         working-directory: ./smithy-kotlin
         shell: bash
         run: |
+          # Bump memory profile for Kotlin daemon
+          cat gradle.properties \
+              | sed -r 's/-Xmx[0-9]+G/-Xmx6G/g' \
+              | sed -r 's/-XX:MaxMetaspaceSize=[0-9]+G/-XX:MaxMetaspaceSize=2G/g' \
```
lauzadis, line 119: Do we also need to apply this to our internal release builds?

ianbotsf: Yes, good point. I'll open up a CR and get it merged before shipping this change.

PR #1449, opened 2025-10-24
misc: add base64 url encode

xinsong-cui:
## Description of changes

### DIFF
--- a/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/text/encoding/Base64.kt
+++ b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/text/encoding/Base64.kt
@@ -54,6 +54,15 @@ public fun String.encodeBase64(): String = encodeToByteArray().encodeBase64().de
 public fun ByteArray.encodeBase64String(): String = encodeBase64().decodeToString()
 
 /**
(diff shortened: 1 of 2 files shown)


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/text/encoding/Base64.kt
```diff
@@ -53,6 +53,15 @@ public fun String.encodeBase64(): String = encodeToByteArray().encodeBase64().de
...
 
+/**
+ * Encode [ByteArray] in base64url format (RFC 4648) as a [String]
+ */
+public fun ByteArray.encodeBase64Url(): String =
+    encodeBase64String()
+        .replace('+', '-')
+        .replace('/', '_')
+        .trimEnd('=')
+
```
lauzadis, line 64: Can we use the stdlib API instead of implementing this ourselves? https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.io.encoding/-base64/-default/-url-safe.html

xinsong-cui: Oh I wasn't aware of that, yeah I think we can just use that API

PR #1454, opened 2025-10-30
feat: sdk source readRemaining

0marperez:
## Issue \#

N/A

## Description of changes

Used by https://github.com/aws/aws-sdk-kotlin/pull/1712


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/io/SdkSource.kt
```diff
@@ -110,3 +110,18 @@ public fun SdkSource.readFully(sink: SdkBuffer, byteCount: Long) {
...
 }
+
+/**
+ * **Caution** Reads the entire contents of the source into [sink].
+ * This function will read until the source is exhausted and no bytes remain
+ *
+ * @param sink the buffer that data read from the source will be appended to
+ */
+@InternalApi
+public fun SdkSource.readRemaining(sink: SdkBuffer) {
```
lauzadis: I think this should return the total bytes read

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/io/SdkSource.kt
```diff
@@ -110,3 +110,23 @@ public fun SdkSource.readFully(sink: SdkBuffer, byteCount: Long) {
...
+    var totalReadBytes: Long = 0
+    var readBytes: Long
+
+    do {
+        // ensure any errors are propagated by attempting to read at least once
+        readBytes = read(sink, Long.MAX_VALUE)
+        totalReadBytes += readBytes
+    } while (readBytes != -1L)
+
+    return totalReadBytes + 1L // Account for last -1 read
```
lauzadis: I don't think we should be counting the `-1L`. Can you write some unit tests for this function?

PR #1457, opened 2025-11-04
misc: ecdsa raw signature

xinsong-cui:
## Description of changes

Adds support for generating ECDSA signatures in raw r||s format (64 bytes) in addition to the existing ASN.1 DER format.

### DIFF
--- a/runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/hashing/EcdsaJVMTest.kt
+++ b/runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/hashing/EcdsaJVMTest.kt
@@ -102,6 +102,7 @@ class EcdsaJVMTest {
     }
 
     @Test
+    // SHA256withECDSAinP1363Format algorithm was introduced in Java 11, skip on Java 8
     @EnabledForJreRange(min = JRE.JAVA_11)
     fun testVerifyRawSignature() {
         val keyGen = KeyPairGenerator.getInstance("EC")


runtime/runtime-core/api/runtime-core.api
```diff
@@ -726,7 +726,16 @@ public final class aws/smithy/kotlin/runtime/hashing/Crc32cKt {
 }
 
 public final class aws/smithy/kotlin/runtime/hashing/EcdsaJVMKt {
-	public static final fun ecdsaSecp256r1 ([B[B)[B
```
xinsong-cui, line 729: Existing ```ecdsaSecp256r1(key, message)``` calls work unchanged, should not be a breaking change

ianbotsf: This is definitely a breaking change because you've altered the method signature. Older consumers of **runtime-core** will encounter a `NoSuchMethodError` exception at runtime. The only way to do this safely without breaking binary compatibility is to maintain the old method signature and add a new one.

I recommend one of the following approaches:
* Expose a new method `ecdsaSecp256r1Rs` which invokes `ecdsaSecp256r1Rs` and extracts the `r` and `s` values. This could be implemented in **common** since it's just byte manipulation
* Expose a new method `asn1DerToRs` which converts any `ByteArray` from ASN.1 DER format to r||s. Callers would then invoke `ecdsaSecp256r1Rs` first and then `asn1DerToRs`. This could also be implemented in **common**.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/hashing/Ecdsa.kt
```diff
@@ -4,7 +4,16 @@
...
+
 /**
  * ECDSA on the SECP256R1 curve.
  */
-public expect fun ecdsaSecp256r1(key: ByteArray, message: ByteArray): ByteArray
+public expect fun ecdsaSecp256r1(
+    key: ByteArray,
+    message: ByteArray,
+    signatureType: EcdsaSignatureType = EcdsaSignatureType.ASN1_DER,
+): ByteArray
```
ianbotsf: **Nit**: We should improve the KDocs here to document the parameters and the expected formats of the byte arrays used for input/output.

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/hashing/EcdsaJVM.kt
```diff
@@ -28,10 +32,42 @@ public actual fun ecdsaSecp256r1(key: ByteArray, message: ByteArray): ByteArray
...
+
+    // Remove leading zero bytes and pad to 32 bytes
+    val rFixed = r.dropWhile { it == 0.toByte() }.toByteArray()
+    val sFixed = s.dropWhile { it == 0.toByte() }.toByteArray()
+
+    val rPadded = if (rFixed.size < 32) ByteArray(32 - rFixed.size) + rFixed else rFixed
+    val sPadded = if (sFixed.size < 32) ByteArray(32 - sFixed.size) + sFixed else sFixed
+
+    return rPadded + sPadded
+}
```
ianbotsf: **Question**: Where did we find this algorithm? Can we link to a reliable source?

ianbotsf (changes requested):
**Correctness**: Please add/update tests which verify actual values.

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/hashing/EcdsaJVM.kt
xinsong-cui: I don't really find any available algorithm, I find this is helpful: [the structure of DER](https://bitcoin.stackexchange.com/questions/12554/why-the-signature-is-always-65-13232-bytes-long/12556#12556)  (which tells where is r and s located in DER)

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/hashing/Ecdsa.kt
```diff
@@ -5,6 +5,41 @@
...
  */
 public expect fun ecdsaSecp256r1(key: ByteArray, message: ByteArray): ByteArray
+
+/**
+ * ECDSA on the SECP256R1 curve returning raw r||s format.
+ */
+public fun ecdsaSecp256r1Rs(key: ByteArray, message: ByteArray): ByteArray {
+    val derSignature = ecdsaSecp256r1(key, message)
+    return parseDerSignature(derSignature)
+}
```
lauzadis, line 18: Can we create some unit tests for this?

lauzadis (changes requested):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/hashing/Ecdsa.kt
xinsong-cui: Yes, tests added.
Note: I have to disable signature verification test on JVM8 since ```SHA256withECDSAinP1363Format``` is not supported

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/hashing/EcdsaJVMTest.kt
```diff
@@ -81,4 +100,23 @@ class EcdsaJVMTest {
 
         assertTrue(verifier.verify(signature))
     }
+
+    @Test
+    @EnabledForJreRange(min = JRE.JAVA_11)
```
lauzadis, line 106: Can you add a comment explaining why this is necessary?

PR #1551, opened 2026-04-06
feat: multiplatform file and dir apis

0marperez:
## Issue \#

N/A

## Description of changes

New multiplatform file and dir apis that were discussed internally.

⚠️ Old APIs will be deprecated in upcoming minor version bump


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
```diff
@@ -22,21 +32,135 @@ public interface Filesystem {
      * @param path fully qualified path encoded specifically to the target platform's filesystem.
      * @return contents of file or null if error (file does not exist, etc.)
      */
+    // TODO: Deprecate
+    // @Deprecated("Use read() instead", replaceWith = ReplaceWith("read(path, readAll = true)"))
```
lauzadis, line 36: We'll also need to add our `@PlannedRemoval` annotation

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
```diff
@@ -22,21 +32,135 @@ public interface Filesystem {
...
+    public fun atomicMove(source: String, destination: String, mustExist: Boolean = true, overwrite: Boolean = false) {
+        val sourcePath = Path(source)
+        val destinationPath = Path(destination)
+
+        if (!SystemFileSystem.exists(sourcePath) && mustExist) {
+            throw FileNotFoundException("$sourcePath does not exist and mustExist is set to true")
+        }
+
+        if (SystemFileSystem.exists(destinationPath) && !overwrite) {
+            throw IOException("$destinationPath already exists and overwrite is set to false")
```
lauzadis: I believe you're actually throwing `kotlinx.io.IOException` here, let's wrap all 3P exception types

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
```diff
@@ -22,21 +32,135 @@ public interface Filesystem {
...
+    // TODO: Deprecate
+    // @Deprecated("Use write() instead", replaceWith = ReplaceWith("write(path, data, WriteType.OVERWRITE)"))
     public suspend fun writeFile(path: String, data: ByteArray)
 
     /**
      * Check if a file exists at the [path].
      * @param path fully qualified path encoded specifically to the target platform's filesystem
      */
+    // TODO: Deprecate
+    // @Deprecated("Use kotlinx.io.files.SystemFileSystem.exists(Path(path)) instead")
```
lauzadis: Are we not providing our own wrapped implementation of `exists` anymore?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
```diff
@@ -22,21 +32,135 @@ public interface Filesystem {
      * @param path fully qualified path encoded specifically to the target platform's filesystem.
      * @return contents of file or null if error (file does not exist, etc.)
      */
+    // TODO: Deprecate
```
lauzadis, line 35: Will you ship these deprecations in a follow-up PR to the minor version branch

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
lauzadis, line 21: Should we add `@InternalApi` to discourage customers from using this file system API directly?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
```diff
@@ -22,21 +32,135 @@ public interface Filesystem {
...
+        val path = path.toPath()
+        return OkioFileSystem.SYSTEM.source(path).buffer().use {
+            if (readAll) {
+                it.readByteArray()
+            } else {
+                it.skip(offset)
+                it.readByteArray(amount)
+            }
+        }
+    }
```
ianbotsf, line 190: **Correctness**: This gives callers who specify `mustExist = false` no way to differentiate between an empty file or a missing file.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
```diff
@@ -22,21 +32,135 @@ public interface Filesystem {
...
+
+    /**
+     * Atomically move a file from [source] to [destination].
+     *
+     * @param source fully qualified path of the file to move
+     * @param destination fully qualified path of the target location
+     * @param mustExist if true, throws [FileNotFoundException] when [source] does not exist
+     * @param overwrite if false, throws [IOException] when [destination] already exists
+     */
+    public fun atomicMove(source: String, destination: String, mustExist: Boolean = true, overwrite: Boolean = false) {
```
ianbotsf, line 78: **Nit**: We shouldn't throw `kotlinx.io.IOException` from one method and `aws.smithy.kotlin.runtime.io.IOException` from another. Let's stay consistent.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
```diff
@@ -22,21 +32,135 @@ public interface Filesystem {
...
+            throw FileNotFoundException("$sourcePath does not exist and mustExist is set to true")
+        }
+
+        if (SystemFileSystem.exists(destinationPath) && !overwrite) {
+            throw IOException("$destinationPath already exists and overwrite is set to false")
+        }
+
+        SystemFileSystem.atomicMove(sourcePath, destinationPath)
+    }
+
```
ianbotsf, line 89: **Correctness**: This is not a safe default implementation. No-op/test filesystems should not touch `SystemFileSystem` but currently `MapFilesystem` will do just that because it's not required to override `atomicMove`. Applies to other methods in this interface too.

ianbotsf (changes requested):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
0marperez: I'd be in favor of that. I'm not sure why this is public to begin with. Maybe it's because it's used elsewhere in our projects. Also it would be a breaking change I believe so I can add it in the follow up PR.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
0marperez: Yep, I'll do the deprecations in a separate PR

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
0marperez: Right, I'll add those in the follow up PR

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
0marperez: I was planning on replacing this with `SystemFileSystem.exists(path)` in the follow up

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
0marperez: I'm confused here. Should the function do something else if the file is empty? Do you want to add parameters like `mustHaveData` e.g.?

Empty files still show as existing.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
0marperez: Good catch, any concerns with fixing this by making `atomicMove` and `exists` part of the interface?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
lauzadis: Ok, so this function won't be deprecated, we'll just delegate to SystemFileSystem?

0marperez: It will be deprecated, and replaced with `SystemFileSystem.exists`

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
ianbotsf: I'm saying that reading a file and finding it empty is different from attempting to read a file and finding it missing. There are many times the _presence_ of a file is significant (e.g., `__init.py__` is an empty file that declares a Python module, `.gitkeep` is an empty file that indicates an empty directory should still be tracked by Git, `*.lock` is a common pattern for lock files which are often empty, etc.).

Now, in situations where we _need_ the file to exist we have `mustExist = true` which throws an exception as expected. That's great! I'm worried about the `mustExist = false` case. The function always returns a byte array (vs `null`) so a caller has no way to determine if a file _actually_ exists and is empty or does not exist. This could have implications for tests which try to read a local file to see if it was created successfully, uploads via S3TM which may send an empty file instead of no file or throwing an error, etc.

I think we should either:
* Modify the function to return `ByteArray?` and use `null` to indicate a missing file. Callers that wish to treat it the same as an empty file can use `?: byteArrayOf()`. I don't love this option since, in the `mustExist = true` case, the function should never return `null` but the compiler cannot enforce that for us.
* Add a new `readOrNull` function which returns `ByteArray?`. Move the `mustExist` parameter out of `read` and into `readOrNull`. Then the `read` function _always_ assumes a file must exist and can safely return `ByteArray` (non-null). I like this better but it's a more complex API.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
ianbotsf: No concerns. I think that's the right call.

0marperez:
Using https://github.com/smithy-lang/smithy-kotlin/pull/1561 instead

PR #1553, opened 2026-04-10
Replace junit with kotlin test

luigi617:
### Description

Replaces JUnit-specific APIs with kotlin.test and cross-platform utilities in `runtime:testing`, as part of the effort to standardize on `kotlin-test` across our repositories.

#### New cross-platform utilities in runtime:testing

- **parameterized()** — cross-platform alternative to JUnit's `@ParameterizedTest`. Runs a test block for each case, collects all failures, and reports them at the end with case-level detail.
- **withTempDir()** — cross-platform alternative to JUnit's `@TempDir`. Creates a temporary directory using `kotlinx-io`, runs a block, and cleans up based on `TempDirCleanupMode` (ALWAYS, ON_SUCCESS, NEVER). Default is ON_SUCCESS (matches JUnit's `CleanupMode.ON_SUCCESS`).
- Added `kotlinx-io-core` dependency for cross-platform filesystem operations.

#### JUnit removals

- **@ParameterizedTest / @MethodSource** — Tests converted to use `parameterized()`.
- **@ParameterizedTest / @MethodSource / assumeTrue** — Provider-specific tests (canonical request, signature, string-to-sign) moved from the base class into `DefaultSigningSuiteTest` directly, eliminating the need for assumeTrue to skip tests when providers are null in `CrtSigningSuiteTest`.
- **@RegisterExtension** — Replaced JUnit's OpenTelemetryExtension with manual InMemoryMetricReader + SdkMeterProvider + OpenTelemetrySdk setup. Metrics are reset via `@BeforeTest`.
- **@EnabledForJreRange** — Removed from EcdsaJVMTest. Replaced with a runtime check that returns early if `SHA256withECDSAinP1363Format` is unavailable (Java 8). Also fixed the comment: this algorithm was 
introduced in Java 9, not Java 11.
- **@TestInstance(PER_CLASS)** — Removed from DokkaSmithyTest (unnecessary, no shared state).

#### Kept as-is (documented)

- **@EnabledIfSystemProperty** in ProxyTest.kt — Required for class-level skip of Docker-dependent proxy tests. Gradle sets the property based on environment (not CodeBuild + Linux). Replacing would require Gradle test filtering or runtime guards in @BeforeAll.

### DIFF
--- a/runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/hashing/EcdsaJVMTest.kt
+++ b/runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/hashing/EcdsaJVMTest.kt
@@ -4,8 +4,6 @@
  */
 package aws.smithy.kotlin.runtime.hashing
 
-import org.junit.jupiter.api.condition.EnabledForJreRange
-import org.junit.jupiter.api.condition.JRE
 import java.security.*
 import java.security.interfaces.*
 import java.security.spec.*
@@ -102,9 +100,10 @@ class EcdsaJVMTest {
     }
 
     @Test
-    // SHA256withECDSAinP1363Format algorithm was introduced in Java 11, skip on Java 8
-    @EnabledForJreRange(min = JRE.JAVA_11)
+    // SHA256withECDSAinP1363Format algorithm was introduced in Java 9, skip on Java 8
     fun testVerifyRawSignature() {
+        if (runCatching { Signature.getInstance("SHA256withECDSAinP1363Format") }.isFailure) return
+
         val keyGen = KeyPairGenerator.getInstance("EC")
         keyGen.initialize(ECGenParameterSpec("secp256r1"))
         val keyPair = keyGen.generateKeyPair()
--- a/runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/hashing/EcdsaJVMTest.kt
+++ b/runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/hashing/EcdsaJVMTest.kt
@@ -99,8 +99,8 @@ class EcdsaJVMTest {
         assertTrue(verifier.verify(signature))
     }
 
-    @Test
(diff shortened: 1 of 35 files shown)


runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/hashing/EcdsaJVMTest.kt
```diff
@@ -101,10 +99,11 @@ class EcdsaJVMTest {
         assertTrue(verifier.verify(signature))
     }
 
+    // SHA256withECDSAinP1363Format algorithm was introduced in Java 9, skip on Java 8
     @Test
-    // SHA256withECDSAinP1363Format algorithm was introduced in Java 11, skip on Java 8
-    @EnabledForJreRange(min = JRE.JAVA_11)
     fun testVerifyRawSignature() {
+        if (runCatching { Signature.getInstance("SHA256withECDSAinP1363Format") }.isFailure) return
```
ianbotsf: **Nit**: There's no reason to get the signature instance twice. You should capture it here so it can be used later:

```kotlin
val verifier = runCatching { Signature.getInstance("SHA256withECDSAinP1363Format") }.getOrNull() ?: return
```

runtime/testing/common/src/aws/smithy/kotlin/runtime/testing/Parameterized.kt
```diff
@@ -0,0 +1,62 @@
...
+ *     listOf(
+ *         Triple("String", "\"foo\"", "\"foo\""),
+ *         Triple("Boolean", "false", "false"),
+ *     ),
+ * ) { (type, modelDefault, expectedDefault) ->
+ *     // ...
+ * }
+ * ```
+ */
+public fun <T> parameterized(cases: List<T>, test: (T) -> Unit) {
```
ianbotsf, line 48: **Comment**: Very nice KDocs!

ianbotsf (approved):
Minor feedback only. Fix and ship!

PR #1572, opened 2026-04-23
Retry standard strategy

luigi617:

## Issue \#

## Description of changes
###  Exponential backoff                                                                                                                                                                                                                                                                                                                                                     

  - New delay provider with service-aware base delays: 1000ms for throttling errors, 25ms for DynamoDB/DynamoDB Streams, 50ms for all others
  - Scale factor of 2x with `MAX_BACKOFF` of 20s applied before jitter
  - Introduced `RetryAwareDelayProvider` interface that accepts error type, service name, and retry-after duration

###  `x-amz-retry-after` header

  - Parses the header as integer milliseconds from HTTP responses
  - Clamps the value to `[t_i, t_i + 5s]` where `t_i` is the computed exponential backoff; `MAX_BACKOFF` does not apply
  - Invalid values are ignored with DEBUG-level logging

###  Retry quotas

  - Updated default constants: `RETRY_COST`=14, `THROTTLING_RETRY_COST`=5, `NO_RETRY_INCREMENT`=1, `INITIAL_RETRY_TOKENS`=500
  - Only throttling errors use `THROTTLING_RETRY_COST`; all other error types (including transient) use `RETRY_COST`

###  Service-specific behavior

  - DynamoDB and DynamoDB Streams default to `maxAttempts=4` (auto-detected from service name); all others default to 3                                                                                                                                                                                                                                                     
  - Explicit `maxAttempts` configuration overrides auto-detection

###  Tests

  - YAML-driven integration test runner covering all 15 standard mode test cases: single-invocation, multi-invocation (quota recovery across sequential calls), and multi-threaded (shared token bucket across concurrent calls)
  - Feature flag tests verifying correct types are wired based on flag state
  - Unit tests for backoff scaling, retry-after clamping, and token bucket cost behavior

###  Not included (deferred)

  - Long-polling backoff (`IsLongPollingOperation`)
  - Smithy retryable trait integration
  - AWS_MAX_ATTEMPTS / AWS_RETRY_MODE env var support 


runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/middleware/RetryMiddleware.kt
```diff
@@ -44,7 +44,7 @@ internal class RetryMiddleware<I, O>(
         val result = if (modified.subject.isRetryable) {
             // FIXME this is the wrong span/context because we want the fresh one from inside each attempt but there's no way to
             // wire that through without changing the `RetryPolicy` interface
-            val wrappedPolicy = PolicyLogger(policy, coroutineContext)
+            val wrappedPolicy = PolicyLogger(policy, currentCoroutineContext())
```
0marperez, line 47: What is the difference between `currentCoroutineContext()` and `coroutineContext`?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/StandardRetryStrategy.kt
```diff
@@ -252,8 +282,8 @@ public open class StandardRetryStrategy(override val config: Config = Config.def
              * Configure a new exponential backoff delayer
              * @param block A DSL block which sets the parameters for the exponential backoff delayer
              */
-            public fun delayProvider(block: ExponentialBackoffWithJitter.Config.Builder.() -> Unit) {
-                delayProviderProperty.dsl(ExponentialBackoffWithJitter, block)
+            public fun delayProvider(block: StandardExponentialBackoffWithJitter.Config.Builder.() -> Unit) {
```
0marperez: Why was this change necessary?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/StandardExponentialBackoffWithJitter.kt
```diff
@@ -0,0 +1,146 @@
...
+     * the [errorType] and [serviceName]. If [retryAfterMillis] is provided, the delay is clamped to
+     * `[t_i, t_i + 5000ms]` where `t_i` is the computed exponential backoff. `MAX_BACKOFF` does not apply to this value.
+     */
+    override suspend fun backoff(
+        attempt: Int,
+        errorType: RetryErrorType,
+        serviceName: String?,
+        retryAfterMillis: Long?,
+    ) {
+        require(attempt > 0) { "attempt was $attempt but must be greater than 0" }
```
0marperez, line 57: Why is this require needed?

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/retries/NewRetriesFeatureFlagTest.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    @Test
+    fun testFlagOnUsesStandardDelayProvider() {
+        System.setProperty(sysPropKey, "true")
+        val strategy = StandardRetryStrategy()
+        assertIs<StandardExponentialBackoffWithJitter>(strategy.config.delayProvider)
+    }
+
+    @Test
+    fun testFlagOnUsesStandardTokenBucket() {
```
0marperez, line 36: Nit: We don't test what token bucket is used with the flag off.

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries/delay/StandardExponentialBackoffWithJitterTest.kt
```diff
@@ -0,0 +1,144 @@
...
+    }
+}
+
+private suspend fun TestScope.backoffSeries(
+    times: Int,
+    delayer: StandardExponentialBackoffWithJitter,
+    errorType: RetryErrorType,
+    serviceName: String? = null,
+): List<Int> = (1..times)
+    .map { idx -> measure { delayer.backoff(idx, errorType, serviceName) } }
```
0marperez, line 143: Is the `measure` function measuring how long the delayer backs off for?

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/retries/impl/NewStandardRetryIntegrationTestResources.kt
0marperez, line 1: These YAML tests are from the SEP right?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/StandardRetryStrategy.kt
```diff
@@ -269,15 +299,35 @@ public open class StandardRetryStrategy(override val config: Config = Config.def
...
             /**
-             * The maximum number of attempts to make (including the first attempt)
+             * The maximum number of attempts to make (including the first attempt).
              */
             public override var maxAttempts: Int = DEFAULT_MAX_ATTEMPTS
+                set(value) {
+                    field = value
+                    isMaxAttemptsSet = true
+                }
+            internal var isMaxAttemptsSet = false
```
0marperez, line 313: Can you explain this a bit more?

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/middleware/RetryMiddleware.kt
luigi617: It is preferred to use `currentCoroutineContext()` because it does not bring umbiguity. I think funcionally they are the same. See [suggestion from Jetbrain](https://www.jetbrains.com/help/inspectopedia/PreferCurrentCoroutineContextToCoroutineContext.html)

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/retries/impl/NewStandardRetryIntegrationTestResources.kt
luigi617: Yes

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/StandardRetryStrategy.kt
luigi617: the `maxAttempt` has 3 conditions in the following ordering:
1. User set it
2. If unset and the service is DynamoDB, the default value is 4
3. All other cases, the value is 3.

And the way the value are set follows these steps:
1. Builder is created with default value (here service name is unknown)
2. User can sets values
3. Config is built (service name is available).

Now without the `isMaxAttemptsSet`, we can set the correct default value based on the service name in step 3. However, the edge case is if the service is DynamoDB and user wants to set `maxAttempt` to be 3. Without knowing if the `maxAttempt` is set, we can't know if the default value 3 is the value that user wants or the initial default value that we set in step 1.

Alternative is to have maxAttempt a nullable variable, but that would be an API break

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/retries/delay/StandardExponentialBackoffWithJitterTest.kt
luigi617: yes

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/retries/NewRetriesFeatureFlagTest.kt
luigi617: In this test, the flag is actually on with `System.setProperty(sysPropKey, "true")`

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/StandardExponentialBackoffWithJitter.kt
luigi617: 1. It was used before for previous exponential backoff with jitter.
2. This is a public API and we intentionally want attempt to be greater than 0.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/StandardRetryStrategy.kt
luigi617: Otherwise it is always created an `ExponentialBackoffWithJitter`, which only implements `DelayProvider` — not `RetryAwareDelayProvider`. But I found out a case that this might not be convinient. Will change this

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/StandardRetryStrategy.kt
```diff
@@ -29,6 +32,12 @@ public open class StandardRetryStrategy(override val config: Config = Config.def
         public override operator fun invoke(block: Config.Builder.() -> Unit): StandardRetryStrategy = StandardRetryStrategy(Config(Config.Builder().apply(block)))
     }
 
+    /**
+     * The server-specified retry-after duration in milliseconds from the `x-amz-retry-after` response header.
+     * Set by the HTTP retry middleware after each HTTP call, before the delay provider computes backoff.
+     */
+    public var retryAfterMillis: Long? = null
```
ianbotsf, line 39: **Correctness**: By default, retry strategies are instantiated once per client and shared across all client invocations so they draw from the same token bucket. Mutating a property will affect _all_ in progress calls on that client, which is not what we want. Retry strategies must be immutable so they can function in multiple contexts.

Fortunately, the `RetryStrategy.retry` method is `suspend`, which means you have access to the coroutine context which should include the `HttpOperationContext.RetryAfterMillis` for the active response.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/operation/HttpOperationContext.kt
```diff
@@ -78,6 +78,12 @@ public object HttpOperationContext {
      * The name of the default algorithm to be used for computing a checksum of the request.
      */
     public val DefaultChecksumAlgorithm: AttributeKey<String> = AttributeKey("aws.smithy.kotlin#DefaultChecksumAlgorithm")
+
+    /**
+     * The parsed value of the `x-amz-retry-after` response header in milliseconds, if present and valid.
+     * Set by [RetryMiddleware] after each attempt and consumed by the retry strategy's delay provider.
+     */
+    public val RetryAfterMillis: AttributeKey<Long> = AttributeKey("aws.smithy.kotlin#RetryAfterMillis")
```
ianbotsf, line 86: **Style**: Generally, we try to avoid passing around numbers for time values. Prefer [`Duration`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.time/-duration/) instead:

```kotlin
public val RetryAfter: AttributeKey<Duration> = AttributeKey("aws.smithy.kotlin#RetryAfter")
```

Then you can parse the raw value into a number and use [`milliseconds`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.time/-duration/-companion/milliseconds.html) to convert that to duration:

```
val rawMsValue = raw.toLongOrNull()?.takeIf { it >= 0 }
request.context[HttpOperationContext.RetryAfterMillis] = rawMsValue.milliseconds
```

runtime/runtime-core/api/runtime-core.api
```diff
@@ -1973,23 +1984,34 @@ public final class aws/smithy/kotlin/runtime/retries/delay/ExponentialBackoffWit
...
-	public final fun setScaleFactor (D)V
+	public fun getInitialDelay-UwyO8pc ()J
+	public fun getJitter ()D
+	public fun getMaxBackoff-UwyO8pc ()J
+	public fun getScaleFactor ()D
+	public fun setInitialDelay-LRDsOJo (J)V
+	public fun setJitter (D)V
+	public fun setMaxBackoff-LRDsOJo (J)V
+	public fun setScaleFactor (D)V
 }
```
ianbotsf, line 1997: **Correctness**: This change is backwards-incompatible and cannot be shipped in a patch version. We should investigate other ways to change/commonize the config for these two strategies or whether we want to wait until our next minor version bump.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/RetryAwareDelayProvider.kt
```diff
@@ -0,0 +1,38 @@
...
+     * @param retryAfterMillis The optional server-specified retry-after duration in milliseconds from the
+     * `x-amz-retry-after` response header. When present, the delay is clamped to `[t_i, t_i + 5s]` where `t_i` is
+     * the computed exponential backoff. `MAX_BACKOFF` does not apply to this value.
+     */
+    public suspend fun backoff(
+        attempt: Int,
+        errorType: RetryErrorType,
+        serviceName: String? = null,
+        retryAfterMillis: Long? = null,
+    )
```
ianbotsf, line 37: **Question**: Are the error type and service name in the call context? If so, the `suspend` function can already access them without explicit parameters.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/StandardExponentialBackoffWithJitter.kt
```diff
@@ -0,0 +1,146 @@
...
+        }
+        val exp = baseMs * config.scaleFactor.pow(attempt - 1)
+        val capped = min(exp, config.maxBackoff.toDouble(DurationUnit.MILLISECONDS))
+        val jitterProportion = if (config.jitter > 0.0) random.nextDouble(config.jitter) else 0.0
+        val tI = capped * (1.0 - jitterProportion)
+
+        // Clamp retry-after to [t_i, t_i + 5s]. MAX_BACKOFF does not apply.
+        val delayMs = retryAfterMillis?.toDouble()?.coerceIn(tI, tI + 5000.0) ?: tI
+
+        delay(delayMs.toLong().milliseconds)
```
ianbotsf, line 71: **Nit**: Each of these numbers (e.g., `1000.0`, `25.0`, `5000.0`) should be configurable and all time values should be typed as `Duration`.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/StandardRetryTokenBucket.kt
ianbotsf, line 117: **Correctness**: This is a behavioral change in retry policy, which this update seeks to avoid outside of the opt-in case. We must not change how `Transient` exceptions are handled in the existing retry logic.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/RetrySettings.kt
```diff
@@ -0,0 +1,18 @@
...
+import aws.smithy.kotlin.runtime.config.resolve
+
+private val NEW_RETRIES_SETTING = boolEnvSetting("smithy.newRetries2026", "SMITHY_NEW_RETRIES_2026").orElse(false)
+
+/**
+ * Returns `true` when the `SMITHY_NEW_RETRIES_2026` environment variable (or `smithy.newRetries2026` system property)
+ * is set to `true`, enabling the standard retry strategy behavior. Defaults to `false`.
+ */
+@InternalApi
+public fun newRetriesEnabled(): Boolean = NEW_RETRIES_SETTING.resolve() ?: false
```
ianbotsf, line 18: **Style**: We generally prefer to put settings and configs in container objects for namespacing (e.g., [`ClientSettings`](https://github.com/smithy-lang/smithy-kotlin/blob/b1c9276e85d8d2d9a4c801f57eb1913a73a9d316/runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/config/ClientSettings.kt#L14-L38). Also, there's no need to hide the `val` behind `fun`—just make the value `public`:

```kotlin
@InternalApi
public object CoreSettings {
    public val NewRetriesEnabled = boolEnvSetting("smithy.newRetries2026", "SMITHY_NEW_RETRIES_2026")
        .orElse(false)
        .resolve()
        ?: false
}
```

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/StandardRetryStrategy.kt
```diff
@@ -16,6 +17,8 @@ import aws.smithy.kotlin.runtime.util.DslBuilderProperty
 import aws.smithy.kotlin.runtime.util.DslFactory
 import kotlinx.coroutines.CancellationException
 
+private val DYNAMODB_SERVICES = setOf("dynamodb", "dynamodb streams")
+
```
ianbotsf, line 21: **Correctness**: We should avoid AWS-specific service logic in **smithy-kotlin**. We should specify the higher retry values in **aws-sdk-kotlin** instead.

Rather than tightly coupling service logic to the retry strategy, it may be better to use a codegen customization to change the default retry parameters on these two service clients.

ianbotsf (changes requested):

PR #1577, opened 2026-04-24
Retry standard strategy

luigi617:

## Issue \#

## Description of changes
###  Exponential backoff                                                                                                                                                                                                                                                                                                                                                     

  - New delay provider with service-aware base delays: 1000ms for throttling errors, 25ms for DynamoDB/DynamoDB Streams, 50ms for all others
  - Scale factor of 2x with `MAX_BACKOFF` of 20s applied before jitter
  - Introduced `RetryAwareDelayProvider` interface that accepts error type, service name, and retry-after duration

###  `x-amz-retry-after` header

  - Parses the header as integer milliseconds from HTTP responses
  - Clamps the value to `[t_i, t_i + 5s]` where `t_i` is the computed exponential backoff; `MAX_BACKOFF` does not apply
  - Invalid values are ignored with DEBUG-level logging

###  Retry quotas

  - Updated default constants: `RETRY_COST`=14, `THROTTLING_RETRY_COST`=5, `NO_RETRY_INCREMENT`=1, `INITIAL_RETRY_TOKENS`=500
  - Only throttling errors use `THROTTLING_RETRY_COST`; all other error types (including transient) use `RETRY_COST`

###  Service-specific behavior

  - DynamoDB and DynamoDB Streams default to `maxAttempts=4` (auto-detected from service name); all others default to 3                                                                                                                                                                                                                                                     
  - Explicit `maxAttempts` configuration overrides auto-detection

###  Tests

  - YAML-driven integration test runner covering all 15 standard mode test cases: single-invocation, multi-invocation (quota recovery across sequential calls), and multi-threaded (shared token bucket across concurrent calls)
  - Feature flag tests verifying correct types are wired based on flag state
  - Unit tests for backoff scaling, retry-after clamping, and token bucket cost behavior

###  Not included (deferred)

  - Long-polling backoff (`IsLongPollingOperation`)
  - Smithy retryable trait integration
  - AWS_MAX_ATTEMPTS / AWS_RETRY_MODE env var support 


runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/middleware/RetryMiddleware.kt
```diff
@@ -81,6 +91,9 @@ internal class RetryMiddleware<I, O>(
         next: Handler<SdkHttpRequest, O>,
         attempt: Int,
     ): Result<O> {
+        // Clear any previous retry-after value
+        request.context.remove(HttpOperationContext.RetryAfter)
+
```
ianbotsf: **Nit**: Do we _need_ to remove this? Every attempt should be operating on a fresh copy of the original request (see lines 43 & 66) and the original request cannot have `HttpOperationContext.RetryAfter` set because there wouldn't have been any attempts yet.

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/middleware/RetryMiddleware.kt
```diff
@@ -93,6 +106,16 @@ internal class RetryMiddleware<I, O>(
...
 
+        // Parse x-amz-retry-after header (integer milliseconds). Invalid values are ignored.
+        call?.response?.headers?.get("x-amz-retry-after")?.let { raw ->
+            val ms = raw.toLongOrNull()?.takeIf { it >= 0 }
+            if (ms != null) {
+                request.context[HttpOperationContext.RetryAfter] = ms.milliseconds
+            } else {
+                currentCoroutineContext().debug<RetryMiddleware<*, *>> { "ignoring invalid x-amz-retry-after header: $raw" }
+            }
+        }
```
ianbotsf, line 114: **Nit**: The "ignoring invalid" message is emitted even when there's _no_ `x-amz-retry-after` header in the response. I think we don't want a log message in that case.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/ExponentialBackoffWithJitter.kt
```diff
@@ -17,18 +21,23 @@ import kotlin.time.Duration.Companion.seconds
...
+ * [Config.initialDelay] as the base.
  *
- * For instance, a jitter
- * configuration of 0.5 means that up to 50% of the max delay time could be reduced. A jitter configuration of 1.0 means
- * that 100% of the max delay time could be reduced (potentially down to 0). A jitter configuration of 0.0 means jitter
- * is disabled.
+ * The delay for a given attempt is calculated as:
+ * ```
+ * delay = random(0, 1) * min(base * scaleFactor^(attempt - 1), maxBackoff)
+ * ```
```
ianbotsf: **Nit**: The calculation uses `random(0, 1)` only if `jitter = 1.0`. I think `random(1 - jitter, 1)` is more correct.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/delay/ExponentialBackoffWithJitter.kt
```diff
@@ -17,18 +21,23 @@ import kotlin.time.Duration.Companion.seconds
...
 /**
- * A [DelayProvider] that implements exponentially increasing delays and jitter (i.e., randomization of delay amount).
- * This delayer calculates a maximum delay time from the initial delay amount, the scale factor, and the attempt number.
- * It then randomly reduces that time down to something less based on the jitter configuration.
+ * A [DelayProvider] that implements exponential backoff with jitter
+ * (i.e., randomization of delay amount).
+ * When a [RetryContext] is present in the coroutine context, the delay provider uses
+ * [RetryContext.errorType] to select the base delay and [RetryContext.retryAfter] to
+ * clamp the computed delay. Without a [RetryContext], the provider falls back to
+ * [Config.initialDelay] as the base.
```
ianbotsf: **Correctness**: A `RetryContext` is _always_ present in the coroutine context since your changes to `RetryMiddleware` are unconditional.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/RetrySettings.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+package aws.smithy.kotlin.runtime.retries
+
+import aws.smithy.kotlin.runtime.InternalApi
+import aws.smithy.kotlin.runtime.config.boolEnvSetting
+import aws.smithy.kotlin.runtime.config.resolve
+
+@InternalApi
+public object CoreSettings {
+    private val newRetriesSetting = boolEnvSetting("smithy.newRetries2026", "SMITHY_NEW_RETRIES_2026").orElse(false)
```
ianbotsf: **Nit**: This object is called `CoreSettings` but it's in the file **RetrySettings.kt**. Generally the name of the file should match the name of the top-level type/object inside of it.

Also, this is in the `aws.smithy.kotlin.runtime.retries` namespace but I think a true `CoreSettings` object would live one level higher (i.e., `aws.smithy.kotlin.runtime`).

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/RetrySettings.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    /**
+     * Whether the standard retry strategy behavior is enabled.
+     * Controlled by the `SMITHY_NEW_RETRIES_2026` environment variable or `smithy.newRetries2026` system property.
+     *
+     * Uses a computed property (`get()`) instead of a stored `val` so the value is re-evaluated on each access,
+     * allowing tests to toggle the system property between test methods within the same JVM.
+     */
+    public val NewRetriesEnabled: Boolean
+        get() = newRetriesSetting.resolve() ?: false
```
ianbotsf: **Style**: I see your comments about unit tests but we should use `PlatformProvider` to inject system properties and environment variables into our tests rather than affecting _real_ system properties. That typically means writing some `resolve` method that accepts the platform provider as an argument (e.g., similar to [`resolveRegion`](https://github.com/aws/aws-sdk-kotlin/blob/f9e5b4919d75a90473544aa19e3a982f9b049eed/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/region/ResolveRegion.kt#L22-L26)).

ianbotsf (changes requested):

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/middleware/RetryMiddleware.kt
luigi617: We can't remove this because HttpOperationContext is not part of the copy, see [here](https://github.com/smithy-lang/smithy-kotlin/blob/a0201da767ac22615f266cf94056042464959d1c/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/operation/OperationRequest.kt#L26-L30). But I think it would be better to use directly `RetryContext` avoiding setting the value to `HttpOperationContext`. Both ways needs to clean the value

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/middleware/RetryMiddleware.kt
luigi617: `call?.response?.headers?.get("x-amz-retry-after")?.let` returns `null` when the header is missing.

ianbotsf: You're correct, I missed the `?` in `?.let`.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/retries/RetryContext.kt
```diff
@@ -0,0 +1,42 @@
...
+import kotlin.coroutines.AbstractCoroutineContextElement
+import kotlin.coroutines.CoroutineContext
+import kotlin.time.Duration
+
+/**
+ * A coroutine context element that carries per-call retry state. Installed by the retry middleware
+ * so that the delay provider can read retry metadata without mutable state on the shared strategy instance.
+ *
+ * Uses a computed property (`get()`) instead of a stored `val` so the value is re-evaluated on each access,
+ * allowing tests to toggle the system property between test methods within the same JVM.
```
ianbotsf: **Nit**: I think this comment is obsolete.

runtime/runtime-core/jvm/test/aws/smithy/kotlin/runtime/retries/NewRetriesFeatureFlagTest.kt
```diff
@@ -0,0 +1,71 @@
...
+import aws.smithy.kotlin.runtime.retries.policy.RetryErrorType
+import aws.smithy.kotlin.runtime.util.TestPlatformProvider
+import kotlinx.coroutines.test.runTest
+import kotlin.test.*
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.seconds
+
+class NewRetriesFeatureFlagTest {
+    private val flagOff = TestPlatformProvider()
+    private val flagOn = TestPlatformProvider(props = mapOf("smithy.newRetries2026" to "true"))
```
ianbotsf, line 20: **Nit**: You should also test the environment variable (e.g., `TestPlatformProvider(env = mapOf("SMITHY_NEW_RETRIES_2026" to "true")`)

.changes/8ff144dd-952c-46e3-b59a-40c8eb673ead.json
```diff
@@ -0,0 +1,5 @@
+{
+  "id": "8ff144dd-952c-46e3-b59a-40c8eb673ead",
+  "type": "feature",
+  "description": "Add standard retry strategy with updated exponential backoff, retry quotas, `x-amz-retry-after` header support, and service-specific behavior for DynamoDB. Gated behind the `SMITHY_NEW_RETRIES_2026` feature flag"
```
ianbotsf: **Nit**: Post the GitHub discussion you've been drafting and link to it from this description. (e.g., `Add [new implementation of standard retry strategy](https://github.com/aws/aws-sdk-kotlin/discussions/0000) with updated...`)

ianbotsf (approved):
Minor feedback only; fix and ship!

PR #1578, opened 2026-04-26
chore: add test filesystem capabilities

ianbotsf:
## Description of changes

This change adds some new capabilities to `TestPlatformProvider` to facilitate verifying permissions on files in unit tests.

**Downstream PR**: https://github.com/aws/aws-sdk-kotlin/pull/1877

### DIFF
--- a/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/TestPlatformProvider.kt
+++ b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/TestPlatformProvider.kt
@@ -6,6 +6,7 @@
 package aws.smithy.kotlin.runtime.util
 
 import aws.smithy.kotlin.runtime.InternalApi
+import kotlin.jvm.JvmName
 
 /**
  * An implementation of [PlatformProvider] meant for testing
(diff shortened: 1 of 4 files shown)


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/TestPlatformProvider.kt
```diff
@@ -15,15 +16,35 @@ import aws.smithy.kotlin.runtime.InternalApi
  * @param os Operating system info to emulate
  */
 @InternalApi
-public class TestPlatformProvider(
+public class TestPlatformProvider
```
luigi617, line 19: Question: is there any reason why this class is not under `test` dir if it is meant for testing?

luigi617 (approved):
Just have a question and looks good to me

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/TestPlatformProvider.kt
ianbotsf: `TestPlatformProvider` is one of our utility classes. It's used in **runtime-core** but also in other modules (e.g., in [**aws-http**](https://github.com/aws/aws-sdk-kotlin/blob/f9e5b4919d75a90473544aa19e3a982f9b049eed/aws-runtime/aws-http/common/test/aws/sdk/kotlin/runtime/http/operation/CustomUserAgentMetadataTest.kt#L18), [**smithy-client**](https://github.com/smithy-lang/smithy-kotlin/blob/a0201da767ac22615f266cf94056042464959d1c/runtime/smithy-client/common/test/aws/smithy/kotlin/runtime/client/config/LogModeClientSettingTest.kt#L23-L25), etc.). Dependencies must be in a **main** source set (e.g., **commonMain**, **jvmMain**, etc.) to be visible to other modules.

An argument could be made this belongs in **:runtime:testing** instead but moving it would be a breaking change so it'll probably stay here for the time being.

PR #1611, opened 2026-06-01
fix: limit the parsing depth of JSON, XML, and CBOR parsers

ianbotsf:
## Issue \#

(none)

## Description of changes

This change introduces a new limit of 1,000 levels of recursion when parsing JSON, XML, and CBOR documents, providing a way to gracefully handle malformed documents which nest too deeply. While this trades one throwable (`StackOverflowError`) for another (`DeserializationRecursionException`), it prevents stack exhaustion that might otherwise bring down the whole runtime and can be caught by calling code while there's still a chance to recover.

### DIFF
--- a/runtime/serde/serde-cbor/common/src/aws/smithy/kotlin/runtime/serde/cbor/encoding/Value.kt
+++ b/runtime/serde/serde-cbor/common/src/aws/smithy/kotlin/runtime/serde/cbor/encoding/Value.kt
@@ -8,6 +8,7 @@ import aws.smithy.kotlin.runtime.io.SdkBuffer
 import aws.smithy.kotlin.runtime.io.SdkBufferedSink
 import aws.smithy.kotlin.runtime.io.SdkBufferedSource
 import aws.smithy.kotlin.runtime.serde.DeserializationException
+import aws.smithy.kotlin.runtime.serde.DeserializationRecursionException
 
 /**
  * Represents an encodable / decodable CBOR value.
@@ -23,8 +24,11 @@ internal interface Value {
         /**
          * Decode a [Value] from the given [buffer]
          * @param buffer the [SdkBufferedSource] to read the next [Value] from
+         * @param depth the current recursion depth
          */
-        fun decode(buffer: SdkBufferedSource): Value {
+        fun decode(buffer: SdkBufferedSource, depth: Int = 0): Value {
+            DeserializationRecursionException.assertDepth(depth)
+
             val major = peekMajor(buffer)
             val minor = peekMinorByte(buffer)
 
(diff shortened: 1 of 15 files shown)


ianbotsf:
ℹ️ This PR has the **acknowledge-api-break** label attached because the API dump has this change:

```diff
- public final class aws/smithy/kotlin/runtime/serde/DeserializationException : aws/smithy/kotlin/runtime/ClientException {
+ public class aws/smithy/kotlin/runtime/serde/DeserializationException : aws/smithy/kotlin/runtime/ClientException {
```

Despite being a red/green change, this _is_ safe from binary- and source-compatibility standpoints. Changing a class from `final` to `open` only affects subclasses, of which there were previously none.

runtime/serde/serde-cbor/common/src/aws/smithy/kotlin/runtime/serde/cbor/encoding/Value.kt
```diff
@@ -33,33 +37,31 @@ internal interface Value {
...
-                        Map.decode(buffer)
-                    }
+
+                Major.MAP -> if (minor == Minor.INDEFINITE.value) {
+                    IndefiniteMap.decode(buffer, depth)
+                } else {
+                    Map.decode(buffer, depth)
                 }
+
                 Major.TAG -> Tag.decode(buffer)
```
luigi617: CBOR tag can also be nested, e.g. `Tag(0, Tag(0, Tag(0, ...))`, and everytime it goes to `.decode`, depth is set to 0 which can cause `StackOverflowError`. Extremely rare case but possible

runtime/serde/serde-cbor/common/test/aws/smithy/kotlin/runtime/serde/cbor/CborDeserializerTest.kt
```diff
@@ -27,4 +30,112 @@ class CborDeserializerTest {
...
+        val p = ByteArray(n + 1)
+        for (i in 0..<n) p[i] = 0x81.toByte() // array(1)
+        p[n] = 0x00 // uint 0
+
+        val buffer = SdkBuffer().apply { write(p) }
+        aws.smithy.kotlin.runtime.serde.cbor.encoding.Value.decode(buffer)
+    }
+
+    @Test
+    fun testFindNextFieldIndexRecursion() {
```
luigi617: Nit: This test is duplicate of `findNextFieldIndexNullSkipDoesNotStackOverflow`

luigi617 (changes requested):

runtime/serde/serde-cbor/common/src/aws/smithy/kotlin/runtime/serde/cbor/encoding/Value.kt
ianbotsf: Good catch! I'll add test cases and fixes for that too.

luigi617 (approved):

PR #1612, opened 2026-06-01
fix: validate OkHttp header values do not contain CR or LF

ianbotsf:
## Issue \#

(none)

## Description of changes

Adds verification and validation of CRLF handling in OkHttp headers. Note that the CRT engine already rejects CRLF values and does not need any changes.

### DIFF
--- a/runtime/protocol/http-client-engines/test-suite/jvmAndNative/src/aws/smithy/kotlin/runtime/http/test/suite/Header.kt
+++ b/runtime/protocol/http-client-engines/test-suite/jvmAndNative/src/aws/smithy/kotlin/runtime/http/test/suite/Header.kt
@@ -16,5 +16,16 @@ internal fun Application.headerTests() {
                 call.respond(HttpStatusCode.OK)
             }
         }
+
+        route("echo-headers") {
+            get {
+                // Echo back all request headers as response headers prefixed with "x-echo-"
+                // and return the count of headers whose name starts with "x-]]" (injection marker)
+                val injectedCount = call.request.headers.names()
+                    .count { it.startsWith("x-injected", ignoreCase = true) }
+                call.response.header("x-injected-count", injectedCount.toString())
+                call.respond(HttpStatusCode.OK)
+            }
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


runtime/protocol/http-client-engines/test-suite/jvmAndNative/src/aws/smithy/kotlin/runtime/http/test/suite/Header.kt
```diff
@@ -16,5 +16,16 @@ internal fun Application.headerTests() {
                 call.respond(HttpStatusCode.OK)
             }
         }
+
+        route("echo-headers") {
+            get {
+                // Echo back all request headers as response headers prefixed with "x-echo-"
+                // and return the count of headers whose name starts with "x-]]" (injection marker)
```
luigi617, line 23: Nit: it should be `x-injected` instead of `x-]]`

luigi617 (approved):
Looks good, just have a question: should this http injection check also be applied to http header key?

ianbotsf:
> Looks good, just have a question: should this http injection check also be applied to http header key?

OkHttp already prevents characters outside the ASCII range 33-127, which would include CR and LF. For reference, note we call [`okHeaders.addUnsafeNonAscii`](https://github.com/smithy-lang/smithy-kotlin/blob/258ab44c341b74245c5280169225cc3f63d93a18/runtime/protocol/http-client-engines/http-client-engine-okhttp/jvm/src/aws/smithy/kotlin/runtime/http/engine/okhttp/OkHttpUtils.kt#L90) which invokes [`headersCheckName`](https://github.com/square/okhttp/blob/0c5a45b1174b8f735cc251eb19faaed87b65d868/okhttp/src/commonJvmAndroid/kotlin/okhttp3/Headers.kt#L257) which [validates the characters](https://github.com/square/okhttp/blob/0c5a45b1174b8f735cc251eb19faaed87b65d868/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/-HeadersCommon.kt#L143-L148).

PR #1613, opened 2026-06-01
fix: limit exponent expansion to prevent OOM

ianbotsf:
## Issue \#

(none)

## Description of changes

This change limits exponent expansion to prevent out-of-memory errors on numerical types.

### DIFF
--- a/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/content/BigDecimal.kt
+++ b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/content/BigDecimal.kt
@@ -108,3 +108,11 @@ public expect class BigDecimal(value: String) :
      */
     public override operator fun compareTo(other: BigDecimal): Int
 }
+
+internal fun assertExponent(value: Int) {
+    require(value in -MAX_DECIMAL_FRACTION_EXPONENT..MAX_DECIMAL_FRACTION_EXPONENT) {
+        "BigDecimal exponent $value exceeds maximum allowed magnitude of $MAX_DECIMAL_FRACTION_EXPONENT"
+    }
+}
+
+internal fun assertExponent(value: Long) = assertExponent(value.toInt())
--- a/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/content/BigDecimal.kt
+++ b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/content/BigDecimal.kt
@@ -115,4 +115,8 @@ internal fun assertExponent(value: Int) {
     }
 }
 
-internal fun assertExponent(value: Long) = assertExponent(value.toInt())
+internal fun assertExponent(value: Long) {
+    require(value in -MAX_DECIMAL_FRACTION_EXPONENT..MAX_DECIMAL_FRACTION_EXPONENT) {
+        "BigDecimal exponent $value exceeds maximum allowed magnitude of $MAX_DECIMAL_FRACTION_EXPONENT"
+    }
+}
--- a/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/content/BigDecimal.kt
+++ b/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/content/BigDecimal.kt
@@ -5,6 +5,12 @@
 package aws.smithy.kotlin.runtime.content
 
 /**
+ * Maximum allowed magnitude for a [BigDecimal] exponent. Exponents beyond this limit would cause
+ * excessive memory allocation when the value is materialized as a plain string.
+ */
+public const val MAX_DECIMAL_FRACTION_EXPONENT: Int = 1000
+
+/**
  * A floating point decimal number with arbitrary precision.
  * @param value the [String] representation of this decimal number
  */
--- a/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigDecimalJVM.kt
+++ b/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigDecimalJVM.kt
@@ -22,10 +22,17 @@ public actual class BigDecimal private constructor(private val delegate: JvmBigD
             right.delegate -> right
             else -> BigDecimal(value)
         }
+
+        fun checkedCreate(mantissa: BigInteger, exponent: Int): JvmBigDecimal {
+            require(exponent in -MAX_DECIMAL_FRACTION_EXPONENT..MAX_DECIMAL_FRACTION_EXPONENT) {
+                "BigDecimal exponent $exponent exceeds maximum allowed magnitude of $MAX_DECIMAL_FRACTION_EXPONENT"
+            }
+            return JvmBigDecimal(mantissa.delegate, exponent)
+        }
     }
 
     public actual constructor(value: String) : this(JvmBigDecimal(value))
-    public actual constructor(mantissa: BigInteger, exponent: Int) : this(JvmBigDecimal(mantissa.delegate, exponent))
+    public actual constructor(mantissa: BigInteger, exponent: Int) : this(checkedCreate(mantissa, exponent))
 
     public actual fun toPlainString(): String = delegate.toPlainString()
     public actual override fun toString(): String = delegate.toString()
--- a/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigDecimalJVM.kt
+++ b/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigDecimalJVM.kt
@@ -22,15 +22,15 @@ public actual class BigDecimal private constructor(private val delegate: JvmBigD
             right.delegate -> right
             else -> BigDecimal(value)
         }
-
-        fun checkedCreate(mantissa: BigInteger, exponent: Int): JvmBigDecimal {
-            assertExponent(exponent)
-            return JvmBigDecimal(mantissa.delegate, exponent)
-        }
     }
 
     public actual constructor(value: String) : this(JvmBigDecimal(value))
-    public actual constructor(mantissa: BigInteger, exponent: Int) : this(checkedCreate(mantissa, exponent))
+    public actual constructor(mantissa: BigInteger, exponent: Int) : this(
+        JvmBigDecimal(
+            mantissa.delegate,
+            mantissa.delegate.abs().toString().length - 1 - exponent,
+        ),
+    )
 
     init {
         assertExponent(delegate.scale())
@@ -45,14 +45,19 @@ public actual class BigDecimal private constructor(private val delegate: JvmBigD
     public actual override fun toLong(): Long = delegate.toLong()
     public actual override fun toShort(): Short = delegate.toShort()
 
-    public actual override fun equals(other: Any?): Boolean = other is BigDecimal && delegate == other.delegate
+    public actual override fun equals(other: Any?): Boolean = other is BigDecimal && (delegate.compareTo(other.delegate) == 0)
+
     public actual override fun hashCode(): Int = 31 + delegate.hashCode()
 
+    private val stripped: JvmBigDecimal by lazy {
+        if (delegate.signum() == 0) JvmBigDecimal.ZERO else delegate.stripTrailingZeros()
+    }
+
     public actual val mantissa: BigInteger
-        get() = BigInteger(delegate.unscaledValue())
+        get() = BigInteger(stripped.unscaledValue())
 
     public actual val exponent: Int
-        get() = delegate.scale()
+        get() = if (delegate.signum() == 0) 0 else stripped.unscaledValue().abs().toString().length - 1 - stripped.scale()
 
     public val value: String = delegate.toString()
 
@@ -61,10 +66,4 @@ public actual class BigDecimal private constructor(private val delegate: JvmBigD
     public actual operator fun minus(other: BigDecimal): BigDecimal = coalesceOrCreate(delegate - other.delegate, this, other)
 
     actual override fun compareTo(other: BigDecimal): Int = delegate.compareTo(other.delegate)
-}
-
-private fun assertExponent(value: Int) {
-    require(value in -MAX_DECIMAL_FRACTION_EXPONENT..MAX_DECIMAL_FRACTION_EXPONENT) {
-        "BigDecimal exponent $value exceeds maximum allowed magnitude of $MAX_DECIMAL_FRACTION_EXPONENT"
-    }
 }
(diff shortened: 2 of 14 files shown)


runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigDecimalJVM.kt
```diff
@@ -22,10 +22,17 @@ public actual class BigDecimal private constructor(private val delegate: JvmBigD
...
+
+        fun checkedCreate(mantissa: BigInteger, exponent: Int): JvmBigDecimal {
+            require(exponent in -MAX_DECIMAL_FRACTION_EXPONENT..MAX_DECIMAL_FRACTION_EXPONENT) {
+                "BigDecimal exponent $exponent exceeds maximum allowed magnitude of $MAX_DECIMAL_FRACTION_EXPONENT"
+            }
+            return JvmBigDecimal(mantissa.delegate, exponent)
+        }
     }
 
     public actual constructor(value: String) : this(JvmBigDecimal(value))
```
luigi617, line 27: In `constructor(mantissa: BigInteger, exponent: Int)`, the creation is checked, but in `constructor(value: String)`, user can also pass a value that cause OOM. Is this situation prevented from other places that I am missing?

runtime/serde/serde-cbor/common/test/aws/smithy/kotlin/runtime/serde/cbor/CborDeserializerErrorTest.kt
```diff
@@ -766,4 +767,24 @@ class CborDeserializerErrorTest {
...
+     * `BigDecimal`.
+     */
+    @Test
+    fun testDecimalFractionExponentDoesNotOom() {
+        val payload = "0xc4821a3b9ac9ff01".toByteArray()
+
+        val buffer = SdkBuffer().apply { write(payload) }
+        val deserializer = CborPrimitiveDeserializer(buffer)
+
+        assertFailsWith<Exception> {
```
luigi617: I think the use of `assertFailsWith<Exception>` is too broad, if we expect some exception, we should specify the exact exception to understand the correct behaviour, e.g. `assertFailsWith<SpecificException>`

runtime/serde/serde-json/common/test/aws/smithy/kotlin/runtime/serde/json/JsonDeserializerTest.kt
```diff
@@ -820,4 +821,16 @@ class JsonDeserializerTest {
...
     }
+
+    /**
+     * tt/V2228436368/F20: "1e999999999" in an epoch-seconds field should not be expanded and cause OOM
+     */
+    @Test
+    fun testEpochExponentDoesNotOOM() {
+        val payload = "\"1e999999999\"".encodeToByteArray()
+        val deserializer = JsonDeserializer(payload)
+        assertFailsWith<Exception> {
```
luigi617: See previous comment

runtime/serde/serde-xml/common/test/aws/smithy/kotlin/runtime/serde/xml/XmlTagReaderTest.kt
```diff
@@ -134,6 +137,20 @@ class XmlTagReaderTest {
...
+    /**
+     * tt/V2228436368/F20: "1e999999999" in an epoch-seconds XML text node should not be expanded and cause OOM
+     */
+    @Test
+    fun testEpochExponentDoesNotOOM() {
+        // F20: "1e999999999" in an epoch-seconds XML text node triggers expandExponent() → padEnd(999999999, '0') → OOM
+        val payload = "<Timestamp>1e999999999</Timestamp>".encodeToByteArray()
+        val fieldDescriptor = SdkFieldDescriptor(SerialKind.Timestamp, XmlSerialName("Timestamp"))
+        val deserializer = XmlPrimitiveDeserializer(payload, fieldDescriptor)
+        assertFailsWith<Exception> {
```
luigi617: See previous comment

luigi617 (changes requested):

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigDecimalJVM.kt
ianbotsf: Nope, that was a miss. Adding!

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigDecimalJVM.kt
```diff
@@ -22,10 +22,19 @@ public actual class BigDecimal private constructor(private val delegate: JvmBigD
...
-    public actual constructor(mantissa: BigInteger, exponent: Int) : this(JvmBigDecimal(mantissa.delegate, exponent))
+    public actual constructor(mantissa: BigInteger, exponent: Int) : this(checkedCreate(mantissa, exponent))
+
+    init {
+        assertExponent(delegate.scale())
```
luigi617, line 36: This also applies to `constructor(mantissa: BigInteger, exponent: Int) ` which becomes redundant, I suggest:
1) rely only on `init {}` block, but the cons is `JvmBigDecimal` will be constracted before rejecting it.
2) use a similar approach to checkedCreate for `constructor(value: String)`, but I think it would be more complex and overengineer for this class.
3) accept the redundancy, but in the future we might be confused of why we have this redundant check.

I would suggest 1.

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/content/BigDecimalNative.kt
```diff
@@ -22,12 +22,18 @@ public actual class BigDecimal private constructor(private val delegate: IonSpin
...
+
+        fun checkedCreate(mantissa: BigInteger, exponent: Int): IonSpinBigDecimal {
+            require(exponent in -MAX_DECIMAL_FRACTION_EXPONENT..MAX_DECIMAL_FRACTION_EXPONENT) {
+                "BigDecimal exponent $exponent exceeds maximum allowed magnitude of $MAX_DECIMAL_FRACTION_EXPONENT"
+            }
+            return IonSpinBigDecimal.fromBigIntegerWithExponent(mantissa.delegate, exponent.toLong())
+        }
     }
 
     public actual constructor(value: String) : this(IonSpinBigDecimal.parseString(value, 10))
```
luigi617: The same solution should be applied also in native implementation. Sorry for not pointing out in the first review

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonDeserializer.kt
```diff
@@ -144,11 +145,15 @@ public class JsonDeserializer(payload: ByteArray) :
...
-        else -> throw DeserializationException("unknown timestamp format: $format")
+    override fun deserializeInstant(format: TimestampFormat): Instant = try {
+        when (format) {
+            TimestampFormat.EPOCH_SECONDS -> deserializeString().let { Instant.fromEpochSeconds(it) }
+            TimestampFormat.ISO_8601 -> deserializeString().let { Instant.fromIso8601(it) }
+            TimestampFormat.RFC_5322 -> deserializeString().let { Instant.fromRfc5322(it) }
+            else -> throw DeserializationException("unknown timestamp format: $format")
+        }
+    } catch (pe: ParseException) {
+        throw DeserializationException("Cannot deserialize unsupported timestamp format", pe)
```
luigi617: I think the error message is misleading if the error is not coming from format but from limit exponent expansion. I suggest:
`throw DeserializationException("Cannot deserialize timestamp: ${pe.message}", pe)`
or
`throw DeserializationException("Cannot deserialize timestamp value", pe)`

luigi617 (changes requested):

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigDecimalJVM.kt
ianbotsf: Went with option 1. 👍

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/content/BigDecimalNative.kt
ianbotsf: You're right and it turns out there were some serious bugs with the `BigDecimal` implementations for both JVM and Native, which are also fixed in this PR.

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/content/BigDecimalNative.kt
```diff
@@ -43,13 +56,13 @@ public actual class BigDecimal private constructor(private val delegate: IonSpin
...
+        get() = BigInteger(normalized.significand)
 
     public actual val exponent: Int
-        get() = delegate.exponent.toInt()
+        get() = normalized.exponent.toInt()
 
     actual override fun compareTo(other: BigDecimal): Int = delegate.compare(other.delegate)
-    actual override fun equals(other: Any?): Boolean = other is BigDecimal && other.delegate == delegate
+    public actual override fun equals(other: Any?): Boolean = other is BigDecimal && (delegate.compareTo(other.delegate) == 0)
     actual override fun hashCode(): Int = 31 + delegate.hashCode()
```
luigi617: Correctness: If two objects are equal according to `equals()`, they must produce the identical `hashCode()`. Currently `BigDecimal("1.0") == BigDecimal("1.00")`, but `BigDecimal("1.0").hashCode() != BigDecimal("1.00").hashCode()`.  I would suggest:
```
public actual override fun hashCode(): Int = normalized.hashCode()
```
or
```
public actual override fun hashCode(): Int {
      val n = normalized 
      return 31 * n.significand.hashCode() + n.exponent.hashCode()
  } 
```

runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/content/BigDecimalJVM.kt
```diff
@@ -45,14 +45,19 @@ public actual class BigDecimal private constructor(private val delegate: JvmBigD
     public actual override fun toLong(): Long = delegate.toLong()
     public actual override fun toShort(): Short = delegate.toShort()
 
-    public actual override fun equals(other: Any?): Boolean = other is BigDecimal && delegate == other.delegate
+    public actual override fun equals(other: Any?): Boolean = other is BigDecimal && (delegate.compareTo(other.delegate) == 0)
+
     public actual override fun hashCode(): Int = 31 + delegate.hashCode()
```
luigi617: Correctness: If two objects are equal according to `equals()`, they must produce the identical `hashCode()`. Currently `BigDecimal("1.0") == BigDecimal("1.00")`, but `BigDecimal("1.0").hashCode() != BigDecimal("1.00").hashCode()`.  I would suggest:
```
public actual override fun hashCode(): Int = stripped.hashCode()
```
or
```
public actual override fun hashCode(): Int {
      val n = stripped 
      return 31 * n.significand.hashCode() + n.exponent.hashCode()
  } 
```

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/content/BigDecimal.kt
```diff
@@ -108,3 +108,11 @@ public expect class BigDecimal(value: String) :
...
     public override operator fun compareTo(other: BigDecimal): Int
 }
+
+internal fun assertExponent(value: Int) {
+    require(value in -MAX_DECIMAL_FRACTION_EXPONENT..MAX_DECIMAL_FRACTION_EXPONENT) {
+        "BigDecimal exponent $value exceeds maximum allowed magnitude of $MAX_DECIMAL_FRACTION_EXPONENT"
+    }
+}
+
+internal fun assertExponent(value: Long) = assertExponent(value.toInt())
```
luigi617: Correctness: `Long.toInt()` silently wraps on overflow. A malicious exponent like 4294967296L (2^32) truncates to 0. I suggest:
```
internal fun assertExponent(value: Long) {
    require(value in -MAX_DECIMAL_FRACTION_EXPONENT.toLong()..MAX_DECIMAL_FRACTION_EXPONENT.toLong()) {
        "BigDecimal exponent $value exceeds maximum allowed magnitude of $MAX_DECIMAL_FRACTION_EXPONENT"
    }
}

internal fun assertExponent(value: Int) = assertExponent(value.toLong())
```

runtime/serde/serde-cbor/common/test/aws/smithy/kotlin/runtime/serde/cbor/CborDeserializerErrorTest.kt
```diff
@@ -777,15 +777,24 @@ class CborDeserializerErrorTest {
...
-            // If deserialization doesn't reject it, toPlainString() will OOM
-            result.toPlainString()
+        val payloads = listOf(
+            "0xc48219270f01", // 1E9999
+            "0xc4821a0001869f01", // 1E99999
+            "0xc4821a000f423f01", // 1E999999
+            "0xc4821a0098967f01", // 1E9999999
+            "0xc4821a05f5e0ff01", // 1E99999999
+            "0xc4821a3b9ac9ff01", // 1E999999999
+        ).map { it.toByteArray() }
```
luigi617, line 787: Question: `toByteArray()` treat the input as plain text, is it more accurate to write:
```
val payloads = listOf(
    "c48219270f01", // 1E9999
    "c4821a0001869f01", // 1E99999
    "c4821a000f423f01", // 1E999999
    "c4821a0098967f01", // 1E9999999
    "c4821a05f5e0ff01", // 1E99999999
    "c4821a3b9ac9ff01", // 1E999999999
).map { it.hexToByteArray() }
```

ianbotsf: I can see how this is confusing. In this case `toByteArray()` is _not_ [stdlib's `String.toByteArray()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/to-byte-array.html), which does indeed treat the input as plain text. Instead it's [the `String.toByteArray()` defined in this same namespace](https://github.com/smithy-lang/smithy-kotlin/blob/main/runtime/serde/serde-cbor/common/test/aws/smithy/kotlin/runtime/serde/cbor/CborDeserializerSuccessTest.kt#L18-L24), which treats input as a hex-encoded byte sequence in a string.

The creation of this helper function predates the existence of stdlib's `String.hexToByteArray()` and could probably be replaced by it, especially since the name shadows a stdlib function that does something different. I'll look into that in a separate PR.

ianbotsf: Separate PR: https://github.com/smithy-lang/smithy-kotlin/pull/1620

luigi617 (approved):

PR #1614, opened 2026-06-02
feat: add support for `@requiresLength` trait

### LINKED ISSUE #363: support requiresLength trait
The [requiresLength trait](https://awslabs.github.io/smithy/1.0/spec/core/stream-traits.html#requireslength-trait) helps clarify what streaming operations require a known `Content-Length` even though they take a streaming input. Not many models seem to support this yet but nevertheless we should predicate off of it and either customize serialization or add a middleware that validates the request can be satisfied.

ianbotsf:
## Issue \#

Closes #363 

## Description of changes

This change adds support for [Smithy's `@requiresLength` trait](https://smithy.io/2.0/spec/streaming.html#smithy-api-requireslength-trait) by way of a new interceptor that ensures the body `ByteStream` has a non-null `contentLength`.

### DIFF
--- a/codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/rendering/protocol/RequiresLengthIntegration.kt
+++ b/codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/rendering/protocol/RequiresLengthIntegration.kt
@@ -1,0 +1,48 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.codegen.rendering.protocol
+
+import aws.smithy.kotlin.codegen.KotlinSettings
+import aws.smithy.kotlin.codegen.core.KotlinWriter
+import aws.smithy.kotlin.codegen.core.RuntimeTypes
+import aws.smithy.kotlin.codegen.integration.KotlinIntegration
+import aws.smithy.kotlin.codegen.model.hasTrait
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.model.shapes.StructureShape
+import software.amazon.smithy.model.traits.RequiresLengthTrait
+import software.amazon.smithy.model.traits.StreamingTrait
+import kotlin.jvm.optionals.getOrNull
+
+/**
+ * Adds the [RequiresLengthInterceptor] to operations whose input contains a streaming member
+ * targeting a blob with the `@requiresLength` trait.
+ */
+class RequiresLengthIntegration : KotlinIntegration {
+    override fun enabledForService(model: Model, settings: KotlinSettings): Boolean = model
+        .isTraitApplied(RequiresLengthTrait::class.java)
+
+    override fun customizeMiddleware(
+        ctx: ProtocolGenerator.GenerationContext,
+        resolved: List<ProtocolMiddleware>,
+    ): List<ProtocolMiddleware> = resolved + RequiresLengthMiddleware
+}
+
+private object RequiresLengthMiddleware : ProtocolMiddleware {
+    override val name: String = "RequiresLengthMiddleware"
+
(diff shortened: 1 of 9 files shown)


codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/rendering/protocol/RequiresLengthIntegration.kt
```diff
@@ -0,0 +1,48 @@
...
+import aws.smithy.kotlin.codegen.core.KotlinWriter
+import aws.smithy.kotlin.codegen.core.RuntimeTypes
+import aws.smithy.kotlin.codegen.integration.KotlinIntegration
+import aws.smithy.kotlin.codegen.model.hasTrait
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.model.shapes.StructureShape
+import software.amazon.smithy.model.traits.RequiresLengthTrait
+import software.amazon.smithy.model.traits.StreamingTrait
+import kotlin.jvm.optionals.getOrNull
```
luigi617, line 17: Nit: this import is never used, weird that ktlint did not catch it. It happens several times, maybe worth to investigate when dealing with * import

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/RequiresLengthInterceptor.kt
```diff
@@ -0,0 +1,25 @@
...
+import aws.smithy.kotlin.runtime.client.ProtocolRequestInterceptorContext
+import aws.smithy.kotlin.runtime.http.request.HttpRequest
+
+/**
+ * An interceptor that validates the request body has a known content length.
+ * This is used for operations with streaming input members marked with the `@requiresLength` trait.
+ */
+@InternalApi
+public class RequiresLengthInterceptor : HttpInterceptor {
+    override suspend fun modifyBeforeSigning(context: ProtocolRequestInterceptorContext<Any, HttpRequest>): HttpRequest {
```
luigi617, line 18: Question: Another reasonable place would be `modifyBeforeRetryLoop` since contentLength won't change in each attempts. But `modifyBeforeSigning` might be better to avoid signing if content length is not given. Is this the reason why you chose `modifyBeforeSigning`?

luigi617 (approved):
Just 2 comments, code looks good to me

runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/RequiresLengthInterceptor.kt
ianbotsf: While it's probably not a good practice, I believe nothing prevents `Content-Length` from being set/changed by users' interceptors on each attempt. I chose this hook because `Content-Length` _should_ be signed and that makes `modifyBeforeSigning` a suitable enforcement point.

codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/rendering/protocol/RequiresLengthIntegration.kt
ianbotsf: Apparently, [**ktlint** deprecated and disabled the `no-unused-imports` rule in **1.7.0**](https://github.com/ktlint/ktlint/issues/3038) and they plan to remove it in **2.0.0**.

JetBrains recently announced that it's working with Meta to [fold **ktfmt** into Kotlin as the standard linter/formatter](https://blog.jetbrains.com/kotlin/2026/05/kotlinconf26-keynote-highlights/#kotlin-ecosystem). We should begin thinking about migrating away from **ktlint** in favor of [**ktfmt**](https://github.com/facebook/ktfmt). It's going to be obtrusive though—**ktfmt** is intentionally designed for very minimal configurability and completely rewrites the source code on every pass so I expect a high number of cosmetic changes when we switch.

PR #1622, opened 2026-06-15
chore: silence Dependabot alerts for Netty because it's a test-only dependency

ianbotsf:
## Issue \#

(none)

## Description of changes

This is a _speculative_ change to see if we can ignore Netty security alerts from Dependabot while still getting version update PRs. 🤞

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,16 @@
 version: 2
 updates:
+  # Gradle but only non-runtime dependencies
+  - package-ecosystem: gradle
+    directory: /
+    open-pull-requests-limit: 0 # Disable alerts
+    groups:
+      non-runtime-dependencies:
+        applies-to: security-updates
+        patterns:
+          - "io.netty:*"
+
+  # All runtime dependencies
   - package-ecosystem: gradle
     directory: /
     schedule:


.github/dependabot.yml
```diff
@@ -1,5 +1,16 @@
 version: 2
 updates:
+  # Gradle but only non-runtime dependencies
+  - package-ecosystem: gradle
```
luigi617, line 4: Under each `package-ecosystem` entry, [schedule.interval](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#schedule-) is required.

.github/dependabot.yml
```diff
@@ -1,5 +1,16 @@
...
+    directory: /
```
luigi617, line 5: Both entries in the PR use `package-ecosystem: gradle + directory: /` targeting the same default branch, which violates [this constraint](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#directories-or-directory--)

luigi617 (changes requested):
Left some comments.

.github/dependabot.yml
ianbotsf: This is indeed a blocker and means we cannot use **dependabot.yml** to silence these alerts without also silencing update PRs.

ianbotsf:
Closing this PR in favor of [Dependabot auto-triage rules](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/auto-triage-dependabot-alerts).

PR #1624, opened 2026-06-23
V1.7 main changes

luigi617:
## Description of changes
- Update kotlin version to 2.4.0
- Removed deprecated functions planned to be removed in 1.7
- Add PlannedRemoval(1, 8) to deprecated functions

### DIFF
--- a/.changes/bec7ce83-b7f0-4973-ac8a-ac62ec559ea3.json
+++ b/.changes/bec7ce83-b7f0-4973-ac8a-ac62ec559ea3.json
@@ -1,6 +1,6 @@
 {
     "id": "bec7ce83-b7f0-4973-ac8a-ac62ec559ea3",
     "type": "misc",
-    "description": "**Breaking**: Update Kotlin version from 2.3.20 to 2.4.0. Remove deprecated `AwsChunkedSource` constructor overload, `HttpRequestBuilder.setAwsChunkedBody` overload, and `Flow<SdkBuffer>.asEventStreamHttpBody(CoroutineScope)` overload that were scheduled for removal in 1.7. Deprecate `Filesystem.readFileOrNull`, `Filesystem.writeFile`, `Filesystem.fileExists`, `Filesystem.Companion.fromMap(Map<String, ByteArray>)`, `OkHttpEngineConfig.connectionIdlePollingInterval`, and `TestPlatformProvider` legacy constructors for removal in 1.8",
+    "description": "**Breaking**: Update Kotlin version from 2.3.20 to 2.4.0. Remove deprecated `AwsChunkedSource` constructor overload, `HttpRequestBuilder.setAwsChunkedBody` overload, and `Flow<SdkBuffer>.asEventStreamHttpBody(CoroutineScope)` overload that were scheduled for removal in 1.7. Deprecate `Filesystem.readFileOrNull`, `Filesystem.writeFile`, `Filesystem.fileExists`, `Filesystem.Companion.fromMap(Map<String, ByteArray>)`, and `TestPlatformProvider` legacy constructors for removal in 1.8",
     "requiresMinorVersionBump": true
 }
--- a/.changes/bec7ce83-b7f0-4973-ac8a-ac62ec559ea3.json
+++ b/.changes/bec7ce83-b7f0-4973-ac8a-ac62ec559ea3.json
@@ -1,0 +1,6 @@
+{
+    "id": "bec7ce83-b7f0-4973-ac8a-ac62ec559ea3",
+    "type": "misc",
+    "description": "**Breaking**: Update Kotlin version from 2.3.20 to 2.4.0. Remove deprecated `AwsChunkedSource` constructor overload, `HttpRequestBuilder.setAwsChunkedBody` overload, and `Flow<SdkBuffer>.asEventStreamHttpBody(CoroutineScope)` overload that were scheduled for removal in 1.7. Deprecate `Filesystem.readFileOrNull`, `Filesystem.writeFile`, `Filesystem.fileExists`, `Filesystem.Companion.fromMap(Map<String, ByteArray>)`, `OkHttpEngineConfig.connectionIdlePollingInterval`, and `TestPlatformProvider` legacy constructors for removal in 1.8",
+    "requiresMinorVersionBump": true
+}
--- a/.changes/bec7ce83-b7f0-4973-ac8a-ac62ec559ea3.json
+++ b/.changes/bec7ce83-b7f0-4973-ac8a-ac62ec559ea3.json
@@ -1,6 +1,6 @@
 {
     "id": "bec7ce83-b7f0-4973-ac8a-ac62ec559ea3",
     "type": "misc",
-    "description": "**Breaking**: Update Kotlin version from 2.3.20 to 2.4.0. Remove deprecated `AwsChunkedSource` constructor overload, `HttpRequestBuilder.setAwsChunkedBody` overload, and `Flow<SdkBuffer>.asEventStreamHttpBody(CoroutineScope)` overload that were scheduled for removal in 1.7. Deprecate `Filesystem.readFileOrNull`, `Filesystem.writeFile`, `Filesystem.fileExists`, `Filesystem.Companion.fromMap(Map<String, ByteArray>)`, and `TestPlatformProvider` legacy constructors for removal in 1.8",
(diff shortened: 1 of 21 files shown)


.changes/bec7ce83-b7f0-4973-ac8a-ac62ec559ea3.json
```diff
@@ -0,0 +1,6 @@
+{
+    "id": "bec7ce83-b7f0-4973-ac8a-ac62ec559ea3",
+    "type": "misc",
+    "description": "**Breaking**: Update Kotlin version from 2.3.20 to 2.4.0. Remove deprecated `AwsChunkedSource` constructor overload, `HttpRequestBuilder.setAwsChunkedBody` overload, and `Flow<SdkBuffer>.asEventStreamHttpBody(CoroutineScope)` overload that were scheduled for removal in 1.7. Deprecate `Filesystem.readFileOrNull`, `Filesystem.writeFile`, and `Filesystem.fileExists` for removal in 1.8",
+    "requiresMinorVersionBump": true
+}
```
ianbotsf, line 6: **Correctness**:

This single changelog entry should be broken up into three:
* Updating the Kotlin version
* Removing all the deprecated APIs
* Deprecating new APIs for removal in 1.8

All of these are breaking changes so they all need the `"requiresMinorVersionBump": true` flag.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/TestPlatformProvider.kt
```diff
@@ -25,10 +26,10 @@ private constructor(
 ) : PlatformProvider,
     Filesystem by fs {
 
-    // TODO deprecate
+    // TODO Deprecate
     public constructor() : this(emptyMap(), emptyMap(), MapFilesystem(), OperatingSystem(OsFamily.Linux, "test"))
 
-    // TODO deprecate
+    // TODO Deprecate
     public constructor(
```
ianbotsf, line 35: **Nit**: Please actually deprecate these and plan for removal in 1.8.

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/util/SystemDefaultProviderBase.kt
```diff
@@ -16,6 +17,8 @@ import platform.posix.*
 public abstract class SystemDefaultProviderBase : PlatformProvider {
     override fun getenv(key: String): String? = platform.posix.getenv(key)?.toKString()
 
+    @OptIn(PlannedRemoval::class)
+    @Deprecated("Use readOrNull() instead", replaceWith = ReplaceWith("readOrNull(path, readAll = true)"))
     override suspend fun readFileOrNull(path: String): ByteArray? = withContext(SdkDispatchers.IO) {
```
ianbotsf: **Correctness**: When deprecating a method, we should replace its implementation with a call to the new version—in this case, `readOrNull()`. That should eliminate the need to opt into the `@PlannedRemoval` annotation too.

ianbotsf: **Correctness**: When deprecating a method, we should add `@PlannedRemoval` too. In the case of `PlatformProvider`, we do not expect our users to be touching that directly so we don't need a long incubation time. Removing in 1.8 should be fine.

ianbotsf (changes requested):

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/TestPlatformProvider.kt
luigi617: I tried before, but some functions still uses these to-be-deprecated contructor. And since we have `allWarningsAsErrors = true` in `build.gradle.kts`. This make compile fails. I think that's also the reason why it is left as TODO. Should I replace all usage of this constructor to the correct one and deprecate these function?

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/util/SystemDefaultProviderBase.kt
luigi617: I think the behavior for old `readFileOrNull()` is not the same as `readOrNull()`, `readFileOrNull` return `null` even if error is thrown (while the new function just throw the error). I think replacing the implementation with the new function will be _surprising_ for customers. Do we still want to replace it with the new functions?

Nonetheless, the `@OptIn(PlannedRemoval::class)` is required because we're overriding an interface method (`readFileOrNull`) that's marked with `@PlannedRemoval`. It is unrelated to the implementation

luigi617: It has the `@OptIn(PlannedRemoval::class)`. Maybe you commented on the wrong lines? If I understand it correctly in the interfact we need to add `@PlannedRemoval(1, 8)`, but for overring function, we only need to add `@OptIn(PlannedRemoval::class)`. Is my understanding correct?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/TestPlatformProvider.kt
ianbotsf: > Should I replace all usage of this constructor to the correct one and deprecate these function?

Yes please!

runtime/runtime-core/native/src/aws/smithy/kotlin/runtime/util/SystemDefaultProviderBase.kt
ianbotsf: > I think the behavior for old `readFileOrNull()` is not the same as `readOrNull()`, `readFileOrNull` return `null` even if error is thrown (while the new function just throw the error). I think replacing the implementation with the new function will be surprising for customers. Do we still want to replace it with the new functions?

Customers should not be using this function directly in their code. IT's probably a miss that we didn't decorate these interfaces with `@InternalApi`. I'm okay with changing the behavior as part of this minor version bump.

> Nonetheless, the @OptIn(PlannedRemoval::class) is required because we're overriding an interface method (readFileOrNull) that's marked with @PlannedRemoval. It is unrelated to the implementation

Oh right, forgot about how that works with implementing methods. A default implementation (that delegates to the preferred method) in the interface would eliminate the need to override it in the class.

ianbotsf: Yes you're right. As above, I'd forgotten that overrides need to opt into their parent's opt-in annotations. Disregard!

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Filesystem.kt
```diff
@@ -206,7 +206,8 @@ public interface Filesystem {
         /**
          * Construct a fake filesystem from a mapping of paths to contents
          */
-        // TODO deprecate
+        @Deprecated("Use fromMap(Map<String, TestFile>) instead", replaceWith = ReplaceWith("fromMap(data.mapValues { (_, v) -> TestFile(v) }, filePathSeparator)"))
+        @PlannedRemoval(major = 1, minor = 8)
         public fun fromMap(data: Map<String, ByteArray>, filePathSeparator: String = "/"): Filesystem = MapFilesystem(
             TestFile.transformMap(data).toMutableMap(),
             filePathSeparator,
```
luigi617: Comment: did not change the implementation because I get this error if changed:
```
Overload resolution ambiguity between candidates:                                                                                                                                                                                                 
  fun fromMap(data: Map<String, ByteArray>, filePathSeparator: String = ...): Filesystem                                                                                                                                                            
  fun fromMap(data: Map<String, TestFile>, filePathSeparator: String = ...): Filesystem
```

ianbotsf: I tested this and it built without error:

```kotlin
@Deprecated("Use fromMap(Map<String, TestFile>) instead", replaceWith = ReplaceWith("fromMap(data.mapValues { (_, v) -> TestFile(v) }, filePathSeparator)"))
@PlannedRemoval(major = 1, minor = 8)
public fun fromMap(data: Map<String, ByteArray>, filePathSeparator: String = "/"): Filesystem =
    fromMap(TestFile.transformMap(data).toMutableMap(), filePathSeparator)
```

Ultimately, if you can't get it to work it's fine. Changing this implementation isn't critical since it mostly delegates to the `MapFilesystem` constructor. I was more concerned with other methods that had far more code in them.

ianbotsf (approved):

PR #1640, opened 2026-07-01
feat(http): add logRedactedHeaders config property for sanitizing sensitive headers in debug logging

ashishdhingra:
## Issue \#
https://github.com/aws/aws-sdk-kotlin/issues/1460

## Description of changes
Adds a new `logRedactedHeaders` configuration property to `SdkClientConfig` that allows users to specify HTTP header names whose values should be replaced with `<REDACTED>` in request/response debug logging. Matching is case-insensitive and the set is empty by default (no headers redacted).

When `LogMode` includes request or response logging, all HTTP headers are dumped verbatim, including sensitive values like `Authorization` or custom tokens. This property gives SDK users a way to opt-in to redacting specific headers without disabling debug logging entirely. The redaction set is empty by default so there is no behavior change for existing users.

___

### DIFF
--- a/codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt
+++ b/codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt
@@ -132,18 +132,19 @@ object RuntimeConfigProperty {
 
     val LogRedactedHeaders = ConfigProperty {
         name = "logRedactedHeaders"
+        val defaultValue = "${KotlinTypes.Collections.mutableSetOf.fullName}()"
         symbol = KotlinTypes.Collections.set(KotlinTypes.String, isNullable = false)
-        builderSymbol = KotlinTypes.Collections.set(KotlinTypes.String, isNullable = true)
-        toBuilderExpression = ""
-        propertyType = ConfigPropertyType.RequiredWithDefault("emptySet()")
+        builderSymbol = KotlinTypes.Collections.mutableSet(KotlinTypes.String, isNullable = false, default = defaultValue)
+        toBuilderExpression = ".toMutableSet()"
 
         baseClass = RuntimeTypes.SmithyClient.LogRedactionConfig
         useNestedBuilderBaseClass()
 
         documentation = """
-        Set of HTTP header names whose values will be replaced with "<REDACTED>" in
+        Set of HTTP header names whose values will be replaced with "*** Sensitive Data Redacted ***" in
         request/response debug logging. Matching is case-insensitive. Empty by default
-        (no headers are redacted).
+        (no headers are redacted). See [PotentiallySensitiveHeaders][aws.smithy.kotlin.runtime.http.PotentiallySensitiveHeaders] for
+        a pre-built set of commonly-sensitive headers.
         """.trimIndent()
     }
 
--- a/codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt
+++ b/codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt
@@ -130,6 +130,23 @@ object RuntimeConfigProperty {
         """.trimIndent()
     }
 
+    val LogRedactedHeaders = ConfigProperty {
+        name = "logRedactedHeaders"
+        symbol = KotlinTypes.Collections.set(KotlinTypes.String, isNullable = false)
+        builderSymbol = KotlinTypes.Collections.set(KotlinTypes.String, isNullable = true)
+        toBuilderExpression = ""
+        propertyType = ConfigPropertyType.RequiredWithDefault("emptySet()")
+
+        baseClass = RuntimeTypes.SmithyClient.LogRedactionConfig
+        useNestedBuilderBaseClass()
+
+        documentation = """
+        Set of HTTP header names whose values will be replaced with "<REDACTED>" in
+        request/response debug logging. Matching is case-insensitive. Empty by default
+        (no headers are redacted).
+        """.trimIndent()
+    }
+
     var TelemetryProvider = ConfigProperty {
         symbol = RuntimeTypes.Observability.TelemetryApi.TelemetryProvider
         baseClass = RuntimeTypes.Observability.TelemetryApi.TelemetryConfig
--- a/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HeaderRedaction.kt
+++ b/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HeaderRedaction.kt
@@ -4,14 +4,35 @@
  */
 package aws.smithy.kotlin.runtime.http
 
+import aws.smithy.kotlin.runtime.collections.AttributeKey
+
 /**
- * Commonly-sensitive HTTP headers that users may choose to redact from debug logging.
+ * HTTP headers that are potentially sensitive and may be redacted from debug logging.
  * This set is NOT applied by default — users must explicitly opt in via the
  * `logRedactedHeaders` client config property.
+ *
+ * Example usage:
+ * ```kotlin
+ * val client = MyServiceClient {
+ *     logRedactedHeaders += PotentiallySensitiveHeaders.Default
+ * }
+ * ```
  */
-public object SensitiveHeaders {
+public object PotentiallySensitiveHeaders {
+    /**
+     * The default set of potentially sensitive headers, including:
+     * - `Authorization`
+     * - `X-Amz-Security-Token`
+     */
     public val Default: Set<String> = setOf(
         "Authorization",
         "X-Amz-Security-Token",
     )
 }
+
+/**
+ * The string used to replace sensitive header values in debug logging output.
+ */
+internal const val SENSITIVE_DATA_REDACTED = "*** Sensitive Data Redacted ***"
+
+internal val LOG_REDACTED_HEADERS_KEY = AttributeKey<Set<String>>("aws.smithy.kotlin#LogRedactedHeaders")
--- a/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HeaderRedaction.kt
+++ b/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HeaderRedaction.kt
@@ -1,0 +1,17 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.http
+
+/**
+ * Commonly-sensitive HTTP headers that users may choose to redact from debug logging.
+ * This set is NOT applied by default — users must explicitly opt in via the
+ * `logRedactedHeaders` client config property.
+ */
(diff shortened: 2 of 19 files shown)


codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt
```diff
@@ -130,6 +130,23 @@ object RuntimeConfigProperty {
...
+
+        baseClass = RuntimeTypes.SmithyClient.LogRedactionConfig
+        useNestedBuilderBaseClass()
+
+        documentation = """
+        Set of HTTP header names whose values will be replaced with "<REDACTED>" in
+        request/response debug logging. Matching is case-insensitive. Empty by default
+        (no headers are redacted).
+        """.trimIndent()
+    }
```
ianbotsf, line 149: **Nit**: Generally when we have collection config properties we want the builder to use _mutable_ variants. That is, the **built** config should have `val logRedactedHeaders: Set<String>` but the **builder** config should have `var logRedactedHeaders: MutableSet<String>`. This allows users to either completely replace the whole set _or_ to append to it:

```kotlin
MyServiceClient {
    // Replace the whole set of headers
    logRedactedHeaders = mutableSetOf("foo", "bar")
    
    // And/or update the existing list of headers
    logRedactedHeaders += "baz"
}
```

See [the `HttpInterceptors` property](https://github.com/smithy-lang/smithy-kotlin/blob/fbd9945a999790c3335d9d6a4e28a7432aa98cbb/codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt#L147-L164) for an example.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/request/HttpRequestBuilder.kt
```diff
@@ -95,7 +95,25 @@ public fun HttpRequestBuilder.header(name: String, value: String): Unit = header
...
+ * replaced.
+ * @param redactedHeaders Set of header names whose values will be replaced with "<REDACTED>" in the output.
+ * Matching is case-insensitive.
+ */
+@InternalApi
+public suspend fun dumpRequest(
+    request: HttpRequestBuilder,
+    dumpBody: Boolean,
+    redactedHeaders: Set<String>,
+): String {
```
ianbotsf, line 122: **Style**: If we're already adding a new overload then I think we can pick a better API signature that's more future-proof. Rather than taking `request: HttpRequestBuilder` and `redactedHeaders: Set<String>`, we could take `request: SdkHttpRequest` instead. It's already available at the call site and contains both the _actual_ request (`SdkHttpRequest.subject`) and the context (`SdkHttpRequest.context`) from which we can retrieve the redacted headers.

The declaration becomes:

```kotlin
@InternalApi
public suspend fun dumpRequest(
    requestWrapper: SdkHttpRequest,
    dumpBody: Boolean,
): String {
    val request = requestWrapper.subject
    val redactedHeaders = requestWrapper.context.logRedactedHeaders
    // ...rest of method continues as-is
}
```

And the call site becomes:

```kotlin
private suspend fun httpTraceMiddleware(request: SdkHttpRequest, next: Handler<SdkHttpRequest, HttpCall>): HttpCall {
    // ...
    dumpRequest(request, logMode.isEnabled(LogMode.LogRequestWithBody))
    // ...
}
```

If in the future we need more context to dump requests/responses, we'll already have it because we're passing the whole `SdkHttpRequest` instead of smaller pieces of it.

Applies to `dumpResponse` as well.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/request/HttpRequestBuilder.kt
```diff
@@ -111,7 +129,12 @@ public suspend fun dumpRequest(request: HttpRequestBuilder, dumpBody: Boolean):
     request.headers.entries()
         .filterNot { it.key in skip }
         .forEach { (key, values) ->
-            buffer.writeUtf8(values.joinToString(separator = ";", prefix = "$key: ", postfix = "\r\n"))
+            val headerValue = if (redactedHeaders.any { it.equals(key, ignoreCase = true) }) {
+                "<REDACTED>"
+            } else {
```
ianbotsf, line 141: **Nit**: For consistency, let's use [the same redaction string we already have](https://github.com/smithy-lang/smithy-kotlin/blob/fbd9945a999790c3335d9d6a4e28a7432aa98cbb/codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/rendering/StructureGenerator.kt#L30) for `@sensitive` fields in structure shapes (`"*** Sensitive Data Redacted ***"`). While making the change, please centralize the string as a constant somewhere so it's not duplicated in `dumpRequest` and `dumpResponse`. Ideally we'd just reuse the declaration we already have but it's in a codegen module that isn't available at runtime.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/request/HttpRequestBuilder.kt
```diff
@@ -95,7 +95,25 @@ public fun HttpRequestBuilder.header(name: String, value: String): Unit = header
  * replaced.
  */
 @InternalApi
-public suspend fun dumpRequest(request: HttpRequestBuilder, dumpBody: Boolean): String {
+public suspend fun dumpRequest(
+    request: HttpRequestBuilder,
+    dumpBody: Boolean,
+): String = dumpRequest(request, dumpBody, emptySet())
```
ianbotsf: **Nit**: We should prefer usage of the new overload everywhere and deprecate this method. Please add `@Deprecated("...")` and `@PlannedForRemoval(major = 1, minor = 8)` and then replace all calls with calls to the new overload.

Applies to `dumpResponse` as well.

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HeaderRedaction.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Commonly-sensitive HTTP headers that users may choose to redact from debug logging.
+ * This set is NOT applied by default — users must explicitly opt in via the
+ * `logRedactedHeaders` client config property.
+ */
+public object SensitiveHeaders {
+    public val Default: Set<String> = setOf(
+        "Authorization",
+        "X-Amz-Security-Token",
+    )
+}
```
ianbotsf, line 31: **Nit**: We need to message this accurately to users. To be clear, [AWS _does not_ consider these values sensitive](https://github.com/aws/aws-sdk-kotlin/issues/1460#issuecomment-4652102008) although users might want them redacted anyway. I think we should reflect that in the KDocs and also in the object name itself,  which I suggest renaming to `PotentiallySensitiveHeaders` or similar.

ianbotsf, line 31: **Nit**: Providing this prepopulated set is good but it's not easily discoverable. Please link to this class in the KDocs for the config property. Potentially something like:

```kotlin
/**
 * Set of HTTP header names whose values will be replaced with `"*** Sensitive Data Redacted ***"`
 * in request/response debug logging. Matching is case-insensitive. This parameter is empty by
 * default, meaning no headers are redacted. You may choose a preselected set of headers such as
 * [PotentiallySensitiveHeaders][aws.smithy.kotlin.runtime.http.PotentiallySensitiveHeaders] or
 * provide your own custom set.
 */
public val logRedactedHeaders: Set<String>
```

ianbotsf (changes requested):

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/request/HttpRequestBuilder.kt
ashishdhingra: I'm assuming that here we are mentioning to define a new declaration for `dumpRequest` and `dumpResponse` in the same file `SdkOperationExecution.kt` in `http-client` module. The `request: SdkHttpRequest` is defined in this `http-client` module (Same for `dumpResponse` as well). Else if we define it in `http` module then it will create circular dependency.

  - `dumpRequest` lives in the `http` module (`runtime/protocol/http`)
  - `SdkHttpRequest` is a typealias defined in the `http-client` module (`runtime/protocol/http-client`)
  - The `http-client` module already depends on `http`
  - Adding `SdkHttpRequest` as a parameter to a function in `http` would require `http` to depend on `http-client`

codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt
```diff
@@ -130,6 +130,24 @@ object RuntimeConfigProperty {
...
+        toBuilderExpression = ".toMutableSet()"
+
+        baseClass = RuntimeTypes.SmithyClient.LogRedactionConfig
+        useNestedBuilderBaseClass()
+
+        documentation = """
+        Set of HTTP header names whose values will be replaced with "*** Sensitive Data Redacted ***" in
+        request/response debug logging. Matching is case-insensitive. Empty by default
+        (no headers are redacted). See [aws.smithy.kotlin.runtime.http.PotentiallySensitiveHeaders] for
+        a pre-built set of commonly-sensitive headers.
```
ianbotsf, line 147: **Style**: Using the fully-qualified name without an alias (e.g., `[aws.smithy.kotlin.runtime.http.PotentiallySensitiveHeaders]`) renders verbatim in KDocs. Please use aliases when linking to declarations which aren't in the `import` list:

```kotlin
[PotentiallySensitiveHeaders][aws.smithy.kotlin.runtime.http.PotentiallySensitiveHeaders]
```

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HeaderRedaction.kt
```diff
@@ -0,0 +1,38 @@
...
+ * `logRedactedHeaders` client config property.
+ *
+ * Example usage:
+ * ```kotlin
+ * val client = MyServiceClient {
+ *     logRedactedHeaders = PotentiallySensitiveHeaders.Default.toMutableSet()
+ * }
+ * ```
+ */
+public object PotentiallySensitiveHeaders {
```
ianbotsf, line 21: **Suggestion**: The code example you give in KDocs is functionally correct but awkward since you have to convert the set to mutable. I suggest using set addition in your example instead:

```kotlin
val client = MyServiceClient {
    logRedactedHeaders += PotentiallySensitiveHeaders.Default
}
```

ianbotsf (approved):
Minor feedback left. Fix and ship.

codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt
ashishdhingra: Done

runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/HeaderRedaction.kt
ashishdhingra: Done

PR #1642, opened 2026-07-01
fix: correctly generate default values for floating point numbers given in scientific notation

ianbotsf:
## Issue \#

(none)

## Description of changes

This change refactors parsing of defaults for double values in Smithy models to correctly extract numbers given in scientific notation. Previously, a Smithy model default such as `"smithy.api#default": 1.0E-8` would be codegenned as `1.0E-8.0`, which is invalid because Kotlin double literals may not contain decimals in the exponent.


ashishdhingra (approved):
Assuming it is fine that it fails for value `1.0E-8.0` in Smithy model, looks good. Expression `"1.0E-8".toDoubleOrNull()` correctly returns value `1.0E-8` as `Double?`.

ianbotsf:
> Assuming it is fine that it fails for value `1.0E-8.0` in Smithy model, looks good. Expression `"1.0E-8".toDoubleOrNull()` correctly returns value `1.0E-8` as `Double?`.

Yes, the value `1.0E-8.0` would be invalid in a Smithy model and would fail validation before codegen.

luigi617 (approved):

PR #1649, opened 2026-07-07
expose maxIdleConnection and set 50 as default

luigi617:
## Summary

- Expose `maxIdleConnections` as a configurable parameter on `OkHttpEngineConfig`
- Change the default from OkHttp's built-in default of 5 to 50

## Motivation

OkHttp's default connection pool retains only 5 idle connections. The Kotlin SDK's default `maxConcurrency` is 64, meaning under moderate-to-high concurrency the pool constantly tears down and re-establishes TCP/TLS connections instead of reusing them. This mismatch adds latency from repeated handshakes and increases CPU overhead from unnecessary connection churn.

Raising the default to 50 aligns the idle pool with the SDK's concurrency model and exposes the knob so customers with unusual workloads can tune it themselves.

## Benchmark Results

- **Instances:** 5 per variant
- **Warmup batches:** 3
- **Measurement batches:** 5

#### S3 Upload
| Concurrency | Base (Gbps) | Target (Gbps) | Change | Base CPU% | Target CPU% | Base Mem (MB) | Target Mem (MB) |
| 16 | 3.88 ± 0 | 3.90 ± 0 | +0.5% | 29.1 | 29.2 | 662.0 | 728.4 |
| 32 | 7.07 ± 0.20 | 7.31 ± 0.14 | +3.4% | 58.2 | 60.0 | 702.5 | 884.2 |
| 64 | 10.18 ± 0.78 | 10.47 ± 0.62 | +2.8% | 91.8 | 89.4 | 947.9 | 999.4 |
| 128 | 10.36 ± 0.80 | 10.70 ± 0.70 | +3.3% | 93.1 | 92.3 | 1001.6 | 1002.4 |

#### S3 Download
| Concurrency | Base (Gbps) | Target (Gbps) | Change | Base CPU% | Target CPU% | Base Mem (MB) | Target Mem (MB) |
| 16 | 5.79 ± 0.10 | 5.85 ± 0.14 | +1.0% | 49.3 | 51.4 | 986.6 | 987.4 |
| 32 | 9.40 ± 0.24 | 8.90 ± 0.34 | -5.3% | 83.9 | 84.3 | 1071.8 | 1236.1 |
| 64 | 10.05 ± 0.40 | 10.03 ± 0 | -0.2% | 87.5 | 86.2 | 1237.6 | 1322.9 |
| 128 | 9.33 ± 0.53 | 10.41 ± 0.48 | **+11.6%** | 89.7 | 83.9 | 1365.5 | 1442.8 |

**Throughput observations:**
- Higher concurrency now yields higher throughput
- Uploads improve consistently across all concurrency levels (+0.5% to +3.4%)
- The -5.3% at download-32 falls within standard deviation overlap (8.90 ± 0.34 vs 9.40 ± 0.24) and is likely noise
- CPU usage remains comparable across both variants — no regression
- Memory usage increases slightly (up to ~160 MB at download-32) due to retaining more idle connections, well within acceptable bounds

#### DynamoDB PutItem
| Metric | Base (ms) | Target (ms) | Change | Base CPU% | Target CPU% | Base Mem (MB) | Target Mem (MB) |
| Average | 3.82 ± 0.76 | 3.34 ± 0.41 | **-12.6%** | 4.2 | 4.1 | 62.3 | 62.3 |
| P50 | 3.68 ± 0.75 | 3.16 ± 0.42 | **-14.1%** | — | — | — | — |
| P90 | 4.34 ± 0.88 | 3.84 ± 0.41 | **-11.5%** | — | — | — | — |
| P99 | 6.41 ± 1.12 | 6.52 ± 0.95 | +1.7% | — | — | — | — |

#### DynamoDB GetItem
| Metric | Base (ms) | Target (ms) | Change | Base CPU% | Target CPU% | Base Mem (MB) | Target Mem (MB) |
| Average | 2.64 ± 0.63 | 2.14 ± 0.17 | **-18.9%** | 5.9 | 5.9 | 62.3 | 62.6 |
| P50 | 2.48 ± 0.66 | 1.94 ± 0.14 | **-21.8%** | — | — | — | — |
| P90 | 3.15 ± 0.64 | 2.65 ± 0.20 | **-15.9%** | — | — | — | — |
| P99 | 4.99 ± 0.91 | 4.60 ± 0.91 | -7.8% | — | — | — | — |

**Latency observations:**
- DynamoDB GetItem P50 latency drops by **21.8%** (2.48 → 1.94 ms) — connections are reused instead of re-established
- DynamoDB PutItem P50 latency drops by **14.1%** (3.68 → 3.16 ms)
- Improvements are consistent across average, P50, and P90 for both operations
- P99 remains within noise (overlapping standard deviations), as expected — tail latency is dominated by service-side variance


ianbotsf (commented):
The code looks correct if this is the feature we want to implement but I'd like to zoom out first:

Should we even offer a setting for this or would it make more sense to set it to the same value as `maxConcurrency`? Are there benefits to controlling both parameters independently?  It seems simpler to make this Just Work™ without exposing more knobs.

If we _do_ decide to keep the new config parameter, would a default of `null` that gets coerced to `builder.maxConcurrency` be better than hardcoded 50?

luigi617:
Ideally, I think we should not set `maxIdleConnections` equal to `maxConcurrency`. It seems reasonable for the default value (we use 64 for `maxConcurrency`), but there are some cases we want fewer `maxIdleConnections` than `maxConcurrency`:
- Memory-constrained environments (Lambda, Android): each idle connection holds socket buffers + TLS state.
- By settimg them equal means if customer set `maxConcurrency` to 128, there will be at most 128 idle connections opened which are wasted.

Fundamentely, `maxConcurrency` and `maxIdleConnections` are 2 conceptually different parameters which I prefer to not tie them together. Moreover, I prefer to give customer the option to customize it if they need it.

For the second question: for the reason above, I think hardcoded 50 is fine, if customer change `maxConcurrency` to a lower number (e.g. 16) then the 50 does not hurt, if they change `maxConcurrency` to a high number (e.g. 128 or 256) they can have the option to tune `maxIdleConnections` as they want

ianbotsf:
OK, having separate parameters for `maxIdleConnections` and `maxConcurrency` makes sense. Let's just make `maxIdleConnections` a nullable field in the builder and then default it to `maxConcurrency` in the built object unless the caller specified a non-null value. So, for example, if a caller sets `maxConcurrency` to `16` and doesn't configure `maxIdleConnections`, then it'll also inherit `16` when the config object is built.

ianbotsf (approved):

Issue #1656, opened 2026-07-13
`Uuid.random()` uses non-cryptographic `kotlin.random.Random`; duplicate UUIDs (idempotency tokens, invocation IDs) under Lambda SnapStart


agarc8:
### Describe the bug

`aws.smithy.kotlin.runtime.util.Uuid.random()` generates v4 UUIDs using the non-cryptographic
`kotlin.random.Random` PRNG. Its seed is ordinary JVM heap state, so it is captured in a Lambda
SnapStart snapshot during the init phase and then **replayed identically in every restored
execution environment**. As a result, `Uuid.random()` returns the *same* UUID (and the same
subsequent sequence) across independent, concurrent invocations and across many days of restores.

This is a uniqueness/correctness defect, not a cryptographic-strength concern. A v4 UUID's contract
is "unique with very high probability"; under SnapStart that guarantee collapses. The single generator
backs multiple runtime consumers, so the collision surfaces in more than one place:

- `DefaultIdempotencyTokenProvider.generateToken()` → `Uuid.random().toString()` — DynamoDB
  `ClientRequestToken` for `TransactWriteItems`. Duplicate tokens across different transactions inside
  DynamoDB's idempotency window throw `IdempotentParameterMismatchException`.
- `SdkHttpOperation`'s `init { context[HttpOperationContext.SdkInvocationId] = Uuid.random().toString() }`
  — the `amz-sdk-invocation-id` retry-correlation header, duplicated across unrelated operations.

Relevant source:
- `Uuid`: https://github.com/smithy-lang/smithy-kotlin/blob/main/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Uuid.kt
- `IdempotencyTokenProvider`: https://github.com/smithy-lang/smithy-kotlin/blob/main/runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/IdempotencyTokenProvider.kt
- `SdkHttpOperation`: https://github.com/smithy-lang/smithy-kotlin/blob/main/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/operation/SdkHttpOperation.kt

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected behavior

[...]

PR #1657, opened 2026-07-13
use stable UUID api

luigi617:
## Issue \#

[Uuid.random() uses non-cryptographic kotlin.random.Random; duplicate UUIDs (idempotency tokens, invocation IDs) under Lambda SnapStart](https://github.com/smithy-lang/smithy-kotlin/issues/1656)

## Description of changes

Previosly, we use `kotlin.random.Random` to generate UUID, this is a bug because whenever they use it in SnapStart, the same random UUID will be generated.

Since from Kotlin 2.4.0, UUID standard library api is stabilized, we should use it.

### DIFF
--- a/runtime/protocol/aws-event-stream/common/src/aws/smithy/kotlin/runtime/awsprotocol/eventstream/HeaderValue.kt
+++ b/runtime/protocol/aws-event-stream/common/src/aws/smithy/kotlin/runtime/awsprotocol/eventstream/HeaderValue.kt
@@ -77,9 +77,9 @@ public sealed class HeaderValue {
     public data class Timestamp(val value: Instant) : HeaderValue()
 
     @InternalApi
-    @Suppress("DEPRECATION")
     @Deprecated("Use Uuid4 instead", ReplaceWith("Uuid4"))
     @PlannedRemoval(1, 8)
+    @Suppress("DEPRECATION")
     public data class Uuid(val value: aws.smithy.kotlin.runtime.util.Uuid) : HeaderValue()
 
     @InternalApi
@@ -202,10 +202,10 @@ public fun HeaderValue.expectByteArray(): ByteArray = checkNotNull((this as? Hea
 @InternalApi
 public fun HeaderValue.expectTimestamp(): Instant = checkNotNull((this as? HeaderValue.Timestamp)?.value) { "expected HeaderValue.Bool, found: $this" }
 
-@OptIn(PlannedRemoval::class)
 @InternalApi
-@Suppress("DEPRECATION")
 @Deprecated("Use expectUuid4 instead", ReplaceWith("expectUuid4()"))
+@PlannedRemoval(1, 8)
+@Suppress("DEPRECATION")
 public fun HeaderValue.expectUuid(): Uuid = checkNotNull((this as? HeaderValue.Uuid)?.value) { "expected HeaderValue.Uuid, found: $this" }
 
 @InternalApi
--- a/runtime/protocol/aws-event-stream/common/src/aws/smithy/kotlin/runtime/awsprotocol/eventstream/HeaderValue.kt
+++ b/runtime/protocol/aws-event-stream/common/src/aws/smithy/kotlin/runtime/awsprotocol/eventstream/HeaderValue.kt
@@ -88,8 +88,8 @@ public sealed class HeaderValue {
     /**
      * Encode a header value to [dest]
      */
+
     @OptIn(PlannedRemoval::class)
-    @Suppress("DEPRECATION")
     public fun encode(dest: SdkBufferedSink): Unit = when (this) {
         is Bool -> {
             val type = if (value) HeaderType.TRUE else HeaderType.FALSE
@@ -128,7 +128,10 @@ public sealed class HeaderValue {
             dest.writeHeader(HeaderType.TIMESTAMP)
             dest.writeLong(value.epochMilliseconds)
         }
-        is Uuid -> {
+        is
+        @Suppress("DEPRECATION")
+        Uuid,
+        -> {
             dest.writeHeader(HeaderType.UUID)
             dest.writeLong(value.high)
             dest.writeLong(value.low)
--- a/runtime/protocol/aws-event-stream/common/src/aws/smithy/kotlin/runtime/awsprotocol/eventstream/HeaderValue.kt
+++ b/runtime/protocol/aws-event-stream/common/src/aws/smithy/kotlin/runtime/awsprotocol/eventstream/HeaderValue.kt
@@ -6,6 +6,7 @@
 package aws.smithy.kotlin.runtime.awsprotocol.eventstream
 
 import aws.smithy.kotlin.runtime.InternalApi
+import aws.smithy.kotlin.runtime.PlannedRemoval
 import aws.smithy.kotlin.runtime.io.*
 import aws.smithy.kotlin.runtime.time.Instant
 import aws.smithy.kotlin.runtime.time.epochMilliseconds
@@ -78,6 +79,7 @@ public sealed class HeaderValue {
     @InternalApi
     @Suppress("DEPRECATION")
     @Deprecated("Use Uuid4 instead", ReplaceWith("Uuid4"))
+    @PlannedRemoval(1, 8)
     public data class Uuid(val value: aws.smithy.kotlin.runtime.util.Uuid) : HeaderValue()
 
     @InternalApi
@@ -86,6 +88,7 @@ public sealed class HeaderValue {
     /**
      * Encode a header value to [dest]
      */
+    @OptIn(PlannedRemoval::class)
     @Suppress("DEPRECATION")
     public fun encode(dest: SdkBufferedSink): Unit = when (this) {
         is Bool -> {
@@ -199,6 +202,7 @@ public fun HeaderValue.expectByteArray(): ByteArray = checkNotNull((this as? Hea
 @InternalApi
 public fun HeaderValue.expectTimestamp(): Instant = checkNotNull((this as? HeaderValue.Timestamp)?.value) { "expected HeaderValue.Bool, found: $this" }
 
+@OptIn(PlannedRemoval::class)
 @InternalApi
 @Suppress("DEPRECATION")
 @Deprecated("Use expectUuid4 instead", ReplaceWith("expectUuid4()"))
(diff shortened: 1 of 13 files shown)


runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Uuid.kt
```diff
@@ -5,79 +5,20 @@
...
-            var destIndex = destOffset
-
-            repeat(length * 2) {
-                shiftBits -= 4
-                val nibble = src shr shiftBits and 0xf
-                dest[destIndex++] = nibbleChars[nibble.toInt()]
-            }
-        }
+        @Suppress("DEPRECATION")
+        public fun random(): Uuid = KotlinUuid.random().toLongs { high, low -> Uuid(high, low) }
```
ashishdhingra, line 22: Since we are deprecating the class, unsure if we need to replace implementation to use `KotlinUuid`

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/util/UuidTest.kt
```diff
@@ -5,32 +5,20 @@
...
-    fun `it should generate random UUIDs`() {
-        var seenUuids = mutableSetOf<Uuid>()
+    fun `it should generate a valid V4 UUID`() {
+        val uuid = Uuid.random()
 
-        repeat(100) {
-            // randomness is unpredictable so do this a bunch to lower risk of false positive
-            val uuid = Uuid.random()
+        uuid.toLongs { msb, lsb ->
+            val version = (msb ushr 12) and 0xF
```
ashishdhingra, line 17: Is the version embedded in `Uuid` class in standard Kotlin library. Or we came up with formula to check for `Uuid4` data class that we created in the other file?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Uuid.kt
luigi617: There are 2 reason why the implementation needs to be replaced:

1. The previous implementation was wrong becuase it uses `kotlin.random.Random` which isn't random in Lambda SnapStart case. This is the main reason of [this bug reported](https://github.com/smithy-lang/smithy-kotlin/issues/1656)
2. Generally, we want to maintain only 1 function/class. So we change the deprecated function/class to use the new function/class (in this case is Kotlin standard library UUID api).

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/util/UuidTest.kt
luigi617: The previous test was mainly for testing if the generated UUID was actually random (using the old UUID class which we define how UUID is generated). The new test is more a safeguard to test if the Kotlin standard library UUID api is generating a V4 UUID.

The reason of this test is because `Uuid.random()` function should generate v4 UUID according their [doc](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/random.html) and it also mention that `random` function is synonimous to `generateV4` function. But for whatever reason the `generateV4` function is still [Experimental Api](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/generate-v4.html). So this test is just to make sure the behavior of `random` function does not change.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Uuid.kt
ashishdhingra: Normally during deprecation, we won't change the implementation. I'm on the edge with this case since we need to fix the existing customers which still use the old API.

For customer who reported the issue, we can request them to use new API. Older customers will get deprecation warning and can also refer changelog to use the new API.

runtime/runtime-core/common/test/aws/smithy/kotlin/runtime/util/UuidTest.kt
ashishdhingra: Acknowledged since the `generateV4` function is still experimental.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Uuid.kt
luigi617: > Normally during deprecation, we won't change the implementation. I'm on the edge with this case since we need to fix the existing customers which still use the old API.

This is an `@InternalAPI` meaning that we don't expect customer to use it directly (even if they can), it is more of a internal function that we use. I agree that we should not change the implementation during deprecation, but after [discussing](https://github.com/smithy-lang/smithy-kotlin/pull/1624#discussion_r3469334590) with @ianbotsf, I think sometimes replacing with the new implementation is fine. Because it reduces our mantainance workload and we are sure that they behave the same.

> For customer who reported the issue, we can request them to use new API. Older customers will get deprecation warning and can also refer changelog to use the new API.

I already suggested them a workaround which requires them to overwrite the way random number is generated using the correct API. But the fix is more for new customers that do not know the workaround. New customers will get the same error if we do not change the old implementation.

runtime/protocol/aws-event-stream/common/src/aws/smithy/kotlin/runtime/awsprotocol/eventstream/HeaderValue.kt
```diff
@@ -76,11 +76,17 @@ public sealed class HeaderValue {
...
+    @Deprecated("Use Uuid4 instead", ReplaceWith("Uuid4"))
     public data class Uuid(val value: aws.smithy.kotlin.runtime.util.Uuid) : HeaderValue()
 
+    @InternalApi
+    public data class Uuid4(val value: kotlin.uuid.Uuid) : HeaderValue()
+
     /**
      * Encode a header value to [dest]
      */
+    @Suppress("DEPRECATION")
```
ianbotsf: **Nit**: Move the `@Suppress` annotation to the most specific point you can to avoid over-suppressing future deprecations. In this case the deprecated declaration is `Uuid` right? If so, I think you can move this to L128.

runtime/protocol/aws-event-stream/common/src/aws/smithy/kotlin/runtime/awsprotocol/eventstream/HeaderValue.kt
```diff
@@ -76,11 +76,17 @@ public sealed class HeaderValue {
     public data class Timestamp(val value: Instant) : HeaderValue()
 
     @InternalApi
+    @Suppress("DEPRECATION")
+    @Deprecated("Use Uuid4 instead", ReplaceWith("Uuid4"))
     public data class Uuid(val value: aws.smithy.kotlin.runtime.util.Uuid) : HeaderValue()
 
+    @InternalApi
+    public data class Uuid4(val value: kotlin.uuid.Uuid) : HeaderValue()
```
ianbotsf, line 86: **Question**: Why a new class (`Uuid4`) instead of reusing the class we have by adding a new constructor? The actual version of the UUID is not important to our code or our users so I don't see why we'd need separate types for internally-generated UUIDs vs stdlib-generated UUIDs.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/Uuid.kt
ianbotsf: We should replace deprecated API implementations with calls to the preferred API whenever possible and when doing so does not cause a breaking behavior change for existing users. We want to avoid maintaining old implementations because that's extra code we have to manage and may lead to subtle behavior differences/drift which complicate debugging and upgrading.

One could argue that changing this implementation may be breaking, particularly if someone was _relying_ on Lambda SnapStart initialization for non-stochastic UUID generation across multiple instances. I consider it to be unlikely that most users would _want_ that behavior, much less be presently relying on it. I think in this case replacing the old implementation is the right call both for our users and for us.

runtime/protocol/aws-event-stream/common/src/aws/smithy/kotlin/runtime/awsprotocol/eventstream/HeaderValue.kt
```diff
@@ -76,11 +76,17 @@ public sealed class HeaderValue {
     public data class Timestamp(val value: Instant) : HeaderValue()
 
     @InternalApi
+    @Suppress("DEPRECATION")
+    @Deprecated("Use Uuid4 instead", ReplaceWith("Uuid4"))
     public data class Uuid(val value: aws.smithy.kotlin.runtime.util.Uuid) : HeaderValue()
```
ianbotsf, line 83: **Correctness**: Add `@PlannedRemoval` annotation to any newly-`@Deprecated` API. Since this is an internal API, we don't need to give a lot of advance notice for removal so targeting **v1.9** is fine.

ianbotsf (changes requested):

runtime/protocol/aws-event-stream/common/src/aws/smithy/kotlin/runtime/awsprotocol/eventstream/HeaderValue.kt
luigi617: Yes the deprecated declaration is `Uuid`, however, it is used in a `when` claude (I can't add `@Suppress("DEPRECATION")` before `is Uuid -> {...`, I can only add it before `when` clause which is basically what the function is doing. Do you think it is better to add before `when` or keep it above the function?

runtime/protocol/aws-event-stream/common/src/aws/smithy/kotlin/runtime/awsprotocol/eventstream/HeaderValue.kt
luigi617: I agree that users don't care the implementation of UUID, but reusing the `Uuid` data class is a breaking change because `value` parameter has diffrent type.

runtime/protocol/aws-event-stream/common/src/aws/smithy/kotlin/runtime/awsprotocol/eventstream/HeaderValue.kt
luigi617: `@PlannedRemoval` Added to v1.8 (since smithy-kotlin is v1.7 now)

runtime/protocol/aws-event-stream/common/src/aws/smithy/kotlin/runtime/awsprotocol/eventstream/HeaderValue.kt
ianbotsf: The suppression cannot go _before_ the `is` keyword since that's not a declaration, statement, or expression. Instead, put it _after_ the `is` since what we're suppressing on is the type itself, not a branch in the `when` statement.

```kotlin
when (this) {
    ...
    is @Suppress("DEPRECATION") Uuid -> ...
    is Uuid4 -> ...
}
```

runtime/protocol/aws-event-stream/common/src/aws/smithy/kotlin/runtime/awsprotocol/eventstream/HeaderValue.kt
ianbotsf: Oh right, got confused with **aws-sdk-kotlin**.

ianbotsf (approved):

PR #1658, opened 2026-07-15
date optimization

luigi617:
## Summary

- Pre-compute formatted signing dates in `AwsSigningConfig` to avoid repeated `Instant.format()` calls during SigV4 signing
- Suppress `fillInStackTrace` on internal parser exceptions to eliminate expensive JVM stack walks during RFC-5322 date parsing

## Motivation

Profiling sequential DynamoDB latency benchmarks revealed two date-related hotspots in the per-request signing and response handling paths:

1. **Repeated date formatting**: `Instant.format(TimestampFormat.ISO_8601_CONDENSED)` was called 6 times per signing operation (in `stringToSign`, `credentialScope`, `signingKey`, and `canonicalRequest`), each time producing the same string from the same `Instant`.

2. **Exception-driven date parsing**: The `ClockSkewInterceptor` parses the response `Date` header on every request via `Instant.fromRfc5322()`. The parser combinator's `alt` function uses exceptions for backtracking — matching the day name "Thu" requires throwing and catching 3 `ParseException` instances (for "Mon", "Tue", "Wed"). Each exception triggers `Throwable.fillInStackTrace()`, a native JNI call that walks the entire thread stack (~50+ frames deep in coroutine + middleware context), costing ~5-15µs per exception.

### What changed

**Date formatting (signing path):**
`AwsSigningConfig` now exposes `formattedSigningDate` and `formattedSigningDateShort` properties, computed once at construction. All 6 call sites in `BaseSigV4SignatureCalculator`, `Canonicalizer`, `RequestMutator`, and `SigV4SignatureCalculator` use the pre-computed values.

**Date parsing (response path):**
Introduced `NoTraceParseException` as an `expect/actual` class:
- JVM actual: overrides `fillInStackTrace()` to return `this` (no-op)
- Native actual: plain subclass (stack traces are not expensive on Native)

All internal parser combinator throw sites (`char()`, `tag()`, `oneOf()`, `takeMNDigitsT()`, `mnDigitsInRange()`, `fraction()`) now throw `NoTraceParseException`. The final `alt` failure that may escape to user code retains the full `ParseException` with stack trace for diagnosability.

## Benchmark Results

- **Warmup batches:** 3
- **Measurement batches:** 5

### E2E Latency (DynamoDB)

14 paired instances (base and target on same host).

#### DynamoDB PutItem
| Metric | Base (ms) | Target (ms) | Change |
|---|---|---|---|
| Average | 4.36 | 4.34 | -0.50% |
| P50 | 4.20 | 4.19 | -0.27% |
| P90 | 4.93 | 4.88 | -1.02% |
| P99 | 7.23 | 7.01 | **-3.01%** |

#### DynamoDB GetItem
| Metric | Base (ms) | Target (ms) | Change |
|---|---|---|---|
| Average | 3.35 | 3.18 | **-4.92%** |
| P50 | 3.20 | 3.05 | **-4.53%** |
| P90 | 3.84 | 3.69 | **-3.90%** |
| P99 | 6.40 | 5.98 | **-6.50%** |

**Latency observations:**
- GetItem shows consistent improvement across all percentiles (**-4.5% to -6.5%**) — the date parsing optimization is directly visible since response handling (where `ClockSkewInterceptor` parses the `Date` header) represents a larger fraction of the shorter GET latency
- PutItem shows modest tail-latency improvement (**-3% at P99**) — the signing date formatting optimization benefits both operations equally, but PUT latency is dominated by network round-trip for the larger request body

### E2E Throughput (S3)

This optimization targets per-request fixed costs (date format/parse), which are negligible relative to the I/O and crypto costs that dominate high-concurrency throughput workloads. No meaningful throughput change expected or observed.

---

### DIFF
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
@@ -6,6 +6,7 @@ package aws.smithy.kotlin.runtime.auth.awssigning
 
 import aws.smithy.kotlin.runtime.auth.awscredentials.Credentials
 import aws.smithy.kotlin.runtime.time.Instant
+import aws.smithy.kotlin.runtime.time.TimestampFormat
 import kotlin.time.Duration
 
 public typealias ShouldSignHeaderPredicate = (String) -> Boolean
@@ -92,6 +93,18 @@ public class AwsSigningConfig(builder: Builder) {
      * construction time.
      */
     public val signingDate: Instant = builder.signingDate ?: Instant.now()
+
+    /**
+     * The signing date formatted as a condensed ISO-8601 timestamp (e.g., "20260714T163045Z").
+     * Pre-computed to avoid repeated formatting.
+     */
+    public val formattedSigningDate: String = signingDate.format(TimestampFormat.ISO_8601_CONDENSED)
+
+    /**
+     * The signing date formatted as a condensed ISO-8601 date only (e.g., "20260714").
+     * Pre-computed to avoid repeated formatting.
+     */
+    public val formattedSigningDateShort: String = signingDate.format(TimestampFormat.ISO_8601_CONDENSED_DATE)
 
     /**
      * A predicate to control which headers are a part of the canonical request. Note that skipping auth-required
--- a/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
+++ b/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
@@ -104,7 +104,7 @@ public class AwsSigningConfig(builder: Builder) {
      * The signing date formatted as a condensed ISO-8601 date only (e.g., "20260714").
      * Pre-computed to avoid repeated formatting.
      */
-    public val formattedSigningDateShort: String = signingDate.format(TimestampFormat.ISO_8601_CONDENSED_DATE)
+    public val formattedSigningDateOnly: String = signingDate.format(TimestampFormat.ISO_8601_CONDENSED_DATE)
 
     /**
      * A predicate to control which headers are a part of the canonical request. Note that skipping auth-required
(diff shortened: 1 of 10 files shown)


runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
```diff
@@ -93,6 +94,18 @@ public class AwsSigningConfig(builder: Builder) {
...
+     * The signing date formatted as a condensed ISO-8601 timestamp (e.g., "20260714T163045Z").
+     * Pre-computed to avoid repeated formatting.
+     */
+    public val formattedSigningDate: String = signingDate.format(TimestampFormat.ISO_8601_CONDENSED)
+
+    /**
+     * The signing date formatted as a condensed ISO-8601 date only (e.g., "20260714").
+     * Pre-computed to avoid repeated formatting.
+     */
+    public val formattedSigningDateShort: String = signingDate.format(TimestampFormat.ISO_8601_CONDENSED_DATE)
```
ashishdhingra: nit: Should we rename it to `formattedSigningDateOnly`?

runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/AwsSigningConfig.kt
```diff
@@ -93,6 +94,18 @@ public class AwsSigningConfig(builder: Builder) {
      */
     public val signingDate: Instant = builder.signingDate ?: Instant.now()
 
+    /**
+     * The signing date formatted as a condensed ISO-8601 timestamp (e.g., "20260714T163045Z").
+     * Pre-computed to avoid repeated formatting.
+     */
+    public val formattedSigningDate: String = signingDate.format(TimestampFormat.ISO_8601_CONDENSED)
+
+    /**
```
ashishdhingra, line 103: We should add test validating pre-computed == format-on-demand value. The existing tests presumably still pass (verifying the same output), but there's no test that specifically validates the pre-computed values match the
  format-on-demand values, which would catch regressions if signingDate or format logic changes in the future.

luigi617: I don't understand this comment. Why do we need to validate pre-computed value == format-on-demand value? `signingDate` (the original value) is a `val` meaning it cannot be changed. So the pre-computed signing date will always be the intended format logic. If someone changes the initializer (different format, different input), the test would need to change too.

Anyways, we have end-to-end signing test which will catch any regressions if anything is changed: [BasicSigningTestBase.kt](https://github.com/smithy-lang/smithy-kotlin/blob/73bbae3ccf8bfcf3081852241630e4821247e25c/runtime/auth/aws-signing-tests/common/src/aws/smithy/kotlin/runtime/auth/awssigning/tests/BasicSigningTestBase.kt)

ianbotsf (approved):

PR #1664, opened 2026-07-17
Optimize CRT response handler: eliminate writer coroutine + intermediate channel

luigi617:
# Optimize CRT response handler: eliminate writer coroutine + intermediate channel

## Summary

Rewrites how the CRT HTTP engine moves response body bytes from the native CRT callback to the SDK consumer. The previous design buffered body chunks into an unbounded `Channel<SdkBuffer>` and drained it with a dedicated `GlobalScope` writer coroutine. This PR removes both the intermediate channel and the coroutine, writing body bytes **directly** into a single window-sized `SdkByteChannel` from the CRT callback and driving flow-control window replenishment from the read side as the consumer drains bytes.

Net effect: fewer allocations, one fewer thread hop per body chunk, no per-response coroutine, and a hard cap on body-buffer memory — while preserving consumption-driven backpressure.

## Before vs After

| Aspect | Before (`main`) | After (this PR) |
|---|---|---|
| Body buffering | Unbounded `Channel<SdkBuffer>(UNLIMITED)` **+** a second `SdkByteChannel` | Single `SdkByteChannel(true, windowSizeBytes)` sized to the CRT window |
| Writer path | `GlobalScope.launch` coroutine loops the channel, copies each buffer into the `SdkByteChannel`, calls `incrementWindow` per buffer | `onResponseBody` writes directly via non-suspending `tryWrite`; no coroutine, no second channel |
| Window replenishment | Write-side, per buffer received, on the writer's dispatcher thread | Read-side, per actual `read()`, on the consumer thread via `WindowManagedReadChannel` → `onDataConsumed` → `incrementWindow` |
| Coroutines / response | 1 `GlobalScope` Job + completion handler | 0 |
| Extra channel / response | 2nd `SdkByteChannel` | 0 |
| Thread hops (native → consumer) | event-loop → channel → writer coroutine (dispatcher) → channel | event-loop writes directly into the channel the consumer reads |
| Steady-state body memory | ≈ `window + 16 KiB` (≈2× window at defaults) | 1 × `windowSizeBytes` (16 KiB default) |

### Before

```
onResponseBody (event-loop thread)
   └─ bodyChan.trySend(buffer)         // unbounded Channel<SdkBuffer>
                                        // ── thread hop ──
writer coroutine (GlobalScope, dispatcher thread)
   └─ for (buffer in bodyChan) { ch.write(buffer); onDataConsumed(wc) }
consumer
   └─ reads `ch`
```

### After

```
onResponseBody (event-loop thread)
   └─ bodyChan.tryWrite(buffer)         // window-sized SdkByteChannel, non-suspending
consumer
   └─ WindowManagedReadChannel.read → onDataConsumed(bytesRead) → crtStream.incrementWindow
```

## The core invariant (why this is safe)

The whole design rests on one coupling: **`bodyChan` capacity == CRT `initialWindowSize`** (both derive from
`config.initialWindowSizeBytes`; wired at `CrtHttpEngine.kt:67` and `ConnectionManager.kt:62`).

CRT's flow-control contract (`aws-crt-kotlin` `HttpStream.incrementWindow`) guarantees the stream window shrinks as body data arrives and **halts delivery when it reaches 0**. Window credit is only returned *after* the consumer
drains bytes (freeing buffer space). Therefore CRT never has more than `windowSizeBytes` un-acked bytes outstanding, so the non-suspending `tryWrite` always has room and `written == wc` provably holds.

Two ordering rules protect this and are commented in the code:

1. **Read side:** free buffer space *before* returning window credit (`WindowManagedReadChannel.read`). Returning credit first would let CRT deliver into a still-full buffer.
2. **Write side:** single-writer — CRT delivers body callbacks sequentially per stream, and `onResponseComplete`/`complete()` only ever *close* the channel, never write.

## HTTP/2 behavior

This change is fully applicable to HTTP/2 — it is not an HTTP/1.1-only optimization.

The one HTTP/2-specific consideration is the `written != wc` fail-fast. However, aws-c-http decrements the per-stream flow-control window (`window_size_self`) and rejects any DATA frame that would exceed it before invoking our body callback, so CRT can never hand us more un-acked bytes than thewindow we sized the channel buffer to — making the non-suspending tryWrite always fit.

## Why this approach was chosen (alternatives considered)

**Alternative A — keep the writer coroutine, just bound the channel.**
Rejected. Bounding the intermediate `Channel` addresses memory but keeps the per-response coroutine, the second
channel, and the extra thread hop — i.e. it fixes the smallest of the costs and none of the CPU/allocation overhead.

**Alternative B — make `onResponseBody` suspend and call the existing suspending `write`.**
Not possible. `onResponseBody` is a **non-suspending native CRT callback** — it cannot suspend to await buffer
space. That's precisely why a new non-suspending `tryWrite` primitive is introduced instead of reusing `write`.

**Alternative C — block the event-loop thread in the callback until space frees.**
Rejected. Blocking a CRT event-loop thread risks stalling all streams multiplexed on that loop and can deadlock against the consumer. The window-sizing invariant makes blocking unnecessary: there is always room by construction.

**Chosen — direct non-suspending write + read-driven window replenishment.**
This is the only option that (a) respects the non-suspending native callback, (b) removes the coroutine/channel/thread-hop overhead, and (c) preserves consumption-driven backpressure. The `tryWrite` primitive is the minimal addition needed to make it work, and the "buffer sized == window" invariant is what lets it be
non-suspending safely.

## Tradeoffs

- **(−) New assumption:** correctness now *requires* buffer size to equal the CRT window and CRT to honor its window. The old unbounded channel tolerated any window. This is guarded by `require(windowSizeBytes > 0)` and a deterministic fail-fast if the invariant is ever violated (see Robustness below).

## New public API

Adds `SdkByteWriteChannel.tryWrite(source, byteCount): Long` — a non-suspending, single-writer, partial-write
primitive:

- Default implementation throws `UnsupportedOperationException`; `RealSdkByteChannel` overrides it; delegating
  wrappers (`SdkByteChannel by delegate`, e.g. `JobChannel`) forward it automatically.
- Marked **`@InternalApi`** — it's a sharp-edged flow-control primitive (silent partial writes, caller-enforced
  backpressure) with a single engine-internal caller, consistent with the sibling `@InternalApi`
  `SdkByteChannel(...)` / `SdkByteReadChannel(content)` factories. No `.api` diff; binary-compatible.

## Robustness / safety nets

- **Invariant violation is fail-fast, not silent.** If `written != wc` ever occurs (e.g. buffer size and window were
  decoupled), the stream is torn down and the consumer observes a diagnostic `IllegalStateException` rather than
  silently truncated data.
- **Consumer-closes-early race** (consumer cancels the body between the `isClosedForWrite` guard and `tryWrite`, on different threads): the write is wrapped in a try/catch that classifies by **channel state** (not exception type,
  since a close-with-cause rethrows an arbitrary throwable) and discards the bytes instead of letting the exception
  escape the native callback.
- **Unexpected body on a bodyless response** (body bytes after a 204/304/`Content-Length: 0`): discarded and reported consumed, so the bounded buffer can't fill and stall the stream.
- **Force-close mid-body** wakes any suspended reader with a `CancellationException`.

## Tests

`SdkByteChannelTest` (common — runs on JVM + native):

- `tryWrite` within/partial/full capacity, returns-zero-when-full, round-trip, read-frees-capacity, zero-byte no-op,
  closed-throws, **closed-with-cause rethrows that cause**, delegation forwarding, unsupported default.

`SdkStreamResponseHandlerTest` (jvmAndNative):

- window replenishment on read (full + incremental partial reads), multi-chunk reassembly, force-close wakes reader,
  consumer-closes-early discards, consumer-cancels-with-cause discards, **`written != wc` fails deterministically**,
  **unexpected body on bodyless response discarded**, **non-positive window rejected**.

## Benchmark results

E2E S3 throughput benchmark, paired base-vs-target, 14 instances per point (mean values shown).
`base` = `main`, `target` = this PR. CPU% and Mem(MB) are per-variant process-level means. Only the **download** path
exercises this change; upload is included as a control.

### Download (affected path)

| Concurrency | Base (Gbps) | Target (Gbps) | Δ Throughput | Base CPU% | Target CPU% | Base Mem (MB) | Target Mem (MB) |
| crt-16  | 2.68 | 3.33 | **+24.2%** | 55.1 | 34.7 | 312 | 328 |
| crt-32  | 3.07 | 3.73 | **+21.4%** | 67.7 | 46.5 | 342 | 362 |
| crt-64  | 3.03 | 3.40 | **+12.1%** | 71.9 | 56.3 | 349 | 378 |
| crt-128 | 2.87 | 3.16 | **+10.2%** | 73.5 | 58.5 | 390 | 443 |
| crt-256 | 2.97 | 2.84 | −4.2%      | 72.8 | 62.5 | 513 | 498 |

### Upload (control — not on this path)

| Concurrency | Base (Gbps) | Target (Gbps) | Δ Throughput | Base CPU% | Target CPU% | Base Mem (MB) | Target Mem (MB) |
| crt-16  | 3.91 | 3.91 | −0.0% | 20.3 | 20.5 | 189 | 191 |
| crt-32  | 6.05 | 6.09 | +0.4% | 37.2 | 38.1 | 190 | 191 |
| crt-64  | 5.56 | 5.55 | −0.3% | 38.7 | 39.5 | 198 | 197 |
| crt-128 | 5.58 | 5.56 | −0.5% | 40.3 | 40.5 | 202 | 200 |
| crt-256 | 5.44 | 5.44 | +0.2% | 43.4 | 41.8 | 209 | 212 |

### Reading the results

- **Throughput (download):** clear win from crt-16 through crt-128 (+24% → +10%), converging to roughly break-even at crt-256 (−4.2%).
- **CPU (download):** target uses **materially less CPU at every concurrency** (e.g. 34.7 vs 55.1 at crt-16;
  62.5 vs 72.8 at crt-256) — removing the per-response writer coroutine, per-chunk channel-node allocations, and a
  cross-thread dispatch hop lowers per-byte cost across the board.
- **Memory:** roughly a wash at process RSS (small deltas, mixed sign). The theoretical per-stream body-buffer
  saving (one window-sized buffer instead of two staged buffers) is real but swamped by total process allocation at
  this scale.
- **Upload:** flat on every metric, confirming the change is isolated to the response path.

**Why crt-256 doesn't win:** target has spare CPU there (62.5% vs 72.8%) yet doesn't gain throughput, so the limiter is not CPU — it's reduced pipelining depth. The old design returned CRT window credit when bytes moved into the downstream channel (decoupled from the consumer, ~2× read-ahead per stream); the new design returns credit only when the consumer reads, with a single window-sized buffer. Under ~256 concurrent streams competing for coroutine scheduling, the per-stream window reaches 0 and delivery briefly stalls waiting for a read — visible as *both* lower CPU and slightly lower throughput.

### Small-request latency (control)

E2E latency benchmark, paired base-vs-target, 15 instances per point (mean values shown). DynamoDB `PutItem`/`GetItem`
are small single-shot request/response operations — included to confirm the change does not regress the common
small-payload API path (their tiny bodies barely exercise the response-streaming path this PR touches).

| Op | Metric | Base (ms) | Target (ms) | Δ | Base CPU% | Target CPU% | Base Mem (MB) | Target Mem (MB) |
|---|---|---|---|---|---|---|---|---|
| PutItem | Avg | 3.56 | 3.67 | +3.1% | 4.9 | 4.9 | 180 | 179 |
| PutItem | P50 | 3.39 | 3.51 | +3.5% | 4.9 | 4.9 | 180 | 179 |
| PutItem | P90 | 4.02 | 4.18 | +4.0% | 4.9 | 4.9 | 180 | 179 |
| PutItem | P99 | 6.38 | 6.50 | +1.8% | 4.9 | 4.9 | 180 | 179 |
| GetItem | Avg | 2.85 | 2.79 | −1.8% | 6.4 | 6.2 | 176 | 174 |
| GetItem | P50 | 2.63 | 2.58 | −2.0% | 6.4 | 6.2 | 176 | 174 |
| GetItem | P90 | 3.40 | 3.37 | −0.9% | 6.4 | 6.2 | 176 | 174 |
| GetItem | P99 | 6.37 | 6.16 | −3.2% | 6.4 | 6.2 | 176 | 174 |

All deltas are within run-to-run noise (cross-instance variance exceeds the deltas). As expected, small API calls are
a wash — latency there is dominated by network round-trip and service time, not the intra-process buffering this PR
changes, so there is no per-chunk overhead to amortize. The throughput/CPU wins above materialize on large streaming
downloads (many body chunks), not single-shot small responses. Note this benchmark does not measure inter-chunk
delivery latency on a large download, which is the latency dimension the change actually affects.

### DIFF
--- a/runtime/protocol/http-client-engines/http-client-engine-crt/jvmAndNative/src/aws/smithy/kotlin/runtime/http/engine/crt/SdkStreamResponseHandler.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-crt/jvmAndNative/src/aws/smithy/kotlin/runtime/http/engine/crt/SdkStreamResponseHandler.kt
@@ -38,9 +38,8 @@ internal class SdkStreamResponseHandler(
     // exception from a callback such that AWS_OP_ERROR is returned. Wait for HttpStream to have explicit cancellation
 
     init {
-        // The non-suspending write path in onResponseBody is only safe if the channel buffer is at least as large as
-        // the CRT flow-control window (the two are sized from the same windowSizeBytes below). Guard the precondition
-        // at construction so a misconfiguration fails fast here rather than as a mid-stream body failure.
+        // Channel buffer and CRT window are both sized from windowSizeBytes; fail fast on a bad value here rather
+        // than as a mid-stream body failure.
         require(windowSizeBytes > 0) { "windowSizeBytes must be > 0, was $windowSizeBytes" }
     }
 
@@ -69,9 +68,9 @@ internal class SdkStreamResponseHandler(
     private var streamCompleted = false
     private var receivedBodyData = false
 
-    // set true once a consumable body channel has been handed to the caller (see createHttpResponseBody). When the
-    // response was signalled with no body (HttpBody.Empty) nothing ever drains bodyChan, so any body bytes that still
-    // arrive must be discarded rather than written — otherwise the bounded buffer fills and the stream stalls.
+    // set true once a consumable body channel has been handed to the caller . When the response was signalled
+    // with no body (HttpBody.Empty) nothing ever drains bodyChan, so any body bytes that still arrive must be
+    // discarded rather than written — otherwise the bounded buffer fills and the stream stalls.
     private var bodyConsumable = false
 
     /**
@@ -99,8 +98,7 @@ internal class SdkStreamResponseHandler(
 
     private fun createHttpResponseBody(contentLength: Long?): HttpBody {
         // Reads from [bodyChan] drive CRT window replenishment: as the downstream consumer drains bytes we hand the
-        // same number of bytes back to the flow-control window. This ties backpressure to actual consumption without
-        // an intermediate writer coroutine — body bytes are written straight into the channel from onResponseBody.
+        // same number of bytes back to the flow-control window.
         val ch = WindowManagedReadChannel(bodyChan, ::onDataConsumed)
         bodyConsumable = true
         return object : HttpBody.ChannelContent() {
@@ -178,9 +176,7 @@ internal class SdkStreamResponseHandler(
         }
 
         // short circuit, stop buffering data and discard remaining incoming bytes. This also covers the case where
-        // the downstream consumer closed/cancelled the body channel early: writing to it would throw, so instead we
-        // discard the remaining bytes (returning them as "consumed" to CRT) rather than surfacing an exception from
-        // this native callback.
+        // the downstream consumer closed/cancelled the body channel early.
         if (isCancelled || bodyChan.isClosedForWrite) {
             return abortStream(stream, bodyBytesIn.len)
         }
@@ -202,36 +198,22 @@ internal class SdkStreamResponseHandler(
             write(bytes)
         }
 
-        // Write directly into the channel buffer without suspending. CRT never has more than the flow-control
-        // window of un-acked bytes outstanding and the channel buffer is sized to that window, so there is always
-        // room. Window credit is returned on the read side as the consumer drains the channel (see onDataConsumed).
-        //
-        // PRECONDITION (single writer): this is the only writer to bodyChan and CRT delivers body callbacks
-        // sequentially per stream, so tryWrite is never called concurrently with itself. onResponseComplete/complete
-        // only ever close the channel (never write), which tryWrite tolerates.
+        // Write directly into the channel buffer without suspending.
         val wc = buffer.size
         val written = try {
             bodyChan.tryWrite(buffer, wc)
         } catch (e: Throwable) {
-            // The consumer may have closed/cancelled the body channel between the isClosedForWrite check above and
-            // here (they run on different threads). A write to a closed channel throws either
-            // ClosedWriteChannelException or the consumer's close cause (an arbitrary Throwable, e.g. the
-            // CancellationException from cancel(null) or a caller-supplied cause). We therefore classify by channel
-            // state, not exception type: if the channel is now closed, this is the expected consumer-went-away race,
-            // so discard the bytes rather than letting the exception escape into the CRT native callback. If the
-            // channel is still open the throwable is a genuine bug and is rethrown.
+            // The consumer may have closed the channel between the isClosedForWrite check above and here, in which
+            // case tryWrite throws the (arbitrary) close cause. If now closed, discard the bytes rather than letting
+            // it escape the CRT native callback; otherwise it's a genuine bug, so rethrow.
             if (bodyChan.isClosedForWrite) {
                 return abortStream(stream, bodyBytesIn.len)
             }
             throw e
         }
 
-        // INVARIANT: written == wc, because the channel buffer is sized to the CRT flow-control window and window
-        // credit is only returned after buffer space is freed (see WindowManagedReadChannel), so outstanding bytes
-        // never exceed capacity. If this is ever violated (e.g. buffer size and window size were decoupled) we would
-        // otherwise silently drop body bytes CRT considers delivered. Rather than throw out of this native callback,
-        // fail the stream deterministically: close the channel with a diagnostic cause so the consumer observes a
-        // clean error, and abort the stream so CRT stops delivering.
+        // INVARIANT: written == wc (buffer sized to the CRT window, credit returned only after space frees). If ever
+        // violated, fail deterministically instead of silently dropping bytes or throwing out of the native callback.
         if (written != wc) {
             val cause = IllegalStateException(
                 "response body channel accepted only $written of $wc bytes; the channel buffer " +
@@ -311,10 +293,8 @@ private class WindowManagedReadChannel(
     private val onConsumed: (Int) -> Unit,
 ) : SdkByteReadChannel by delegate {
     override suspend fun read(sink: SdkBuffer, limit: Long): Long {
-        // INVARIANT (ordering): buffer space MUST be freed before window credit is returned. delegate.read
-        // completes the read (freeing availableForWrite) before we call onConsumed, which increments the CRT window
-        // and lets CRT deliver more. Returning credit first would let CRT write into a still-full buffer, breaking
-        // the "channel capacity == window" guarantee that onResponseBody's non-suspending tryWrite relies on.
+        // ORDERING: free buffer space (delegate.read) before returning window credit (onConsumed). Crediting first
+        // would let CRT deliver into a still-full buffer, breaking the "channel capacity == window" guarantee.
         val rc = delegate.read(sink, limit)
         if (rc > 0) onConsumed(rc.toInt())
         return rc
--- a/runtime/protocol/http-client-engines/http-client-engine-crt/jvmAndNative/src/aws/smithy/kotlin/runtime/http/engine/crt/SdkStreamResponseHandler.kt
+++ b/runtime/protocol/http-client-engines/http-client-engine-crt/jvmAndNative/src/aws/smithy/kotlin/runtime/http/engine/crt/SdkStreamResponseHandler.kt
@@ -50,9 +50,7 @@ internal class SdkStreamResponseHandler(
 
     // Body bytes are written directly into this channel from onResponseBody (a non-suspending CRT callback).
     // The channel buffer is sized to the CRT flow-control window, so no more than [windowSizeBytes] un-acked bytes
-    // are ever in-flight and the synchronous write always has room. The CRT window is replenished on the read side
-    // (see WindowManagedReadChannel) as the downstream consumer drains bytes, preserving consumption-driven
-    // backpressure without an intermediate writer coroutine.
+    // are ever in-flight and the synchronous write always has room.
     private val bodyChan = SdkByteChannel(true, windowSizeBytes)
 
     private val lock = reentrantLock() // protects crtStream and cancelled state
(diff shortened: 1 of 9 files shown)


runtime/protocol/http-client-engines/http-client-engine-crt/jvmAndNative/src/aws/smithy/kotlin/runtime/http/engine/crt/SdkStreamResponseHandler.kt
```diff
@@ -33,17 +31,26 @@ import kotlin.coroutines.CoroutineContext
 internal class SdkStreamResponseHandler(
     private val conn: HttpClientConnection,
     private val callContext: CoroutineContext,
+    private val windowSizeBytes: Int,
 ) : HttpStreamResponseHandler {
     // TODO - need to cancel the stream when the body is closed from the caller side early.
     // There is no great way to do that currently without either (1) closing the connection or (2) throwing an
     // exception from a callback such that AWS_OP_ERROR is returned. Wait for HttpStream to have explicit cancellation
 
+    init {
```
ianbotsf, line 40: **Question**: Is this `TODO` still relevant or does this PR effectively address the problem?

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/io/RealSdkByteChannel.kt
```diff
@@ -116,6 +116,25 @@ internal class RealSdkByteChannel(
...
+            if (wc == 0L) return 0L
+
+            synchronized(lock) {
+                buffer.write(source, wc)
+            }
+            afterWrite(wc.toInt())
+            return wc
+        } finally {
+            _writeInProgress.compareAndSet(true, false)
+        }
```
ianbotsf: **Nit**: The calls to `check(_writeInProgress.compareAndSet(false, true))`, `afterWrite(wc.toInt())`, and `_writeInProgress.compareAndSet(true, false)` are already part of the `writing` method. Why can't we just use that here instead?

```kotlin
override fun tryWrite(source: SdkBuffer, byteCount: Long): Long {
    ensureNotClosed()
    if (byteCount == 0L) return 0L

    return writing {
        val wc = minOf(availableForWrite.toLong(), byteCount)

        if (wc > 0L) {
            synchronized(lock) {
                buffer.write(source, wc)
            }
        }

        wc
    }
}
```

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/io/SdkByteWriteChannel.kt
```diff
@@ -51,6 +53,27 @@ public interface SdkByteWriteChannel : Closeable {
...
+     * The default implementation throws [UnsupportedOperationException]; implementations that can service a
+     * non-suspending write override it (delegating wrappers forward it automatically). Low-level flow-control
+     * primitive for engine internals; not part of the stable API.
+     *
+     * @param source the buffer data will be read from and written to this channel
+     * @param byteCount the number of bytes to read from source
+     * @return the number of bytes actually written
+     */
+    @InternalApi
+    public fun tryWrite(source: SdkBuffer, byteCount: Long = source.size): Long = throw UnsupportedOperationException("${this::class.simpleName} does not support non-suspending writes")
```
ianbotsf, line 72: **Question**: Why do we need to make this an `@InternalApi`? What's wrong with allowing external callers to directly invoke `tryWrite`?

runtime/protocol/http-client-engines/http-client-engine-crt/jvmAndNative/src/aws/smithy/kotlin/runtime/http/engine/crt/SdkStreamResponseHandler.kt
luigi617: I think this is still needed because the current version still use the "not great" workaround like before rather than cancelling the stream.

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/io/RealSdkByteChannel.kt
luigi617: Yes, your suggestion is better, the only thing I changed to make this happen is to specify `writing` output type

runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/io/SdkByteWriteChannel.kt
luigi617: I added `@InternalApi` because I thought this function is for optimization and requires understanding how CRT works due to the [assumption added](https://github.com/smithy-lang/smithy-kotlin/pull/1664#:~:text=(%E2%88%92)%20New%20assumption%3A%20correctness%20now%20requires%20buffer%20size%20to%20equal%20the%20CRT%20window%20and%20CRT%20to%20honor%20its%20window.%20The%20old%20unbounded%20channel%20tolerated%20any%20window.%20This%20is%20guarded%20by%20require(windowSizeBytes%20%3E%200)%20and%20a%20deterministic%20fail%2Dfast%20if%20the%20invariant%20is%20ever%20violated%20(see%20Robustness%20below).). So I thought this function will be used only internally.  But now that I am re-thinking about it, anyone who uses this interface they are responsible for the implementation. So probably exposing it is fine

ianbotsf (approved):

Issue #1667, opened 2026-07-28
Inconsistent `context` and `attributes` parameters in metric recordings across OperationTelemetryInterceptor and SdkOperationExecution


ashishdhingra:
### Describe the bug

Several metric recording sites in the SDK do not pass `context` (telemetry context) or `attributes` (dimensions like `rpc.service`, `rpc.method`) to their `record()`/`add()` calls. This results in:

1. **Missing telemetry context** — providers that use `context: Context?` for correlation (e.g., linking metrics to traces) cannot associate these recordings with the active span
2. **Dimensionless metrics** — metrics recorded without `attributes` produce data points with no dimensions, making them impossible to filter by service/operation in monitoring dashboards

#### `OperationTelemetryInterceptor.kt` — 3 sites missing context

| Recording | Issue |
|-----------|-------|
| `rpcAttemptOverheadDuration.recordSeconds(attemptDuration - ttfb, perRpcAttributes)` | Missing `context` parameter (defaults to `null`) |
| `requestPayloadSize.recordPayloadSize(context.protocolRequest.body.contentLength)` | Missing both `attributes` and `context` (defaults to empty attributes + null context) |
| `responsePayloadSize.recordPayloadSize(context.protocolResponse.body.contentLength)` | Missing both `attributes` and `context` (defaults to empty attributes + null context) |

The other 6 recordings in the same file correctly pass both `perRpcAttributes` and `metrics.provider.contextManager.current()`.

#### `SdkOperationExecution.kt` — 3 sites missing context

| Recording | Issue |
|-----------|-------|
| `resolveIdentityDuration.measureSeconds(schemeAttr)` | Missing `context` parameter |
| `resolveEndpointDuration.measureSeconds(request.context.operationAttributes)` | Missing `context` parameter |
| `signingDuration.measureSeconds(schemeAttr)` | Missing `context` parameter |

These pass `attributes` (scheme attributes or operation attributes) but not the telemetry `context`.

### References

- `OperationTelemetryInterceptor.kt`: https://github.com/smithy-lang/smithy-kotlin/blob/main/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/OperationTelemetryInterceptor.kt
- `SdkOperationExecution.kt`: https://github.com/smithy-lang/smithy-kotlin/blob/main/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/operation/SdkOperationExecution.kt

[...]

PR #1668, opened 2026-07-29
feat: add EMF telemetry provider for CloudWatch Embedded Metric Format

ashishdhingra:
## Issue \#
https://github.com/aws/aws-sdk-kotlin/issues/1638

## Description of changes
Adds a new `telemetry-provider-emf` module that emits SDK operational metrics in [CloudWatch Embedded Metric Format (EMF)](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format.html). In environments with built-in CloudWatch Logs integration (AWS Lambda, Amazon ECS), these metrics are automatically extracted and published as CloudWatch metrics without requiring separate `PutMetricData` API calls.

### Motivation

Addresses [aws-sdk-kotlin#1638](https://github.com/aws/aws-sdk-kotlin/issues/1638) — request for EMF integration similar to the Java SDK's `EmfMetricLoggingPublisher`.

### What's included

- New module: `runtime/observability/telemetry-provider-emf` (entirely in `common` source set — fully multiplatform)
- `EmfTelemetryProvider` — public entry point with builder DSL for `namespace` and `logGroupName`
- `CloudWatchUnit` enum — all valid CloudWatch metric units per the EMF specification
- `EmfJsonWriter` — builds EMF JSON documents using `jsonStreamWriter()` from `serde-json`
- Per-record emission: each instrument's `record()`/`add()` call emits one self-contained EMF JSON document via `println()`
- Unit tests covering JSON structure, unit mapping, validation, and instrument behavior

### Design decisions

**Per-record emission (not batched):** Each metric recording produces one EMF document. CloudWatch charges per unique metric series (name + dimension combination), not per document — so cost is identical to batching.

**`println()` for output (not SLF4J):** Keeps the module entirely in `common` with no platform-specific code. In Lambda/ECS, stdout is automatically ingested by CloudWatch Logs. Users who don't want EMF output simply don't configure the provider — there's no need for log-level suppression.

**`@ExperimentalApi` on `EmfTelemetryProvider`:** The `TelemetryProvider` interface members (`meterProvider`, `loggerProvider`, `tracerProvider`, `contextManager`) are all annotated with `@ExperimentalApi`. Any class implementing `TelemetryProvider` must therefore be annotated with `@ExperimentalApi` as well (or opt in), since consumers of the class implicitly access these experimental interface members. This is consistent with `MicrometerTelemetryProvider` and `OpenTelemetryProvider` which are both `@ExperimentalApi`.

### Usage

```kotlin
val client = S3Client {
    region = "us-west-2"
    telemetryProvider = EmfTelemetryProvider {
        namespace = "MyApp"
    }
}
```

---

### Unit tests

```bash
./gradlew :runtime:observability:telemetry-provider-emf:jvmTest
```

Tests cover:
- EMF JSON structure validation (`EmfJsonWriterTest`)
- CloudWatch unit mapping — known, null, unknown units (`CloudWatchUnitTest`)
- Provider validation — namespace/logGroupName constraints (`EmfTelemetryProviderTest`)
- Instrument emission behavior — stdout capture, negative value rejection (`EmfInstrumentsTest`)

### End-to-end integration test (Lambda + S3Client)

Deployed a Lambda function using `EmfTelemetryProvider` with a real `S3Client.listBuckets()` call to verify EMF output is correctly ingested by CloudWatch.

**Prerequisites:**
- AWS CLI configured with credentials
- Java 21 installed locally
- smithy-kotlin published to Maven local (`./gradlew publishToMavenLocal -x test`)

**Project structure:**

```
emf-lambda-test/
├── build.gradle.kts
├── settings.gradle.kts
└── src/main/kotlin/com/example/Handler.kt
```

**`settings.gradle.kts`:**

```kotlin
rootProject.name = "emf-lambda-test"
```

**`build.gradle.kts`:**

```kotlin
plugins {
    kotlin("jvm") version "2.4.0"
}

repositories {
    mavenLocal()
    mavenCentral()
}

dependencies {
    implementation("aws.sdk.kotlin:s3-jvm:<latest-release>")
    implementation("aws.smithy.kotlin:telemetry-provider-emf-jvm:1.7.4-SNAPSHOT")
    implementation("com.amazonaws:aws-lambda-java-core:1.2.3")
    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.10.2")
}

tasks.register<Zip>("buildZip") {
    into("lib") {
        from(tasks.jar)
        from(configurations.runtimeClasspath)
    }
}
```
> **Note on versions:** `1.7.4-SNAPSHOT` is the local development version from `publishToMavenLocal` on this branch. Once merged and released, replace with the actual release version (e.g., `1.7.4`). The `aws.sdk.kotlin:s3-jvm` version should be the latest release that depends on the corresponding smithy-kotlin release.

**`src/main/kotlin/com/example/Handler.kt`:**

```kotlin
package com.example

import aws.sdk.kotlin.services.s3.S3Client
import aws.smithy.kotlin.runtime.telemetry.emf.EmfTelemetryProvider
import com.amazonaws.services.lambda.runtime.Context
import com.amazonaws.services.lambda.runtime.RequestHandler
import kotlinx.coroutines.runBlocking

@OptIn(aws.smithy.kotlin.runtime.ExperimentalApi::class)
class Handler : RequestHandler<Map<String, String>, String> {

    private val s3 = S3Client {
        region = "us-west-2"
        telemetryProvider = EmfTelemetryProvider {
            namespace = "EmfIntegrationTest"
        }
    }

    override fun handleRequest(event: Map<String, String>, context: Context): String = runBlocking {
        s3.listBuckets()
        "OK"
    }
}
```

**Build and deploy:**

```bash
# 1. Build the Lambda ZIP
cd emf-lambda-test
./gradlew buildZip
# Output: build/distributions/emf-lambda-test.zip

# 2. Create IAM role for Lambda
aws iam create-role \
    --role-name emf-lambda-test-role \
    --assume-role-policy-document '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"Service":"lambda.amazonaws.com"},"Action":"sts:AssumeRole"}]}'

aws iam attach-role-policy \
    --role-name emf-lambda-test-role \
    --policy-arn arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole

aws iam attach-role-policy \
    --role-name emf-lambda-test-role \
    --policy-arn arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess

# Wait for IAM propagation
sleep 10

# 3. Create and invoke the Lambda function
aws lambda create-function \
    --function-name emf-integration-test \
    --runtime java21 \
    --handler com.example.Handler::handleRequest \
    --role arn:aws:iam::<ACCOUNT_ID>:role/emf-lambda-test-role \
    --zip-file fileb://build/distributions/emf-lambda-test.zip \
    --memory-size 512 \
    --timeout 30

aws lambda wait function-active-v2 --function-name emf-integration-test

aws lambda invoke \
    --function-name emf-integration-test \
    --payload '{"test": "s3"}' \
    --cli-binary-format raw-in-base64-out \
    output.json && cat output.json

# 4. Verify CloudWatch Logs contain EMF documents
# Check /aws/lambda/emf-integration-test log group in CloudWatch console

# 5. Cleanup (optional)
aws lambda delete-function --function-name emf-integration-test
aws iam detach-role-policy --role-name emf-lambda-test-role --policy-arn arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
aws iam detach-role-policy --role-name emf-lambda-test-role --policy-arn arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess
aws iam delete-role --role-name emf-lambda-test-role
```

**CloudWatch Logs output (warm invocation — 11 EMF documents per call):**

```
START RequestId: 1aab954e-9ffe-4f2b-9bd0-f7a67710b05b Version: $LATEST
{"_aws":{"Timestamp":1785276862107,"LogGroupName":"/aws/lambda/emf-integration-test","CloudWatchMetrics":[{"Namespace":"EmfIntegrationTest","Dimensions":[["rpc.service","rpc.method"]],"Metrics":[{"Name":"smithy.client.call.serialization_duration","Unit":"Seconds"}]}]},"rpc.service":"S3","rpc.method":"ListBuckets","smithy.client.call.serialization_duration":1.00656e-4}
{"_aws":{"Timestamp":1785276862108,"LogGroupName":"/aws/lambda/emf-integration-test","CloudWatchMetrics":[{"Namespace":"EmfIntegrationTest","Dimensions":[["auth.scheme_id"]],"Metrics":[{"Name":"smithy.client.call.auth.resolve_identity_duration","Unit":"Seconds"}]}]},"auth.scheme_id":"aws.auth#sigv4","smithy.client.call.auth.resolve_identity_duration":2.6723e-5}
{"_aws":{"Timestamp":1785276862108,"LogGroupName":"/aws/lambda/emf-integration-test","CloudWatchMetrics":[{"Namespace":"EmfIntegrationTest","Dimensions":[["rpc.service","rpc.method"]],"Metrics":[{"Name":"smithy.client.call.resolve_endpoint_duration","Unit":"Seconds"}]}]},"rpc.service":"S3","rpc.method":"ListBuckets","smithy.client.call.resolve_endpoint_duration":1.26572e-4}
{"_aws":{"Timestamp":1785276862109,"LogGroupName":"/aws/lambda/emf-integration-test","CloudWatchMetrics":[{"Namespace":"EmfIntegrationTest","Dimensions":[["auth.scheme_id"]],"Metrics":[{"Name":"smithy.client.call.auth.signing_duration","Unit":"Seconds"}]}]},"auth.scheme_id":"aws.auth#sigv4","smithy.client.call.auth.signing_duration":6.73428e-4}
{"_aws":{"Timestamp":1785276862109,"LogGroupName":"/aws/lambda/emf-integration-test","CloudWatchMetrics":[{"Namespace":"EmfIntegrationTest","Dimensions":[[]],"Metrics":[{"Name":"smithy.client.call.request_payload_size","Unit":"Bytes"}]}]},"smithy.client.call.request_payload_size":0.0}
{"_aws":{"Timestamp":1785276862211,"LogGroupName":"/aws/lambda/emf-integration-test","CloudWatchMetrics":[{"Namespace":"EmfIntegrationTest","Dimensions":[[]],"Metrics":[{"Name":"smithy.client.call.response_payload_size","Unit":"Bytes"}]}]},"smithy.client.call.response_payload_size":0.0}
{"_aws":{"Timestamp":1785276862232,"LogGroupName":"/aws/lambda/emf-integration-test","CloudWatchMetrics":[{"Namespace":"EmfIntegrationTest","Dimensions":[["rpc.service","rpc.method"]],"Metrics":[{"Name":"smithy.client.call.deserialization_duration","Unit":"Seconds"}]}]},"rpc.service":"S3","rpc.method":"ListBuckets","smithy.client.call.deserialization_duration":0.018314045}
{"_aws":{"Timestamp":1785276862232,"LogGroupName":"/aws/lambda/emf-integration-test","CloudWatchMetrics":[{"Namespace":"EmfIntegrationTest","Dimensions":[["rpc.service","rpc.method"]],"Metrics":[{"Name":"smithy.client.call.attempts","Unit":"Count"}]}]},"rpc.service":"S3","rpc.method":"ListBuckets","smithy.client.call.attempts":1.0}
{"_aws":{"Timestamp":1785276862233,"LogGroupName":"/aws/lambda/emf-integration-test","CloudWatchMetrics":[{"Namespace":"EmfIntegrationTest","Dimensions":[["rpc.service","rpc.method"]],"Metrics":[{"Name":"smithy.client.call.attempt_duration","Unit":"Seconds"}]}]},"rpc.service":"S3","rpc.method":"ListBuckets","smithy.client.call.attempt_duration":0.124928983}
{"_aws":{"Timestamp":1785276862233,"LogGroupName":"/aws/lambda/emf-integration-test","CloudWatchMetrics":[{"Namespace":"EmfIntegrationTest","Dimensions":[["rpc.service","rpc.method"]],"Metrics":[{"Name":"smithy.client.call.attempt_overhead_duration","Unit":"Seconds"}]}]},"rpc.service":"S3","rpc.method":"ListBuckets","smithy.client.call.attempt_overhead_duration":0.092164673}
{"_aws":{"Timestamp":1785276862233,"LogGroupName":"/aws/lambda/emf-integration-test","CloudWatchMetrics":[{"Namespace":"EmfIntegrationTest","Dimensions":[["rpc.service","rpc.method"]],"Metrics":[{"Name":"smithy.client.call.duration","Unit":"Seconds"}]}]},"rpc.service":"S3","rpc.method":"ListBuckets","smithy.client.call.duration":0.126129929}
END RequestId: 1aab954e-9ffe-4f2b-9bd0-f7a67710b05b
```

CloudWatch automatically extracted all metrics under namespace `EmfIntegrationTest` with dimensions `rpc.service=S3, rpc.method=ListBuckets`.

**CloudWatch Dashboard:**
image

> **Note:** Dimension inconsistency in `auth.resolve_identity_duration`/`auth.signing_duration` (using `auth.scheme_id` instead of `rpc.service`/`rpc.method`) and empty dimensions for payload sizes is tracked in [smithy-kotlin#1667](https://github.com/smithy-lang/smithy-kotlin/issues/1667).

---

## Checklist

- [x] New module compiles on all platforms (`common` source set)
- [x] Unit tests pass (`./gradlew :runtime:observability:telemetry-provider-emf:jvmTest`)
- [x] API dump generated (`api/telemetry-provider-emf.api`)
- [x] Module registered in `settings.gradle.kts`
- [x] End-to-end tested in Lambda with real S3 API calls
- [x] CloudWatch successfully extracts metrics from EMF output
___


runtime/observability/telemetry-provider-emf/common/src/aws/smithy/kotlin/runtime/telemetry/emf/EmfConstants.kt
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.smithy.kotlin.runtime.telemetry.emf
+
+import aws.smithy.kotlin.runtime.InternalApi
+
+@InternalApi
```
luigi617: The object is already internal so no need to add `@InternalApi`.

`@InternalApi` is used when we have to expose the class/function/... (e.g. we need a function to be used in another module), but we want to discourage customer to use them.

ashishdhingra: Good callout. Fixed.

runtime/observability/telemetry-provider-emf/common/src/aws/smithy/kotlin/runtime/telemetry/emf/CloudWatchUnit.kt
```diff
@@ -0,0 +1,61 @@
...
+    "By", "bytes" -> CloudWatchUnit.BYTES
+    "KBy" -> CloudWatchUnit.KILOBYTES
+    "MBy" -> CloudWatchUnit.MEGABYTES
+    "GBy" -> CloudWatchUnit.GIGABYTES
+    "{attempt}", "{error}", "{request}" -> CloudWatchUnit.COUNT
+    "%" -> CloudWatchUnit.PERCENT
+    else -> {
+        emfLog("[WARN] Unmapped OTel unit '$unit' — defaulting to None. Consider adding a mapping.")
+        CloudWatchUnit.NONE
+    }
```
ianbotsf, line 60: **Question**: Would it be safer (or more generally correct) to default to `CloudWatchUnit.COUNT` in the case of an unrecognized unit string?

runtime/observability/telemetry-provider-emf/common/src/aws/smithy/kotlin/runtime/telemetry/emf/EmfJsonWriter.kt
```diff
@@ -0,0 +1,86 @@
...
+        writer.endObject()
+        writer.endArray()
+        writer.endObject()
+
+        @Suppress("UNCHECKED_CAST")
+        for (key in dimensionKeys) {
+            val attributeKey = key as? AttributeKey<Any> ?: continue
+            writer.writeName(key.name)
+            writer.writeValue(attributes.getOrNull(attributeKey).toString())
+        }
```
ianbotsf: **Nit**: Suppression can be narrowed to the `val attributeKey = ...` line.

runtime/observability/telemetry-provider-emf/common/src/aws/smithy/kotlin/runtime/telemetry/emf/EmfJsonWriter.kt
```diff
@@ -0,0 +1,86 @@
...
+    ): String {
+        val writer = jsonStreamWriter()
+
+        writer.beginObject()
+
+        writer.writeName(EmfConstants.AWS_METADATA_KEY)
+        writer.beginObject()
+
+        writer.writeName(EmfConstants.TIMESTAMP_KEY)
+        writer.writeValue(Instant.now().epochMilliseconds)
```
ianbotsf: **Style**: This method is too long and too difficult to read through. Please refactor it into smaller chunks that make it more maintainable.

**Optional**: You _could_ make use of Kotlin's extension methods and DSL features to help simplify and improve the code simultaneously.

Sample DSL usage:

```kotlin
val writer = jsonStreamWriter()

writer.withObject {
    withObject(EmfConstants.AWS_METADATA_KEY) {
        writeEntry(EmfConstants.TIMESTAMP_KEY, Instant.now().epochMilliseconds)
        writeEntryIfNotNull(EmfConstants.LOG_GROUP_NAME_KEY, logGroupName)

        withArray(EmfConstants.CLOUDWATCH_METRICS_KEY) {
            withObject {
                writeEntry(EmfConstants.NAMESPACE_KEY, namespace)

                withArray(EmfConstants.DIMENSIONS_KEY) {
                    withArray {
                        attributes
                            .keys
                            .take(EmfConstants.MAX_DIMENSIONS_PER_SET)
                            .forEach { key -> writeValue(key.name) }
                    }
                    ...
                }
                ...
            }
            ...
        }
        ...
    }
    ...
}
```

Sample DSL implementation:

```kotlin
fun JsonStreamWriter.withObject(block: JsonStreamWriter.() -> Unit) {
    beginObject()
    block()
    endObject()
}

fun JsonStreamWriter.withObject(key: String, block: JsonStreamWriter.() -> Unit) {
    writeName(key)
    withObject(block)
}

...
```

This DSL part is just a suggestion, not a requirement.

ianbotsf (changes requested):

runtime/observability/telemetry-provider-emf/common/src/aws/smithy/kotlin/runtime/telemetry/emf/CloudWatchUnit.kt
ashishdhingra: Good question. I think `None` is the safer default, for two reasons:

1. `None` is CloudWatch's own documented value for "this value has no unit" ([MetricDatum.Unit](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_MetricDatum.html) valid values include `None`), so it makes no claim about the data. `Count` is a positive assertion that the value is a cardinality. If an unmapped unit shows up that is *not* a count — e.g. `ns`, `{connection}/s`, `Kib` — labeling it `Count` mislabels it, and CloudWatch surfaces the unit in graph axes, alarm definitions, and Metric Math. A wrong unit is harder to notice and debug than an absent one.

2. It matches what the Java SDK does. [`MetricEmfConverter.getMetricUnit`](https://github.com/aws/aws-sdk-java-v2/blob/master/metric-publishers/emf-metric-logging-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverter.java) returns `Milliseconds` for `Duration` and `null` for everything else, and the `null` case omits the `Unit` field entirely — i.e. its fallback is also "no unit", not `Count`.

The `else` branch also logs a warning naming the unmapped unit, so a new instrument with an unrecognized unit is visible during development rather than silently mislabeled.

Happy to switch if you still prefer `Count` — but wanted to lay out the reasoning first.

runtime/observability/telemetry-provider-emf/common/src/aws/smithy/kotlin/runtime/telemetry/emf/EmfJsonWriter.kt
ashishdhingra: Done — the cast and its suppression are now colocated on a single line inside `writeDimensionValues`.

runtime/observability/telemetry-provider-emf/common/src/aws/smithy/kotlin/runtime/telemetry/emf/EmfJsonWriter.kt
ashishdhingra: Done, and I took the DSL suggestion — thanks, it reads much better.

Added `JsonStreamWriterExt.kt` with `withObject`/`withArray`/`writeEntry`/`writeEntryIfNotNull` extensions, and split `buildEmfDocument` into four focused private helpers, each itself a `JsonStreamWriter` extension so the nesting is structural rather than manual `begin`/`end` pairing:

- `writeMetadata` — the `_aws` node
- `writeDimensionKeys` — the `Dimensions` node
- `writeMetricDefinition` — the `Metrics` node
- `writeDimensionValues` — the top-level dimension target members

`buildEmfDocument` is now down to a 4-line body. Emitted JSON is byte-for-byte unchanged — existing `EmfJsonWriterTest`/`EmfInstrumentsTest` assertions pass without modification.

runtime/observability/telemetry-provider-emf/common/src/aws/smithy/kotlin/runtime/telemetry/emf/CloudWatchUnit.kt
ianbotsf: That makes sense, thanks for looking into the options.

runtime/observability/telemetry-provider-emf/common/src/aws/smithy/kotlin/runtime/telemetry/emf/EmfConstants.kt
```diff
@@ -0,0 +1,31 @@
...
+    const val ENV_LAMBDA_LOG_GROUP = "AWS_LAMBDA_LOG_GROUP_NAME"
+
+    const val MAX_METRICS_PER_DOCUMENT = 100
+    const val MAX_DIMENSIONS_PER_SET = 30
+    const val MAX_VALUES_PER_METRIC = 100
+    const val MAX_NAMESPACE_LENGTH = 1024
+    const val MAX_LOG_GROUP_NAME_LENGTH = 512
+    const val MAX_METRIC_NAME_LENGTH = 1024
+    const val MAX_DIMENSION_KEY_LENGTH = 250
+    const val MAX_DIMENSION_VALUE_LENGTH = 1024
```
luigi617, line 27: There are several `const val` variable defined but not used (`EMF_LOGGER_NAME`, `MAX_METRICS_PER_DOCUMENT`, `MAX_VALUES_PER_METRIC`, `MAX_METRIC_NAME_LENGTH`, `MAX_DIMENSION_KEY_LENGTH`, `MAX_DIMENSION_VALUE_LENGTH`). Since most of them are constraint variable, I was wondering if I should use them to validate the EMF log. It seems we are not validating all the constraits

runtime/observability/telemetry-provider-emf/common/src/aws/smithy/kotlin/runtime/telemetry/emf/EmfMeter.kt
```diff
@@ -0,0 +1,50 @@
...
+) : Meter {
+
+    override fun createUpDownCounter(name: String, units: String?, description: String?): UpDownCounter = EmfUpDownCounter(namespace, logGroupName, name, units)
+
+    override fun createAsyncUpDownCounter(
+        name: String,
+        callback: LongUpDownCounterCallback,
+        units: String?,
+        description: String?,
+    ): AsyncMeasurementHandle = AsyncMeasurementHandle.None
```
luigi617, line 29: Question: EMF does not include any metric about async? Why this function is an no-op?

runtime/observability/telemetry-provider-emf/common/test/aws/smithy/kotlin/runtime/telemetry/emf/EmfInstrumentsTest.kt
```diff
@@ -0,0 +1,117 @@
...
+
+import aws.smithy.kotlin.runtime.ExperimentalApi
+import aws.smithy.kotlin.runtime.collections.attributesOf
+import aws.smithy.kotlin.runtime.collections.emptyAttributes
+import kotlin.test.Test
+import kotlin.test.assertContains
+import kotlin.test.assertTrue
+
+@OptIn(ExperimentalApi::class)
+class EmfInstrumentsTest {
```
luigi617, line 15: Why is this test in `jvm` directory? is there any specific reason (maybe due to `captureEmfOutput` function) that we are testing it only in `jvm` platform?

I think if a test is platform specific, we should also add test to other platform (kotlin native for now) if the same functionality is provided also for other platform. I think EMF Instrument tests are provided also for K/N, right?

runtime/observability/telemetry-provider-emf/common/src/aws/smithy/kotlin/runtime/telemetry/emf/EmfJsonWriter.kt
```diff
@@ -0,0 +1,92 @@
...
+     */
+    fun buildEmfDocument(
+        namespace: String,
+        logGroupName: String?,
+        metricName: String,
+        metricValue: Double,
+        metricUnit: CloudWatchUnit,
+        attributes: Attributes,
+    ): String {
+        val dimensionKeys = attributes.keys.take(EmfConstants.MAX_DIMENSIONS_PER_SET)
```
luigi617: I think `.take(EmfConstants.MAX_DIMENSIONS_PER_SET)` will silently drop the overflow, do you think it is better to give some sort of logging/waring to notify that some dimentsions are not being used?

luigi617 (changes requested):

runtime/observability/telemetry-provider-emf/common/src/aws/smithy/kotlin/runtime/telemetry/emf/EmfConstants.kt
ashishdhingra: Good catch — there were actually 7 unused (`METRIC_STORAGE_RESOLUTION_KEY` too). Handled in three ways:

- **Deleted as dead**: `EMF_LOGGER_NAME` (leftover from an earlier SLF4J design that `emfLog`/`println` replaced) and `METRIC_STORAGE_RESOLUTION_KEY` (we never emit `StorageResolution`).
- **Deleted as unreachable**: `MAX_METRICS_PER_DOCUMENT` and `MAX_VALUES_PER_METRIC`. In the per-record design each document holds exactly one metric with a scalar value, so nothing can violate them. (Java has the equivalents and *does* use them, because it batches.)
- **Now enforced**: `MAX_METRIC_NAME_LENGTH`, `MAX_DIMENSION_KEY_LENGTH`, `MAX_DIMENSION_VALUE_LENGTH` — you were right that these were a real gap.

Enforcement truncates-and-warns rather than throwing, since these are per-record and a telemetry provider should not fail the operation it is measuring. Same approach as Java's `MetricEmfConverter` (`subList` + `logger.warn`). Namespace/logGroup stay as `require()` because those are validated once in `Builder.build()`.

Worth noting these are unreachable with today's instruments — longest SDK metric name is 49 chars vs the 1024 limit — so this is hardening for custom/future instruments, not a live bug.

runtime/observability/telemetry-provider-emf/common/src/aws/smithy/kotlin/runtime/telemetry/emf/EmfJsonWriter.kt
ashishdhingra: Agreed, fixed. Overflow now logs a warning naming the dropped keys — silently discarding them is especially bad here since dimensions define a metric's identity in CloudWatch, so the metric would land on an unexpected series with no indication why.

While doing this I restructured slightly: dimensions are now resolved once into a normalized list that feeds both the `Dimensions` array and the root target members. The spec requires those two to agree, and applying the new key truncation independently at each write site could have desynchronized them into an invalid document.

Added tests for the drop, for each truncation, and one asserting the `Dimensions`/target-member invariant directly. I mutation-tested that last one (writing mismatched key names) to confirm it actually fails.

runtime/observability/telemetry-provider-emf/common/test/aws/smithy/kotlin/runtime/telemetry/emf/EmfInstrumentsTest.kt
ashishdhingra: Good catch, fixed in 22fa56a8. It was JVM-only only because `captureEmfOutput` used `System.setOut` — nothing under test is platform-specific.

Rather than adding an `expect`/`actual` stdout capturer (K/N has no portable equivalent), I made the EMF output destination overridable within the module and capture there. The test now lives in `commonTest` and runs on both JVM and mingwX64 (33 tests each). Capturing into a `List<String>` also lets the tests assert how many documents were emitted, which the old trimmed-stdout assertions couldn't distinguish.

One note: `emfSink` is unsynchronized mutable global state. Only tests write to it and each restores the prior value, so I didn't pull in atomicfu — happy to if you'd prefer.

runtime/observability/telemetry-provider-emf/common/src/aws/smithy/kotlin/runtime/telemetry/emf/EmfMeter.kt
ashishdhingra: OTel and Micrometer both implement these, but by delegating to a backend that already owns a collection loop (the OTel SDK's MetricReader, Micrometer's MeterRegistry) — and the user constructs and shuts that backend down, so   lifecycle comes for free. EMF just writes documents to stdout, so async would mean EMF owning a scheduler: TelemetryProvider has no close() to stop it, and on Lambda the runtime freezes between invocations, so periodic collection would sample at the wrong times. Both those providers are also JVM-only; EMF is common and has to work on K/N.

luigi617 (approved):
Look goods to me

runtime/observability/telemetry-provider-emf/README.md
```diff
@@ -0,0 +1,111 @@
+# EMF Telemetry Provider
+
+A telemetry provider that emits SDK metrics in [CloudWatch Embedded Metric Format (EMF)](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format.html).
+
+Metrics are written as structured JSON to stdout. In environments with built-in CloudWatch Logs
+integration, those logs are ingested automatically and CloudWatch extracts the metrics from them — no
```
ianbotsf: **Nit**:

> In environments with built-in CloudWatch Logs integration, those logs are ingested automatically...

I think it's more correct to say "environments running CloudWatch Agent". A fresh EC2 instance is "integrated" with CloudWatch Logs but doesn't get EMF support for free—the user has to install CloudWatch Agent.

runtime/observability/telemetry-provider-emf/README.md
```diff
@@ -0,0 +1,111 @@
...
+This must be called exactly once, before any client is constructed — a second call throws.
+`GlobalTelemetryProvider` lives in the `telemetry-defaults` module, which you will need on your
+classpath.
+
+### Options
+
+| Option | Default | Notes |
+|---|---|---|
+| `namespace` | `AwsSdk/KotlinSdk` | The CloudWatch namespace metrics are published under. 1–1024 characters. |
+| `logGroupName` | `AWS_LAMBDA_LOG_GROUP_NAME` | Included as the `LogGroupName` field. Omitted from the document if null. See below. |
```
ianbotsf: **Nit**: This line and several others in the **README.md** refer to the emitted JSON data as "documents", which may be technically correct but likely to cause confusion without first qualifying them as "JSON documents". I recommend replacing the word "document" in user-facing documentation with "metrics" or "metrics JSON".

runtime/observability/telemetry-provider-emf/README.md
```diff
@@ -0,0 +1,111 @@
...
+`logGroupName` only labels the document; it does not create or route to a log group. Where the output
+lands is determined entirely by whatever collects your process's stdout.
+
+## What gets emitted
+
+One JSON document per recorded value, on its own line. A single call duration looks like:
+
+```json
+{"_aws":{"Timestamp":1753920000000,"LogGroupName":"/aws/lambda/my-fn","CloudWatchMetrics":[{"Namespace":"MyApp","Dimensions":[["rpc.service","rpc.method"]],"Metrics":[{"Name":"smithy.client.call.duration","Unit":"Seconds"}]}]},"rpc.service":"S3","rpc.method":"GetObject","smithy.client.call.duration":0.042}
+```
```
ianbotsf: **Style**: Please pretty-print the JSON so it's easier for users to read. You can call out that the _real_ metrics JSON is minified.

ianbotsf (approved):
Minor feedback only; fix and ship!

runtime/observability/telemetry-provider-emf/README.md
ashishdhingra: Done

runtime/observability/telemetry-provider-emf/README.md
ashishdhingra: Done

runtime/observability/telemetry-provider-emf/README.md
ashishdhingra: Done

PR #1686, opened 2026-08-14
perf(cbor): deserialize directly from the input buffer

luigi617:
Decode CBOR arguments, integers, floats, string/byte-string payloads, and container/head bytes by reading **directly** from the `SdkBufferedSource` (`readByte`/`readShort`/`readInt`/`readLong`/`readByteArray`) instead of copying each value through an intermediate `SdkBuffer`, allocating throwaway `ByteArray`s, and folding them back into numbers (which also boxed a `ULong` per byte). This is the read-path counterpart to the serialize-path optimization in #1683. Decoding is value-for-value identical (guarded by the CBOR deserializer hex-vector tests).

## Description of changes

- **`decodeArgument`** (Minor.kt) now reads the CBOR argument with fixed-width big-endian primitives (`readByte`/`readShort`/`readInt`/`readLong`) instead of `readFully`-ing into a temporary `SdkBuffer`, calling `readByteArray()`, and folding the bytes back to a `ULong`. This is on the hottest path — it runs for every integer, every length prefix (string/bytes/list/map), and every tag id.
- **Floats** (Numbers.kt): `Float16`/`Float32`/`Float64` decode via `readShort`/`readInt`/`readLong` + `fromBits` instead of `readByteArray(n)` + a boxed big-endian fold.
- **Strings / byte strings** (Collections.kt): definite-length `TextString`/`ByteString` read straight into a single `ByteArray` with `readByteArray(length)`, dropping the intermediate `SdkBuffer` and the redundant second copy. Length semantics (`toInt()`) and `decodeToString()` are preserved exactly.
- **Head-byte peeking** (Major.kt, Minor.kt, Value.kt, CborUtils.kt): added a shared `peekHead` primitive plus `majorOf`/`minorOf` extractors so a single peek of the head byte serves both major and minor extraction. `peekMajor` now indexes `Major.entries` directly instead of a linear `firstOrNull` scan, and `Value.decode` / `nextValueIsIndefiniteBreak` / `nextValueIsNull` peek the head byte once instead of 2–3 times.

No public API changes — all edits are to `internal` code, and `apiCheck` is unchanged. The `Value` wrapper objects (`UInt`, `TextString`, etc.) are intentionally left as-is; only the buffer reads underneath them were made direct.

## Benchmark results

Measured with the serde CBOR benchmarks (`rpcv2Cbor`), 5 paired instances per benchmark. Negative `Δ%` means the target (this change) is faster.

```
GetItemOutputBinary_L                      135701.4   102737.9   -24.29% -11.01***      5
GetItemOutputBinary_M                       22566.5    16871.7   -25.24% -10.65***      5
GetItemOutputBinary_S                        7191.4     5335.6   -25.81% -20.06***      5
GetItemOutput_Baseline                        782.5      734.9    -6.08%     -0.96      5
GetItemOutput_L                            137723.9   106186.3   -22.90% -15.90***      5
GetItemOutput_M                             22713.4    17130.6   -24.58% -14.31***      5
GetItemOutput_S                              7736.5     5813.0   -24.86% -15.02***      5
PutItemRequest_Baseline                      1818.4     1816.6    -0.10%     -0.06      5
PutItemRequest_BinaryData_L                 17632.4    17429.1    -1.15%     -1.00      5
PutItemRequest_BinaryData_M                  5993.5     5986.2    -0.12%     -0.07      5
PutItemRequest_BinaryData_S                  2324.1     2341.9    +0.76%      1.64      5
PutItemRequest_MixedItem_L                  53597.4    53599.4    +0.00%      0.01      5
PutItemRequest_MixedItem_M                  19367.4    19252.3    -0.59%     -0.76      5
PutItemRequest_MixedItem_S                   5076.1     5166.4    +1.78%    2.65**      5
PutItemRequest_Nested_L                      9831.0     9864.7    +0.34%      0.28      5
PutItemRequest_Nested_M                      3820.8     3945.4    +3.26%      1.36      5
PutItemRequest_ShallowMap_L                 57185.7    56113.6    -1.87%    -2.06*      5
PutItemRequest_ShallowMap_M                 20204.1    19993.9    -1.04%     -0.79      5
PutItemRequest_ShallowMap_S                  5134.7     5097.0    -0.74%     -0.55      5
```

### Highlights

- Deserialization-dominated read paths benefit most: all `GetItemOutput*` / `GetItemOutputBinary*` variants improve **−22% to −26%** (highly significant, `***`), directly reflecting the eliminated per-value buffer/array allocations and copies on the decode path.
- Serialization-dominated `PutItemRequest_*` paths are essentially flat (mostly within noise, `Δ%` near 0 and statistically insignificant), as expected — this change targets the deserialize path and leaves encoding untouched. The two small positive deltas (`MixedItem_S` +1.78%, `Nested_M` +3.26%) fall on serialize-dominated benchmarks unaffected by these changes and are within run-to-run variance.


runtime/serde/serde-cbor/common/src/aws/smithy/kotlin/runtime/serde/cbor/encoding/Minor.kt
```diff
@@ -33,10 +29,11 @@ internal enum class Minor(val value: UByte) {
 
 internal val MINOR_BYTE_MASK: UByte = 0b11111u
 
-internal fun peekMinorByte(buffer: SdkBufferedSource): UByte {
-    val byte = buffer.peek().readByte().toUByte()
-    return byte and MINOR_BYTE_MASK
-}
+internal fun peekHead(buffer: SdkBufferedSource): UByte = buffer.peek().readByte().toUByte()
```
ianbotsf, line 32: **Nit**: This function is in **Minor.kt** but it's used by **Major.kt** so it doesn't seem _specific_ to minor values. It probably belongs somewhere more general, such as **CborUtils.kt**.

runtime/serde/serde-cbor/common/src/aws/smithy/kotlin/runtime/serde/cbor/encoding/Minor.kt
```diff
@@ -45,17 +42,11 @@ internal fun decodeArgument(buffer: SdkBufferedSource): ULong {
...
-        Minor.ARG_1.value -> 1L
-        Minor.ARG_2.value -> 2L
-        Minor.ARG_4.value -> 4L
-        Minor.ARG_8.value -> 8L
+    return when (minor) {
+        Minor.ARG_1.value -> buffer.readByte().toUByte().toULong()
+        Minor.ARG_2.value -> buffer.readShort().toUShort().toULong()
+        Minor.ARG_4.value -> buffer.readInt().toUInt().toULong()
+        Minor.ARG_8.value -> buffer.readLong().toULong()
         else -> throw DeserializationException("Unsupported minor value $minor, expected one of ${Minor.ARG_1.value}, ${Minor.ARG_2.value}, ${Minor.ARG_4.value}, ${Minor.ARG_8.value}")
```
ianbotsf, line 50: **Nit**: The `.toULong()` tail call is common to all branches so move it outside the `when` block:

```kotlin
return when (minor) {
    Minor.ARG_1.value -> buffer.readByte().toUByte()
    Minor.ARG_2.value -> buffer.readShort().toUShort()
    Minor.ARG_4.value -> buffer.readInt().toUInt()
    Minor.ARG_8.value -> buffer.readLong()
    else -> throw DeserializationException(...)
}.toULong()
```

ianbotsf (approved):

runtime/serde/serde-cbor/common/src/aws/smithy/kotlin/runtime/serde/cbor/encoding/Minor.kt
luigi617: `CborUtils.kt` lives in the outer `...serde.cbor` package and already depends on the inner `...cbor.encoding` package. So, I think it would be incorrect to add it in CborUtils. Do you think it is better to create a tiny neutral file in the encoding package that include this function or leave it here?

runtime/serde/serde-cbor/common/src/aws/smithy/kotlin/runtime/serde/cbor/encoding/Minor.kt
luigi617: This won't compile because for a `when` expression to chained to an extention function (`.toULong()`), all branches must evaluate to the same type (or a common supertype that supports that function). So, for the code example you gave, the compiler will say `Any` does not have `toULong` function

PR #1687, opened 2026-08-14
perf(cbor): peek the head byte without allocating a peek-source chain

luigi617:
Peek at the next CBOR head byte during deserialization via okio's O(1) indexed buffer access
instead of `buffer.peek().readByte()`. Each `SdkBuffer.peek()` allocated a ~3–4 object okio
peek-source chain (okio `PeekSource` + `SdkSource` adapter + buffered wrapper) just to read one
byte and immediately discard the chain. `peekHead` is the single choke point behind `peekMajor`,
`peekMinorByte`, `nextValueIsNull`, and `nextValueIsIndefiniteBreak` — each invoked multiple times
per field/element/value before anything is decoded — so this was the dominant per-value allocation
on the decode path. The peek remains read-only and consumes nothing; decoded output is unchanged
(guarded by the CBOR deserializer success/error test vectors).

## Description of changes

- **`SdkBuffer.headByte()`** (new `@InternalApi` accessor in `runtime-core`) reads the
  next-to-be-read byte via okio's O(1) indexed access (`inner.require(1L); inner[0L]`), wrapped in
  `wrapOkio` so it throws the runtime's `EOFException` on an empty buffer. This keeps okio fully
  encapsulated inside `runtime-core` — the accessor returns a plain `Byte`, and no okio type leaks
  across the module boundary into `serde-cbor`.
- **`peekHead`** now routes concrete `SdkBuffer` sources through `headByte()` instead of
  `buffer.peek().readByte()`. During CBOR deserialization the source is always a concrete
  `SdkBuffer` — the deserializer constructs one and threads it through every decode call — so the
  fast path is taken universally. The original `peek().readByte()` is retained as a fallback for
  any non-`SdkBuffer` source.
- **EOF semantics are preserved.** The old `peek().readByte()` threw on an empty buffer; the new
  path throws the same translated `EOFException` via `require(1L)`, which also guarantees the
  subsequent indexed read is in-bounds.

## Benchmark results

Measured with the serde CBOR benchmarks (`rpcv2Cbor`), 5 paired instances per benchmark.
Negative `Δ%` means the target (this change) is faster. **0 regressions; 9 statistically
significant improvements**, all on the deserialize/read path.

```
 BASE vs TARGET Comparison
 Instances w/ base: 5 | w/ target: 5 | paired instances: 5

GetItemOutputBinary_L                      105413.8    51792.3   -50.87% -36.35***      5
GetItemOutputBinary_M                       17463.1     9130.5   -47.72% -33.69***      5
GetItemOutputBinary_S                        5413.0     3161.8   -41.59% -18.86***      5
GetItemOutput_Baseline                        839.4      614.2   -26.82%  -8.25***      5
GetItemOutput_L                            110559.8    55578.1   -49.73% -23.89***      5
GetItemOutput_M                             17742.0     9385.8   -47.10% -30.69***      5
GetItemOutput_S                              5870.5     3519.4   -40.05% -32.59***      5
PutItemRequest_Baseline                      1728.8     1720.4    -0.49%     -0.27      5
PutItemRequest_BinaryData_L                 14767.7    14373.0    -2.67%  -3.71***      5
PutItemRequest_BinaryData_M                  4953.0     4903.2    -1.01%     -0.55      5
PutItemRequest_BinaryData_S                  2014.1     1956.0    -2.88%   -2.62**      5
PutItemRequest_MixedItem_L                  36929.4    36862.4    -0.18%     -0.17      5
PutItemRequest_MixedItem_M                  13933.7    13950.6    +0.12%      0.13      5
PutItemRequest_MixedItem_S                   4206.4     4254.2    +1.14%      1.09      5
PutItemRequest_Nested_L                      8319.1     8275.1    -0.53%     -0.38      5
PutItemRequest_Nested_M                      3647.8     3747.0    +2.72%      0.90      5
PutItemRequest_ShallowMap_L                 44039.3    44007.1    -0.07%     -0.08      5
PutItemRequest_ShallowMap_M                 16263.2    16297.0    +0.21%      0.18      5
PutItemRequest_ShallowMap_S                  4739.6     4722.2    -0.37%     -0.25      5
```

### Highlights

- Deserialization-heavy read paths benefit most: `GetItemOutputBinary_L` **−50.87%**,
  `GetItemOutput_L` **−49.73%**, `GetItemOutputBinary_M` **−47.72%**, `GetItemOutput_M`
  **−47.10%** — all highly significant (`***`). Even the tiny `GetItemOutput_Baseline` improves
  **−26.82%**, reflecting how much of the small-payload cost was the per-peek allocation.
- Serialization-dominated write paths (`PutItemRequest*`) move within noise (**−2.9% to +2.7%**,
  none significant beyond the two small binary-data cases), as expected — this change targets the
  deserialize path only.
- No regressions: the four small positive deltas are all statistically insignificant
  (|t| ≤ 1.09) and confined to the untouched serialize path.


runtime/serde/serde-cbor/common/src/aws/smithy/kotlin/runtime/serde/cbor/encoding/Minor.kt
```diff
@@ -29,7 +31,10 @@ internal enum class Minor(val value: UByte) {
 
 internal val MINOR_BYTE_MASK: UByte = 0b11111u
 
-internal fun peekHead(buffer: SdkBufferedSource): UByte = buffer.peek().readByte().toUByte()
+internal fun peekHead(buffer: SdkBufferedSource): UByte = when (buffer) {
+    is SdkBuffer -> buffer.headByte().toUByte()
+    else -> buffer.peek().readByte().toUByte()
+}
```
ianbotsf: **Nit**: `.toUByte()` can be extracted from the `when` block

ianbotsf: **Question**: Seeing this `else` case makes me wonder—can we add `peekByte` directly to the `SdkBufferedSource` interface so that `BufferedSourceAdapter` can benefit too?

ianbotsf (commented):
**Style**: I don't like the word "head" in these names. All reads/writes on sources/sinks happen at the "head". I'd prefer something like `peekByte`, which eliminates the unnecessary word "head" and conveys that the source isn't modified without needing to read the KDocs.

ianbotsf (approved):

runtime/serde/serde-cbor/common/src/aws/smithy/kotlin/runtime/serde/cbor/encoding/Minor.kt
luigi617: Agree, I think adding in the interface is cleaner, here is the result after adding into the interface so `BufferedSourceAdapter` has also the `peekByte` function. Overall a ~2% of improvement comparing to the previous version (this can be just variance but a consistent -2% in all benchmarks make me think that this is real)

```
GetItemOutputBinary_L                      110353.4    52840.9   -52.12% -41.58***      5
GetItemOutputBinary_M                       18274.1     9253.8   -49.36% -23.71***      5
GetItemOutputBinary_S                        5673.1     3129.4   -44.84% -17.16***      5
GetItemOutput_Baseline                        728.4      620.9   -14.76%    -2.08*      5
GetItemOutput_L                            112875.0    57095.7   -49.42% -33.81***      5
GetItemOutput_M                             18574.1     9604.1   -48.29% -31.20***      5
GetItemOutput_S                              6136.8     3526.7   -42.53% -23.49***      5
PutItemRequest_Baseline                      1794.3     1782.0    -0.69%     -1.08      5
PutItemRequest_BinaryData_L                 15130.0    15408.1    +1.84%      1.49      5
PutItemRequest_BinaryData_M                  5156.9     5064.9    -1.78%     -0.76      5
PutItemRequest_BinaryData_S                  2051.9     2060.2    +0.40%      0.18      5
PutItemRequest_MixedItem_L                  37676.1    37505.1    -0.45%     -0.56      5
PutItemRequest_MixedItem_M                  14384.4    14177.8    -1.44%     -1.71      5
PutItemRequest_MixedItem_S                   4328.4     4366.3    +0.87%      0.96      5
PutItemRequest_Nested_L                      8454.8     8533.1    +0.93%      0.84      5
PutItemRequest_Nested_M                      3869.2     3783.2    -2.22%     -1.45      5
PutItemRequest_ShallowMap_L                 45080.4    45457.1    +0.84%      1.16      5
PutItemRequest_ShallowMap_M                 16731.6    16869.5    +0.82%      0.98      5
PutItemRequest_ShallowMap_S                  4912.4     4903.0    -0.19%     -0.15      5
```

PR #1693, opened 2026-08-20
perf(serde): cache field names

luigi617:
Pre-encode the generated serde **field names** once on their field-name trait objects and reuse the cached result on every serialize call, instead of re-encoding each field name on every field write. Previously each `field(...)`/`structField`/`listField`/`mapField`/`nullField` invocation re-derived the on-the-wire field name from scratch — CBOR re-emitted the string's major-type header + UTF-8 bytes, JSON re-ran `escape()` and re-quoted the name, and form-url re-encoded the name (and, for `QueryLiteral`, re-percent-encoded the literal and rebuilt a throwaway descriptor). After this change the encoded form is computed **lazily once** (at first use, then memoized on the trait) and written directly via `buffer.write(...)`/`append(...)` on every subsequent call, so the name-encoding work is amortized across all fields and all requests/responses. The generated wire output is **byte-for-byte identical** — only the *when/how-often* of field-name encoding changes, not the names, field indices, ordering, escaping/encoding rules, or serialization logic.

## Description of changes

- **CBOR** (`CborFieldTraits.kt`, `CborSerializer.kt`): `CborSerialName` now lazily computes `encoded: ByteArray` — the complete CBOR-encoded field name (STRING major-type header written via `writeArgument` + the UTF-8 name bytes). A new internal `SdkFieldDescriptor.serialNameBytes` exposes it. `CborSerializer` adds `writeFieldName(descriptor) = buffer.write(descriptor.serialNameBytes)` and every field/struct/list/map/null override now calls it and then serializes the value directly, replacing the old `entry(descriptor.serialName, value)` / `buffer.write(TextString(descriptor.serialName))` calls that re-encoded the name on each write.
- **JSON** (`JsonFieldTraits.kt`, `JsonEncoder.kt`, `JsonSerializer.kt`): `JsonSerialName` now lazily computes `escaped: String` — the pre-escaped, quoted name (`"…"`). A new internal `SdkFieldDescriptor.escapedSerialName` exposes it. `JsonEncoder.writeFieldName(descriptor)` appends the cached string, skipping the per-call `escape()` + quoting; `writeName(name)` and `writeFieldName` are refactored around a shared inline `writeName { … }` helper so the state-machine/comma/indent logic is not duplicated. `JsonSerializer` now uses `JsonEncoder()` directly and reads `jsonWriter.bytes` without the empty-payload null check.
- **Form-URL / Query** (`Traits.kt`, `FormUrlSerializer.kt`): `FormUrlSerialName` lazily computes `encoded: ByteArray` (UTF-8 name bytes) and `QueryLiteral` lazily computes `encoded: ByteArray` (the full `key=percentEncoded(value)` pair). A new internal `SdkFieldDescriptor.serialNameBytes` exposes the name bytes. `writeField` and empty-list serialization write the cached bytes instead of `writeUtf8(descriptor.serialName)`; the `QueryLiteral` init block writes `literal.encoded` directly, and the throwaway `QueryLiteral.toDescriptor()` helper is removed.
- **XML** is intentionally untouched — this change only affects the CBOR, JSON, and form-url/query serialization paths; the XML serializer and both XML deserialization paths are unchanged.
- **No wire-format change**: the cached bytes/string are exactly what the old per-call code produced, so serialized output is identical; only the encoding lifetime (once, memoized) changes.

## Memory
I benchmarked memory usage **before** (`cbor-optimization-main`) vs **after** (this PR) by invoking `n` distinct operations. I benchmarked `n = 5, 10, 30, 50, 100`. This is the result I am getting:

| distinct ops | before (main) | after (this PR) | difference (after − before) |
|---|---|---|---|
| 5   | 2,198,984 bytes | 2,231,312 bytes  | **+32,328 bytes** (~31.6 KB) |
| 10  | 2,861,344 bytes | 2,896,616 bytes  | **+35,272 bytes** (~34.4 KB) |
| 30  | 7,353,312 bytes | 7,399,104 bytes  | **+45,792 bytes** (~44.7 KB) |
| 50  | 10,017,064 bytes | 10,065,208 bytes | **+48,144 bytes** (~47.0 KB) |
| 100 | 16,571,344 bytes | 16,629,792 bytes | **+58,448 bytes** (~57.1 KB) |

The change does cause the pre-encoded field names to be retained (0 before → ~57 KB at 100 distinct ops, growing sub-linearly). But the retention is bounded by the number of **distinct operation types** a client invokes.

## Benchmark results

Measured across the serde benchmarks for all five protocols, 5 paired instances per benchmark. Negative `Δ%` means the target (this change) is faster.

```
--- Protocol: awsJson1  (mean ns; Δ%<0 = target faster) ---
0_GetItemInput_Baseline                       848.5      845.4    -0.37%     -0.28      5
0_GetItemOutputBinary_L                     81226.2    81113.4    -0.14%     -0.03      5
0_GetItemOutputBinary_M                     12011.6    11991.3    -0.17%     -0.03      5
0_GetItemOutputBinary_S                      3625.7     3566.4    -1.64%     -0.34      5
0_GetItemOutput_Baseline                      549.2      505.1    -8.02%     -1.13      5
0_GetItemOutput_L                           75685.8    75571.9    -0.15%     -0.02      5
0_GetItemOutput_M                           11582.0    11076.3    -4.37%     -0.56      5
0_GetItemOutput_S                            3649.8     3452.5    -5.40%     -0.73      5
0_HealthcheckRequest_Example                  292.2      303.8    +3.99%      1.35      5
0_HealthcheckResponse_Example                 189.8      198.6    +4.64%      0.74      5
0_PutItemRequest_Baseline                     856.5      863.5    +0.82%      0.35      5
0_PutItemRequest_BinaryData_L              191352.1   195973.1    +2.41%    3.23**      5
0_PutItemRequest_BinaryData_M               21715.6    22299.5    +2.69%     2.06*      5
0_PutItemRequest_BinaryData_S                1349.5     1311.8    -2.79%     -1.12      5
0_PutItemRequest_MixedItem_L                21549.3    21047.5    -2.33%    -2.22*      5
0_PutItemRequest_MixedItem_M                 7738.4     7615.1    -1.59%     -1.38      5
0_PutItemRequest_MixedItem_S                 2196.6     2222.0    +1.16%      1.36      5
0_PutItemRequest_Nested_L                    3531.2     3443.6    -2.48%   -3.35**      5
0_PutItemRequest_Nested_M                    1652.9     1661.5    +0.52%      0.78      5
0_PutItemRequest_ShallowMap_L               23832.2    22933.7    -3.77%     -1.90      5
0_PutItemRequest_ShallowMap_M                8035.4     7826.6    -2.60%     -1.19      5
0_PutItemRequest_ShallowMap_S                1950.6     1958.8    +0.42%      0.60      5

--- Protocol: awsQuery  (mean ns; Δ%<0 = target faster) ---
GetMetricDataRequest_L                      29003.4    29306.8    +1.05%      0.39      5
GetMetricDataRequest_M                      16396.5    15814.8    -3.55%   -3.19**      5
GetMetricDataRequest_S                       4254.7     3965.0    -6.81%  -7.45***      5
GetMetricDataResponse_L                    732280.6   756082.9    +3.25%   4.05***      5
GetMetricDataResponse_M                     59673.1    60172.5    +0.84%      0.47      5
GetMetricDataResponse_S                     15411.7    15786.8    +2.43%      1.32      5
PutMetricDataRequest_Baseline                1001.1      770.2   -23.07%  -6.63***      5
PutMetricDataRequest_L                     120110.4   120050.5    -0.05%     -0.06      5
PutMetricDataRequest_M                      15498.4    15230.2    -1.73%    -2.04*      5
PutMetricDataRequest_S                       2609.4     2359.3    -9.58%  -6.28***      5

--- Protocol: restJson1  (mean ns; Δ%<0 = target faster) ---
CopyObjectOutput_Baseline                     152.3      149.4    -1.88%     -0.31      5
CopyObjectOutput_M                           4730.4     4639.7    -1.92%     -0.47      5
CopyObjectRequest_Baseline                    934.1      947.1    +1.39%      0.42      5
CopyObjectRequest_M                          5103.3     5071.4    -0.63%     -0.68      5
GetObject_L                                  4120.8     4204.1    +2.02%      1.66      5
GetObject_M                                  4122.8     4158.4    +0.86%      0.59      5
GetObject_S                                  4157.3     4134.8    -0.54%     -0.36      5
PutObject_L                                  1127.5     1145.5    +1.60%      0.39      5
PutObject_M                                   942.7      948.4    +0.61%      0.59      5
PutObject_S                                   948.2      936.9    -1.19%     -1.44      5

--- Protocol: restXml  (mean ns; Δ%<0 = target faster) ---
CopyObjectOutput_Baseline                     215.7      201.8    -6.42%     -1.49      5
CopyObjectOutput_M                           6449.3     6511.7    +0.97%      0.82      5
CopyObjectRequest_Baseline                    959.0      957.6    -0.14%     -0.09      5
CopyObjectRequest_M                           5148.0     5192.3    +0.86%      0.41      5
GetObject_L                                  4092.0     4067.9    -0.59%     -0.36      5
GetObject_M                                  4072.8     4074.7    +0.05%      0.04      5
GetObject_S                                  4086.0     4106.7    +0.51%      0.31      5
PutObject_L                                  1170.0     1184.6    +1.25%      0.73      5
PutObject_M                                   976.2      954.3    -2.24%     -0.62      5
PutObject_S                                   967.1      953.3    -1.42%     -1.47      5

GetItemOutputBinary_L                       20589.7    20691.5    +0.49%      0.49      5
GetItemOutputBinary_M                        4255.4     4271.9    +0.39%      0.31      5
GetItemOutputBinary_S                        1715.8     1724.1    +0.48%      0.30      5
GetItemOutput_Baseline                        440.2      464.3    +5.47%      0.44      5
GetItemOutput_L                             23820.0    23289.7    -2.23%     -0.87      5
GetItemOutput_M                              4470.5     4473.5    +0.07%      0.07      5
GetItemOutput_S                              1829.6     1803.3    -1.44%     -0.87      5
PutItemRequest_Baseline                      1234.1     1205.9    -2.28%    -2.26*      5
PutItemRequest_BinaryData_L                 12174.1    11954.1    -1.81%     -0.93      5
PutItemRequest_BinaryData_M                  2895.6     2735.3    -5.54%   -3.39**      5
PutItemRequest_BinaryData_S                  1383.4     1319.1    -4.65%    -2.10*      5
PutItemRequest_MixedItem_L                  11704.0    11070.2    -5.41%   -3.05**      5
PutItemRequest_MixedItem_M                   5137.3     4762.4    -7.30%  -6.59***      5
PutItemRequest_MixedItem_S                   2178.2     2016.9    -7.41%  -4.31***      5
PutItemRequest_Nested_L                      3224.7     2713.7   -15.85%  -6.48***      5
PutItemRequest_Nested_M                      1729.5     1645.3    -4.87%   -3.28**      5
PutItemRequest_ShallowMap_L                 12938.1    12096.7    -6.50%   -2.93**      5
PutItemRequest_ShallowMap_M                  5110.2     4796.3    -6.14%  -4.19***      5
PutItemRequest_ShallowMap_S                  1968.1     1872.8    -4.84%  -5.12***      5

 Rollup: mean Δ% by protocol (average of per-benchmark mean deltas)
Protocol          Benchmarks    MeanΔ%   MedianΔ%  Regressions  Improvements
---------------- ----------- --------- ---------- ------------ -------------
awsJson1                  22     -0.87      -0.27            8            14
awsQuery                  10     -3.72      -0.89            4             6
restJson1                 10     +0.03      +0.04            5             5
restXml                   10     -0.72      -0.05            5             5
rpcv2Cbor                 19     -3.65      -4.65            5            14
---------------- ----------- --------- ---------- ------------ -------------
OVERALL                   71     -1.87      -0.63           27            44
```

### Highlights

- **Overall −1.87% mean / −0.63% median** across 71 benchmarks, with 44 improvements vs 27 (mostly small) regressions. This is an incremental win layered on top of the descriptor hoisting already present in the baseline — the remaining per-call cost removed here is field-name encoding/escaping, a smaller slice of total serde time, so the deltas are more modest than the earlier hoisting change.
- **`rpcv2Cbor` and `awsQuery` show the largest mean improvements** (−3.65% and −3.72%). The wins concentrate on CBOR **request/serialize** paths that write many field names: `PutItemRequest_Nested_L` **−15.85%** (`***`), `MixedItem_S`/`MixedItem_M` **−7.4%/−7.3%** (`***`), `ShallowMap_*` **−4.8% to −6.5%** (`**`/`***`), and `BinaryData_M` **−5.54%** (`**`). Nested/map/shallow-map shapes benefit most because they emit the most field names per serialization.
- **CBOR `GetItem*Output*` paths are flat** (~+0.4%), as expected — those are **deserialization** benchmarks and this change touches only serializers, so there is nothing to amortize there.
- **`awsQuery` is dominated by request-side wins**: `PutMetricDataRequest_Baseline` **−23.07%** (`***`), plus `GetMetricDataRequest_S` −6.81% and `PutMetricDataRequest_S` −9.58% (both `***`). Its XML-based **responses** (`GetMetricDataResponse_*`) are flat-to-slightly-slower, including a statistically-significant `GetMetricDataResponse_L` **+3.25%** (`***`) — response parsing goes through the XML reader, which this change does not touch, so that delta reflects environmental variance on a ~730 µs benchmark rather than added serde work.
- **`awsJson1` is a modest net win** (mean −0.87%, median −0.27%): deserialize-heavy `GetItemOutput*` shapes improve a few percent, but there are small statistically-significant regressions on `PutItemRequest_BinaryData_L`/`_M` (**+2.41%** `**`, **+2.69%** `*`) — large binary payloads where field-name encoding is a negligible share of the work and the deltas reflect payload-copy variance, not added serde cost, since this change removes work rather than adding any.
- **`restXml` is roughly flat** (mean −0.72%, no significant results): the XML serializer is not modified by this change, so its sub-±3% deltas are run-to-run variance.
- **`restJson1` is essentially flat** (mean +0.03%, median +0.04%, no significant results). These are small-payload, header-bound S3 operations with minimal structured JSON body, so the JSON field-name caching has little to amortize; every delta is within ±2% and none is statistically significant.


.changes/522a9e08-4fb4-4a13-a1e5-ddaf6201f4d2.json
```diff
@@ -0,0 +1,6 @@
+{
+  "id": "522a9e08-4fb4-4a13-a1e5-ddaf6201f4d2",
+  "type": "feature",
+  "description": "Optimize CBOR serialization by caching each field's pre-encoded CBOR name bytes on its descriptor instead of re-encoding the name on every field write",
+  "module": "serde-cbor"
+}
```
ianbotsf, line 6: **Nit**: There's not much value in enumerating these changes repeatedly across three different changelog entries. Let's consolidate into a single entry.

runtime/serde/serde-cbor/common/src/aws/smithy/kotlin/runtime/serde/cbor/CborSerializer.kt
```diff
@@ -158,47 +158,85 @@ public class CborSerializer :
...
+    private fun writeFieldName(descriptor: SdkFieldDescriptor) = buffer.write(descriptor.serialNameBytes)
+
+    override fun field(descriptor: SdkFieldDescriptor, value: Boolean) {
+        writeFieldName(descriptor)
+        serializeBoolean(value)
+    }
+    override fun field(descriptor: SdkFieldDescriptor, value: Byte) {
+        writeFieldName(descriptor)
+        serializeByte(value)
+    }
```
ianbotsf, line 175: **Nit**: The `serializeEntry` and `entry` methods are now unused and should be deleted.

ianbotsf, line 175: **Style**: Please add vertical whitespace between multi-line declaration blocks.

runtime/serde/serde-cbor/common/src/aws/smithy/kotlin/runtime/serde/cbor/CborSerializer.kt
```diff
@@ -158,47 +158,85 @@ public class CborSerializer :
...
-    override fun field(descriptor: SdkFieldDescriptor, value: Byte): Unit = entry(descriptor.serialName, value)
-    override fun field(descriptor: SdkFieldDescriptor, value: Short): Unit = entry(descriptor.serialName, value)
-    override fun field(descriptor: SdkFieldDescriptor, value: Char): Unit = entry(descriptor.serialName, value)
-    override fun field(descriptor: SdkFieldDescriptor, value: Int): Unit = entry(descriptor.serialName, value)
-    override fun field(descriptor: SdkFieldDescriptor, value: Long): Unit = entry(descriptor.serialName, value)
-    override fun field(descriptor: SdkFieldDescriptor, value: Float): Unit = entry(descriptor.serialName, value)
-    override fun field(descriptor: SdkFieldDescriptor, value: Double): Unit = entry(descriptor.serialName, value)
-    override fun field(descriptor: SdkFieldDescriptor, value: String): Unit = entry(descriptor.serialName, value)
-    override fun field(descriptor: SdkFieldDescriptor, value: Instant, format: TimestampFormat): Unit = entry(descriptor.serialName, value, format)
+    private fun writeFieldName(descriptor: SdkFieldDescriptor) = buffer.write(descriptor.serialNameBytes)
```
ianbotsf: **Nit**: Please name this method `serializeFieldName` (or maybe `serializeBytes`?) for consistency with other `serialize*` methods.

runtime/serde/serde-form-url/common/src/aws/smithy/kotlin/runtime/serde/formurl/FormUrlSerializer.kt
```diff
@@ -84,9 +84,9 @@ private class FormUrlStructSerializer(
     init {
         structDescriptor.traits.mapNotNull { it as? QueryLiteral }
             .forEach { literal ->
-                writeField(literal.toDescriptor()) {
-                    serializeString(literal.value)
-                }
+                if (buffer.size > 0L) buffer.writeUtf8("&")
+                if (prefix.isNotBlank()) buffer.writeUtf8(prefix)
+                buffer.write(literal.encoded)
```
ianbotsf: **Question**: Can't we also memoize the prefix encoding and accelerate this path even more?

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonSerializer.kt
```diff
@@ -25,9 +25,9 @@ public class JsonSerializer :
         private val floatsToStringify = setOf(Float.NEGATIVE_INFINITY, Float.POSITIVE_INFINITY, Float.NaN)
     }
 
-    private val jsonWriter = jsonStreamWriter()
+    private val jsonWriter = JsonEncoder()
```
ianbotsf, line 28: **Question**: Why is this change necessary? `jsonStreamWriter()` is already a factory method that returns `JsonEncoder()`.

ianbotsf (commented):
**Question**: Can these optimizations not be applied to **serde-xml** as well?

runtime/serde/serde-cbor/common/src/aws/smithy/kotlin/runtime/serde/cbor/CborSerializer.kt
luigi617: `entry` methods are `MapSerializer` interface methods, and they are used (e.g. [here](https://github.com/smithy-lang/smithy-kotlin/blob/581ecc9d048b6abb7c6bd0d72efafc315129ebf1/codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/rendering/serde/SerializeStructGenerator.kt#L334)). `serializeEntry` is a private function that`entry` will use, so they can't be removed.

runtime/serde/serde-json/common/src/aws/smithy/kotlin/runtime/serde/json/JsonSerializer.kt
luigi617: Because the optimized function that `jsonWriter` needs to call is `writeFieldName` which I did not choose to add it in the `JsonStreamWriter` interface. `writeFieldName` is just an internal method of `JsonEncoder`. If we were using `jsonStreamWriter`, the compiler wouldn't know the existance of`writeFieldName`.

runtime/serde/serde-form-url/common/src/aws/smithy/kotlin/runtime/serde/formurl/FormUrlSerializer.kt
luigi617: Good catch!

luigi617:
**Answer to why not optimize serde-xml**: `BufferingXmlStreamWriter` never writes bytes. It accumulates a tree of `LazyTagWriter` objects, renders that into a `StringBuilder`, and UTF-8-encodes the whole document once at the end ([BufferingXmlStreamWriter.kt](https://github.com/smithy-lang/smithy-kotlin/blob/581ecc9d048b6abb7c6bd0d72efafc315129ebf1/runtime/serde/serde-xml/common/src/aws/smithy/kotlin/runtime/serde/xml/serialization/BufferingXmlStreamWriter.kt)). There is no per-field encode to eliminate, so a cached `ByteArray` on `XmlSerialName` would be dead weight. If we really want to optimize, we might need to rewrite how `BufferingXmlStreamWriter` works but that's out of scope for this PR.

ianbotsf (approved):

PR #1695, opened 2026-08-20
perf(serde): deserialization lookup field O(1) instead of O(n)

luigi617:
Make per-field lookup during **deserialization** O(1) instead of O(N). Previously every decoded field name was resolved by a linear scan over the struct's field descriptors — `descriptor.fields.firstOrNull { it.serialName == name }` (CBOR) / `descriptor.fields.find { it.serialName == propertyName }` (JSON) — run once per decoded field, i.e. O(fields × decoded-fields) per struct. This change precomputes a cached `FieldIndex` on the `SdkObjectDescriptor` (a serial-name → field-index map plus an in-order cursor fast path) and uses it from the CBOR and JSON field iterators, so name resolution is O(1) amortized per field. Because field descriptors are file-level singletons in the baseline, the index is built **once** per descriptor and reused across every deserialization. Field resolution is **behavior-identical** — same first-wins handling of duplicate serial names, same `UNKNOWN_FIELD` result for unmatched names, same `IgnoreKey` handling — so deserialized output is unchanged; only the lookup cost changes.

## Description of changes

- **New `FieldIndex` + `SdkObjectDescriptor.fieldIndex(...)`** (`SdkObjectDescriptor.kt`, `serde.api`): `SdkObjectDescriptor` gains a lazily-built, cached `FieldIndex` (built once per descriptor; since descriptors are hoisted to file/top-level scope in the baseline, that is effectively once per process, reused across all deserializations). `FieldIndex` precomputes a `serialNames` array and a `fieldIndices` `IntArray` keyed by **list position** (always dense/in-bounds), plus a `byName` map (serial name → field index) that keeps the **first** field on a duplicated serial name — exactly matching the old linear `firstOrNull`/`find` scan. It deliberately indexes by list position rather than `SdkFieldDescriptor.index`, which is a mutable `var` and can fall outside `0..size-1` when a field is shared across object descriptors.
- **`FieldIndex.lookup(serialName, expectedPosition)`**: checks the field at `expectedPosition` first (an in-order guess — fields are almost always decoded in schema order, so the field after the last match is the most likely next hit), then falls back to the O(1) `byName` map, returning `Deserializer.FieldIterator.UNKNOWN_FIELD` when no field matches.
- **CBOR deserialize** (`CborDeserializer.kt`): `CborFieldIterator` obtains `descriptor.fieldIndex { it.serialName }` and keeps an `expectedFieldIndex` cursor. For each decoded field name it calls `fieldIndex.lookup(name, expectedFieldIndex)`, advancing the cursor to `index + 1` on a hit, replacing the previous `descriptor.fields.firstOrNull { it.serialName == name }?.index` scan.
- **JSON deserialize** (`JsonDeserializer.kt`): same treatment for `JsonFieldIterator`, replacing `descriptor.fields.find { it.serialName == propertyName }`; the existing `IgnoreKey` skip is preserved ahead of the lookup.
- **XML and serialization intentionally untouched**: XML deserialization streams via `XmlTagReader` and never scans `SdkObjectDescriptor` fields, so there is nothing to index; all serializers are unchanged. Deserialized results are identical (same field resolution, duplicate-name, and unknown-field semantics).

## Memory
I benchmarked memory usage **before** (`cbor-optimization-main`) vs **after** (this PR) by invoking `n` distinct operations. I benchmarked `n = 5, 10, 30, 50, 100`. This is the result I am getting:

| distinct ops | before (main) | after (this PR) | difference (after − before) |
|---|---|---|---|
| 5   | 2,198,984 bytes | 2,220,840 bytes  | **+21,856 bytes** (~21.3 KB) |
| 10  | 2,861,344 bytes | 2,886,160 bytes  | **+24,816 bytes** (~24.2 KB) |
| 30  | 7,353,312 bytes | 7,381,464 bytes  | **+28,152 bytes** (~27.5 KB) |
| 50  | 10,017,064 bytes | 10,045,952 bytes | **+28,888 bytes** (~28.2 KB) |
| 100 | 16,571,344 bytes | 16,608,800 bytes | **+37,456 bytes** (~36.6 KB) |

The change does cause a `FieldIndex` to be retained per struct descriptor (0 before → ~37 KB at 100 distinct ops, growing sub-linearly). But the retention is bounded by the number of **distinct operation types** a client invokes.

## Benchmark results

Measured across the serde benchmarks for all five protocols, 5 paired instances per benchmark. Negative `Δ%` means the target (this change) is faster.

```
--- Protocol: awsJson1  (mean ns; Δ%<0 = target faster) ---
0_GetItemInput_Baseline                       861.3      856.1    -0.61%    -2.53*      5
0_GetItemOutputBinary_L                     88321.2    80088.0    -9.32%  -4.30***      5
0_GetItemOutputBinary_M                     13039.4    11818.4    -9.36%  -4.28***      5
0_GetItemOutputBinary_S                      3911.9     3566.1    -8.84%   -3.47**      5
0_GetItemOutput_Baseline                      570.9      557.1    -2.42%     -0.42      5
0_GetItemOutput_L                           83430.6    69685.3   -16.48%  -4.34***      5
0_GetItemOutput_M                           12532.6    10709.5   -14.55%   -2.89**      5
0_GetItemOutput_S                            3852.5     3351.4   -13.01%   -3.25**      5
0_HealthcheckRequest_Example                  292.6      287.3    -1.81%     -1.05      5
0_HealthcheckResponse_Example                 178.5      194.2    +8.78%      1.21      5
0_PutItemRequest_Baseline                     879.4      872.2    -0.82%     -0.65      5
0_PutItemRequest_BinaryData_L              195754.7   195594.1    -0.08%     -0.27      5
0_PutItemRequest_BinaryData_M               22088.8    22067.9    -0.09%     -0.13      5
0_PutItemRequest_BinaryData_S                1346.0     1320.2    -1.92%     -1.77      5
0_PutItemRequest_MixedItem_L                22105.6    22018.2    -0.40%     -0.47      5
0_PutItemRequest_MixedItem_M                 7890.6     7886.2    -0.06%     -0.11      5
0_PutItemRequest_MixedItem_S                 2258.6     2239.3    -0.85%     -0.62      5
0_PutItemRequest_Nested_L                    3590.1     3560.0    -0.84%     -1.15      5
0_PutItemRequest_Nested_M                    1676.6     1664.2    -0.74%     -0.55      5
0_PutItemRequest_ShallowMap_L               24336.3    24425.4    +0.37%      0.67      5
0_PutItemRequest_ShallowMap_M                8155.2     8226.7    +0.88%      1.07      5
0_PutItemRequest_ShallowMap_S                1982.3     2010.5    +1.42%      1.01      5

--- Protocol: awsQuery  (mean ns; Δ%<0 = target faster) ---
GetMetricDataRequest_L                      29636.5    29441.8    -0.66%     -0.47      5
GetMetricDataRequest_M                      16680.7    16637.8    -0.26%     -0.16      5
GetMetricDataRequest_S                       4349.6     4379.7    +0.69%      0.57      5
GetMetricDataResponse_L                    760877.4   751652.7    -1.21%     -0.86      5
GetMetricDataResponse_M                     61594.6    60524.6    -1.74%   -3.08**      5
GetMetricDataResponse_S                     16025.6    15962.0    -0.40%     -0.45      5
PutMetricDataRequest_Baseline                1074.1     1047.1    -2.52%     -1.91      5
PutMetricDataRequest_L                     121322.6   121947.1    +0.51%      0.84      5
PutMetricDataRequest_M                      15765.2    15590.1    -1.11%    -2.55*      5
PutMetricDataRequest_S                       2691.0     2653.7    -1.38%     -1.55      5

--- Protocol: restJson1  (mean ns; Δ%<0 = target faster) ---
CopyObjectOutput_Baseline                     149.6      151.2    +1.06%      0.20      5
CopyObjectOutput_M                           4794.2     4448.5    -7.21%  -8.00***      5
CopyObjectRequest_Baseline                    960.3      955.6    -0.49%     -0.39      5
CopyObjectRequest_M                          5280.8     5197.4    -1.58%     -1.37      5
GetObject_L                                  4231.7     4153.4    -1.85%     -1.85      5
GetObject_M                                  4247.3     4085.1    -3.82%    -2.25*      5
GetObject_S                                  4202.3     4100.1    -2.43%     -1.84      5
PutObject_L                                  1122.4     1154.1    +2.83%      1.59      5
PutObject_M                                   969.4      974.7    +0.55%      0.28      5
PutObject_S                                   966.2      960.5    -0.59%     -0.58      5

--- Protocol: restXml  (mean ns; Δ%<0 = target faster) ---
CopyObjectOutput_Baseline                     202.1      219.9    +8.77%      0.64      5
CopyObjectOutput_M                           6558.6     6601.3    +0.65%      0.69      5
CopyObjectRequest_Baseline                    976.0      979.9    +0.40%      0.33      5
CopyObjectRequest_M                           5236.3     5290.2    +1.03%      0.69      5
GetObject_L                                  4117.2     4101.2    -0.39%     -0.21      5
GetObject_M                                  4176.5     4115.5    -1.46%     -0.69      5
GetObject_S                                  4146.8     4091.0    -1.34%     -0.53      5
PutObject_L                                  1142.6     1193.8    +4.48%   4.36***      5
PutObject_M                                   997.9     1001.7    +0.38%      0.14      5
PutObject_S                                   971.0     1000.1    +3.00%      1.71      5

GetItemOutputBinary_L                       20537.1    20336.9    -0.97%     -0.77      5
GetItemOutputBinary_M                        4244.0     4249.6    +0.13%      0.10      5
GetItemOutputBinary_S                        1745.3     1820.1    +4.28%      1.91      5
GetItemOutput_Baseline                        425.7      433.3    +1.79%      0.15      5
GetItemOutput_L                             23427.2    21077.2   -10.03%  -8.57***      5
GetItemOutput_M                              4512.9     4325.3    -4.16%     -1.45      5
GetItemOutput_S                              1838.0     1821.3    -0.91%     -0.48      5
PutItemRequest_Baseline                      1269.3     1296.3    +2.12%      1.19      5
PutItemRequest_BinaryData_L                 12574.2    12708.8    +1.07%      0.53      5
PutItemRequest_BinaryData_M                  3009.0     2979.3    -0.99%     -0.32      5
PutItemRequest_BinaryData_S                  1395.1     1388.1    -0.50%     -0.27      5
PutItemRequest_MixedItem_L                  11904.3    12121.6    +1.83%      0.85      5
PutItemRequest_MixedItem_M                   5161.3     5257.0    +1.86%      1.34      5
PutItemRequest_MixedItem_S                   2222.3     2201.7    -0.93%     -0.76      5
PutItemRequest_Nested_L                      3198.5     3235.5    +1.16%      0.48      5
PutItemRequest_Nested_M                      1781.4     1791.9    +0.58%      0.37      5
PutItemRequest_ShallowMap_L                 12989.9    13035.1    +0.35%      0.65      5
PutItemRequest_ShallowMap_M                  5227.6     5307.3    +1.53%      1.48      5
PutItemRequest_ShallowMap_S                  2028.4     2034.1    +0.28%      0.21      5

 Rollup: mean Δ% by protocol (average of per-benchmark mean deltas)
Protocol          Benchmarks    MeanΔ%   MedianΔ%  Regressions  Improvements
---------------- ----------- --------- ---------- ------------ -------------
awsJson1                  22     -3.22      -0.83            2             9
awsQuery                  10     -0.81      -0.88            0             5
restJson1                 10     -1.35      -1.08            2             5
restXml                   10     +1.55      +0.53            4             2
rpcv2Cbor                 19     -0.08      +0.35            8             2
---------------- ----------- --------- ---------- ------------ -------------
OVERALL                   71     -1.10      -0.49           16            23
```

> Regressions/Improvements count benchmarks whose |Δ%| ≥ 1%, classified by sign.

### Highlights

- The wins concentrate entirely on the **deserialization** side, which is exactly what this change touches: replacing the per-field O(N) descriptor scan with an O(1) `FieldIndex` lookup. Serialization benchmarks and all XML paths are flat, as expected.
- **`awsJson1` responses see the largest wins** (`***`/`**`): `GetItemOutput_L` **−16.48%**, `_M` **−14.55%**, `_S` **−13.01%**, and `GetItemOutputBinary_*` **−8.8% to −9.4%**. These are DynamoDB responses — a map of `AttributeValue` unions — where the old code linear-scanned the union's members for every decoded field; O(1) resolution removes that per-field cost, and the benefit scales with the number of fields decoded. The corresponding `PutItemRequest_*` **serialize** benchmarks are flat (±1%, not significant), since serialization is untouched.
- **`rpcv2Cbor` is ~flat overall** (mean **−0.08%**) only because 12 of its 19 benchmarks are `PutItemRequest_*` **serialize** (unchanged code). Its one large deserialize benchmark, `GetItemOutput_L`, shows the expected **−10.03%** (`***`); the small CBOR deserialize shapes sit in the noise, and the scattered +1–2% serialize deltas are not statistically significant (t < 1.5), reflecting run-to-run variance rather than added work.
- **`restJson1` deserialize improves** where there is structured content: `CopyObjectOutput_M` **−7.21%** (`***`) and `GetObject_M` **−3.82%** (`*`); the small header-bound operations are flat.
- **`awsQuery` is roughly flat** (mean −0.81%): its requests serialize (untouched) and its responses are **XML** (untouched, streamed via `XmlTagReader`). The lone significant result, `GetMetricDataResponse_M` **−1.74%** (`**`), is a sub-2% delta on an unmodified path and within measurement noise.
- **`restXml` shows small apparent regressions** (mean +1.55%), including a statistically-significant `PutObject_L` **+4.48%** (`***`). XML deserialization does not use `SdkObjectDescriptor` field lookups and is not modified by this change, so these deltas are on unchanged code paths and reflect environmental variance, not a regression introduced here.
- **Net:** overall **−1.10% mean / −0.49% median** across 71 benchmarks, driven by large, highly-significant gains on multi-field JSON/CBOR deserialization, with serialization and XML paths flat and the few significant "regressions" landing on untouched code.


.changes/2a9f1b35-de19-4953-bbce-e0220f90cc4e.json
```diff
@@ -0,0 +1,6 @@
+{
+  "id": "2a9f1b35-de19-4953-bbce-e0220f90cc4e",
+  "type": "feature",
+  "description": "Optimize JSON deserialization by resolving struct field names through a cached O(1) field index with an in-order cursor instead of a linear scan over the field list",
+  "module": "serde-json"
+}
```
ianbotsf, line 5: **Nit**: Let's consolidate the changelog entries into one.

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/SdkObjectDescriptor.kt
```diff
@@ -17,6 +17,14 @@ public class SdkObjectDescriptor private constructor(builder: Builder) :
...
     public val fields: List<SdkFieldDescriptor> = builder.fields
 
+    private var cachedFieldIndex: FieldIndex? = null
+
+    /**
+     * Returns a cached [FieldIndex] that provides O(1) lookup of a field's index by its serial name.
+     */
+    @InternalApi
+    public fun fieldIndex(serialNameOf: (SdkFieldDescriptor) -> String): FieldIndex = cachedFieldIndex ?: FieldIndex(fields, serialNameOf).also { cachedFieldIndex = it }
+
```
ianbotsf, line 32: **Nit**: This code is not thread-safe. Fortunately, I think the worst possible outcome from unsynchronized access is computing the field index multiple times. Nevertheless, we should either make this thread-safe **or** document that it's not and why that's okay in new code comments.

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/SdkObjectDescriptor.kt
```diff
@@ -17,6 +17,14 @@ public class SdkObjectDescriptor private constructor(builder: Builder) :
     ) {
     public val fields: List<SdkFieldDescriptor> = builder.fields
 
+    private var cachedFieldIndex: FieldIndex? = null
+
+    /**
+     * Returns a cached [FieldIndex] that provides O(1) lookup of a field's index by its serial name.
+     */
+    @InternalApi
+    public fun fieldIndex(serialNameOf: (SdkFieldDescriptor) -> String): FieldIndex = cachedFieldIndex ?: FieldIndex(fields, serialNameOf).also { cachedFieldIndex = it }
```
ianbotsf: **Style**: This method has confusing idempotency semantics. If a caller invokes `descriptor.fieldIndex { it.toString() }` and then `descriptor.fieldIndex { it.toString().reversed() }`, they receive the same field index with the same name derivation both times, even though the argument is different. Is there a cleaner way to embed the decoding logic into the `FieldIndex` that doesn't rely on an argument passed to `fieldIndex`?

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/SdkObjectDescriptor.kt
```diff
@@ -40,3 +48,40 @@ public class SdkObjectDescriptor private constructor(builder: Builder) :
...
+    // Field index at each list position — the value returned to callers.
+    private val fieldIndices: IntArray = IntArray(fields.size) { fields[it].index }
+
+    // Serial name -> field index, for O(1) fallback when the in-order guess misses.
+    private val byName: Map<String, Int> = buildMap(fields.size) {
+        for (position in fields.indices) {
+            // Keep the FIRST field for a duplicated serial name, matching the old linear scan.
+            if (serialNames[position] !in this) put(serialNames[position], fieldIndices[position])
+        }
+    }
```
ianbotsf: **Question**: This class tries to maintain the notion of original list index but is that useful? It seems to me it could be abstracted away so we deal with only field indices and serial names.

```kotlin
// Given serial name, return field index
val byName: Map<String, Int> = fields.associate { serialNameOf(it) to it.index }

// Given field index, return serial name
val serialNames: Array<String> = run {
    // Swap keys and values
    val byIndex = byName.associate { (key, value) -> value to key }

    // Build to array for faster O(1) access
    Array(fields.size) { byIndex[it] }
}
```

ianbotsf (commented):
**Question**: Can these optimizations not be applied to **serde-xml** as well?

luigi617:
XML deserialization is generated as direct string switch over the tag name, so the same optimization cannot be applied to XML directly. Might be other way to optimize XML deserialization specifically, but out of scope of this PR

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/SdkObjectDescriptor.kt
luigi617: Yes, I changed it so that it doesn't have an argument to depends on

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/SdkObjectDescriptor.kt
luigi617: You're right, I use your version but with serialNames being `Array<String?>` because `byName` deliberately drops duplicate serial names (this is more a defensive, as they shouldn't)

ianbotsf (approved):

PR #1698, opened 2026-08-20
# perf(cbor): decode/encode primitives without allocating `Value` wrappers

luigi617:
Removes the per-value `Value`-wrapper allocations on the CBOR serialize/deserialize hot path. Every
scalar (de)serialize previously allocated a short-lived `Value` object (`UInt`, `NegInt`, `TextString`,
`Float32/64`, `Boolean`, `ByteString`, `Tag`/`Timestamp`, …) whose only purpose was to carry a value one
frame up. This change decodes/encodes those primitives directly, cutting the dominant source of
short-lived garbage on the serde path and improving throughput on realistic DynamoDB payloads by up to
**~4.8%** (statistically significant on the map-heavy and small/medium shapes).

The wire format is unchanged (byte-for-byte identical) and there is no public API change.

## What changed

The encode/decode bodies are extracted from the `Value` wrapper classes into allocation-free free
functions and called directly from `CborPrimitiveDeserializer` / `CborSerializer`. The wrapper classes
are retained for the generic `Value.decode` / document path.

- **Decode:** `UInt.decode(buffer).value` → `decodeUInt(buffer)`; `NegInt` → `decodeNegInt`;
  `Float16/32/64` → `decodeFloat16/32/64`; `TextString`/`ByteString` → `decodeTextStringValue` /
  `decodeByteStringValue`; `Boolean` → `decodeBooleanValue`. Tagged types read the tag id with
  `decodeArgument` then call `decodeInstant` / `decodeBigNum` / `decodeNegBigNum` /
  `decodeDecimalFraction` directly — no `Tag`/`Timestamp` wrapper and no `as` cast.
- **Encode:** `serializeBoolean/Float/Double/ByteArray/Instant/BigInteger/BigDecimal` and the field-name
  writes call `writeBoolean` / `writeFloat32/64` / `writeByteString` / `writeTimestamp` / `writeBigNum` /
  `writeNegBigNum` / `writeDecimalFraction` / `writeTextString` directly instead of constructing a
  wrapper and invoking `.encode(...)`.
- **`skipValue`:** unknown/skipped fields advance the cursor via an allocation-free `skipValue(buffer,
  depth)` (head-walk + `buffer.skip(len)`, recursing on list/map/tag, honoring the recursion-depth
  guard) instead of `Value.decode(buffer)`, which materialized and discarded the entire skipped subtree.
- **BigInteger sign:** a hoisted `ZERO_BIG_INTEGER` replaces the `toString().startsWith("-")` sign test,
  avoiding a full decimal-string materialization per `BigInteger` serialized.

## Benchmark results

JMH, 5 paired runs, DynamoDB shapes over **rpcv2Cbor**. `Δ%` is target (this branch) vs base
(`cbor-optimization-main`); **Δ% < 0 means the target is faster**. Significance from the paired t-stat:
`*` p<0.05, `**` p<0.01, `***` p<0.001.

```
GetItemOutputBinary_L                       20984.8    21237.0    +1.20%      0.68      5
GetItemOutputBinary_M                        4295.6     4271.7    -0.56%     -0.25      5
GetItemOutputBinary_S                        1744.3     1716.1    -1.62%     -0.74      5
GetItemOutput_Baseline                        417.1      372.1   -10.79%     -0.70      5
GetItemOutput_L                             23716.8    24527.7    +3.42%     2.13*      5
GetItemOutput_M                              4534.3     4427.1    -2.36%   -3.23**      5
GetItemOutput_S                              1851.6     1765.9    -4.63%   -2.71**      5
PutItemRequest_Baseline                      1186.0     1167.7    -1.55%    -2.53*      5
PutItemRequest_BinaryData_L                 12268.1    12481.5    +1.74%      1.00      5
PutItemRequest_BinaryData_M                  2810.2     2805.9    -0.15%     -0.19      5
PutItemRequest_BinaryData_S                  1317.6     1290.4    -2.06%     -1.61      5
PutItemRequest_MixedItem_L                  11928.5    12064.7    +1.14%      0.44      5
PutItemRequest_MixedItem_M                   5088.3     5024.5    -1.25%     -0.39      5
PutItemRequest_MixedItem_S                   2117.5     2052.2    -3.08%   -3.48**      5
PutItemRequest_Nested_L                      3093.0     3033.5    -1.92%     -0.98      5
PutItemRequest_Nested_M                      1681.1     1630.5    -3.01%  -3.74***      5
PutItemRequest_ShallowMap_L                 13181.5    12674.0    -3.85%  -5.52***      5
PutItemRequest_ShallowMap_M                  5183.5     5007.0    -3.41%     -1.60      5
PutItemRequest_ShallowMap_S                  1875.2     1784.7    -4.82%  -4.78***      5
```

The gains concentrate where the payload carries many scalars/keys per byte (shallow/nested maps, small
and medium items) — exactly the shapes that allocated the most wrappers per unit of work. Large,
binary-dominated payloads are bound by bulk `byte[]`/UTF-8 copying rather than per-value allocation, so
they land within a few percent of the baseline as expected.

## Testing

`CborDeserializerTest` is extended, and the existing byte-exact CBOR serialize/deserialize suites
(`CborSerializerTest`, `CborDeserializerSuccessTest`, `CborDeserializerErrorTest`) plus the rpcv2Cbor
protocol tests pass unchanged, confirming identical wire output and decode behavior.


ianbotsf (approved):
**Question**: Do we still need/use the `Value` class or could it be removed entirely?

luigi617:
You're right, They were used but only in test. I removed them

ianbotsf (approved):

PR #1705, opened 2026-08-29
feat(SSD): add the schema for SSD

luigi617:
## Description of changes
This PR adds **only the runtime schema model**. It has no consumers: nothing on the existing serde path references any of it, and no generated code changes.

* `Schema` (sealed) and its `StructureSchema`, `UnionSchema`, `ListSchema`, `MapSchema`, `MemberSchema` and `SimpleSchema` variants, each with a `@SchemaDsl` builder.
* `ShapeId` / `MemberShapeId` and the `shapeId(...)` factories.
* `ShapeType`, and `PreludeSchemas` for the Smithy prelude shapes.
* `Trait` plus the serde-relevant trait types in `schema/trait/Traits.kt`.

Traits are held as a `Collection<Trait>` with a linear `getTrait(ShapeId)`. The array-indexed `TraitKey` lookup from the design doc is deliberately **not** implemented here.


runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/schema/ShapeId.kt
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * The immutable identifier of a Smithy shape.
+ */
+public sealed interface ShapeId {
+    public val namespace: String
+    public val name: String
+
+    /** The absolute string form of this id. */
+    public val absoluteId: String
+}
```
ianbotsf, line 16: **Question**: Are all shape IDs required to be absolute? [The Smithy spec](https://smithy.io/2.0/spec/model.html#shape-id) distinguishes between absolute IDs which include a namespace and relative IDs which do not.

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/schema/ShapeId.kt
```diff
@@ -0,0 +1,83 @@
...
+    override fun toString(): String = absoluteId
+}
+
+/**
+ * Parse a shape id from its absolute form `namespace#name` (or `namespace#name$member`, yielding a
+ * [MemberShapeId]).
+ *
+ * @throws IllegalArgumentException if [absoluteId] is not a well-formed shape id.
+ */
+public fun shapeId(absoluteId: String): ShapeId {
```
ianbotsf: **Style**: Factory functions in Kotlin are typically named identically to the class they return (e.g., `fun ShapeId(...)` in PascalCase) or implemented as `invoke` methods on the class's companion object:

```kotlin
sealed interface ShapeId {
    val name: String
    val namespace: String
    val absoluteId: String

    companion object {
        operator fun invoke(absoluteId: String): ShapeId {
            // ...
        }
    }
}

val myShapeId = ShapeId("com.foo.bar#FooBarService")
```

Both techniques make the function appear more like a constructor, which streamlines reading the code. We've adopted this pattern in various places in the SDK (e.g., [`TlsContext`](https://github.com/smithy-lang/smithy-kotlin/blob/48fc1b33cdf53c74ef15b9259dac6b78dc6838e6/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/engine/TlsContext.kt#L17)) and we should maintain the pattern for these new types.

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/schema/Schema.kt
```diff
@@ -0,0 +1,49 @@
...
+     * The traits associated with this shape
+     */
+    public val traits: Collection<Trait>
+}
+
+/**
+ * The trait identified by [id] declared directly on this shape, or `null` if absent.
+ */
+@Suppress("UNCHECKED_CAST")
+public fun <T : Trait> Schema.getTrait(id: ShapeId): T? = traits.find { it.id == id } as T?
```
ianbotsf: **Style**: Kotlin methods which return `null` to indicate "no results" are typically suffixed "OrNull" to distinguish them from methods which throw an exception to indicate "no results". See for example [`first`] (https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/first.html) (throws) and [`firstOrNull`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/first-or-null.html) (returns `null`).

Since it can return `null`, we should rename this method `getTraitOrNull`. Consider also whether you'll eventually want/need a version of this method that assumes/requires the trait exists, which would be the better method to name `getTrait`.

Applies to `getEffectiveTrait` as well.

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/schema/trait/Traits.kt
ianbotsf, line 1: **Style**: Suggest naming this file **SerdeTraits.kt** to clarify what kinds of traits are present.

Additionally, I like the namespacing you did for the prelude schemas and I think we should consider the same thing here:

```kotlin
public object SerdeTraits {
    public class JsonNameTrait(...) : ...
    public class XmlNameTrait(...) : ...
    public object XmlAttributeTrait : ...
    // etc.
}
```

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/schema/ListSchema.kt
```diff
@@ -0,0 +1,54 @@
...
+}
+
+internal class ListSchemaImpl(
+    override val shapeId: ShapeId,
+    override val traits: Collection<Trait>,
+    elementProvider: () -> MemberSchema,
+) : ListSchema {
+    override val type: ShapeType = ShapeType.LIST
+    override val element: MemberSchema by lazy(elementProvider)
+    override fun toString(): String = "ListSchema($shapeId)"
```
ianbotsf: **Question**: Why override `toString()` on all of your schema implementations? The default `toString()` contains the same info your impls do plus extra.

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/schema/ListSchema.kt
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+
+    internal fun build(): ListSchema {
+        val e = requireNotNull(element) { "list schema $shapeId is missing its element" }
+        return ListSchemaImpl(shapeId, traits.toList(), e)
+    }
+}
+
+public fun ListSchema(id: ShapeId, block: ListSchemaBuilder.() -> Unit): ListSchema = ListSchemaBuilder(id).apply(block).build()
```
ianbotsf: **Correctness**: I _do_ love a good DSL builder but I think it's not the best fit here for a few reasons:
* `element` is effectively a required field but nothing in the type signatures verifies that one is set at compile time
* `traits` is a list, which can be better handled with a `vararg` parameter or `listOf` at construction time

I recommend we scale back the factory function to something like this:

```kotlin
public fun ListSchema(
    id: ShapeId,
    elementSchema: MemberSchema,
    vararg traits: Trait,
): ListSchema = ...
```

Applies to `MapSchema` as well.

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/schema/ListSchema.kt
ianbotsf: **Question**: Why are we making `MemberSchema.target` _and_ `ListSchemaImpl.element` lazy values? I understand we need to account for recursively-defined shapes but I believe we should only need `lazy` at a single layer, not multiple layers.

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/schema/ShapeType.kt
```diff
@@ -0,0 +1,44 @@
...
+    OPERATION,
+    ;
+
+    public val isSimple: Boolean
+        get() = when (this) {
+            BLOB, BOOLEAN, STRING, TIMESTAMP, BYTE, SHORT, INTEGER, LONG,
+            FLOAT, DOUBLE, BIG_INTEGER, BIG_DECIMAL, DOCUMENT, ENUM, INT_ENUM,
+            -> true
+            else -> false
+        }
```
ianbotsf: **Nit**: This is less efficient and less maintainable than making `isSimple` a constructor parameter:

```kotlin
enum class ShapeType(val isSimple: Boolean) {
    // simple
    BLOB(true),
    BOOLEAN(true),
    STRING(true),
    // etc.

    // aggregate
    LIST(false),
    MAP(false),
    // etc.

    // ...
}
```

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/schema/ListSchema.kt
```diff
@@ -0,0 +1,54 @@
...
+    public fun element(target: Lazy<Schema>, vararg traits: Trait) {
+        val memberTraits = traits.toList()
+        element = {
+            MemberSchemaImpl(
+                shapeId = shapeId.withMember("member"),
+                traits = memberTraits,
+                targetProvider = { target.value },
+            )
+        }
+    }
```
ianbotsf: **Question**: (wasn't sure where best to leave this comment, so I'm attaching it here but it applies to most of the complex schema types)

We seem to be going to a lot of trouble to avoid requiring callers to pass `MemberSchema` values for list elements, map keys/values, structure members, etc. But we do have a `MemberSchema` and it is independently instantiable. Moreover, when we're parsing these models, we'll be traversing member shapes before we traverse their targets—we should already be able to build the member schema ourselves. This has added a lot of extra boilerplate to the schema implementations for what seems like negligible value.

Why not simply define `public var element: MemberSchema` and make the caller pass in a built member schema that already contains its own traits?

ianbotsf (changes requested):

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/schema/ShapeId.kt
luigi617: My understanding is that all shape ID requires to be absolute. If a model uses only one namespace, it means that a shape can be identified using only its relative ID but it doesn't mean that it doesn't have absoluteId.

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/schema/ShapeId.kt
luigi617: My initial thought on this is to have one function that can create both `ShapeId` and `MemberShapeId`. But I forgot why I wanted this and I will use `fun ShapeId` and `fun MemberShapeId` now which is preferred style.

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/schema/ListSchema.kt
luigi617: The default `toString()` will return something like _aws.smithy.kotlin.runtime.serde.schema.ListSchemaImpl@21b8d17c_, so it is better to override`toString`

luigi617: Yeah, it was a useless double lazy layer. Will remove it

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/schema/ListSchema.kt
luigi617: Agree

runtime/serde/common/src/aws/smithy/kotlin/runtime/serde/schema/ListSchema.kt
ianbotsf: Oh I missed that these aren't `data class`es. We should make them `data class`es so that we get nice default `toString`, `equals`, `hashcode`, etc.

ianbotsf (approved):
Looks pretty good. Just one lingering comment about `data class`es—fix and ship!

PR #1707, opened 2026-08-31
chore: add PR workflow for Catapult builds

ianbotsf:
## Issue \#

(n/a)

## Description of changes

Adds a new GitHub PR workflow for Catapult builds.

### DIFF
--- a/.github/workflows/pull-request-build.yaml
+++ b/.github/workflows/pull-request-build.yaml
@@ -18,7 +18,7 @@ env:
   ROLE_SESSION_DURATION_SECONDS: 7200
 
 jobs:
-  aws-sdk-pr-build:
+  pr-build:
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
     permissions:
--- a/.github/workflows/pull-request-build.yaml
+++ b/.github/workflows/pull-request-build.yaml
@@ -1,0 +1,54 @@
+name: Catapult build
+on:
+  pull_request:
+    types: [opened, synchronize, ready_for_review]
+    branches: [main]
+
+concurrency:
+  group: start-pull-request-build-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  # constants
+  DOWNLOAD_FOLDER: '.build-scripts/'
+  SCRIPT_LOCATION: 'workflows/start-pull-request-build/pull-request-build-v1.sh'
+
+  # custom variables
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request-build.yaml
```diff
@@ -0,0 +1,54 @@
...
+    if: github.event.pull_request.draft == false
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      issues: write
+      pull-requests: write
+      contents: read
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@main
```
luigi617, line 31: Nit: I'd suggest pinning to a version `@6` because otherwise an upstream change silently alters this action and makes it non-reproducible.

.github/workflows/pull-request-build.yaml
```diff
@@ -0,0 +1,54 @@
...
+  group: start-pull-request-build-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  # constants
+  DOWNLOAD_FOLDER: '.build-scripts/'
+  SCRIPT_LOCATION: 'workflows/start-pull-request-build/pull-request-build-v1.sh'
+
+  # custom variables
+  IAM_ROLE_ARN: 'arn:aws:iam::769372916334:role/SmithyKotlinGitHubRole'
```
luigi617, line 17: Is it better to move this variable to secret? like secrets.CI_AWS_ROLE_ARN that we already have?

.github/workflows/pull-request-build.yaml
```diff
@@ -0,0 +1,54 @@
...
+    types: [opened, synchronize, ready_for_review]
+    branches: [main]
+
+concurrency:
+  group: start-pull-request-build-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  # constants
+  DOWNLOAD_FOLDER: '.build-scripts/'
```
luigi617, line 13: Nit: when using this variable, an additional / is added after the variable (so we have .build-scripts//.... I'd suggest remove / here

luigi617 (approved):
few comments, fix and ship!

.github/workflows/pull-request-build.yaml
ianbotsf: ARNs are not confidential and do not need to be stored in a secret.

PR #1716, opened 2026-09-10
feat(telemetry): add backend-agnostic metrics aggregation module

ashishdhingra:
## Issue \#

Related to aws/aws-sdk-kotlin#984 - CloudWatch `MetricsProvider`. This is the backend-agnostic half; the CloudWatch exporter that consumes it is aws/aws-sdk-kotlin#1998.

## Description of changes

Adds `telemetry-metrics-aggregation`, an in-process pipeline that aggregates SDK instruments and hands them to a pluggable `MetricExporter`, plus a codegen change so generated clients manage the telemetry provider's lifetime. Aggregation is kept separate from publishing rather than layered on EMF, which writes to stdout and needs the surrounding platform to ingest it - ruling out EC2 without the agent, on-premises, and Kotlin/Native. With no CloudWatch dependency here, other exporters (Prometheus, OTLP) can reuse the module.

`feat(telemetry)`, the module:

- `AggregatingTelemetryProvider`, a `TelemetryProvider` that is additively `Closeable`. Ownership follows creation: a provider the caller builds is theirs to close and is safe to share across clients.
- Delta aggregation for counters and histograms, absolute for gauges and async up/down counters - backends like CloudWatch aggregate within a period, so a cumulative total would report the same sum every interval.
- `Summary` (count/sum/min/max) by default, a fixed footprint per attribute set; full `Distribution` is opt-in per instrument, since its memory grows with the number of distinct values.
- A cardinality guard that buckets overflow instead of growing without bound, because attribute values influenced by remote responses are unbounded in principle.
- `PeriodicMetricReader`, which serialises collect-then-export in one coroutine and switches to on-demand flushing on Lambda, where a timer may never fire.
- Tracing and context inherited as `None`; claiming to provide tracing would silently discard spans.

`feat(codegen)`: generated clients register `config.telemetryProvider` with their managed-resource group, so an SDK-created provider is reference counted and released with the last client using it. A caller-supplied provider is not `SdkManaged`, so `addIfManaged` skips it. Without this the SDK-created provider was never closed, leaking its collection coroutine and any connections its exporter held.

`chore(telemetry)`: review feedback - the `Sdk*` renames, the abstract base collapsed into one `final` class, public `data class`es replaced with plain classes, `MetricExporter.onAttach()` removed, KDoc trimmed.

### Merge ordering

**This PR must merge and release before aws/aws-sdk-kotlin#1998**, whose version catalog references `aws.smithy.kotlin:telemetry-metrics-aggregation` - a module in no released version yet, so that PR's CI cannot resolve it until this ships in 1.7.11.

### Testing

`:runtime:observability:telemetry-metrics-aggregation:build` passes on JVM and Kotlin/Native (mingwX64) - 29 tests per target, plus `apiCheck` against the committed baseline. End-to-end verification through a real exporter is on aws/aws-sdk-kotlin#1998.
___


runtime/observability/telemetry-metrics-aggregation/common/src/aws/smithy/kotlin/runtime/telemetry/metrics/aggregation/internal/ManagedTelemetryProvider.kt
```diff
@@ -0,0 +1,62 @@
...
+ * @param closeable the same instance as [delegate], viewed as [Closeable]. Two parameters rather than one
+ *   type-parameterized constructor: `TelemetryProvider` has no closeable sub-interface to constrain against
+ *   (unlike `CloseableHttpClientEngine`), and inferring a type argument from a smart cast to an intersection
+ *   type is not something to rely on. Both parameters are always the same object — see [manage].
+ */
+private class ManagedTelemetryProvider(
+    delegate: TelemetryProvider,
+    closeable: Closeable,
+) : SdkManagedCloseable(closeable),
+    TelemetryProvider by delegate
```
ianbotsf, line 34: **Question**: Two parameters that always hold the same value is kinda gross. Why not simply create `ClosableTelemetryProvider`?

runtime/observability/telemetry-metrics-aggregation/common/src/aws/smithy/kotlin/runtime/telemetry/metrics/aggregation/AggregatingTelemetryProvider.kt
```diff
@@ -0,0 +1,104 @@
...
+ * consumers' results to be comparable.
+ *
+ * ### Why this class is not part of the supported API
+ *
+ * An abstract class pins its protected surface: every `protected` member becomes something third-party
+ * subclasses may depend on and that cannot change without a breaking release. `@InternalApi` keeps that
+ * surface free to change — the class must be `public` only because [SdkTelemetryProvider], which is
+ * supported, extends it, and Kotlin does not allow a public subclass of an internal class. Promoting it to
+ * supported later is a compatible change; retracting it would not be. The extension point for callers and
+ * for other backends is [MetricExporter].
```
ianbotsf, line 47: **Nit**: Your agent is very verbose and leaves lots of messy moment-in-time context in the docs. Please scale it back to the essentials a consumer would need to know and leave the design rationale in a design doc.

(applies broadly to this PR)

runtime/observability/telemetry-metrics-aggregation/common/src/aws/smithy/kotlin/runtime/telemetry/metrics/aggregation/AggregatingTelemetryProvider.kt
```diff
@@ -0,0 +1,104 @@
...
+    Closeable {
+    /**
+     * Diagnostics for the pipeline itself — cardinality overflow, export failures, dropped datums. Routed
+     * through the caller's [LoggerProvider] so they land wherever the application's logs go.
+     *
+     * `protected` because subclasses must pass it to whatever drives collection; [loggerName] is a
+     * constructor parameter so each provider's diagnostics are attributable to it rather than all appearing
+     * under a shared base-class name.
+     */
+    protected val logger: Logger = loggerProvider.getOrCreateLogger(loggerName)
```
ianbotsf, line 68: **Question**: Why are we eagerly initializing a `Logger` instance?

runtime/observability/telemetry-metrics-aggregation/common/src/aws/smithy/kotlin/runtime/telemetry/metrics/aggregation/AggregatingTelemetryProvider.kt
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * @param config dimension allowlist, aggregation fidelity, and the growth bounds.
+ * @param loggerProvider where the pipeline's own diagnostics go.
+ * @param loggerName names the logger so diagnostics are attributable to the concrete provider.
+ */
+@InternalApi
+public abstract class AggregatingTelemetryProvider(
+    config: AggregationConfig,
+    final override val loggerProvider: LoggerProvider,
+    loggerName: String,
```
ianbotsf, line 57: **Question**: Why require subclasses to pass a `loggerName` value during instantiation? Why not just use `this::class.qualifiedName` or something similar?

runtime/observability/telemetry-metrics-aggregation/common/src/aws/smithy/kotlin/runtime/telemetry/metrics/aggregation/Aggregation.kt
ianbotsf, line 1: **Nit**: We should name this file **Aggregator.kt** since that's the root type in the file.

runtime/observability/telemetry-metrics-aggregation/common/src/aws/smithy/kotlin/runtime/telemetry/metrics/aggregation/InstrumentRegistry.kt
```diff
@@ -0,0 +1,296 @@
...
+ *   instrument exists. Empty by default: percentiles cost memory and requests, so they are opted into
+ *   per instrument. See [DistributionAggregator].
+ * @param maxCardinality distinct attribute sets per instrument before overflow bucketing begins. 1000
+ *   by default — high enough that a normal service/operation breakdown never approaches it, low enough
+ *   to cap worst-case heap and spend at a survivable level. See [CardinalityGuard].
+ * @param maxDistinctValues distinct values per attribute set for instruments in [detailedMetrics].
+ *   The second, independent unboundedness axis: [maxCardinality] bounds how many series exist, this
+ *   bounds how large one series' accumulator can get.
+ */
+public data class AggregationConfig(
```
ianbotsf, line 46: **Correctness**: Do not expose `data class`es as public APIs unless they wrap a single value and will never wrap more than that. `data class`es cannot evolve to include new fields without [breaking binary compatibility](https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api).

runtime/observability/telemetry-metrics-aggregation/common/src/aws/smithy/kotlin/runtime/telemetry/metrics/aggregation/SdkInstruments.kt
```diff
@@ -0,0 +1,110 @@
...
+ * Every instrument here is functional. There are no no-op implementations: an instrument that accepts
+ * measurements and silently discards them is the specific defect this module exists to avoid.
+ */
+
+/**
+ * `MonotonicCounter` backed by a delta [SumAggregator].
+ *
+ * @see MetricValue.Sum for why the aggregation is delta rather than cumulative.
+ */
+internal class SdkMonotonicCounter(private val instrument: SyncInstrument) : MonotonicCounter {
```
ianbotsf, line 33: **Question**: Why are these classes named "SDK"?

ianbotsf (changes requested):

runtime/observability/telemetry-metrics-aggregation/common/src/aws/smithy/kotlin/runtime/telemetry/metrics/aggregation/internal/ManagedTelemetryProvider.kt
ashishdhingra: Done — one type-parameterized parameter now: `private class ManagedTelemetryProvider<T>(delegate: T) : SdkManagedCloseable(delegate), TelemetryProvider by delegate where T : TelemetryProvider, T : Closeable`.

runtime/observability/telemetry-metrics-aggregation/common/src/aws/smithy/kotlin/runtime/telemetry/metrics/aggregation/AggregatingTelemetryProvider.kt
ashishdhingra: Point taken, and thanks for the direct feedback. Trimmed the KDoc and comments across both PRs down to what a consumer needs and dropped the moment-in-time justifications; the rationale lives in the design doc instead.

runtime/observability/telemetry-metrics-aggregation/common/src/aws/smithy/kotlin/runtime/telemetry/metrics/aggregation/AggregatingTelemetryProvider.kt
ashishdhingra: It's consumed two lines later by the `AggregatingMeterProvider` built in the same initializer, so `by lazy` would resolve during construction anyway and only add a synchronization wrapper.

runtime/observability/telemetry-metrics-aggregation/common/src/aws/smithy/kotlin/runtime/telemetry/metrics/aggregation/AggregatingTelemetryProvider.kt
ashishdhingra: Gone along with the abstract base — the single `final` `AggregatingTelemetryProvider` now calls `loggerProvider.getLogger<AggregatingTelemetryProvider>()` itself.

runtime/observability/telemetry-metrics-aggregation/common/src/aws/smithy/kotlin/runtime/telemetry/metrics/aggregation/Aggregation.kt
ashishdhingra: Renamed to `Aggregator.kt`.

runtime/observability/telemetry-metrics-aggregation/common/src/aws/smithy/kotlin/runtime/telemetry/metrics/aggregation/InstrumentRegistry.kt
ashishdhingra: Done — the nine public `data class`es in `MetricData.kt` plus `FlushMode.Periodic` are now plain classes with hand-written `equals`/`hashCode`/`toString`. Kept equality where it's load-bearing (`InstrumentDescriptor` keys the instrument registry, `Dimension` feeds `DimensionKey`); nothing used `copy()` or destructuring, so no callers changed.

runtime/observability/telemetry-metrics-aggregation/common/src/aws/smithy/kotlin/runtime/telemetry/metrics/aggregation/SdkInstruments.kt
ashishdhingra: Fair — they described who wrote them, not what they do. Renamed to `AggregatingTelemetryProvider`, `AggregatingMeterProvider`, `AggregatingMeter`, and `SdkInstruments.kt` -> `Instruments.kt`.

ianbotsf (approved):

